PR #8, opened 2026-04-29
fix(ui) Rendering inline formating deletes prefixed spaces.

wviana:
## What

While using Reasonix I noticed that inline code renderings were missing the space before the opening backtick — text like ``use `DEEPSEEK_API_KEY` env var.`` displayed as `useDEEPSEEK_API_KEY env var.`, with the word before the code span welded to the code content. The same happened with bold, italic, and link spans: `this is **bold** text` rendered as `this isbold text`.

This PR adds 5 unit tests that expose this rendering bug. **No fix is included yet** — this is a diagnostic first step to lock down the expected behaviour before changing the renderer.

Captura de Tela 2026-04-29 às 15 06 23

Notice it before the parseInt in blue

## Why

The `segmentsToFrame` function in `src/cli/ui/markdown-frame.ts` renders each inline segment separately via `text()` at a wide width, then strips trailing unstyled spaces from the rendered row to remove **padding cells** (which `text()` adds to fill the requested row width) before concatenating atoms into a flat list for word-wrap.

The stripping loop (`markdown-frame.ts:93-99`) walks backwards from the end of the row and drops every cell where `char === " "` and `fg === undefined && bg === undefined`. This correctly removes the padding — but it **also** removes content spaces that happen to sit at the end of a segment's rendered content, because those content spaces have the same unstyled appearance as padding cells.

Every inline span triggers this: `parseInline` splits `` "use `foo()`" `` into the segments ``["use ", "foo()", ""]``, where the first segment ends with a real space. That space gets stripped together with the padding, producing atoms `['u','s','e']` instead of `['u','s','e',' ']`. The same applies to `**bold**`, `*italic*`, `[link](url)`, and `~~strike~~`.

| Source text | What renders (bug) | What should render |
|---|---|---|
| ``use `foo()` for output`` | `usefoo() for output` | `use foo() for output` |
| `this is **bold** text` | `this isbold text` | `this is bold text` |
| `this is *italic* text` | `this isitalic text` | `this is italic text` |
| `see [here](…) for details` | `seehere for details` | `see here for details` |
| ``Set it via `KEY` env var.`` | `Set it viaKEY env var.` | `Set it via KEY env var.` |

### Scope

- **Affected**: the Frame-based rendering path (`markdownToFrame` / `markdown-frame.ts`), used in the TUI log for warning/status rows, tool body content, and plan event bodies.
- **Not affected**: the React-based `InlineMd` component (`markdown.tsx`), which passes plain text segments directly to Ink's `<Text>` and preserves whitespace naturally.
- **All inline span types** are affected (code, bold, italic, links, strikethrough) because they all flow through the same `parseInline` → `segmentsToFrame` pipeline.

### Why wasn't this caught before?

Existing tests only verified that styled cells *exist* — e.g. bold cells have `bold: true`, code cells have `bg: "#0f172a"`. None of them checked that the surrounding whitespace is preserved. The test `"preserves bold styling on inline markup"` asserts `boldChars.toContain("b")` — it never looks at whether there's a space before that `b`.

## How to verify

```sh
npm test -- tests/markdown-frame.test.ts
```

5 tests will fail with assertions showing the missing space:

```
 FAIL  tests/markdown-frame.test.ts > markdownToFrame > preserves spaces before inline code spans
AssertionError: expected 'usefoo() for output …' to contain 'use foo()'
```

Run `npx vitest run` (the full suite) to confirm no other tests are broken — all existing 15 still pass.

## Checklist

- [ ] CHANGELOG updated under `## [Unreleased]` if user-visible
- [ ] Comments follow CLAUDE.md (no module-essay headers, no incident history)

### DIFF
--- a/src/cli/ui/markdown-frame.ts
+++ b/src/cli/ui/markdown-frame.ts
@@ -7,6 +7,7 @@ import {
   borderLeft,
   empty,
   pad,
+  stringWidth,
   text,
   vstack,
 } from "../../frame/index.js";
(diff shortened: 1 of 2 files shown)


esengine:
@wviana Diagnosis is spot-on — I ran the 5 tests locally, all 5 fail with exactly the symptoms you describe (`isitalic`, `seehere`, `viaDEEPSEEK_API_KEY`, etc.), and the root cause at `markdown-frame.ts:90-97` is correct: that strip loop has no way to tell padding spaces apart from a segment's trailing content space when the segment is unstyled.

Tests look good and the writeup is excellent. Per repo convention I'd like to land the fix in the same PR (keeps main green and gives the test something to lock down in one step). When you have a moment, want to push the fix on top of this branch?

Hint if useful: the cleanest fix is to skip the strip entirely — call `text(seg.text, { ...seg.opts, width: stringWidth(seg.text) })` so `text()` never adds padding in the first place. `string-width` is already a dep. The trailing-space-strip loop (and the equivalent one in the multi-line branch right below it) can then be deleted.

Once it's green I'll do the review pass and merge. Thanks again for the careful diagnosis — this kind of "failing tests first" PR is genuinely the right way to start.

wviana:
> Per repo convention I'd like to land the fix in the same PR (keeps main green and gives the test something to lock down in one step). When you have a moment, want to push the fix on top of this branch?

Yes, this is also the flow I'm used to. It was not my intention to get it merged in RED state. I just didn't have the time to make the fix at the same moment, so I've decided to open the MR as WIP (work in progress)


Here is a evidence of the fix working

image

PR #12, opened 2026-04-30
feat: Timesteamped session allowing mutiple sessions by project.

wviana:
## What

Every launch of Reasonix showed a spurious "RESUMED PLAN 0/3" banner -- even on brand-new sessions created with `--new`, the session picker "new"/"delete", or after `/forget`. The structured plan state file (`<session>.plan.json`) was never cleaned up when a session was reset or deleted, so on the next launch `loadPlanState` found the stale file and the UI faithfully displayed the orphaned plan.

## Root cause

Two problems, one superficial and one structural:

**A) Orphan sidecars** -- three session-wipe paths all failed to remove the plan-state sidecar:
1. **`deleteSession()`** -- cleaned up `.pending.json` sidecars but not `.plan.json`. Called by `/forget`.
2. **SessionPicker "new"/"delete" + `--new` flag** -- called `rewriteSession(name, [])` to truncate the JSONL log, but never called `clearPlanState(name)` to remove the plan file.
3. **No defense-in-depth in the UI** -- `App.tsx` restored plan state unconditionally whenever a `.plan.json` file existed, even when the session had zero prior messages (truly fresh).

**B) Destructive "new"** -- even with sidecar cleanup, picking "Start new conversation" from the SessionPicker *truncated* the existing session's `.jsonl` file. The old conversation was unrecoverable. The session name never changed: `code-reasonix` today, `code-reasonix` tomorrow. Every "new" destroyed the old session.

### Round 1: sidecar cleanup (surgical)

Three surgical changes, one per wipe path:

| File | Change |
|---|---|
| `src/memory/session.ts` | `deleteSession` now removes `.plan.json` sidecars alongside `.pending.json` |
| `src/cli/commands/chat.tsx` | `clearPlanState` is called alongside `rewriteSession([], [])` in both the SessionPicker handler and the `forceNew` block |
| `src/cli/ui/App.tsx` | Plan restoration is gated on `loop.resumedMessageCount > 0` -- a stale plan file on an empty session is never displayed |

### Round 2: timestamped session names (structural)

Instead of truncating the existing session on "new", generate a **fresh timestamped name** like `code-reasonix-20260430T154500`. Old session files stay intact on disk.

| File | Change |
|---|---|
| `src/memory/session.ts` | Added `findSessionsByPrefix(prefix)` -- lists `.jsonl` files matching a prefix, sorted by filename (alpha-reverse), so the most recently **created** session comes first. A `TODO` comment flags the alternative: `statSync(f).mtimeMs` would show the most recently **used** session instead, at the cost of `O(n)` extra I/O. |
| `src/cli/commands/chat.tsx` | **SessionPicker "new"** -- creates timestamped name instead of truncating. **`--new` flag** -- same. **Launch detection** -- checks for prefixed sessions first, falls back to exact name for backward compat. **`--resume`** -- resolves prefix to latest timestamped session. **"Delete" option removed** -- `/forget` covers deletion; "new" is now non-destructive. |
| `src/cli/ui/SessionPicker.tsx` | Removed the "Delete and start new" option -- only "Start new conversation" and "Resume" remain. |

What changed behaviorally:

| Scenario | Before | After |
|---|---|---|
| SessionPicker "Start new conversation" | Truncated `.jsonl` to empty (destructive) | Creates new session with timestamped name. Old session preserved. |
| `--new` flag | Truncated `.jsonl` to empty | Same -- generates fresh timestamped name. |
| Normal launch (no flags) | Only checked exact session name | Checks for prefixed sessions first, then falls back to exact name. Shows picker for most recent session found. |
| `--resume code-reasonix` | Resumed exact `code-reasonix` session | Resolves to latest `code-reasonix-*` timestamped session (if any), falls back to exact name. |

On-disk layout after the change:

```
~/.reasonix/sessions/
  code-reasonix.jsonl                       <- first-ever session (preserved)
  code-reasonix.plan.json
  code-reasonix-20260430T143200.jsonl       <- first "new" session
  code-reasonix-20260430T143200.plan.json
  code-reasonix-20260430T154500.jsonl       <- second "new" session (current)
  code-reasonix-20260430T154500.plan.json
```

No more orphan plans -- each `.plan.json` stays with its own `.jsonl`, and a "new" never destroys an old session.

### What stayed the same

- `/forget` -- still deletes all sidecars for a given session name (unchanged)
- `/prune-sessions` -- untouched; naturally cleans up old timestamped sessions by age
- `reasonix chat` (non-code mode) -- unchanged

### Future work

- **`/new` slash command (in-session)** -- still truncates the current session instead of creating a timestamped name. Fixing this requires:
  - Adding `setSessionName` to `CacheFirstLoop` so the in-memory log is flushed to a new file
  - Re-opening the event sink (`events.jsonl`) against the new name
  - This is more involved than the launch-time changes because the loop, event sink, and App state are all mid-session.

## How to verify

```sh
npm test -- tests/session.test.ts tests/plan-store.test.ts tests/pending-edits.test.ts
```

The new test `deleteSession removes the plan-state sidecar too` in `session.test.ts` fails before the Round-1 fix and passes after. The `findSessionsByPrefix` tests verify prefix matching, alpha-reverse sorting, and sidecar exclusion. Full suite:

```sh
npm test
```

## Checklist

- [x] Tests pass (`npm test` -- targeted suite green; 4 pre-existing locale-formatting failures unchanged)
- [x] Typecheck passes (`npx tsc --noEmit`)
- [ ] CHANGELOG updated if user-visible

### DIFF
--- a/src/cli/commands/chat.tsx
+++ b/src/cli/commands/chat.tsx
@@ -1,4 +1,3 @@
-import { existsSync, statSync } from "node:fs";
 import { render } from "ink";
 import React, { useState } from "react";
 import { loadApiKey, searchEnabled } from "../../config.js";
@@ -10,11 +9,7 @@ import { parseMcpSpec } from "../../mcp/spec.js";
 import { SseTransport } from "../../mcp/sse.js";
 import { type McpTransport, StdioTransport } from "../../mcp/stdio.js";
 import { StreamableHttpTransport } from "../../mcp/streamable-http.js";
-import {
-  loadSessionMessages,
-  rewriteSession,
-  sessionPath as sessionPathOf,
-} from "../../memory/session.js";
+import { resolveSession, timestampSuffix } from "../../memory/session.js";
 import { ToolRegistry } from "../../tools.js";
 import { registerChoiceTool } from "../../tools/choice.js";
 import { registerMemoryTools } from "../../tools/memory.js";
@@ -99,6 +94,13 @@ interface RootProps extends ChatOptions {
   progressSink: { current: ((info: ProgressInfo) => void) | null };
   /** Present when the session has prior messages; drives the picker. */
   sessionPreview?: { messageCount: number; lastActive: Date };
+  /**
+   * Original base session name before any timestamp resolution. Used
+   * by the "new" handler to generate fresh timestamped names without
+   * nesting (e.g. `code-reasonix` stays the base even when the
+   * effective session is `code-reasonix-20260430T143200`).
+   */
+  baseSession?: string;
 }
 
 function Root({
@@ -114,6 +116,9 @@ function Root({
   // `null` once the picker is resolved (or was never needed). Starts as
   // the preview so we can render the picker once before mounting App.
   const [pending, setPending] = useState<typeof sessionPreview>(sessionPreview);
+  // Effective session name — starts as the prop (resolved by chatCommand)
+  // but changes when the user picks "new" (creates a timestamped name).
+  const [effectiveSession, setEffectiveSession] = useState<string | undefined>(appProps.session);
 
   if (!key) {
     return (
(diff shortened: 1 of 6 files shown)


wviana:
@esengine the summary is: creating new session files with the timestamp. Please take a look on the code and write comments on things out of the project styles and I can get to fixed it later.

PR #62, opened 2026-05-01
test: add unit tests for clipboard.ts

### LINKED ISSUE #18: Add unit tests for `src/cli/ui/clipboard.ts`
## Background
`src/cli/ui/clipboard.ts` is a 31-line pure module that powers the TUI's "copy to clipboard" actions. It has two output paths — OSC 52 escape sequence to the terminal, and a fallback file write for content past a size budget — and zero direct test coverage. A good first contribution: write the unit tests.

## Current behaviour
- File: `src/cli/ui/clipboard.ts` (31 lines)
- Public surface: a single export (`writeClipboard`)
- Two branches:
  - **Short content** → write OSC 52 (`\x1b]52;c;<base64>\x1b\\`) to stdout
  - **Long content** (over the size budget, ~75K chars) → write to a file in `tmpdir` and return the path so the caller can show it

There's no `tests/clipboard.test.ts` today.

## Expected behaviour
A new test file `tests/clipboard.test.ts` that covers both paths.

## Files to touch
- `tests/clipboard.test.ts` (new)
- (no source changes — if the module needs adjusting to be testable, ask in the issue first)

## Acceptance criteria
- [ ] Test: short ASCII text → asserts the exact OSC 52 escape was written to a mock stdout, with correct base64 of the input
- [ ] Test: text containing UTF-8 (e.g. "你好世界 · ¥48.20") → encodes correctly (base64 of UTF-8 bytes, not UTF-16 surrogates)
- [ ] Test: empty string → either no-ops or writes a valid empty OSC 52 (whichever the implementation does today; document it)
- [ ] Test: text over size budget → writes a tmp file, return value points at it, file contents match the input
- [ ] Test: cleanup — tests don't leave orphan tmp files (use `os.tmpdir()` + a unique dir, clean in `afterEach`)
- [ ] All tests pass under `npm run test -- clipboard`

## Hints
- Look at `tests/paste-collapse.test.ts` for the project's vitest style — small, descriptive `it()` blocks, no test fixtures, plain assertions.
- Mock stdout via `vi.spyOn(process.stdout, "write")` — restore in `afterEach`.
- The base64 you assert against can be computed inline: `Buffer.from(input, "utf8").toString("base64")`.
- Don't read environment variables that affect the size budget — use whatever the source defines.

## Out of scope
- Refactoring `clipboard.ts` to add new features
- Cross-platform abstractions (Windows clipboard via PowerShell, etc.) — the current impl is OSC-only
- Mocking specific terminal emulators

## Difficulty
2-3 hours.

KuaaMU:
## Summary
Add unit tests for `src/cli/ui/clipboard.ts`

## Motivation
Closes #18 — the clipboard module has two output paths (OSC 52 and tmp file fallback) but zero test coverage.

## Changes
- New test file `tests/clipboard.test.ts` with 5 tests:
  - Short ASCII text → OSC 52 escape with correct base64
  - UTF-8 text (你好世界 · ¥48.20) → correct encoding
  - Empty string → valid OSC 52 escape
  - Content over 75K size budget → tmp file write with correct contents
  - Return size matches input for both paths

## Verification
```sh
npm test -- clipboard
```
All 5 tests pass. Full test suite (1684 tests) also passes.

## Notes
- Tests mock `process.stdout.write` to capture OSC 52 output
- No source changes — only adds test coverage
- Follows project's vitest style from `tests/paste-collapse.test.ts`

### DIFF
--- a/tests/clipboard.test.ts
+++ b/tests/clipboard.test.ts
@@ -1,0 +1,76 @@
+import { existsSync, mkdirSync, readFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { writeClipboard } from "../src/cli/ui/clipboard.js";
+
+describe("writeClipboard", () => {
+  const testDir = join(tmpdir(), `reasonix-clip-test-${process.pid}`);
+  let stdoutWriteSpy: ReturnType<typeof vi.spyOn>;
+
+  beforeEach(() => {
+    mkdirSync(testDir, { recursive: true });
+    stdoutWriteSpy = vi.spyOn(process.stdout, "write").mockImplementation(() => true);
+  });
+
+  afterEach(() => {
+    stdoutWriteSpy.mockRestore();
+    if (existsSync(testDir)) {
+      rmSync(testDir, { recursive: true, force: true });
+    }
+  });
+
+  it("writes OSC 52 escape for short ASCII text", () => {
+    const input = "hello world";
+    const result = writeClipboard(input);
+
+    expect(result.osc52).toBe(true);
+    expect(result.size).toBe(11);
+
+    const expectedB64 = Buffer.from(input, "utf8").toString("base64");
+    expect(stdoutWriteSpy).toHaveBeenCalledWith(`\x1b]52;c;${expectedB64}\x1b\\`);
+  });
+
+  it("encodes UTF-8 correctly in OSC 52", () => {
+    const input = "你好世界 · ¥48.20";
+    const result = writeClipboard(input);
+
+    expect(result.osc52).toBe(true);
+    expect(result.size).toBe(input.length);
+
+    const expectedB64 = Buffer.from(input, "utf8").toString("base64");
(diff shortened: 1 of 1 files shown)


esengine:
Thanks for the contribution — coverage is solid and style matches the project. Codex flagged the only blocker: the teardown removes `testDir` but `writeClipboard` writes directly under `tmpdir()`, so the suite leaves orphan `reasonix-clip-*.txt` files (AC #5). Easiest fix is tracking `result.filePath` from each call and `rmSync`-ing in `afterEach`. Once that's in I'll merge.

KuaaMU:
Fixed! Pushed a commit that tracks `result.filePath` from each `writeClipboard` call and `rmSync`s them in `afterEach`.

Verified locally: all 5 tests pass and no orphan `reasonix-clip-*.txt` files remain in `tmpdir()` after the suite runs.

Changes:
- Added `createdFiles: string[]` array to track temp file paths
- Each test pushes `result.filePath` to the array when present
- `afterEach` iterates and removes all tracked files before cleaning `testDir`

esengine:
Merged — thanks for picking this up, @KuaaMU! Clean coverage of both branches (OSC 52 + tmp fallback), and the UTF-8 case is exactly the kind of thing that catches encoding regressions later. Appreciate it.

PR #86, opened 2026-05-02
feat(code): --system-append / --system-append-file for append-only system prompt customization

ctharvey:
## What

Adds `--system-append <prompt>` and `--system-append-file <path>` CLI options to `reasonix code`. Both append user-provided instructions to the default Reasonix Code system prompt under a `# User System Append` heading, without replacing or weakening the base prompt.

## Why

Users want to layer personal workflow rules (always inspect files before editing, always run linter after changes, etc.) on top of Reasonix Code's built-in tool-use and edit-protocol instructions. The `chat` command has `--system` for full overrides, but `code` mode's prompt is carefully engineered for cache stability — a full override would break that. Append-only is the safe middle ground.

## How to verify

```
npm run verify
```

Manual smoke test:
```
npx reasonix code --system-append "Test instruction"
npx reasonix code --system-append-file ./prompt.md
npx reasonix code --system-append "First" --system-append-file ./prompt.md
```

## Checklist

- [x] CHANGELOG updated under `## [Unreleased]`
- [x] Comments follow CLAUDE.md (no module-essay headers, no incident history)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 All notable changes to Reasonix. The format follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/);
 this project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+- feat(code): `--system-append` / `--system-append-file` CLI options append user instructions to the code system prompt without replacing the default prompt. Both can be used together (inline first, then file contents). File read errors surface with their OS error code.
+
 ## [0.17.1] — 2026-04-29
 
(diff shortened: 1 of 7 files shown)


esengine (approved):
Thanks for sending this — and for working *with* the cache-first invariant rather than around it. The append-only framing is exactly right for `code` mode; nice to see the PR body argue it from Pillar 1 directly.

One heads-up for next time: per CONTRIBUTING.md, new features should land an issue first so we can align on scope before code is written. This PR is small and well-scoped enough that I'm merging as-is, but the convention exists because cold "we could add X" PRs occasionally need rescoping mid-review, and that's no fun for either side. Two paragraphs in an issue is plenty.

Approving and merging.

PR #135, opened 2026-05-02
Add multi-language support with Chinese/English toggle

kabaka9527:
- Add /language command for language switching
- Implement language state management and persistence
- Update UI components to support multi-language display
- Add translation files and Chinese/English resources
- Update test cases to support multi-language scenarios
- Refactor prompt texts into translatable format
- Fix MCP spec in slash test to prevent stderr leak

What
Add i18n (internationalization) support across both the CLI and web dashboard. Introduce a /language command to switch between English and Chinese at runtime. All UI strings go through a t() function, and the language preference is persisted to disk so it survives restarts.

Why
Feature enhancement. Enables Chinese-speaking users to operate Reasonix in their native language, lowering the barrier to entry.

How to verify
npm run verify — 116 test files, 1830 tests all pass
Launch the CLI, run /language zh-CN to switch to Chinese, /language EN to switch back
Open the Dashboard, go to Settings and toggle the language — all panels should refresh immediately without a page reload
Restart the CLI or Dashboard and confirm the language preference persists

Checklist
npm run verify passes locally (lint + typecheck + tests + comment-policy gate)
No Co-Authored-By: Claude trailer in commits
Comments follow CONTRIBUTING.md (no module-essay headers, no incident history)
No edits to CHANGELOG.md — release notes are maintainer-written at release time


esengine:
Thanks for taking this on, @kabaka9527 — i18n is a real ask and the dashboard side is laid out cleanly (the `useLang()` hook + `tabSections()` re-eval pattern is the right call). I'm going to keep reviewing the rest, but a few things need to be reworked before I can merge, and I want to flag them up front so you're not waiting.

**The big one — please don't merge as-is.** The new i18n catalog stores model-facing system prompts as truncated stubs:

In [`src/i18n/EN.ts`](https://github.com/esengine/reasonix/blob/main/src/i18n/EN.ts) the values for `prompts.defaultSystem`, `prompts.formattingRules`, `prompts.escalationContract`, `prompts.negativeClaimRule`, and `prompts.commitSystem` all literally end in `…` — the body of each rule was summarized away. For example `formattingRules` becomes just `"Formatting (rendered in a TUI with a real markdown renderer):..."`, and `defaultSystem` cuts off mid-sentence after "Every factual claim about a codebase must be backed by evidence...". Same in [`zh-CN.ts`](https://github.com/esengine/reasonix/blob/main/src/i18n/zh-CN.ts).

Those strings are not UI labels — they're sent to the model as the system prompt. After this PR, even an English-locale user loses the cite-or-shut-up rules, the negative-claim rule, the escalation contract, and the markdown-link citation format. That's a behavior regression for every user, not just zh-CN. (My guess is an LLM-assisted summarization step truncated them; easy mistake, but we have to restore the full text.)

**Architectural ask, even after the strings are restored:** I'd like system prompts to stay out of the i18n catalog entirely. Two reasons:

1. The original `prompt-fragments.ts` had a comment — `Embedded literally — no interpolation, so prefix-cache hash stays stable across sessions` — that got removed in the diff. Routing prompts through `t()` makes the prefix cache key language-dependent and splits the cache.
2. Translating model instructions changes how the agent reasons. Reasonix's behavior contract (citations, negatives, escalation) should be one canonical English spec, not two parallel specs that can drift.

The fix is to keep `TUI_FORMATTING_RULES`, `ESCALATION_CONTRACT`, `NEGATIVE_CLAIM_RULE`, `DEFAULT_SYSTEM`, and the commit prompt as plain English string constants in `prompt-fragments.ts` / `cli/index.ts` (their original form), and only translate the **user-visible** strings (slash descriptions, info messages, banners, dashboard labels). Same approach Cursor / Claude Code take.

**Smaller things I noticed so far:**

- `src/cli/ui/state/events.ts` and `state/reducer.ts` add a `language.change` event type but [`handlers/language.ts`](https://github.com/esengine/reasonix/blob/main/src/cli/ui/slash/handlers/language.ts) never dispatches it — re-render works only because `notifyLanguageChange()` bumps `languageVersion` on a side channel. Either wire the dispatch or drop the unused event type.
- `src/config.ts` types the new field as `lang?: string`; please narrow to `lang?: LanguageCode` so a typo in the config file doesn't silently fall through to English.
- The `notion=node nope` → `notion=tail -f /dev/null` test cleanup is a nice fix but it's unrelated to i18n — happy to take it, just calling out scope.
- `commitSystem` is in the catalog but I can't find a caller. If it's not wired yet, drop the key and add it in the PR that actually uses it.

**Process suggestion — this PR is hard to review at 4300/1450.** Once the prompt-fragments revert is done, would you mind splitting the rest into two PRs?

1. CLI i18n framework (`src/i18n/*`, `/language` handler, `WelcomeBanner`, slash command descriptions, state plumbing, tests)
2. Dashboard i18n (`dashboard/src/i18n/*`, panel changes)

They're independent, and at the current size it's hard for me to give the review the care it deserves. Each one on its own would be very mergeable.

Genuinely appreciate the work — the test coverage you added (`tests/i18n-notify.test.ts` + the `/language` block in `slash.test.ts`) is solid and the slash-handler shape matches the existing handlers cleanly. Get the system-prompt question sorted and we're most of the way there.

PR #152, opened 2026-05-03
fix(plan-mode): sharpen bounce + show write_file size while running

### LINKED ISSUE #150: plan-mode write UX: bounces feel post-hoc + write_file has no progress feedback
### Symptom

With `/plan` mode on, the user sees what looks like multiple `write_file` attempts before the model converges on `submit_plan`. The visual sequence is: tool card shows "writing…", then a brief error, then another tool card, then another error — the bounce reads as if it happened *after* a write actually went out.

### Actual code path

The plan-mode gate is at dispatch time, before `tool.fn()` ever runs:

```ts
// src/tools.ts:134-141
if (this._planMode && !isReadOnlyCall(tool, args)) {
  return JSON.stringify({
    error: `${name}: unavailable in plan mode — this is a read-only exploration phase. ...`,
  });
}
```

So no disk write happens. But each bounced attempt is still a full model round-trip — the model generates the call, gets the error, generates another call, gets bounced again, eventually settles on `submit_plan`. From the user's seat that's N spinner frames where it looks like writes are landing.

### Proposed fixes (pick one or stack)

- **a.** Sharpen the bounce message at `src/tools.ts:139`. Current copy reads *"Call submit_plan when you're ready for the user's review"*, which is permissive enough that the model treats it as advice. Tighten to imperative: *"STOP. Call submit_plan now. Do not retry write tools this turn."* The matching system-prompt line at `src/code/prompt.ts:57` already says *"Don't retry them"* — the per-bounce error should echo that.
- **b.** Surface the bounce as a banner / live card so the user can see "plan-mode rejected write" and stop interpreting the spinner as a write in progress.
- **c.** Short-circuit after the first bounce in a turn — intercept subsequent write attempts cheaply, without a full model round-trip, and roll up to "awaiting submit_plan".

(a) is the cheapest and probably enough on its own.

---

### Symptom

`write_file` (outside plan mode, normal flow) runs silently — the tool card shows the standard "running" spinner with no detail until completion. For large generated content or slow file systems (network mounts, WSL → Windows), the wait can be long enough to look hung.

### Code

```ts
// src/tools/filesystem.ts:498-515
fn: async (args: { path: string; content: string }) => {
  const abs = safePath(args.path);
  await fs.mkdir(pathMod.dirname(abs), { recursive: true });
  await fs.writeFile(abs, args.content, "utf8");
  return `wrote ${args.content.length} chars to ${pathMod.relative(rootDir, abs)}`;
},
```

The byte count is in the result, but only after the write completes. Nothing surfaces while it's in flight.

### Proposed fixes

- **a.** Surface `{content.length} chars → {path}` in the `ToolCard` subtitle the moment the call starts, not just on success. Cheap — `ToolCard` already has a subtitle slot for running tools, and the args are available at dispatch time.
- **b.** Stream the write in chunks and emit `tool.progress` events. More invasive — there's no `progress` channel on tools today. Defer.

(a) is a few lines in `ToolCard.tsx` (or wherever args render for in-progress tools). (b) is RFC territory.

---

## Out of scope

- Generic per-tool `progress` channel — would benefit other slow tools too; warrants its own design issue.
- Plan-mode policy changes (allowing a single "safe" write before requiring `submit_plan`, etc.).

## Refs

- Plan mode toggle wiring: `src/cli/ui/App.tsx:1644-1645`
- Plan-mode dispatch gate: `src/tools.ts:134-141`
- Bounce-message copy: `src/tools.ts:139`, mirrored in `src/code/prompt.ts:57`
- `write_file` tool: `src/tools/filesystem.ts:498-515`

esengine:
Closes #150.

### 1. Sharpen plan-mode bounce copy

`src/tools.ts:139` previously read like advice:

> *unavailable in plan mode — this is a read-only exploration phase. ... Call submit_plan with your proposed plan when you're ready for the user's review.*

"When you're ready" left room for the model to attempt more writes first. Tightened to an imperative:

> *BLOCKED in plan mode. Call submit_plan now with your proposed plan — writes execute only after user approval. Do NOT retry write tools this turn.*

System-prompt mirror at `src/code/prompt.ts:57` updated to reference the new opening (`"BLOCKED in plan mode"`) so the model recognizes the bounce marker.

Existing tests asserted the literal old phrase (`/unavailable in plan mode/`). Switched to case-insensitive `/plan mode/i` so future copy edits don't drag the test suite.

### 2. `write_file` size in the running subtitle

`ToolCard`'s generic args summary picked the first key, so for `write_file({ path, content })` users saw `"foo.ts  +1"` — nothing about the actual write size. The spinner ran silent until completion.

Per-tool summary in `formatArgsSummary`:

```ts
if (name === "write_file" && args && typeof args === "object") {
  const a = args as Record<string, unknown>;
  if (typeof a.path === "string" && typeof a.content === "string") {
    return `${a.path}  · ${a.content.length.toLocaleString()} chars`;
  }
}
```

Now the running tool card reads `▣ write_file  foo.ts · 1,240 chars` from dispatch — no more "did it hang?" moment.

## Test plan

- [x] `npm run verify` — 1851/1851 (existing assertions relaxed to wording-agnostic `/plan mode/i`)
- [ ] CI green
- [ ] Manual: with `/plan` on, ask the model to edit a file → bounce reads "BLOCKED in plan mode...". Outside plan mode, ask it to `write_file` a long string → ToolCard subtitle shows path + char count while spinner runs.

## Out of scope

- Generic per-tool `progress` channel (RFC territory — would need a tool primitive change).
- Surfacing the bounce as a banner / live card (option (b) in #150 — deferred; (a) sharpen + (a) write subtitle are the cheap wins).

### DIFF
--- a/src/cli/ui/cards/ToolCard.tsx
+++ b/src/cli/ui/cards/ToolCard.tsx
@@ -25,7 +25,7 @@ export function ToolCard({ card }: { card: ToolCardData }): React.ReactElement {
   const { stdout } = useStdout();
   const cols = stdout?.columns ?? 80;
   const lineCells = Math.max(20, cols - BODY_INDENT_CELLS - 1);
-  const argsLabel = formatArgsSummary(card.args);
+  const argsLabel = formatArgsSummary(card.name, card.args);
   const meta = formatMeta(card);
   const allLines = card.output.length > 0 ? card.output.split("\n") : [];
   const tail = tailLinesFor(card.name);
@@ -71,7 +71,15 @@ export function ToolCard({ card }: { card: ToolCardData }): React.ReactElement {
   );
 }
 
-function formatArgsSummary(args: unknown): string {
+function formatArgsSummary(name: string, args: unknown): string {
+  // write_file: surface byte count alongside the path so the user sees progress
+  // intent the moment dispatch starts, instead of staring at a silent spinner.
+  if (name === "write_file" && args && typeof args === "object") {
+    const a = args as Record<string, unknown>;
+    if (typeof a.path === "string" && typeof a.content === "string") {
+      return `${a.path}  · ${a.content.length.toLocaleString()} chars`;
+    }
+  }
   if (typeof args === "string") return args.length > 60 ? `${args.slice(0, 60)}…` : args;
   if (args && typeof args === "object") {
     const keys = Object.keys(args as Record<string, unknown>);
--- a/src/code/prompt.ts
+++ b/src/code/prompt.ts
@@ -54,7 +54,7 @@ Each option: short stable id (A/B/C), one-line title, optional summary. \`allowC
 # Plan mode (/plan)
 
 The user can ALSO enter "plan mode" via /plan, which is a stronger, explicit constraint:
-- Write tools (edit_file, write_file, create_directory, move_file) and non-allowlisted run_command calls are BOUNCED at dispatch — you'll get a tool result like "unavailable in plan mode". Don't retry them.
+- Write tools (edit_file, write_file, create_directory, move_file) and non-allowlisted run_command calls are BOUNCED at dispatch — you'll get a tool result starting with "BLOCKED in plan mode". Don't retry them.
 - Read tools (read_file, list_directory, search_files, directory_tree, get_file_info) and allowlisted read-only / test shell commands still work — use them to investigate.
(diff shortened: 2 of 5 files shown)


esengine:
Closing — direction not pursued. Reopening will require fresh discussion.

Issue #100, opened 2026-05-02
Add unit tests for `src/cli/ui/presets.ts`


esengine:
## Background

`src/cli/ui/presets.ts` defines the three model bundles (`auto` / `flash` / `pro`) and two pure helpers — `resolvePreset` and `canonicalPresetName` — used wherever the CLI maps a preset name to model + reasoning settings. The helpers also keep three legacy aliases alive (`fast` → `flash+high`, `smart` → `auto`, `max` → `pro`) so older `config.json` files don't break.

There are zero tests for this file. A drift on the alias table or a typo in the legacy mapping would silently change every user's session settings — exactly the kind of thing a small unit-test file pins down forever.

## Current state

- `src/cli/ui/presets.ts` (~80 lines) — pure functions only, no I/O, no React
- `tests/presets.test.ts` — does not exist

## Expected behaviour

A new `tests/presets.test.ts` exercising the public surface:

- `resolvePreset("auto")` / `"flash"` / `"pro"` — returns the canonical bundle
- `resolvePreset("fast")` — returns flash but with `reasoningEffort: "high"` (alias rewires effort)
- `resolvePreset("smart")` — returns the auto bundle byte-for-byte
- `resolvePreset("max")` — returns the pro bundle byte-for-byte
- `resolvePreset(undefined)` and unknown names — fall through to auto
- `canonicalPresetName(...)` — only returns `"auto" | "flash" | "pro"`; legacy and unknown values normalize to `"auto"`
- Branch / harvest invariant — every preset has `harvest: false` and `branch: 1` (the project rule that branch and harvest are NEVER auto-enabled)

## Files to touch

- `tests/presets.test.ts` (new, ~50-80 lines)

## Acceptance criteria

- [ ] All 6+ behaviours above covered with `it(...)` blocks
- [ ] An invariant test asserts that **every** entry in `PRESETS` has `harvest === false && branch === 1` (so the rule is enforced even if a future preset is added)
- [ ] No new dependencies; uses existing vitest setup
- [ ] `npm run verify` passes

## Hints

- Look at `tests/clipboard.test.ts` (closed via #18) for the shape of a small pure-function test file in this repo.
- Use `describe("resolvePreset", () => { ... })` / `describe("canonicalPresetName", () => { ... })` blocks.
- For the invariant test: `for (const [name, p] of Object.entries(PRESETS))` + a single `expect(p.harvest).toBe(false)` plus `expect(p.branch).toBe(1)`.
[...]

PR #180, opened 2026-05-03
test(gitignore): add unit tests for src/gitignore.ts

### LINKED ISSUE #139: Add unit tests for `src/gitignore.ts`
## Background

`src/gitignore.ts` (added in #129) is a small shared util — `ignoredByLayers`, `loadGitignoreAt`, `loadGitignoreAtSync` — that walks a layered `.gitignore` chain (root + every subdir) the way git itself does. It has two callers:

- `src/at-mentions.ts` — the `@`-picker walker
- `src/index/semantic/chunker.ts` — the semantic indexer

Both callers have end-to-end tests that exercise it indirectly, but the util has no **direct** unit tests. That means future refactors can break path-relativization or the directory-vs-file flag without anything failing visibly until someone notices their `.gitignore` is being misapplied. This is the same gap the closed GFIs #18 (`clipboard.ts`) and #100 (`presets.ts`) closed — small pure-ish util, real consumers, no test file.

## Current behaviour

- `src/gitignore.ts:14` — `loadGitignoreAt(dirAbs)`: async, reads `<dirAbs>/.gitignore`, returns an `Ignore` instance or `null` if the file doesn't exist.
- `src/gitignore.ts:22` — `loadGitignoreAtSync(dirAbs)`: same as above, sync.
- `src/gitignore.ts:31` — `ignoredByLayers(layers, abs, isDir)`: walks each layer outermost → innermost; for each layer, computes the absolute path relative to that layer's `dirAbs`, skips the layer if the path is outside it (`rel` empty or starts with `..`), then asks the `Ignore` instance whether the path is ignored (with a trailing `/` when `isDir=true`).

No `tests/gitignore.test.ts` exists.

## Expected behaviour

A focused unit-test file — `tests/gitignore.test.ts` — that locks in:

1. **`ignoredByLayers` branches**
   - Empty layers → returns `false`
   - Single layer matches a literal pattern (`node_modules/`)
   - Single layer with intra-file negation (`*.log\n!keep.log` — `keep.log` should NOT be ignored, others should)
   - `isDir` flag toggles dir-only matching: pattern `build/` matches `build` when `isDir=true`, does NOT match when `isDir=false`
   - Path outside the layer's `dirAbs` (e.g. layer rooted at `/proj/sub`, query is `/proj/other/file`) → that layer is skipped; falls through to next layer or returns `false`

2. **`loadGitignoreAt` / `loadGitignoreAtSync`**
   - Missing `.gitignore` → returns `null` (no throw)
   - Existing `.gitignore` → returns an `Ignore` whose `.ignores()` honors the patterns

A handful of tests (8–12) is the right size. Don't try to test the `ignore` library itself.

## Files to touch

- New file: `tests/gitignore.test.ts`
- (No source changes expected.)

## Acceptance criteria

- [ ] All branches in `ignoredByLayers` are exercised (empty layers, match, no-match, `..` skip, `isDir` toggle, intra-file negation)
- [ ] `loadGitignoreAt` and `loadGitignoreAtSync` each have at least: missing-file → `null`, existing-file → working `Ignore`
- [ ] Any FS-touching tests use `node:fs.mkdtempSync` (or `os.tmpdir()`) — never write into the real project root
- [ ] `npm run verify` passes (lint + typecheck + tests + comment-policy gate)
- [ ] No new dependencies (`ignore` is already in `package.json`)

## Hints

- `tests/presets.test.ts` (closed PR #100) is the closest stylistic parallel — small, pure, focused on locking invariants.
- For `ignoredByLayers`, build `Ignore` instances in-memory — no FS needed:
  ```ts
  import ignore from "ignore";
  const ig = ignore().add("*.log\n!keep.log");
  const layers = [{ dirAbs: "/proj", ig }];
  expect(ignoredByLayers(layers, "/proj/foo.log", false)).toBe(true);
  expect(ignoredByLayers(layers, "/proj/keep.log", false)).toBe(false);
  ```
  Only the two `loadGitignoreAt*` tests need a real tmpdir.
- `dirAbs` is intentionally absolute — your test layers should pass absolute paths (e.g. `path.resolve()` or `os.tmpdir()`).
- The contract for `ignoredByLayers` is **"any layer that ignores wins"** — DON'T write tests asserting cross-layer negation override (where an inner `!keep.log` un-ignores a path the outer layer ignored). That's a separate semantics question and would fail under the current implementation. Keep tests aligned with the current contract.

## Out of scope

- Adding new behavior to `gitignore.ts` (no API changes)
- Cross-layer negation-precedence semantics (see hints)
- Mocking `node:fs` — use a real tmpdir
- Performance / benchmarks

## Difficulty

2–3 hours.

mvanhorn:
## Problem

`src/gitignore.ts` (the layered `.gitignore` walker added in #129) has no direct unit tests. Both callers (`src/at-mentions.ts` and `src/index/semantic/chunker.ts`) exercise it indirectly, so a refactor that breaks path-relativization or the `isDir` flag wouldn't fail visibly until someone notices their `.gitignore` is being misapplied.

Same shape as the gaps closed by #18 (`clipboard.ts`) and #100 (`presets.ts`).

## Change

New file `tests/gitignore.test.ts`. 11 tests, no source changes, no new deps.

`ignoredByLayers`:
- Empty layers returns `false`
- Literal directory pattern under a single layer
- Intra-file negation (`*.log` + `!keep.log`)
- `isDir` flag toggles dir-only matching for `build/`
- Layer whose `dirAbs` does not contain the queried path is skipped
- Fall-through to a later layer when an earlier layer is out of scope
- No-match returns `false`

`loadGitignoreAt` / `loadGitignoreAtSync`:
- Missing `.gitignore` returns `null` (no throw)
- Existing `.gitignore` returns an `Ignore` whose `.ignores()` honors patterns

FS-touching tests use `mkdtempSync(tmpdir())` with `afterEach` cleanup; they never write into the project root.

## Validation

```
npx vitest run tests/gitignore.test.ts   # 11/11 pass
npm run lint                             # only pre-existing warnings (no new ones from this file)
npm run typecheck                        # clean
npm test --silent                        # 2064 tests pass, 3 skipped
```

Closes #139

### DIFF
--- a/tests/gitignore.test.ts
+++ b/tests/gitignore.test.ts
@@ -1,0 +1,91 @@
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import ignore from "ignore";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import {
+  type GitignoreLayer,
+  ignoredByLayers,
+  loadGitignoreAt,
+  loadGitignoreAtSync,
+} from "../src/gitignore.js";
+
+describe("ignoredByLayers", () => {
+  it("returns false when there are no layers", () => {
+    expect(ignoredByLayers([], "/proj/foo.log", false)).toBe(false);
+  });
+
+  it("matches a literal directory pattern under a single layer", () => {
+    const layers: GitignoreLayer[] = [{ dirAbs: "/proj", ig: ignore().add("node_modules/") }];
+    expect(ignoredByLayers(layers, "/proj/node_modules", true)).toBe(true);
+    expect(ignoredByLayers(layers, "/proj/src/index.ts", false)).toBe(false);
+  });
+
+  it("honors intra-file negation within a single layer", () => {
+    const layers: GitignoreLayer[] = [{ dirAbs: "/proj", ig: ignore().add("*.log\n!keep.log") }];
+    expect(ignoredByLayers(layers, "/proj/foo.log", false)).toBe(true);
+    expect(ignoredByLayers(layers, "/proj/keep.log", false)).toBe(false);
+  });
+
+  it("toggles dir-only matching via the isDir flag", () => {
+    const layers: GitignoreLayer[] = [{ dirAbs: "/proj", ig: ignore().add("build/") }];
+    expect(ignoredByLayers(layers, "/proj/build", true)).toBe(true);
+    expect(ignoredByLayers(layers, "/proj/build", false)).toBe(false);
+  });
+
+  it("skips a layer whose dirAbs does not contain the queried path", () => {
+    const layers: GitignoreLayer[] = [
+      { dirAbs: "/proj/sub", ig: ignore().add("*") }, // would match anything inside /proj/sub
+    ];
+    expect(ignoredByLayers(layers, "/proj/other/file.ts", false)).toBe(false);
+  });
+
+  it("falls through to a later layer when an earlier layer is out of scope", () => {
+    const layers: GitignoreLayer[] = [
+      { dirAbs: "/proj/sub", ig: ignore().add("*") }, // out of scope for the query
+      { dirAbs: "/proj", ig: ignore().add("dist/") }, // matches
+    ];
+    expect(ignoredByLayers(layers, "/proj/dist/main.js", false)).toBe(true);
+  });
+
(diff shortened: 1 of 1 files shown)


esengine:
Thanks @mvanhorn — coverage is exactly what #139 asked for, branch-by-branch on `ignoredByLayers` plus both sync/async load paths. Clean, well-scoped, no source churn. Merging.

Issue #196, opened 2026-05-03
Align on modular MCP runtime and UI boundaries


ctharvey:
## What

I’d like to work on some MCP runtime and UI cleanup, and wanted to align on scope before opening PRs.

MCP is already a core part of Reasonix. I’m not proposing a rewrite or trying to reframe it as something new. The thing I’m interested in is making the existing MCP pieces easier to work with: cleaner runtime boundaries, clearer lifecycle/status handling, and better TUI surfaces around what MCP servers are doing.

## Why

The MCP code already covers a lot: client/transport, registry bridging, reconnect behavior, drift handling, slash commands, browser UI, disable/reconnect helpers, and status output.

That is useful, but it also means MCP behavior is spread across runtime code and UI code in a few different places. When working on it, I found myself wanting a cleaner boundary between “what is the MCP server state?” and “how should the TUI show that state?”

I think tightening that seam would make future MCP work easier to land as small PRs.

## Areas I’m looking at

- clearer MCP server lifecycle state
- cleaner separation between MCP runtime logic and TUI formatting
- reusable server summaries or view models for `/mcp`, browser, toast, reconnect, and inspect surfaces
- enable / disable / reconnect / inspect behavior
- registry and tool identity behavior across reconnects
- tests around lifecycle transitions, recovery paths, and UI-facing status summaries

## Proposed approach

I’d like to keep this as a sequence of small PRs rather than one large change.

I do already have some local work in this direction. I’m happy to split it up, reshape it, or start with a smaller first PR depending on what you think the right boundary should be.

A possible first PR would be a no-behavior-change extraction around MCP server summaries or lifecycle view models, used by the existing UI paths. The goal would be to give the TUI a cleaner object to render without changing MCP behavior.

After that, follow-up PRs could improve status display, reconnect/disable feedback, and recovery behavior one piece at a time.

## Non-goals

- no broad rewrite
- no unrelated TUI redesign
- no external server bundle or integration in core
- no MCP protocol behavior changes without discussion
- no `CHANGELOG.md` edits
- no behavior changes without tests

## Question

Does this direction fit where you want MCP to go?

If so, what would you prefer as the first small PR boundary?

esengine:
Direction's a yes — the seam observation is right.

[...]

PR #201, opened 2026-05-04
refactor(mcp): move server summary out of UI types

ctharvey:
## Summary

Moves McpServerSummary from src/cli/ui/slash/types.ts (UI layer) into src/mcp/summary.ts (MCP layer) where it logically belongs, and adds a buildMcpServerSummary() builder function.

Relates to #196 — this is the first boundary (relocation + builder only, zero behavior change).

## Changes

- **New** src/mcp/summary.ts: canonical McpServerSummary interface + buildMcpServerSummary() builder
- **Updated** src/cli/ui/slash/types.ts: imports and re-exports McpServerSummary from the MCP layer; removed the inline interface definition and its now-unused InspectionReport/BridgeEnv/McpClientHost imports
- **Updated** src/cli/commands/chat.tsx: uses buildMcpServerSummary() instead of inline object literal

## Design notes

- slash/types.ts needs both import type (for local SlashContext usage) and export type (re-export for existing consumers) — TypeScript doesn't make re-exported names available locally
- buildMcpServerSummary() is a plain field-copy builder, preserving exact behavior. It enables future field validation/defaults without touching call sites
- Field doc comments from the original interface intentionally dropped per CONTRIBUTING.md (no what comments)

## Verification

- npm run verify passes: build, lint (Biome), typecheck, all 2150 tests
- Diff: +43 / -25 lines (3 files)

### DIFF
--- a/src/cli/commands/chat.tsx
+++ b/src/cli/commands/chat.tsx
@@ -9,6 +9,7 @@ import { parseMcpSpec } from "../../mcp/spec.js";
 import { SseTransport } from "../../mcp/sse.js";
 import { type McpTransport, StdioTransport } from "../../mcp/stdio.js";
 import { StreamableHttpTransport } from "../../mcp/streamable-http.js";
+import { buildMcpServerSummary } from "../../mcp/summary.js";
 import {
   deleteSession,
   listSessionsForWorkspace,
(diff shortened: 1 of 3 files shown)


esengine:
Clean execution — exactly the boundary I had in mind. The TS `import type` + `export type` re-export dance is right.

Merging. For the next pass I'd take the `healthBadge()` dedup (currently duplicated between `McpBrowser.tsx` and `slash/handlers/mcp.ts`) — same shape of PR, pure consolidation, no behavior change. Open it whenever you're ready, no need to re-align.

PR #216, opened 2026-05-04
fix(mcp): close client on startup failure

### LINKED ISSUE #215: MCP startup failure can leave a client process running
## What

If an MCP server fails during startup after the client has been created, Reasonix reports the failure and keeps going, but the partially-created MCP client is not closed.

That can leave the child process and read loop alive after the server has already been treated as failed.

This affects `chat`, `run`, and `code` startup paths. `code` goes through the same chat startup flow after registering its native tools, so the cleanup gap applies there too.

## Why

`chat`, `run`, and `code` intentionally treat per-server MCP startup failures as non-fatal. That part is good: one broken MCP server should not take down the whole session.

The gap is cleanup. If `mcp.initialize()` times out or rejects after the client has been constructed, the client has not yet been added to the normal cleanup list, so the catch path needs to close it directly.

There is also a Windows edge case here: stdio process shutdown can throw when given a POSIX-style signal name. Since this cleanup path is supposed to be best-effort, close should not be able to create a second failure while handling the first one.

## Expected behavior

When an MCP server fails during startup:

- Reasonix should still report the server failure and continue where appropriate
- any partially-created MCP client should be closed
- cleanup errors should not mask the original startup error
- stdio process shutdown should be best-effort across platforms

## Proposed fix

- keep a reference to the MCP client once constructed in `chat.tsx` and `run.ts`
- call `mcp?.close().catch(() => undefined)` in the startup failure path
- make `StdioTransport.close()` handle platform-specific process shutdown errors as best-effort cleanup

## Verification

Add regression coverage for `StdioTransport.close()` around:

- fast child exit
- failed spawn
- idempotent close
- thrown process-shutdown errors

Then run:

```sh
npm run verify
```

ctharvey:
## What

Closes a partially-created MCP client when server startup fails in `chat`, `run`, or `code`.

`code` goes through the same chat startup flow after registering its native tools, so this cleanup gap applies there as well.

This also makes stdio transport cleanup best-effort for Windows and failed-startup paths.

Closes #215.

## Why

MCP startup failures are intentionally non-fatal: Reasonix should report the broken server and keep going when it can. But if the client had already been constructed before startup failed, it was not yet in the normal cleanup list.

That left a small gap where the child process/read loop could stay alive after the server had already been marked failed.

## How to verify

```sh
npm run verify
```

Added regression coverage for `StdioTransport.close()` handling:

- fast child exit
- failed spawn
- idempotent close
- thrown process-shutdown errors

## Checklist

### DIFF
--- a/src/cli/commands/chat.tsx
+++ b/src/cli/commands/chat.tsx
@@ -210,6 +210,7 @@ export async function chatCommand(opts: ChatOptions): Promise<void> {
     if (!tools) tools = new ToolRegistry();
     for (const raw of requestedSpecs) {
       let label = "anon";
+      let mcp: McpClient | undefined;
       try {
         const spec = parseMcpSpec(raw);
         label = spec.name ?? "anon";
@@ -230,7 +231,7 @@ export async function chatCommand(opts: ChatOptions): Promise<void> {
             : spec.transport === "streamable-http"
               ? new StreamableHttpTransport({ url: spec.url })
               : new StdioTransport({ command: spec.command, args: spec.args });
-        const mcp = new McpClient({ transport });
+        mcp = new McpClient({ transport });
         await mcp.initialize();
         // Host indirection lets `/mcp reconnect` swap the underlying client
         // without re-bridging the registered tools (closures resolve via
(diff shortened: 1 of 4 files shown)


esengine:
Real fix — the lifted declaration + guarded `mcp?.close().catch(() => undefined)` is exactly the right shape so cleanup never masks the original bootstrap error. The monkey-patch test for the Windows EINVAL path on non-Windows CI is a nice touch.

Merging.

PR #226, opened 2026-05-04
refactor(mcp): deduplicate health badge helpers

ctharvey:
Refs #196.

This pulls the MCP health badge formatting into a small CLI UI helper without changing the existing output.

The two call sites looked almost the same, but they are not quite interchangeable:

- `McpBrowser` keeps `elapsedMs === 0` as `no inspect data`
- `/mcp` text output keeps the old threshold behavior, so `0ms` still renders as healthy

So this keeps two helpers in `mcp-health.ts` instead of forcing both paths through one shared function. The `/mcp` line formatting is left alone as well, including the existing spacing around the label and em dash.

I kept this intentionally narrow: no runtime MCP changes, no threshold changes, no output wording changes, and no tests added for the small extraction since the behavior preservation is visible in the diff.

Verified locally with:

- `npm run verify`

### DIFF
--- a/src/cli/ui/McpBrowser.tsx
+++ b/src/cli/ui/McpBrowser.tsx
@@ -5,6 +5,7 @@ import { Box, Text } from "ink";
 import React, { useState } from "react";
 import { useKeystroke } from "./keystroke-context.js";
 import { toggleMcpDisabled } from "./mcp-disable.js";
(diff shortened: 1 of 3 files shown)


esengine:
Thanks @ctharvey — clean extraction, the two-helper split with the `elapsedMs === 0` note is exactly the right call. Merging.

PR #237, opened 2026-05-05
fix: Model was still doing things while pending to approve

wviana:
## Discovery

The bug was first noticed during a session where the model called a non-allowlisted `run_command`. The tool threw `NeedsConfirmationError`, the TUI showed the ShellConfirm modal asking "Run once / Always allow / Deny" -- but while the user was still reading the modal, the model had already generated a follow-up response in the same turn ("Hmm, let me just write it directly instead... 1 2 3 ..."). The confirmation gate was effectively bypassed: the model answered the question before the user could approve or deny the command.

Later, the same pattern was observed during plan execution: when `mark_step_complete` fired its checkpoint modal (Continue / Revise / Stop), the model kept inferring in the background and produced unwanted output while the plan was paused for user review. The issue wasn't specific to shell commands -- it affected every tool that needed to pause for user input.

## Root cause

All five control-signal errors (`NeedsConfirmationError` in shell.ts, `PlanProposedError` / `PlanCheckpointError` / `PlanRevisionProposedError` in plan-errors.ts, `ChoiceRequestedError` in choice.ts) shared the same pattern: they were **thrown** from tool functions, caught by `ToolRegistry.dispatch()`, serialized to JSON, and appended to the message log as if they were normal tool results. The loop had no way to distinguish "this is a control signal, stop iterating" from a regular tool failure. It advanced to the next model call, and the model -- seeing the error text in its context -- naturally treated it as "that didn't work, try something else" and generated a workaround response.

The error messages literally said *"STOP calling tools now ... Wait for the next user message"* -- but that text was just a string inside a JSON tool result. At the architecture level, it had no effect on the loop's iteration logic.

## Fix

Replace all control-signal throws with a blocking pause gate. `PauseGate` is a generic Promise-based bridge that tool functions call to await user input synchronously (from the model's perspective). The App subscribes via `gate.on()` to show the correct modal, and calls `gate.resolve()` when the user picks.

```
shell.ts -> await gate.ask({ kind: "run_command", payload: { command } })
               |
  gate fires listener synchronously
               |
App.tsx -> pauseGate.on() -> shows ShellConfirm modal
User picks -> pauseGate.resolve(id, choice) -> promise resolves
               |
shell.ts -> runs command -> returns output -> dispatch() returns normally
loop -> continues same iteration -> model sees output inline
```

### Files changed

| File | What |
|---|---|
| `src/core/pause-gate.ts` | New -- `PauseGate` class with typed `ask({ kind, payload })`, `resolve(id, data)`, `on(fn)`, `current` |
| `src/tools/shell.ts` | `run_command`/`run_background` use `gate.ask()` instead of throwing `NeedsConfirmationError` |
| `src/tools/plan-core.ts` | `submit_plan`/`mark_step_complete`/`revise_plan` use `gate.ask()` instead of throwing `Plan*Error` |
| `src/tools/choice.ts` | `ask_choice` uses `gate.ask()` instead of throwing `ChoiceRequestedError` |
| `src/loop.ts` | Constructor accepts optional `PauseGate`, passes through to `dispatch()` |
| `src/tools.ts` | `ToolCallContext.confirmationGate` + `dispatch()` opts carry injectable `PauseGate` |
| `src/cli/ui/App.tsx` | Gate listener switches on `request.kind` to show correct modal; all modal callbacks call `gate.resolve()` |
| `src/cli/ui/hooks/handle-tool-event.ts` | Removed all error-string greps (NeedsConfirmationError, PlanProposedError, etc.) |
| `src/cli/ui/Select.tsx` | Added `onTab` prop for inline deny-context editor |
| `src/cli/ui/ShellConfirm.tsx` | Tab on "deny" opens context editor; updated footer hints |

## How to verify

```sh
npm test
npm run typecheck
npm run lint
```

### DIFF
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -33,6 +33,7 @@ import {
   saveEditMode,
 } from "../../config.js";
 import { Eventizer } from "../../core/eventize.js";
+import { pauseGate } from "../../core/pause-gate.js";
 import { type ResolvedHook, formatHookOutcomeMessage, loadHooks, runHooks } from "../../hooks.js";
 import { onLanguageChange } from "../../i18n/index.js";
 import { CacheFirstLoop, DeepSeekClient, ImmutablePrefix } from "../../index.js";
@@ -510,13 +511,8 @@ function AppInner({
   // project / Deny and we feed the result back as a synthetic user
   // message so the model sees what happened.
   const [pendingShell, setPendingShell] = useState<{
+    id: number;
     command: string;
-    /**
-     * Which tool surfaced the NeedsConfirmationError. Drives post-
-     * approval dispatch: `run_command` uses the synchronous runCommand
-     * (waits for exit); `run_background` spawns via JobRegistry and
-     * returns after a ready-signal / waitSec window.
-     */
     kind: "run_command" | "run_background";
   } | null>(null);
   // Plan text the model submitted via `submit_plan` while plan mode
@@ -542,6 +538,21 @@ function AppInner({
   const [stagedInput, setStagedInput] = useState<{
     plan: string;
     mode: "refine" | "approve" | "reject";
+  } | null>(null);
+  // Mid-execution pause from mark_step_complete — model finished a step
+  // and the loop waits for user to pick Continue / Revise / Stop.
+  const [pendingCheckpoint, setPendingCheckpoint] = useState<{
+    stepId: string;
+    title?: string;
+    completed: number;
+    total: number;
+  } | null>(null);
+  // Staged entry for the Revise feedback input at a checkpoint.
+  const [stagedCheckpointRevise, setStagedCheckpointRevise] = useState<{
+    stepId: string;
+    title?: string;
+    completed: number;
+    total: number;
   } | null>(null);
   // Plan revision proposal from `revise_plan`. Non-null mounts the
   // PlanReviseConfirm picker showing a step-level diff. Accept replaces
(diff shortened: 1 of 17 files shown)


esengine:
Thanks for digging into this — diagnosis is spot on and `PauseGate` is the right architectural direction. Promise-based blocking on `dispatch()` is the clean fix for control-signal-as-tool-result.

One regression that should land before merge:

**Checkpoint revise feedback is dropped.** When the user picks `revise` at a `mark_step_complete` checkpoint and types follow-up text, `handleCheckpointReviseSubmit` only `log.pushInfo`s the marker locally — the trimmed feedback never reaches the model. The old path posted a synthetic user message carrying that text; the new path resolves the gate immediately on the `revise` pick and ignores the staged input. Model has to guess what to revise.

Cleanup while you're there:
- `syntheticSubmit` in `App.tsx` (~2632) is dead code after the refactor — drop the hook
- `addProjectShellAllowed` import in `App.tsx:27` is now unused

For the long-term shape, I'd suggest extending the verdict types to carry user feedback inline rather than re-introducing a synthetic-message channel. e.g.

```ts
type CheckpointVerdict =
  | { type: "continue" }
  | { type: "revise"; feedback?: string }
  | { type: "stop" };
```

and have `mark_step_complete` surface the feedback in its returned string — same shape as how `deny` already carries `denyContext` and `ask_choice` carries `{ type: "text" }`. Everything flows through `tool_result`, which is where the model expects to read the world; no parallel "synthetic user turn" channel needed. Same pattern fits `RevisionVerdict` if we ever want to attach context to accept/reject.

Once the revise-feedback path is wired and the dead code is gone this is good to land.

wviana:
On it, I've also moved it to draft for now.

PR #247, opened 2026-05-05
chore: test coverage and mutation test

wviana:
## Summary

Add code coverage reporting and mutation testing to the development
toolchain. Coverage runs as part of CI and writes a summary table to
the GitHub Actions job page; mutation testing is available locally via
`npm run test:mutation` and helps spot tests that pass by accident
rather than by design.

## Motivation

The test suite is large (2,229 tests across 154 files), but until now
there was no visibility into which code paths are actually exercised.
Two gaps drove this work:

- **Coverage gaps.** A passing test suite doesn't tell you which
  branches are never hit, which error paths are untested, or which
  files have dead code. Coverage answers those questions.
- **Test quality.** Passing tests can be weak -- they might not fail
  when the code they exercise is subtly broken. Mutation testing
  introduces deliberate bugs (mutants) into the source and verifies
  that at least one test catches each one. If a mutant survives,
  either the test is too lenient or the code is dead.

Together, coverage and mutation testing turn the test suite from a
simple pass/fail gate into a tool for assessing *how well* the code is
tested.

### Coverage (`npm run test:coverage`)

- Uses `@vitest/coverage-v8` (V8 native instrumentation -- faster than
  Istanbul for Node-only code).
- Reporters: text (terminal), HTML (`coverage/index.html`), and
  JSON summary (`coverage/coverage-summary.json`).
- Configured in `vitest.config.ts` -- covers `src/**`, excludes
  `src/cli/ui/**` (Ink rendering is expensive to instrument and
  rarely a source of logic bugs).

### CI integration

- `npm run test:coverage` runs in CI (`npm test` step was replaced).
- A new `scripts/coverage-summary.mjs` script reads the JSON summary
  and writes a Statements / Branches / Functions / Lines table to the
  GitHub Actions job summary. No external services or tokens needed.

### Mutation testing (`npm run test:mutation`)

- Uses Stryker 9 with the Vitest runner.
- Configuration in `stryker.config.mjs` is scoped to the
  PauseGate refactor files (`src/core/pause-gate.ts`,
  `src/tools/shell.ts`, `src/tools/plan-core.ts`,
  `src/tools/choice.ts`, `src/loop.ts`) and their corresponding test
  files. This keeps runs fast (~20-40s) rather than full-suite.
- Thresholds: high 80%, low 60%, break 50%.
- `.stryker-tmp/` is git-ignored (sandbox directory Stryker creates
  during runs).
- Kept local-only for now -- mutation testing is too slow for every
  CI run. If a Stryker Dashboard integration is wanted later, add
  `dashboard.project` to `stryker.config.mjs` and an upload step to
  the workflow.

## How to use

```sh
# Run tests with coverage (also runs in CI on every push/PR)
npm run test:coverage

# Open the HTML coverage report in a browser
open coverage/index.html

# Run mutation testing on the focused file set
npm run test:mutation
```

Coverage output looks like:

```
 % Coverage report from v8
-------------------|---------|----------|---------|---------|
                   | % Stmts | % Branch | % Funcs | % Lines |
-------------------|---------|----------|---------|---------|
 src/core/pause-gate.ts |   100 |      100 |     100 |     100 |
 src/tools/shell.ts     |  96.5 |     82.1 |    87.5 |    96.5 |
 ...
-------------------|---------|----------|---------|---------|
```

## Files changed

| File | Change |
|---|---|
| `package.json` | Added `test:coverage` and `test:mutation` scripts; new devDeps |
| `vitest.config.ts` | Added `json-summary` reporter to coverage config |
| `.github/workflows/ci.yml` | `npm test` -> `npm run test:coverage`; added coverage job summary step |
| `scripts/coverage-summary.mjs` | New -- reads `coverage-summary.json` and writes a markdown table to `$GITHUB_STEP_SUMMARY` |
| `stryker.config.mjs` | New -- Stryker config targeting PauseGate + tool files, with thresholds |
| `.gitignore` | Added `.stryker-tmp/` |
| `package-lock.json` | Updated with new devDependencies |

## How to verify

```sh
npm run test:coverage   # should pass all 154 files, 2229 tests, and
                        # output a coverage table + json summary

npm run test:mutation   # should run stryker on the focused file set
                        # and report a mutation score >= 50%

npm run lint            # should pass (no config/lockfile drift)
npm run typecheck       # should pass
```


esengine:
Thanks for this — coverage in CI plus the GitHub Actions job-summary table is the right shape (no external service or token, summary lands directly on the PR page).

One thing to flag before this merges: `stryker.config.mjs` references `src/core/pause-gate.ts` and `tests/pause-gate.test.ts` in its `mutate` and `testFiles` lists. Those files don't exist on `main` yet — they come from #237. So while CI here is green (it only runs `npm run test:coverage`), `npm run test:mutation` would fail against `main` until #237 lands.

Plan: I'll review #237 next. Once that's in, please rebase this PR on top so the Stryker scope resolves cleanly. The coverage half (vitest config + CI step + summary script) is good as-is; only the Stryker file list depends on #237.

No other changes needed from you right now.

wviana:
> One thing to flag before this merges: stryker.config.mjs references src/core/pause-gate.ts and tests/pause-gate.test.ts in its mutate and testFiles lists. Those files don't exist on main yet — they come from https://github.com/esengine/reasonix/pull/237. So while CI here is green (it only runs npm run test:coverage), npm run test:mutation would fail against main until https://github.com/esengine/reasonix/pull/237 lands.

Nice catch. This is wrong then, it should be mutating everything. I'll fix that later.

stryker.config.mjs
```diff
@@ -0,0 +1,48 @@
...
+  ignorePatterns: ["home_sessions", ".reasonix", "node_modules"],
+
+  // Only mutate the files we changed (focus on PauseGate + tool integrations).
+  mutate: [
+    "src/core/pause-gate.ts",
+    "src/tools/shell.ts",
+    "src/tools/plan-core.ts",
+    "src/tools/choice.ts",
+    "src/loop.ts",
+  ],
```
wviana: This should not be only those hardcoded files. Should be pretty much everything.

esengine:
Agreed the PauseGate-only list is too narrow — that was scoped for an in-flight refactor. But I'd push back on "mutate everything":

Stryker generates dozens of mutants per file. Across `src/` (~470 files) that's well into five-digit territory, and each mutant reruns at least the relevant test slice. Even with Stryker's incremental mode, full-tree mutation runs end up in multi-hour territory — at which point nobody runs them and the tool becomes ornamental.

Mutation testing pays off when it's fast enough that contributors actually run it on the area they're touching. The original PR description had this right: "scoped to PauseGate refactor files... keeps runs fast (~20-40s)".

I'd suggest a middle ground — scope to the load-bearing modules where weak tests would be expensive to find late:

```js
mutate: [
  "src/loop.ts",
  "src/context-manager.ts",
  "src/core/**/*.ts",
  "src/tools/shell.ts",
  "src/tools/plan-core.ts",
  "src/tools/choice.ts",
  "src/repair/**/*.ts",
],
testFiles: [
  "tests/loop.test.ts",
  "tests/context-manager.test.ts",
  "tests/shell-tools.test.ts",
  "tests/plan.test.ts",
  "tests/choice.test.ts",
  "tests/repair.test.ts",
],
```

That keeps the run under a few minutes locally — short enough that people will actually use it before pushing — while covering the parts of the codebase where mutation testing genuinely beats coverage as a quality signal. UI rendering, MCP transport plumbing, and TUI primitives are deliberately excluded; they're better tested by snapshot/integration than mutation.

CI keeps running coverage only (your existing `npm run test:coverage` step is the right shape), so the mutation runtime never gates a merge — it's a developer-local diagnostic.

If you want to expand later as test density improves, lean toward the targeted approach over a broad include + exclude list.

PR #272, opened 2026-05-05
Fix: usd currency on status bar

wviana:
## Summary

Fix hardcoded yen symbols in StatusRow and UsageCard so all money-related
display values -- wallet balance, turn cost, and session cost -- render
with the correct currency symbol (`$` for USD wallets, `¥` for CNY
wallets).  Previously a USD wallet showed `¥0.91` instead of `$0.91`,
and turn/session costs always showed `¥` regardless of the user's
currency.

## Symptoms

A user whose DeepSeek account is denominated in USD saw:

```
¥0.0352 turn · ¥0.461 session · wallet ¥0.91
```

All three should adapt to the wallet currency:

```
$0.0308 turn · $0.064 session · wallet $0.91
```

The same bug appeared in UsageCard (full and compact variants) where
`balance ¥0.91` was rendered instead of `balance $0.91`.

Additionally the balance color showed *err* (red) when it should have
been *warn* (yellow): $0.91 ≈ ¥6.55, which falls in the 5-20 CNY
warn band, but the code passed the raw `0.91` straight to
`balanceColor`, which treated it as < 5 CNY.

ChromeBar did NOT have this bug -- it was fixed in v0.14.0 (commit
`33a3a7f`) with `formatBalanceLabel()`, but StatusRow and UsageCard
were never given the same treatment.  Turn/session costs were also
always displayed in CNY with no regard for the wallet currency.

## Root cause

Three problems in the data pipeline:

1. **State + events drop the currency.**  `App.tsx` dispatched only
   the numeric balance, discarding `balance.currency`.  `StatusBar`
   stored `balance?: number` with no currency field.  The Zod schemas
   for `sessionUpdate.patch` and `usageShow` rejected `balanceCurrency`.

2. **Components hardcode ¥ everywhere.**  `StatusRow.tsx:52,56,60`
   and `UsageCard.tsx:74,95` all used the literal `¥` character
   instead of a currency-aware formatter.

3. **Turn/session costs ignore wallet currency.**  `StatusRow.tsx`
   always multiplied costs by `USD_TO_CNY` (7.2) to show CNY,
   regardless of whether the wallet was in USD.

## What changed

| File | Change |
|---|---|
| `tokens.ts` | Exported `formatBalance`, `formatBalanceLabel`, `formatWalletDisplay`, `formatCost`, `balanceColorCny`, `balanceColorForBalance` |
| `state.ts` | Added `balanceCurrency?: string` to `StatusBar` |
| `cards.ts` | Added `balanceCurrency?: string` to `UsageCard` |
| `events.ts` | Added `balanceCurrency` to `sessionUpdate.patch` and `usageShow` schemas |
| `reducer.ts` | Passes `balanceCurrency` through `usage.show` handler |
| `App.tsx:858` | Dispatches `balanceCurrency: balance.currency` alongside `balance.total` |
| `useScrollback.ts` | Added `balanceCurrency` to `showUsageVerbose` args |
| `slash/types.ts` | Added `balanceCurrency` to `postUsage` callback type |
| `StatusRow.tsx` | Wallet uses `formatWalletDisplay` + `balanceColorForBalance`; turn/session costs use `formatCost` |
| `UsageCard.tsx` | Full card and compact row use `formatWalletDisplay` |
| `ChromeBar.tsx` | Replaced inline `formatBalanceLabel` with import from `tokens.ts` |
| `vitest.config.ts` | Removed `src/cli/ui/**` from coverage exclusions |

### How `formatCost` works

Turn/session costs follow the wallet currency:
- USD wallet: `formatCost(0.0308, "USD")` → `"$0.0308"` (no conversion)
- CNY wallet: `formatCost(0.0308, "CNY")` → `"¥0.2218"` (multiplied by 7.2)
- No wallet: `formatCost(0.0308)` → `"¥0.2218"` (backward compat; DeepSeek pricing is natively in CNY)

## Tests

55 new tests across 3 files, covering the full pipeline from pricing
to display:

| File | Tests | Layer |
|---|---|---|
| `tests/ui-reducer.test.ts` | +39 | Event schemas, state types, reducer propagation, all 6 format functions, full turn.end + session.update flow |
| `tests/ui-status-row-balance.test.tsx` | +9 | Renders real StatusRow via Ink: wallet, turn, session for USD/CNY/no-wallet; full turn.end + session.update flow |
| `tests/ui-usage-card-balance.test.tsx` | +5 | Renders real UsageCard via Ink: full card + compact row for USD/CNY/undefined balance |

### Coverage (after removing `src/cli/ui/**` exclusion)

| File | Lines |
|---|---|
| `tokens.ts` | 100% |
| `state.ts` | 100% |
| `events.ts` (state) | 100% |
| `UsageCard.tsx` | 98.85% |
| `StatusRow.tsx` | 54.16% (remaining: recording/network branches unrelated to this bug) |
| `reducer.ts` | 60.24% (remaining: plan/doctor/branch branches unrelated to this bug) |

### What the pure-function tests cover

| Function | Contract |
|---|---|
| `formatBalance({currency, total})` | `"$0.91"` for USD, `"¥6.55"` for CNY, `"EUR 1.23"` for unknown |
| `formatBalanceLabel({currency, total})` | `"w $0.91"` for USD (ChromeBar style) |
| `formatWalletDisplay(balance, currency?)` | `"$0.91"`, `"¥6.55"`, `null` when undefined |
| `formatCost(costUsd, currency?)` | `"$0.0308"` (USD), `"¥0.2218"` (CNY), `"¥0.2218"` (undefined/backward compat) |
| `balanceColorCny(cny)` | `< 5` err, `5-20` warn, `>= 20` brand |
| `balanceColorForBalance(balance, currency)` | Converts USD to CNY before threshold check; $0.91 -> warn, not err |

### What the Ink render tests cover

| Render test | Assertions |
|---|---|
| USD wallet | Wallet: `$0.91`, turn: `$0.0308`, session: `$0.064` |
| CNY wallet | Wallet: `¥6.55`, turn: `¥0.2218`, session: `¥0.461` |
| No wallet | Turn: `¥0.2218`, session: `¥0.461`, no wallet text |
| Full USD flow (turn.end + session.update) | `$0.0001 turn · $0.000 session · wallet $0.71` |
| Full CNY flow (turn.end + session.update) | `¥0.0011 turn · ¥0.001 session · wallet ¥6.55` |
| UsageCard full + compact | Both USD and CNY variants |

### Regression safety

Every test has a counterpart for each currency (USD, CNY, undefined).
The existing `tests/telemetry.test.ts` already covers `costUsd` pricing
calculation and session stats accumulation -- we added the bridge from
pricing through `turn.end` + `session.update` to StatusRow display.

### DIFF
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -862,7 +862,10 @@ function AppInner({
   useEffect(() => {
     balanceRef.current = balance;
     if (balance) {
-      agentStore.dispatch({ type: "session.update", patch: { balance: balance.total } });
+      agentStore.dispatch({
+        type: "session.update",
+        patch: { balance: balance.total, balanceCurrency: balance.currency },
+      });
     }
   }, [balance, agentStore]);
 
--- a/src/cli/ui/ChromeBar.tsx
+++ b/src/cli/ui/ChromeBar.tsx
@@ -5,7 +5,7 @@ import stringWidth from "string-width";
 import type { SessionSummary } from "../../telemetry/stats.js";
 import { ChromeRule } from "./primitives.js";
 import { COLOR, GRADIENT } from "./theme.js";
-import { formatBalanceLabel } from "./theme/tokens.js";
+import { formatBalance } from "./theme/tokens.js";
 
 const COLD_START_TURNS = 3;
 const CACHE_BAR_CELLS = 10;
@@ -54,7 +54,9 @@ export function ChromeBar(props: ChromeBarProps): React.ReactElement {
       ? `[↑ ${Math.round(props.scrollRatio * 100)}%]`
       : "";
   const updateLabel = props.updateAvailable ? `↑ ${props.updateAvailable}` : "";
-  const balanceLabel = props.balance ? formatBalanceLabel(props.balance) : "";
+  const balanceLabel = props.balance
+    ? formatBalance(props.balance.total, props.balance.currency, { label: true })
+    : "";
   const cachePct = Math.round(props.summary.cacheHitRatio * 100);
   // Use the worst-case rendered cache string (3-digit pct, full bar) so the
   // shed decision doesn't oscillate as the percentage changes turn-to-turn.
@@ -198,7 +200,9 @@ function BalancePill({
 }): React.ReactElement {
   const { total } = balance;
   const color = total < 1 ? COLOR.err : total < 5 ? COLOR.warn : COLOR.ok;
-  return <Text color={color}>{formatBalanceLabel(balance)}</Text>;
+  return (
+    <Text color={color}>{formatBalance(balance.total, balance.currency, { label: true })}</Text>
+  );
 }
 
 function CachePill({
--- a/src/cli/ui/ChromeBar.tsx
+++ b/src/cli/ui/ChromeBar.tsx
@@ -5,6 +5,7 @@ import stringWidth from "string-width";
 import type { SessionSummary } from "../../telemetry/stats.js";
 import { ChromeRule } from "./primitives.js";
 import { COLOR, GRADIENT } from "./theme.js";
+import { formatBalanceLabel } from "./theme/tokens.js";
 
 const COLD_START_TURNS = 3;
 const CACHE_BAR_CELLS = 10;
@@ -163,12 +164,6 @@ export function ChromeBar(props: ChromeBarProps): React.ReactElement {
       <ChromeRule />
     </Box>
   );
-}
-
-function formatBalanceLabel(balance: { currency: string; total: number }): string {
-  const sym = balance.currency === "USD" ? "$" : balance.currency === "CNY" ? "¥" : "";
-  const suf = sym ? "" : ` ${balance.currency}`;
-  return `w ${sym}${balance.total.toFixed(2)}${suf}`;
 }
 
 function CostPill({
--- a/src/cli/ui/cards/UsageCard.tsx
+++ b/src/cli/ui/cards/UsageCard.tsx
@@ -4,7 +4,7 @@ import React from "react";
 import { Card } from "../primitives/Card.js";
 import { CardHeader } from "../primitives/CardHeader.js";
 import type { UsageCard as UsageCardData } from "../state/cards.js";
-import { FG, TONE, formatCNY } from "../theme/tokens.js";
+import { FG, TONE, formatCNY, formatWalletDisplay } from "../theme/tokens.js";
 
 const BAR_CELLS = 30;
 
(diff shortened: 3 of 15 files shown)


wviana:
@esengine does your deepseek API returns the wallet balance as CNY? Could you checkout on this branch and check what it shows you and if it's working as you expect as a CNY user?

esengine:
Yes — my account is CNY, API returns `currency: "CNY"` with `total` in yuan. Pulled the branch, ran the new tests (60/60 pass) and traced the CNY render path by hand: turn cost, session cost, wallet amount, wallet color — all byte-identical to current behavior for a CNY user. No regression on my side.

Two notes while you're still iterating:

1. **`balanceCurrency` undefined fallback is inconsistent.** In `StatusRow`, `formatWalletDisplay(balance, undefined)` returns the bare number with no `¥`, but the sibling `balanceColorForBalance(balance, currency ?? "CNY")` defaults to CNY for the color. Either both should default to CNY (matching current behavior — `¥` was unconditional pre-fix) or both should treat "unknown currency" the same way. The transient case (balance arrived, currency not yet) is reachable on the first turn.
2. **`tokens.ts` ends up with 6 overlapping helpers** (`formatBalance` / `formatBalanceLabel` / `formatWalletDisplay`, plus `balanceColorCny` / `balanceColorForBalance`). I'd rather collapse to one `formatBalance(balance, currency, opts?)` + one `balanceColor(balance, currency)` than carry five thin wrappers.

esengine:
@wviana — pushed a fixup (`6d766f2`) to your branch addressing the two review notes:

1. **Undefined-currency fallback unified to CNY.** All three helpers (`formatBalance`, `formatCost`, `balanceColor`) now treat `undefined` currency as CNY — matches the pre-fix unconditional `¥` so the transient first-turn case (balance arrived, currency not yet) renders consistently.
2. **Seven helpers collapsed to three.** `formatCNY` / `formatBalance` (old object form) / `formatBalanceLabel` / `formatWalletDisplay` / `formatCost` / `balanceColorCny` / `balanceColorForBalance` → `formatBalance(amount, currency?, opts?)` + `formatCost(usd, currency?, digits?)` + `balanceColor(amount, currency?)`. The `cost vs amount` distinction is real (cost is internally USD, amount is in target currency) so I kept it as two format functions instead of one with a mode flag — three total still beats seven.

Also fixed a third bug surfaced while collapsing: **UsageCard turn cost (header) and session cost (body) used `formatCNY`**, which was unconditional `¥` regardless of wallet currency. They now use `formatCost(…, balanceCurrency)` like StatusRow does. Render tests added for the USD case.

`SessionPicker.tsx:143` has the same hardcoded `¥` problem but isn't in this PR's scope — opened as a follow-up issue (#312) so it doesn't block this merge.

esengine:
Heads up @wviana — went ahead and squash-merged so the fix lands in the next release. Your original commits + the fixup are all in the squash; thanks for digging into the data-pipeline side, the `balanceCurrency` plumbing through state/events/reducer/App was the right call. SessionPicker variant tracked at #312 if you want to take that one too.

Issue #324, opened 2026-05-06
feat(semantic): OpenAI-compatible remote embedding adapter


esengine:
Driven by discussion #310 — local Ollama on a CPU-only VM is too slow for first-time indexing, and pointing `OLLAMA_URL` at a remote box still requires running Ollama somewhere. A real third-party endpoint is the next step.

## Why

Current state (`src/index/semantic/embedding.ts`):

- Hardcoded to Ollama's `/api/embeddings` shape.
- `OLLAMA_URL` env var supports a *remote Ollama daemon* but not non-Ollama providers.
- `probeOllama` similarly Ollama-specific.

Gemini, OpenAI, vLLM, llama.cpp's server, LM Studio, and most third-party providers all expose `POST /v1/embeddings` with the same request/response shape. One adapter covers them all.

## Scope

**One adapter, OpenAI-compatible.** No per-vendor code paths.

- New env vars (or config keys):
  - `REASONIX_EMBED_PROVIDER` — `ollama` (default) | `openai-compat`
  - `REASONIX_EMBED_BASE_URL` — full base URL when provider is `openai-compat` (e.g. `https://generativelanguage.googleapis.com/v1beta/openai`, `https://api.openai.com/v1`, `http://localhost:8000/v1`)
  - `REASONIX_EMBED_API_KEY` — bearer token, sent as `Authorization: Bearer <key>`
  - `REASONIX_EMBED_MODEL` already exists — reused as the model name passed to the API
- Ollama remains the default. Zero behavior change for existing users.
- Adapter dispatch hidden behind the existing `embed()` / `embedAll()` / `probe*()` surface; callers see one API.

## Constraints

- **Vector dimensions per store.** Different models output different dims (Ollama `nomic-embed-text` = 768, OpenAI `text-embedding-3-small` = 1536, Gemini `text-embedding-004` = 768). The vector store must refuse mixing embeddings from different model identities — index header records `{ provider, model, dim }` and rejects appends that don't match. Caller can rebuild the index when switching providers.
- **No "free tier" framing in docs.** Document that an OpenAI-compatible endpoint *can* be Gemini / OpenAI / etc., without promising free. Free tiers move on someone else's schedule.
- **Errors stay actionable.** ECONNREFUSED keeps Ollama install hint when provider is Ollama; OpenAI-compat path needs its own "check your API key / base URL" hint on 401 / 404.

## Non-goals

[...]

PR #381, opened 2026-05-07
feat: update semantic search integration

kabaka9527:
## What
This pull request introduces support for OpenAI-compatible embedding providers to the semantic search feature. It allows users to configure custom API URLs, API keys, models, and request bodies for embedding generation, replacing the previously hardcoded Ollama-only setup. Additionally, it refactors the semantic search panel UI, adds comprehensive UI text improvements for both English and Chinese translations, and enhances the indexing job handling with more detailed status reporting.

## Why
This change enables users to leverage alternative embedding models or services that are compatible with the OpenAI API standard. This flexibility is crucial for users who may prefer different embedding models for performance, cost, or language support reasons. The update also improves the user experience by providing clearer status messages, better configuration options, and more robust error handling during the indexing process. No pre-existing issue is directly linked, but the changes fulfill a common feature request for broader embedding model support.

## How to verify
Start the application: Run reasonix code to start the main application or launch the dashboard directly.
Navigate to Semantic Search: Open the dashboard (via /dashboard command) and go to the semantic search settings panel.
Select a new provider: In the semantic panel, select "OpenAI-Compatible" from the provider dropdown.
Configure the provider: Enter a valid API URL, API key, and model name. Add any optional custom request body if needed.
Save the configuration: Click the "Save" button. The UI should show a success message without errors.
Build an index: Initiate a new index build. The job should proceed through "scanning", "embedding", and "writing" phases.
Verify functionality: After the index is built, perform a semantic search query. The search should return relevant results using the configured provider.
Check UI text: Verify that the updated English and Chinese text strings appear correctly in the interface (e.g., "incompatible status", "setup failed", "remote provider").

## Checklist


esengine (changes requested):
Thanks for the work on this — semantic-search provider extensibility has been on the wishlist (#324). Before merge there are a few things to settle.

## Blocker: CI is failing on Linux

`tests/server-dashboard.test.ts:235` does a real `mkdir("/test/project/.reasonix/semantic")`. The other tests in this file using `PROJ = "/test/project"` only pass it as a string into the API; they never touch disk. On Linux CI, `/test` is unwritable for a non-root user, so the test bombs with `EACCES`.

The fix is to use the same tmpdir pattern the surrounding tests already use:

```ts
it("GET /api/semantic reports incompatible on-disk index against current config", async () => {
  const proj = mkdtempSync(join(tmpdir(), "reasonix-dash-sem-"));
  try {
    const semanticDir = join(proj, ".reasonix", "semantic");
    await mkdir(semanticDir, { recursive: true });
    // ... rest of the test, with `proj` in place of PROJ
  } finally {
    rmSync(proj, { recursive: true, force: true });
  }
});
```

CI has to be green before this can merge.

## Scope and review process

The PR description reads as a focused semantic-search update, but the diff is **+1908/-545 across 20 files** — provider abstraction, batched `embedAll`, dashboard panel rewrite, API endpoint redesign, and a config schema expansion. That's substantial enough that I'd like to settle a couple of design points first:

1. **Provider shape.** `EmbedOptions` is currently a discriminated union over `"ollama" | "openai-compat"`. That works for two, but if Voyage / Gemini / others follow it'll grow awkward. Have you considered a small registry where each provider plugs in `{ probe, embed, embedBatch }`? Happy to land the union now if there's a follow-up plan, but want to flag it.
2. **Migration for existing Ollama users.** Please add a regression test asserting that an empty config (or a pre-existing `~/.reasonix/config.json` with no `semantic` key) still resolves to `provider: "ollama"` with the old defaults. We have users on Ollama-only setups; their next launch should be a no-op.
3. **API key in plaintext config.** Consistent with how the DeepSeek key is stored, so OK with me, but worth a one-liner note in the panel UI ("API key stored at ~/.reasonix/config.json — do not share that file") so users aren't surprised.

## Smaller things

- Please link `Refs #324` (or `Closes #324`) in the PR description so the issue auto-closes on merge.
- A few of the new tests in `tests/semantic-store.test.ts` and `tests/semantic-bootstrap.test.ts` look great — just want to confirm they exercise both providers, not only `openai-compat`.

Once CI is green and the migration test is in, I'll do a full pass.

Issue #350, opened 2026-05-07
Feature request: add explicit `wait` action/tool for intentional polling without tripping the repeat-loop guard


ctharvey:
## Summary

Reasonix should expose a first-class `wait` action for cases where the agent is intentionally waiting on asynchronous work to advance.

Today, the repeat-loop guard correctly suppresses identical tool calls when the model gets stuck. But it also catches a legitimate workflow: repeated polling while waiting for a long-running process, background job, dev server, MCP server, or similar async state transition.

An explicit `wait` primitive would let the agent yield time intentionally, without having to simulate waiting by repeating the same status/read action until the circuit breaker fires.

## Current behavior

The current repeat-loop suppression is implemented in `src/repair/storm.ts`. It suppresses identical `tool + args` tuples after 3 repeats within a sliding window.

That is generally correct, but it means benign polling patterns can look like runaway loops.

Relevant existing pieces:

- Repeat-loop suppression: `src/repair/storm.ts`
- Background jobs and polling tools: `src/tools/shell.ts`
- Architecture note for call-storm repair: `docs/ARCHITECTURE.md`

## Problem

This shows up when the agent is waiting for:

- `run_background` jobs to become ready
- `job_output` to produce new logs
- a dev server or MCP server to restart
- generated files to appear
- external state to settle before the next step

Without a wait primitive, the model has two bad options:

1. Repeat the same polling action until the storm breaker trips.
2. Guess and continue too early.

Neither is ideal.

## Proposed solution

Add a first-class `wait` action/tool that represents intentional waiting.

Example shape:

```ts
wait({
  durationMs: number,
  reason?: string,
  jobId?: number,
  wakeOn?: "timeout" | "job-exit" | "job-output"
})
```

## Expected behavior

- `wait` should **not count toward** the repeated-action storm breaker.
- `wait` should be **bounded** by harness-enforced limits.
- `wait` should create a **visible structured event** in the transcript / event log.
- If `jobId` is provided, the harness may wake early when the job exits or new output arrives.
- If the waited-on condition is already satisfied, it should return immediately.

## Why this is better than just relaxing the storm breaker

I do not think the best fix is "special-case some repeated tool calls and ignore them."

That would reduce safety clarity and hide intent.

An explicit `wait` action is better because it:

- makes waiting legible in logs and transcripts
- preserves the existing storm breaker for real loops
- gives the model a clear, sanctioned way to yield time
- allows runtime-level budgets and safeguards

[...]

PR #388, opened 2026-05-07
feat(storm): add explicit storm exemptions for safe inspector tools

ctharvey:
This implements the narrow version discussed in #350 rather than adding a generic `wait(...)` tool.

What changed:
- add a `stormExempt` flag on `ToolDefinition`
- skip storm accounting for tools marked `stormExempt`
- mark `read_file`, `list_directory`, `job_output`, and `list_jobs` as exempt
- keep existing repeat-loop behavior for non-exempt tools unchanged
- add regression coverage for exempt vs non-exempt behavior

Why this shape:
- matches the maintainer feedback on #350
- keeps the carve-out explicit and testable
- avoids adding a generic wait primitive that would broaden the surface area

Verification:
- `npm exec vitest run tests/repair/storm.test.ts`
- `npm run typecheck` currently fails on current upstream Ink/type drift in untouched files:
  - `src/cli/commands/chat.tsx`
  - `src/cli/ui/layout/CardStream.tsx`
  - `src/cli/ui/ticker.tsx`
- repo pre-push `verify` hook failed for the same upstream reasons, so this branch was pushed with `--no-verify`

Refs #350

### DIFF
--- a/src/loop.ts
+++ b/src/loop.ts
@@ -226,9 +226,15 @@ export class CacheFirstLoop {
       }
       return def.readOnly !== true;
     };
+    const isStormExempt = (call: ToolCall): boolean => {
+      const name = call.function?.name;
+      if (!name) return false;
+      return registry.get(name)?.stormExempt === true;
+    };
     this.repair = new ToolCallRepair({
       allowedToolNames: allowedNames,
       isMutating,
+      isStormExempt,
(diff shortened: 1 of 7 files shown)


esengine (approved):
Spec-perfect — exactly the narrow shape I asked for in #350. Tests cover both directions (exempt tool stays free, non-exempt still trips). Thanks for the clean implementation.

PR #396, opened 2026-05-07
test(mcp): cover startup summary states

ctharvey:
Part of #196 — MCP cleanup roadmap item 1.

## Summary
- add startup-state coverage at the `chatCommand()` seam
- cover the three startup handoff states the issue thread called out: live bridged, disabled with no live summary, and unbridged startup failure
- preserve configured MCP specs at startup so `/mcp` can still surface configured-but-unbridged servers on the marketplace fallback path

## Why this shape
The issue comment asked for startup summary state coverage before the larger UI/view-model extractions. This keeps the slice narrow and behavior-focused: one small startup handoff fix in `chat.tsx`, plus focused tests for the state transitions the MCP hub depends on.

## Testing
- `npm run verify`

Refs #196

### DIFF
--- a/src/cli/commands/chat.tsx
+++ b/src/cli/commands/chat.tsx
@@ -452,7 +452,9 @@ export async function chatCommand(opts: ChatOptions): Promise<void> {
   if (runtime.size() === 0 && !opts.seedTools) {
     tools = undefined;
   }
-  const mcpSpecs = runtime.specs();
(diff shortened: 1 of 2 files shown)


esengine (approved):
Verified locally — rebased onto current main and ran the full suite (136 files, 2180 passed). Narrow change to chat.tsx, focused tests for the three startup-handoff states. Thanks!

Issue #399, opened 2026-05-07
bug: Esc does not promptly stop recursive filesystem search


ctharvey:
## Repro

1. Start a long recursive filesystem search in `reasonix code`, for example `search_content` over a broad pattern/glob.
2. While the tool card is still showing `running`, press `Esc` repeatedly.
3. The turn eventually aborts, but only after the search finishes enough of its walk to return.

## Expected

`Esc` should abort the in-flight recursive search promptly, not only after the filesystem walk naturally returns.

## Actual

The loop-level abort is set, but `search_content` / `search_files` keep walking because the recursive JS search helpers do not observe the tool `AbortSignal` during traversal.

## Scope

This is not a new feature request. It is an abort-propagation bug in the existing filesystem search tools.

## Fix shape

Thread the existing tool abort signal into the recursive filesystem search helpers and check it during:

- directory walk
- file open/read
- per-line scan

That keeps the change narrow and avoids changing tool behavior apart from making `Esc` stop sooner.

ctharvey:
Opened the fix here: [#400](https://github.com/esengine/DeepSeek-Reasonix/pull/400)

This keeps the change narrow: it just makes the existing recursive filesystem search tools observe the existing abort signal so `Esc` stops them promptly instead of eventually.

esengine:
Resolved by #400 — thanks @ctharvey for the narrow, well-tested fix.

PR #400, opened 2026-05-07
fix(search): honor abort during recursive fs scans

ctharvey:
Refs #399.

## Summary

This fixes an abort-propagation bug in the recursive filesystem search tools.

When `search_content` or `search_files` is walking a large tree, pressing `Esc` in the TUI should stop the in-flight tool promptly. Before this change, the loop abort was set, but the recursive JS search helpers never checked the tool `AbortSignal`, so the turn only stopped once the search had naturally returned.

The fix keeps the scope narrow:

- thread the existing tool abort signal into `search_files` and `search_content`
- check it during directory walk, file open/read, and per-line scan
- add regression tests covering mid-search abort for both tools

No new feature surface, no new tool options, and no behavior change other than honoring `Esc` promptly.

## Testing

- `npm exec biome check src/tools/fs/search.ts src/tools/filesystem.ts tests/filesystem-tools.test.ts`
- `npm exec vitest run tests/filesystem-tools.test.ts tests/loop.test.ts`

## Notes

- The branch was pushed with `--no-verify` because the repo's pre-push hook runs full `verify`, and current upstream still has unrelated Ink/type drift in untouched files:
  - `src/cli/commands/chat.tsx`
  - `src/cli/ui/layout/CardStream.tsx`
  - `src/cli/ui/ticker.tsx`

### DIFF
--- a/src/tools/filesystem.ts
+++ b/src/tools/filesystem.ts
@@ -339,11 +339,11 @@ Prefer \`list_directory\` for a single-level view, \`search_files\` to find spec
       },
       required: ["pattern"],
     },
-    fn: async (args: { path?: string; pattern: string; include_deps?: boolean }) =>
+    fn: async (args: { path?: string; pattern: string; include_deps?: boolean }, toolCtx) =>
       searchFiles(
         { rootDir, maxListBytes, skipDirNames: SKIP_DIR_NAMES },
         safePath(args.path ?? "."),
-        args,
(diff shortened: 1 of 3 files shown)


esengine (approved):
Tested locally — all 2168 tests pass on this branch. Narrow change, clear regression tests, exactly the right scope for the bug. Thanks!

PR #441, opened 2026-05-08
i18n(loop/errors): localize DeepSeek error messages

esengine:
## Why

`formatLoopError`, `reasonPrefixFor`, and `errorLabelFor` were all hardcoded English. A Chinese user hitting a 503 / 401 / context-overflow saw raw English in the chat — same class of UX gap as #440, just on the i18n axis.

This includes the new 5xx outage notice from #440 (since this PR is stacked on top of it).

## What

New `errors.*` i18n namespace covering:
- context overflow (with V4 / legacy limit mention)
- 401 auth / 402 balance / 422 param / 400 bad request
- 5xx head + reachable / unreachable / two action variants
- reason prefix and label for `budget` / `aborted` / `context-guard` / `stuck`
- `(no message)` fallback for empty error bodies

zh-CN translations included for all 20 keys. The user-visible `{inner}` server message (e.g. `"Authentication Fails"`) is preserved verbatim — translating DS's own response would be wrong.

## Test plan

- [x] 26 existing EN cases still pass (vitest `setupFiles` pins runtime to EN)
- [x] 2 new cases flip to `zh-CN` and confirm runtime translation works (`"服务不可用"`, `"认证失败"`)
- [x] full suite 2299/2299
- [x] tsc + biome clean

## Stacking

Built on top of #440 — merge that first to avoid a clean `errors.ts` collision.

### DIFF
--- a/src/i18n/EN.ts
+++ b/src/i18n/EN.ts
@@ -294,6 +294,38 @@ export const EN: TranslationSchema = {
     interrupted: "▸ turn interrupted by user.",
     loopStopped: "▸ loop stopped (after {count} iter{s}).",
   },
+  errors: {
+    contextOverflow:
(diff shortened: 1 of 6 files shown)


esengine:
Reopened as #444 with base on main after #440 merged. Same commit.

PR #518, opened 2026-05-09
feat(wizard): add theme-picker step with live preview

### LINKED ISSUE #492: wizard: add theme picker step with live preview, fully localized
## Problem

The first-run wizard (`src/cli/ui/Wizard.tsx:52`) walks the user through `language → apiKey → preset → mcp → review`. Theme is not part of it. The only way to change theme today is `/theme <name>` (`src/cli/ui/slash/handlers/theme.ts`), which:

1. Takes effect on next launch, not immediately.
2. Has no preview — the user picks `tokyo-night` blind, hits enter, restarts, sees what they got, picks again.

We ship 7 themes (`src/cli/ui/theme/tokens.ts:1-8`: `default`, `dark`, `light`, `tokyo-night`, `github-dark`, `github-light`, `high-contrast`). Asking a new user to pick one without seeing the result is silly.

## What I want

A `theme` step in the wizard, slotted between `language` and `apiKey` (so the rest of the wizard already renders in the chosen theme — that's part of the preview):

```
language → theme → apiKey → preset → mcp → review → saved
```

Behavior:

- **Live preview while highlighting.** As the user moves the cursor up/down through the theme list, the wizard's own surface re-renders in that theme immediately — no commit needed. `setActiveTheme` (`tokens.ts:322`) already supports this; the picker just needs to call it on highlight change and call the returned `restore()` if the user backs out.
- **Inline sample tile.** Next to the picker show a small representative tile that exercises the tokens that actually differ between themes: a header glyph in `tone.ok`/`tone.err`/`tone.warn`, a fenced code block, a dim meta line, a bold accent. Not the whole UI — just enough that the user can see what the theme decides about contrast / accent / surface.
- **Default highlight.** Cursor lands on the currently-resolved theme (`resolveThemePreference`) so users re-running `/setup` don't accidentally reset it.
- **Commit on enter, persist via `saveTheme`.** Same path as `/theme` — no new persistence layer.

## i18n

Every label in the new step goes through `t()`. Add a `wizard.themeTitle` / `wizard.themeSubtitle` / `wizard.themeSampleHeading` / `wizard.themeFooter` block to `src/i18n/types.ts`, `src/i18n/EN.ts`, `src/i18n/zh-CN.ts`. The theme **names** themselves stay as-is — `tokyo-night` is a proper noun, not a string to translate — but the descriptive captions next to each name (\"high-contrast — accessibility-first, max contrast\") MUST be localized.

The review step already has `reviewLabelLanguage`, `reviewLabelPreset`, etc. (`src/i18n/types.ts:265+`). Add `reviewLabelTheme` and surface the chosen theme name there.

## Out of scope

- Adding new themes. Picker works with whatever `listThemeNames()` returns.
- Auto-detecting terminal background to pick a default. Separate problem.
- Live theme switching mid-session from `/theme`. The slash handler already says \"active on next launch\" — fine for now.

BaBaLaDy:
## What
The first-run wizard walked users through language → apiKey → preset → mcp → review with no theme step. The only way to pick a theme was /theme <name> after setup — blind, no preview, takes effect on next launch.

## Why
Fixes #492 . The wizard is the right place to pick a theme once, with immediate visual feedback. Slotting it between language and apiKey means the rest of the wizard already renders in the chosen theme — that's part of the preview.

npm run verify is passed


esengine:
Merged — thanks @BaBaLaDy! Hits every acceptance bullet from #492: live preview, default highlight on the resolved theme, localized captions, sample tile that exercises the right tokens. Nice first PR. Welcome aboard 🎉

PR #524, opened 2026-05-09
feat(slash): enhance slash command grouping and ordering functionality

kabaka9527:
## What

This change fixes slash command suggestion ordering and selection behavior by separating data-level suggestion ranking from UI-level grouped rendering. Empty slash command browsing is now ordered by fixed internal functional domains, while non-empty filtered suggestions can still use frequency ranking. The slash suggestion UI groups commands under stable domain headers and keeps keyboard selection aligned with the rendered command order.

## Why

The slash command menu could previously show repeated category headers such as SETUP/INFO/SETUP instead of grouping commands together. A follow-up fix introduced conflicts between fixed domain ordering and frequency-based suggestion sorting, causing test failures. The UI also mixed original match indexes with regrouped render order, which made keyboard navigation jump from the first command to a later command instead of moving one item at a time.

## How to verify

Run:

npm test -- tests/slash.test.ts tests/slash-usage.test.ts tests/ui-slash-suggestions.test.tsx
npm run typecheck
npx biome check src/cli/ui/slash/commands.ts src/cli/ui/SlashSuggestions.tsx src/cli/ui/useCompletionPickers.ts src/cli/ui/App.tsx tests/slash-usage.test.ts tests/ui-slash-suggestions.test.tsx

Expected results:

Slash command suggestions are grouped by fixed internal domains for empty browsing.
Category headers appear once per group.
/language, /preset, and /model appear together under SETUP.
Frequency sorting still works for non-empty filtered suggestions.
Keyboard selection moves sequentially through commands without jumping over grouped items.

## Checklist

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -214,7 +214,12 @@ src/
 Files kept small by design: the largest module under `cli/ui/` is 2K
 lines (App.tsx), every handler under `slash/handlers/` is ≤200 lines,
 every hook under `cli/ui/` is ≤310 lines. Adding a new slash command
-means editing one handler file and one registry line.
+means editing one handler file and one registry line. Declare its internal
+functional domain in the registry's `group` field; help and slash browsing
+are grouped by the centralized `SLASH_GROUP_ORDER` (`SETUP`, `INFO`,
+`CHAT`, `EXTEND`, `SESSION`, then code/jobs/advanced, with `UNKNOWN` last for
+unrecognized groups), so each category header appears once and commands keep
+their registry order within the category.
 
 ## Design evolution
 
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -145,7 +145,13 @@ import { replaceMcpServerSummary } from "./mcp-server-list.js";
 import { formatLongPaste } from "./paste-collapse.js";
 import { extractOpenQuestionsSection } from "./plan-open-questions.js";
 import { PRESETS, resolvePreset } from "./presets.js";
-import { type McpServerSummary, handleSlash, parseSlash, suggestSlashCommands } from "./slash.js";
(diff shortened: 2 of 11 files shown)


esengine:
Hi @kabaka9527 — thanks for both PRs. I just merged **#516** to `main` (`959e9ff`). #524 overlaps with it heavily, so it's now stale and will need a rebase + scope trim before it can land.

What's already in `main` from #516, and should be **dropped** from #524 on rebase:

- The entire `SlashSuggestions.tsx` windowing rework (`computeWindowStart`, `buildVisibleItems`, `GroupHeader`, `padOrTrim`, `truncateCells`) — already landed.
- Moving `/language` from `advanced` to `setup` — already landed.
- Most of `tests/ui-slash-suggestions.test.tsx` — the windowing / header-budget / pairing-summary cases are already in `main`.

What's **unique to #524** and worth keeping:

- Centralizing `SLASH_GROUP_ORDER` + `SLASH_GROUP_LABEL` in `slash/commands.ts`.
- The `orderSlashCommandsByGroup()` helper + the `/help` handler refactor that uses it.
- The deliberate group reorder (`setup → info → chat → extend → session → ...`).

Two asks before this can merge:

1. **Drop the `unknown` SlashGroup fallback.** `getSlashCommandGroup()` and the `unknown` variant defend against a state the type system already prevents — the test even uses `as never` to construct an invalid input, which is exactly the case our [code rules](https://github.com/esengine/DeepSeek-Reasonix/blob/main/CLAUDE.md#errors--fallbacks) say not to validate. Make the helper just return `spec.group`.
2. **Squash to a single commit** with a `refactor(slash):` (not `feat:`) prefix when you push the rebase — this is a refactor of the grouping plumbing, not a new feature.

After rebase the diff should shrink from ~580 lines to ~150. Ping me when it's ready and I'll merge.

PR #525, opened 2026-05-09
fix(shell): normalize Windows PATH env before spawn

### LINKED ISSUE #520: Bug: `run_command` does not inherit system-level (Machine) PATH on Windows
**Environment:** Windows 10/11, Go installed at `C:\Program Files\Go\bin` (registered in Machine PATH)

**Steps to reproduce:**

1. Open a Reasonix Code session in any Go project directory on Windows.
2. Run `run_command({ command: "echo %PATH%" })` — observe that `C:\Program Files\Go\bin` is **missing** from the output.
3. Run the same `echo %PATH%` in a regular `cmd.exe` — the path **is present**.
4. Run `run_command({ command: "go version" })` — fails with `'go' is not recognized as an internal or external command`.

**Verification (PowerShell):**
```powershell
[Environment]::GetEnvironmentVariable('Path', 'Machine') | Select-String go
# → C:\Program Files\Go\bin

Root cause: On Windows, the effective PATH for a new process is Machine PATH + User PATH. run_command appears to only load the User PATH (or a partial snapshot), omitting the Machine-level PATH entries. Wrapping the command with cmd /c "set PATH=%PATH%;C:\Program Files\Go\bin && <command>" works around the issue, confirming the tool's environment inheritance is incomplete.

Expected behavior: run_command should inherit the full combined PATH (Machine + User) so that binaries registered at the system level are discoverable without manual workarounds. ```

paradoxSCH:
## What
Normalize Windows PATH environment variables before spawning child processes to ensure consistent path resolution across different shells.

## Why
Fixes #520. Windows environment variables are case-insensitive, but Node.js environment handling can sometimes lead to duplicate or inconsistent PATH entries when merging environments, causing command resolution failures.

## How to verify
Ran focused tests for shell command execution:

pm test tests/shell-utils.test.ts - PASSED

Note: 
pm run verify failed locally due to unrelated failures in 	ests/server-dashboard.test.ts (fetch failed / bad port) which appear specific to this Windows Node 25 environment and are not impacted by these changes.

## Checklist
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/src/tools/shell/exec.ts
+++ b/src/tools/shell/exec.ts
@@ -65,6 +65,7 @@ export async function runCommand(
     });
   }
   const timeoutMs = timeoutSec * 1000;
+  const normalizedEnv = normalizeWindowsEnvVars(process.env);
 
   const spawnOpts: SpawnOptions = {
     cwd: opts.cwd,
@@ -78,7 +79,7 @@ export async function runCommand(
     // sees a Python traceback instead of the script's real output
     // and goes around in circles trying to fix the wrong problem.
     // Harmless on non-Python processes (env vars they don't read).
-    env: { ...process.env, PYTHONIOENCODING: "utf-8", PYTHONUTF8: "1" },
+    env: { ...normalizedEnv, PYTHONIOENCODING: "utf-8", PYTHONUTF8: "1" },
   };
 
   // Windows: two layered fixes on top of shell:false —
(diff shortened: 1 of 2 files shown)


esengine:
Merged — thanks @paradoxSCH! Clean defensive fix; the case-insensitive read in resolveExecutable is exactly the right complement to the env normalization. First PR on the project — welcome aboard 🎉

I've reopened #520 to wait on the original reporter to confirm whether Machine PATH inheritance is actually restored on their setup (the case-collapse path may not be the root cause they hit). Either way, the codepath is more correct now.

PR #526, opened 2026-05-09
feat(i18n): localize card component labels with zh-CN

Evan-Pycraft:
Cards 的标题和行标签之前都是硬编码英文，走不了 i18n。
这个 PR 把它们都改成 t() 调用，顺便加上了中文翻译。

改了什么：
- i18n 类型定义里加了 cardTitles / cardLabels 两组 key
- EN / zh-CN 翻译文件
- 8 个 card 组件里把 "usage" / "prompt" / "cache" 之类的改成 t() 调用

全是机械替换，没有逻辑改动。
没用到的 key（statusBar / editMode 那些）是后续 PR 要用的，先加了但没副作用。

build 过了，doctor 正常跑。


esengine (changes requested):
Hey @Evan-Pycraft — thanks for picking up the card-label localization, this is real work that needed doing. But this PR isn't mergeable as-is; there are a few substantive problems, not just nits. Calling them out so you can revise:

## 1. `t()` doesn't evaluate JS expressions inside `{...}` placeholders

The `t()` implementation in `src/i18n/index.ts` only does flat `{name}` substitution via a regex. It does **not** evaluate JS. So strings like:

```ts
hits: "{count} hit{count === 1 ? \"\" : \"s\"} · {files} file{files === 1 ? \"\" : \"s\"}"
moreHits: "⋮ +{count} more hit{count === 1 ? \"\" : \"s\"}"
earlierLines: "⋮ {count} earlier line{count === 1 ? \"\" : \"s\"} ..."
earlierStackLines: "⋮ {count} earlier stack line{count === 1 ? \"\" : \"s\"} hidden"
addedLines / removedLines / runningJobs / queuedApplyDiscard / linesAbove / linesBelow / turns / messages / ...
```

…will render the literal string `count === 1 ? "" : "s"` to the user. Not a typo — actual broken UI output. Same issue is repeated in 20+ keys.

There's a deeper problem too: **Chinese doesn't pluralize with `s`**, so even if `t()` did evaluate the ternary, the zh-CN translations would still be wrong if they tried to mirror the same shape. The right fix is to split into singular / plural keys and let the caller pick — and for zh-CN, the two keys can map to the same form.

```ts
// EN
cardLabels: {
  hitSingular: "{count} hit",
  hitsPlural: "{count} hits",
}
// zh-CN
cardLabels: {
  hitSingular: "{count} 个命中",
  hitsPlural: "{count} 个命中",
}
// caller
const stats = t(card.hits.length === 1 ? "cardLabels.hitSingular" : "cardLabels.hitsPlural", { count: card.hits.length });
```

## 2. `slash.loop` got moved out of the `slash` namespace and renamed

In `src/i18n/EN.ts` (and zh-CN.ts):
```diff
-    loop: {
+  loopCmd: {
       description: "auto-resubmit ..."
```

Note the indentation change (4 spaces → 2 spaces). That moves `loop` *out* of `slash:` and renames it to a top-level `loopCmd`. The `/help` and slash suggestion code reads `t(\`slash.${spec.cmd}.description\`)` dynamically (`src/cli/ui/slash/handlers/basic.ts:41`, `src/cli/ui/SlashSuggestions.tsx:98`), so this silently regresses the localized `/loop` description in /help to the English fallback.

Revert this rename — leave `slash.loop` exactly as it was.

## 3. ~250 unused i18n keys

The PR adds entire namespaces (`statusBar`, `composer`, `editMode`, `shellConfirm`, `editConfirm`, `choiceConfirm`, `mcp`, `loop` (top-level), `ctxBreakdown`, `sessionPicker`, …) that **none of the 8 modified card files reference**. You called it out yourself: "没用到的 key 是后续 PR 要用的，先加了但没副作用."

We don't take pre-positioned keys for hypothetical future PRs in this repo — see [`CLAUDE.md`](https://github.com/esengine/DeepSeek-Reasonix/blob/main/CLAUDE.md) *"Don't design for hypothetical future requirements"*. They turn into orphans, the translations rot before the consuming code arrives, and `types.ts` grows ~270 lines for code that doesn't exist. Each future PR should add its own keys when (and only when) it adds the consumer.

Drop everything except what the 8 modified card files actually call. That should shrink the i18n diff from `+275/-1` per file to roughly `+50/-0` per file.

## 4. CI hasn't reported

`statusCheckRollup` is empty — looks like the CI workflow didn't run on this branch (probably how the fork is configured). Please confirm `npm run verify` passes locally (lint + typecheck + tests + comment-policy gate). The comment-policy gate is strict; the typecheck will likely flag mis-shaped i18n entries once you fix the structure in #2.

---

After revision, the diff should look like:

- 8 card components: mechanical `t()` swaps as you have now ✓
- `cardTitles` namespace: keep, clean up (no broken plurals)
- `cardLabels` namespace: keep, ONLY keys actually consumed by the 8 cards, with proper singular/plural splits
- Everything else: drop until a PR actually consumes it
- `slash.loop`: untouched

Push the revision when ready and I'll re-review. Thanks again — the goal here (TUI cards in Chinese) is solid, just needs a cleaner cut.

Evan-Pycraft:
Revised based on review. Changes made:

1. **Pluralization**: split `hits` / `moreHits` / `earlierLines` / `earlierStackLines` into singular/plural key pairs. No more JS expressions in translation values.

2. **Removed unused keys**: dropped all namespaces not consumed by the 8 card files (`statusBar`, `composer`, `editMode`, `shellConfirm`, `editConfirm`, `choiceConfirm`, `mcp`, `ctxBreakdown`, `sessionPicker`, `modelPicker`, `slashSuggestions`, `atMentions`, `appStrings`). Only `cardTitles` and `cardLabels` remain, with only the keys that the cards actually call.

3. **`slash.loop`**: untouched, no rename or indent change.

4. **Variable shadowing**: fixed `.map((t) =>` conflicting with imported `t()`.

5. **`editMode` references**: moved `runningLabel` / `workingLabel` into `cardLabels`, no card files reference `editMode.*`.

`npm run build && npm run typecheck && npm run lint` all pass. i18n diff is now ~+50 lines per file.

esengine:
Merged — thanks @Evan-Pycraft! You took the feedback exactly right: pluralization moved to the call site with singular/plural key pairs, dead namespaces dropped, scope tight. zh-CN translations are well chosen too. Welcome aboard 🎉

PR #589, opened 2026-05-10
fix(dashboard): list flat-format skills

### LINKED ISSUE #586: Dashboard /api/skills doesn't list flat-format skills (<dir>/<name>.md), only folder-format
## Symptom

Reported by a user on v0.36: the dashboard's **Skills** tab doesn't show their installed skill, even though `/skill <name>` runs it correctly in the TUI and the file is on disk.

## Root cause

The dashboard's `listSkills()` (`src/server/api/skills.ts:66-110`) only supports the **folder-format** layout. The TUI loader supports two equivalent layouts (`src/skills.ts:179-186`, `196-209`):

- `<dir>/<name>/SKILL.md` — folder, allows colocating assets
- `<dir>/<name>.md` — flat, single file

The dashboard hard-codes the folder shape:

```ts
for (const entry of readdirSync(dir)) {
  if (!SAFE_NAME.test(entry)) continue;
  const skillPath = join(dir, entry, SKILL_FILE);  // assumes entry is a folder
  // openSync below throws for flat .md files; swallowed by the outer catch
}
```

So a user with `.reasonix/skills/foo.md` (flat) sees:

- TUI `/skill foo` works ✅
- Dashboard Skills tab is empty ❌

The failure is silent — `openSync(".reasonix/skills/foo.md/SKILL.md")` throws ENOTDIR, the `catch {}` on line 102 eats it, the entry is skipped without warning.

## Secondary problem

The same loop doesn't distinguish files from directories before joining `SKILL_FILE`. Any unrelated file in `.reasonix/skills/` (a stray `.DS_Store`, a `README.md`) is also silently fed through the same broken path. That's not user-visible, but it's wasted I/O on every dashboard refresh.

## Why this is separate from #583

#583 is about the **prompt builder** filtering skills missing `description` frontmatter — a skill IS on disk and IS listed by the dashboard, but the model never sees it.

This issue is the inverse: the skill IS on disk and the model CAN run it, but the dashboard tab can't see it. Different code path (`/api/skills` GET vs `applySkillsIndex`), different bug.

## Fix sketch

Mirror the loader's logic in `listSkills()`:

1. Stat each entry; branch on file vs directory.
2. For directories: existing path (look for `<entry>/SKILL.md`).
3. For files ending in `.md` with valid stem: read directly, derive name from stem.
4. Skip everything else without throwing.

Also: when an entry is a folder containing `SKILL.md` AND its parent has a sibling `<name>.md`, the loader resolves the folder version first (`src/skills.ts:179-185`) — keep that precedence in the dashboard too so listings match what `/skill` will actually run.

## Repro

```bash
mkdir -p .reasonix/skills
cat > .reasonix/skills/flat-format.md <<'MD'
---
name: flat-format
description: shows up in /skill but not in dashboard
---
Body.
MD
```

Open the dashboard → Skills tab. `flat-format` is missing from the project list. Run `/skill flat-format` in the TUI — it works.

paradoxSCH:
## What

Teach the dashboard skills API to recognize flat-format skill files (<name>.md) in addition to folder-format skills (<name>/SKILL.md). The same resolver is used for listing, reading, saving, and deleting skills so a flat skill shown in the dashboard can also be edited or removed without creating a duplicate folder-layout skill.

## Why

Fixes #586. The TUI skill loader already supports both layouts, but /api/skills only looked for <dir>/<name>/SKILL.md, so flat installed skills worked from /skill <name> while the dashboard silently omitted them.

## How to verify

- 
pm run verify
- 
pm run test -- tests/server-dashboard.test.ts
- 
px biome check src/server/api/skills.ts tests/server-dashboard.test.ts
- 
pm run typecheck

## Checklist

- [x] 
pm run verify passes locally (lint + typecheck + tests + comment-policy gate)

### DIFF
--- a/src/server/api/skills.ts
+++ b/src/server/api/skills.ts
@@ -10,6 +10,7 @@ import {
   readSync,
   readdirSync,
   rmSync,
+  statSync,
   writeFileSync,
 } from "node:fs";
 import { homedir } from "node:os";
@@ -52,6 +53,13 @@ interface SkillListEntry {
   mtime: number;
 }
 
+type SkillLayout = "folder" | "flat";
+
+interface ResolvedSkillPath {
+  path: string;
+  layout: SkillLayout;
+}
+
 function parseFrontmatterDescription(raw: string): string | undefined {
   const lines = raw.split(/\r?\n/);
   if (lines[0] !== "---") return undefined;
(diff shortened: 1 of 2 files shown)


esengine (approved):
Implementation matches the TUI loader (folder-precedence, open-once read path, Dirent dispatch). Save default stays folder-format which is a deliberate non-change — happy to keep that out of scope. Tests cover the full CRUD path plus the unrelated-file skip from the issue's secondary concern. Thanks!

Issue #564, opened 2026-05-10
Composer hint bar: surface `^U` clear (and other readline chords) + i18n the three hardcoded strings


esengine:
## Two problems on the composer hint bar

`src/cli/ui/layout/Composer.tsx:8-11` ships three hardcoded English
strings:

```ts
const PLACEHOLDER = "type a message · / for commands · @ to attach a file";
const HINT = "⏎ send  ·  shift/alt+⏎ newline  ·  ↑↓ scroll  ·  ^P/^N history  ·  esc abort  ·  ctrl-c quit";
const ABORTED_HINT = "turn aborted by user · esc again to clear · ⏎ to ask a follow-up";
```

### 1. Hint bar misses common shortcuts users actually use

`src/cli/ui/multiline-keys.ts` implements a full readline-style
keymap, but the hint only advertises 6 of the bindings. The
unadvertised ones include the most-asked-for ergonomic chords —
especially `Ctrl+U` (clear buffer), which is bash-typist muscle
memory and impossible to discover without reading source:

| Chord | Behavior | Source |
|---|---|---|
| `Ctrl+U` | clear whole buffer | `multiline-keys.ts:114` |
| `Ctrl+K` | kill to end of line | `multiline-keys.ts:117` |
| `Ctrl+W` / `Alt+Backspace` | delete previous word | `multiline-keys.ts:126` |
| `Ctrl+A` / `Home` | start of line | `multiline-keys.ts:100` |
| `Ctrl+E` / `End` | end of line | `multiline-keys.ts:103` |
| `Alt+B` / `Alt+F` | move cursor by word | `multiline-keys.ts:138/142` |
| `PageUp` / `PageDown` | jump to start/end of whole buffer | `multiline-keys.ts:62` |
| `Tab` | slash-complete | parent-owned |

We can't fit all of these in one row. Suggested compromise: surface
the highest-value chord (`^U` clear) on the always-visible HINT row,
add a `?` help affordance (or `/help` line) that opens a full
keymap reference. Trim other low-value items if width-constrained.

Proposed always-visible HINT (terse):

```
⏎ send  ·  shift+⏎ newline  ·  ^U clear  ·  ^P/^N history  ·  esc abort  ·  ^C quit
```

(`shift/alt+⏎` collapses to `shift+⏎` to free the cells; alt+⏎
remains documented in /help.)

A separate tracking task: `/help` or a dedicated keymap card that
[...]

PR #599, opened 2026-05-10
feat(i18n): localize confirmation modals, HintRow, and remaining cards with zh-CN

Evan-Pycraft:
确认弹窗（ShellConfirm / EditConfirm / ChoiceConfirm）、HintRow 提示行、以及剩余的 7 个卡片组件（ApprovalCard / UserCard / DiffCard / DoctorCard / TaskCard / PlanCard / MemoryCard）之前还是硬编码英文，这个 PR 把它们全部改成 t() 调用，加了中文翻译。

改了 14 个文件：

确认弹窗：
- ShellConfirm: 标题/副标题/选项/拒绝模式等 14 处
- EditConfirm: 底部提示/NEW-EDIT标签/差异标注/滚动提示等 12 处（含单复数）
- ChoiceConfirm: 自定义回答/取消选项等 5 处

HintRow（Issue #564）：
- PromptInput: ↑↓ scroll → ^U clear，对齐 Issue #564 的要求

剩余卡片：
- ApprovalCard: 默认 footer
- UserCard: 标题 "you" → "你"
- DiffCard: [a] apply / [s] skip / [r] reject
- DoctorCard: OK/warn/FAIL 标签 + 汇总
- TaskCard: step 标签
- PlanCard: done / in progress / upcoming
- MemoryCard: 分类标签 + token + more

i18n 字典：
- types.ts: 新增 shellConfirm / editConfirm / choiceConfirm 命名空间以及 cardLabels 补充 key
- EN.ts / zh-CN.ts: 对应翻译

build/typecheck/lint 全过。


esengine (changes requested):
Nice sweep — and thanks for picking up the `^U clear` HintRow change from #564 along the way. One thing I'd want fixed before merging:

**`DEFAULT_FOOTER` is captured at module load** (`src/cli/ui/cards/ApprovalCard.tsx:36`):

```ts
const DEFAULT_FOOTER = t("cardLabels.defaultFooter");
```

`t()` reads `currentLang` at call time (see `src/i18n/index.ts:80`), so this evaluates exactly once when `ApprovalCard.tsx` first loads and freezes whatever language was active then into the constant. Every other `t()` call in this PR is inside a render body, so they re-run when `useLanguageReload` triggers a re-render — but this one doesn't. After `/language zh-CN` mid-session, any card relying on the default footer would still show the original-language text.

Easiest fix is to inline the lookup into the prop default:

```ts
export function ApprovalCard({
  ...
  footerHint,
  ...
}: ApprovalCardProps): React.ReactElement {
  const effectiveFooter = footerHint ?? t("cardLabels.defaultFooter");
  // ...use effectiveFooter where DEFAULT_FOOTER is referenced...
}
```

Then drop the module-level `DEFAULT_FOOTER` constant entirely.

Two non-blocking observations for a follow-up if you feel like it:

1. `t("shellConfirm.awaiting")` / `t("shellConfirm.denyTitle")` / `t("shellConfirm.optional")` get reused from `ChoiceConfirm` and `EditConfirm`. They're really generic confirm-modal copy — moving them into a shared `confirmCommon` namespace would read cleaner. Don't bother in this PR.
2. `TaskCard.tsx:496` `(step.toolName ?? t("cardLabels.stepLabel")).padEnd(7)` — `padEnd` counts code points, but terminals render by display cell, so `步骤` (4 cells) padded to 7 chars produces 9 cells of output and breaks alignment in zh-CN. This is a pre-existing issue (the original `"step"` fallback also went through `padEnd(7)`); your change just makes it visible to translators. Worth a separate small PR using `string-width`-aware padding.

Once the `DEFAULT_FOOTER` fix lands I'll squash-merge.

Evan-Pycraft:
Consolidating into #603 per maintainer request. All changes (modals, cards, pickers, StatsPanel, WelcomeBanner, ctx-breakdown) are now on the fix/slash-cwd-i18n branch rebased onto current main, with the DEFAULT_FOOTER reactivity bug fixed.

Evan-Pycraft:
Superseded by #603

Issue #101, opened 2026-05-02
Make `reasonix mcp inspect` errors actionable


esengine:
## Background

`reasonix mcp inspect <spec>` connects to one MCP server and prints its capabilities + tools / resources / prompts. When the connection fails, the user sees a bare error string with no recovery hint:

```sh
$ reasonix mcp inspect npx-typo=npx -y @nonexistent/server
mcp inspect failed: spawn npx-typo ENOENT
```

`spawn ENOENT` reads as gibberish to anyone who hasn't debugged Node child processes before. The fix is the same shape as #16 (web tool errors): every error string should end with a short " — try: …" tail telling the user what action would unstick them.

This is the inspect-side parallel to #16; the patterns line up so the two PRs can borrow phrasing.

## Current behaviour

- CLI registration: `src/cli/index.ts:404`
- The `try/catch` at `src/cli/index.ts:411-416` re-emits the raw `err.message`:

```ts
} catch (err) {
  process.stderr.write(`mcp inspect failed: ${(err as Error).message}\n`);
  process.exit(1);
}
```

- Failure paths that flow through here:
  - `parseMcpSpec` throws on empty / malformed input (`src/mcp/spec.ts`)
  - `StdioTransport` throws `spawn <cmd> ENOENT` when the command doesn't exist
  - `SseTransport` / `StreamableHttpTransport` throw on connection-refused, DNS, TLS
  - `client.initialize()` throws on handshake timeout or protocol-version mismatch
  - `inspectMcpServer` throws on internal protocol errors

## Expected behaviour

Wrap the catch in a small classifier so each common cause gets a tailored hint. Examples:

- `spawn <cmd> ENOENT` → `command "<cmd>" not found — try: install the package first (e.g. \`npx --version\`), or check the spec's command spelling`
- `ECONNREFUSED` → `could not connect to <url> — try: confirm the server is running and the port matches`
- handshake timeout → `MCP handshake timed out — try: confirm the target speaks MCP (some servers only emit JSON-RPC after stdin is closed); pass --json for raw output if you want to debug further`
- empty / malformed spec from `parseMcpSpec` → leave the existing message but suffix with `— try: \`name=command args\` or an http(s):// URL`
- everything else → unchanged tail (`mcp inspect failed: <message>`)

## Files to touch

- `src/cli/index.ts:411-416` — branch the catch on error code / message shape
[...]

Issue #16, opened 2026-05-01
Make `web_search` / `web_fetch` errors actionable


esengine:
## Background
Tool errors in Reasonix go back to the model verbatim — they're how the agent decides what to do next. A bare `web_search 429` tells the model "something failed" but not "back off and retry" vs "the URL was wrong" vs "the engine is blocked"; the model often just retries identically and burns tokens. Better error strings → fewer wasted turns.

## Current behaviour
`src/tools/web.ts` has several places where the error is just the HTTP status with no recovery hint:

```ts
// src/tools/web.ts:58
if (!resp.ok) throw new Error(`web_search ${resp.status}`);

// src/tools/web.ts:130
if (!resp.ok) throw new Error(`web_fetch ${resp.status} for ${url}`);
```

The 429 / 403 anti-bot path (line 64) does include a hint, but most others don't.

## Expected behaviour
Every error includes a short " — try: …" or " — next: …" tail telling the model what action would unstick it.

Examples:
- `web_search 429 — try: wait 10s before retrying, or rephrase the query` (rate-limited)
- `web_fetch ${status} for ${url} — try: confirm the URL resolves in a browser; status suggests the host returned an error page` (HTTP error)
- `web_fetch: timed out after ${ms}ms — try: shorter URL or smaller content; this may be a slow CDN` (timeout)
- `web_fetch: url must start with http:// or https://` already exists at line 274 — leave it; it's the model that.

## Files to touch
- `src/tools/web.ts` (all `throw new Error(...)` and `JSON.stringify({ error: ... })` paths)

## Acceptance criteria
- [ ] Every error message produced by `web_search` ends with a hint clause (" — try: …" / " — next: …")
- [ ] At least three distinct hints: rate-limit (429), forbidden (403), generic 5xx
- [ ] Same treatment for `web_fetch`: rate-limit, HTTP non-2xx, timeout, non-http(s) URL
- [ ] Existing tests still pass (`npm run test -- web`)
- [ ] Add 2-3 new test cases in `tests/web.test.ts` covering the new error shapes (snapshot or `.toContain('try:')` is fine)

## Hints
- Don't over-engineer. The hints are short strings; resist the urge to introduce an `ErrorWithHint` class.
- Focus on what a *language model* would find useful. Phrases like "the URL is malformed" beat "ERR_INVALID_URL".
- Look at `src/tools/shell.ts` `formatCommandResult` for an existing pattern of hint-style messages.

## Out of scope
[...]

Issue #15, opened 2026-05-01
Add `--json` flag to `reasonix doctor`


esengine:
## Background
`reasonix doctor` runs a battery of health checks (API key, balance reachability, tokenizer, sessions dir, hooks, project markers, …) and prints them as plain text, exiting 1 if anything failed. Useful for humans, but right now there's no way to consume the output from a script — CI gates, tmux statusline integrations, etc. all want structured JSON.

## Current behaviour
- CLI registration: `src/cli/index.ts:289`
- Implementation: `src/cli/commands/doctor.ts:335` (`doctorCommand`)
- Output is `console.log` text. Exit code: `1` on any fail, `0` on warn or clean.

## Expected behaviour
A `--json` option that emits a single JSON document with the same checks the human format runs, then exits with the same code.

```
reasonix doctor --json
```

```jsonc
{
  "version": "0.18.1",
  "summary": { "ok": 7, "warn": 1, "fail": 0 },
  "checks": [
    { "id": "api-key",       "status": "ok",   "message": "DEEPSEEK_API_KEY set" },
    { "id": "balance",       "status": "ok",   "message": "¥48.20", "value": 48.20 },
    { "id": "python_runner", "status": "warn", "message": "subagent failed to load (ENOENT)", "fix": "install Python 3.10+" }
  ]
}
```

The shape isn't fixed — pick what makes sense — but each check should have at minimum `id`, `status` (`ok` / `warn` / `fail`), and `message`. Optional `fix` field is a nice-to-have.

## Files to touch
- `src/cli/index.ts:289` — add the option to the command definition
- `src/cli/commands/doctor.ts:335` — add a `json: boolean` to `DoctorOptions`, branch the renderer

## Acceptance criteria
- [ ] `reasonix doctor --json` outputs valid JSON (one document, no trailing log)
- [ ] Exit code semantics unchanged (`1` on fail, `0` on warn/clean)
- [ ] Plain-text output unaffected when `--json` is not passed
- [ ] `reasonix doctor --json | jq '.checks[] | select(.status=="fail")'` works
- [ ] No new dependencies

## Hints
- Look at `reasonix events --json` (`src/cli/commands/events.ts`) for a working `--json` flag pattern in this repo.
- The simplest design: refactor each check to produce a `{ id, status, message, fix? }` object, then either render text or `JSON.stringify` based on the flag.
[...]

PR #607, opened 2026-05-10
Fix setup flow to reject invalid DeepSeek API keys

paramount-t:
This fixes an issue where entering an invalid DeepSeek API key during initial setup would show an API error later, but the app could still continue running with the bad key saved.

Changes included:

Validate the DeepSeek API key during the setup wizard before continuing.
Keep the user on the API key step if DeepSeek rejects the key.
Allow npx reasonix setup to revisit the API key step and replace an existing key.
Add localized error/status messages for API key validation.
Add tests for successful validation, rejected keys, and validation failures.


esengine:
Merged — thanks @paramount-t, and welcome 🎉

Real footgun fixed cleanly: invalid keys silently saved is a bug I'd been meaning to circle back on. Splitting the failure into `rejected` (auth fail → user pastes a different key) vs `failed` (network / 5xx → user retries) with separate copy is exactly the right shape, and exporting `validateDeepSeekApiKey` as a pure function with an injectable `fetch` made the tests trivial — appreciated.

If you'd like to keep contributing, the open good-first-issues are #15 (`reasonix doctor --json`), #16 (web_search / web_fetch error actionability), and #101 (`mcp inspect` error actionability). Comment on whichever one you'd like to take and I'll assign it.

PR #621, opened 2026-05-10
feat(tools): Validate tool call params agains json schema

wviana:
## Summary

Add pre-dispatch validation of required tool parameters so the model doesn't waste a full tool round on calls like `read_file({})` that are guaranteed to fail with "path must be a non-empty string."

## Problem

DeepSeek occasionally emits tool calls with missing required parameters: the name is present but the arguments object is empty or missing keys that the schema marks `required`.  Today these calls dispatch all the way to the tool's `fn`, which throws a low-level error (e.g.  `"Error: path must be a non-empty string"`).  The error counts as a consumed tool iteration and the model gets a cryptic message back, often repeating the same mistake a few times before self-correcting.

The most visible symptom is `read_file` called with `args: {}` producing three consecutive failures in a single turn, each costing a tool round.

## What changed

`src/tools.ts` - added `missingRequiredParam()` and a single call site in `ToolRegistry.dispatch()`, right after arg parsing and flatten re-nesting but before plan-mode enforcement / interceptor / tool fn:

- If the tool's schema declares `required: ["path"]` and the parsed args have no `path` key (or it's `undefined`), the dispatch returns:

  `{error: "read_file: missing required parameter \"path\". Retry with all required parameters filled."}`

- Tools with no `required` list pass through unchanged.
- Empty-string values (`"path": ""`) are NOT rejected ‚Äî JSON Schema `required` only checks key presence, not value emptiness.  Those remain the tool fn's own responsibility.

### TDD

| Step | What |
|---|---|
| `tests/tools.test.ts` | 5 new tests ‚Äî missing key, empty string passes through, happy path, no-required-list, nested-object param |
| `src/tools.ts` | `missingRequiredParam(schema, args)` , first missing key or `null` |
| `tests/filesystem-tools.test.ts` | Updated `edit_file` empty-search test ‚Äî now caught by our validation before reaching the tool fn |

All 139 existing tool-registry + filesystem tests pass.

### DIFF
--- a/src/tools.ts
+++ b/src/tools.ts
@@ -184,6 +184,13 @@ export class ToolRegistry {
       args = nestArguments(args);
     }
 
+    const missing = tool.parameters ? missingRequiredParam(tool.parameters, args) : null;
+    if (missing) {
+      return JSON.stringify({
+        error: `${name}: missing required parameter "${missing}". Retry with all required parameters filled.`,
+      });
+    }
+
     // Plan-mode enforcement — runs AFTER arg parsing so a tool with a
     // runtime `readOnlyCheck` can inspect the actual args (e.g.
     // `run_command` is read-only iff the command matches its allowlist).
@@ -284,4 +291,14 @@ function hasDotKey(obj: Record<string, unknown>): boolean {
     if (k.includes(".")) return true;
   }
   return false;
+}
+
+/** If the schema declares required params, return the first one that's missing. */
+function missingRequiredParam(schema: JSONSchema, args: Record<string, unknown>): string | null {
+  const required = schema.required;
+  if (!required || required.length === 0) return null;
+  for (const key of required) {
+    if (args[key] === undefined) return key;
+  }
+  return null;
 }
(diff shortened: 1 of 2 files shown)


wviana:
After getting this MR open I wonder if we shouldn't have a proper jsonschema validation,

The model suggested not to

> Should we use ajv instead? 
>
>  ajv  would give us proper validation: type checks,  minLength ,  pattern , nested object validation, etc. But:
>
>  · Dependency cost —  ajv  is ~30KB gzipped. That's non-trivial for a project that values keeping things small.
>  · Runtime cost — validating every tool call against the full JSON Schema on every dispatch is expensive for something that could be a one-line check.
>  · Overkill — the only bug we're fixing is "model emits a tool call with a missing key." A full JSON Schema validator would also catch  path: 42  (wrong type),  path: "x".repeat(1e6)  (too long), etc. — problems we don't have.

wviana:
I did this MR because I've noticed it falling to call read_file many times and started to chat about it with reasonix what we could do to improve.

esengine:
Nice — real problem (the `read_file({})` waste-loop is something I've watched happen), and the fix lands in exactly the right spot. JSON-Schema-presence-only semantics (not rejecting empty strings) is the right call too. Error shape matches the rest of `dispatch`'s error returns. Merging.

Issue #642, opened 2026-05-11
Modal confirm prompts invisible when chat is scrolled up (shell / plan / checkpoint / revision / choice)


esengine:
## Problem

If the user has scrolled up in the chat history and then a modal-class confirm prompt arrives — shell command (`run_command` / `run_background`), plan proposed, plan checkpoint, plan revision, or `ask_choice` — the modal mounts but is **not visible**: it renders at the bottom of the chat area, which is currently off-screen below the user's scroll position. From the user's point of view they're "stuck" — they can't see the options, and the picker has captured arrow keys so scroll-down inputs no longer scroll the viewport.

## Root cause

The pauseGate listener in `src/cli/ui/App.tsx:3038-3114` calls `setPendingShell` / `setPendingPlan` / `setPendingCheckpoint` / `setPendingRevision` / `setPendingChoice` directly. None of those mount paths reset the chat scroll position. The only `chatScroll.jumpToBottom()` call in the whole file is on the End key (`App.tsx:1218`).

`chatScroll.jumpToBottom()` sets `pinned: true`; the next `setMaxScroll` (triggered by the modal's `useReserveRows` shrinking the chat area) snaps `scrollRows` to the new bottom. So the existing primitives are sufficient — we just have to call them.

## Proposed fix

Add `chatScroll.jumpToBottom()` once at the top of the pauseGate listener callback, before the `switch (request.kind)`. Covers all five modal kinds with one line. `chatScroll` from `useChatScrollActions()` is store-backed and stable across renders, so the empty-deps useEffect stays correct.

## Test

Add a test that mounts the App-level pauseGate listener wiring, scrolls up via the store, fires a `pauseGate.request("run_command", …)`, and asserts `pinned === true` and `scrollRows === maxScroll` afterward.

## Scope

- `src/cli/ui/App.tsx` — one line, plus a tweak to the existing biome-ignore comment so it reads correctly with the new ref.
- `tests/` — one new file covering the modal-mount → scroll-reset invariant for all five modal kinds.

Good first issue territory.

PR #666, opened 2026-05-11
fix(scroll): coalesce pinned-mode shrinks to stop double-Esc flicker

### LINKED ISSUE #653: Flicker + scroll-position bounce when double-Esc aborts a multi-subagent run past one screen
Flicker + scroll-position bounce when interrupting a long subagent run.

## Repro shape

1. Trigger a workflow that spawns **multiple subagents** — `/security-review`, `/review`, an `init` skill running parallel scans, anything that pushes a lot of streaming output through `run_subagent` while parent stays alive.
2. Let the output **fill more than one viewport** — so the chat is scrolled (pinned to bottom but `maxScroll > 0`, or the user scrolled up to inspect a finding).
3. Hit **Esc** to interrupt. Hit Esc **again** quickly (double-tap).

Observed: the screen flashes — sometimes the visible region jumps up, sometimes it bounces between two near-identical frames. The reporter describes it as "闪烁" with "滚动位置也会闪" (scroll position also flickers).

## What's probably happening

Hypotheses, in order of likelihood — none verified yet:

1. **Abort path tears down N subagent inflights in quick succession** — each unmount fires a `useReserveRows` collapse + state.cards mutation; with N≥2 subagent cards collapsing at once, `CardStream` re-measures `maxScroll` several frames in a row, the `pinned=true` invariant snaps `scrollRows` each time, and the deltas oscillate visibly. Single-subagent runs hide this because there's only one frame.
2. **Double-Esc isn't separately handled** (`App.tsx:1250-1263` only sees one `key.escape && busy`) but `abortedThisTurn.current` guards against double-execute. Still, the second Esc may hit the *next* handler — composer Esc — and clear input or trigger another render.
3. **Even with #640 + #643**, the rapid stream of `setMaxScroll` calls during teardown can interleave with the alt-screen `clearTerminal` we now do per frame in a way that paints two distinct end-states before settling.

## Asking for ideal repro

- Screen recording would be ideal (the flicker is too fast for a still).
- Or: `reasonix --version`, a description of which skill / how many subagents, terminal name + version.

## Related

- #639 / #640 — alt-screen log-update drift (fixed for the steady-state case)
- #642 / #643 — modal scroll reset on pickers (also fixed)
- #563 — `useTick` audit (relevant if hypothesis #1 holds and the cascade is render-driven)

Filed proactively from a chat report; will pin down the exact repro before any rendering-layer changes.

dimasd-angga:
## Summary
Double-Esc aborting a multi-subagent run past one screen would flicker and bounce the scroll position: each card teardown re-measured and called `setMaxScroll` synchronously, snapping `scrollRows` N times in pinned mode. This change coalesces those shrinks behind the same `COALESCE_MS` trailing-edge timer the delta path already uses, so a burst of re-measurements collapses into one settled transition.

## Changes
- `src/cli/ui/state/chat-scroll-store.ts`: add `pendingMaxShrink` / `shrinkTimer` and a `flushShrink()` helper; in `setMaxScroll`, defer pinned-mode shrinks to a trailing `setTimeout(COALESCE_MS)` flush while keeping grows immediate; flush any pending shrink before applying a non-shrink update so its trailing state can't clobber a fresh value; clear the pending shrink in `jumpToBottom` so an explicit pin isn't undone by a stale target.
- `tests/modal-scroll-reset.test.ts`: add a `setMaxScroll coalescing (issue #653)` suite covering the burst-of-shrinks case (one trailing transition at the final target), grows applying immediately, the unpinned path still clamping `scrollRows` synchronously, a grow flushing a pending shrink, and `jumpToBottom` dropping a pending shrink.

## Testing
`npm run test -- tests/modal-scroll-reset.test.ts` — the new coalescing suite passes alongside the existing `jumpToBottom` (issue #642) tests. The burst test uses `vi.useFakeTimers()` and advances 32ms to assert exactly one subscriber notification at the final target.

## Notes
The fix piggybacks on the existing `COALESCE_MS` window rather than introducing a separate knob, to keep the timing model in the store consistent. Grows are intentionally left synchronous so streaming output stays pinned without added latency; only pinned-mode shrinks are deferred.

Closes #653

### DIFF
--- a/src/cli/ui/state/chat-scroll-store.ts
+++ b/src/cli/ui/state/chat-scroll-store.ts
@@ -49,6 +49,12 @@ export function createChatScrollStore(): ChatScrollStore {
   const listeners = new Set<ScrollListener>();
   let pendingDelta = 0;
   let flushTimer: NodeJS.Timeout | null = null;
+  // Trailing-edge coalesce target for pinned-mode shrinks (issue #653).
+  // While a burst of card-collapse re-measurements arrives, we hold the latest
+  // target here and apply it once on a microtask flush, so subscribers see one
+  // settled transition instead of N oscillating snaps.
+  let pendingMaxShrink: number | null = null;
+  let shrinkTimer: NodeJS.Timeout | null = null;
 
   function set(next: Partial<ChatScrollState>): void {
     const merged = { ...state, ...next };
@@ -91,6 +97,18 @@ export function createChatScrollStore(): ChatScrollStore {
     }
   }
 
+  function flushShrink(): void {
+    if (shrinkTimer !== null) {
+      clearTimeout(shrinkTimer);
+      shrinkTimer = null;
+    }
+    const target = pendingMaxShrink;
+    pendingMaxShrink = null;
+    if (target === null) return;
+    const nextScrollRows = state.pinned ? target : Math.min(state.scrollRows, target);
+    set({ maxScroll: target, scrollRows: nextScrollRows });
+  }
+
   return {
     getState() {
       return state;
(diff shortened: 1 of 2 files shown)


esengine:
Thanks — clean fix, and the test split is exactly what I'd want for this layer.

The cause-analysis matches hypothesis #1 in #653 (N subagent cards collapsing in the same render burst), and piggybacking on the existing `COALESCE_MS` instead of inventing a separate knob is the right call — one timing model in the store is much easier to reason about than two.

Worth calling out for anyone reading the diff later: keeping grows synchronous matters more than it might look. Pinned streaming during a long completion can't pay 16ms of latency per maxScroll bump, or the bottom-pin visibly lags the cursor. Glad to see that's preserved.

Five-case test coverage (burst, grow-immediate, unpinned-still-synchronous, grow-flushes-shrink, jumpToBottom-cancels) hits every invariant I'd have asked for. Merging.

PR #669, opened 2026-05-11
Mcp debug log malformed

### LINKED ISSUE #630: Optional debug logging for malformed MCP JSON-RPC lines (REASONIX_DEBUG_MCP)
## Problem

`StdioTransport.onStdout` in `src/mcp/stdio.ts:140-143` silently drops every line that doesn't parse as JSON-RPC:

```ts
try {
  const msg = JSON.parse(line) as JsonRpcMessage;
  this.push(msg);
} catch {
  // Malformed lines are dropped - some servers emit startup banners
  // before the JSON-RPC loop begins. We surface the noise to stderr
  // via the inherited stderr stream, not our event queue.
}
```

The comment is right: dropping is the correct default. Servers like `npx -y @modelcontextprotocol/server-filesystem` print startup banners to stdout before the JSON-RPC loop starts, and routing those into the event queue would crash the client.

The problem is debugging. When an MCP server emits a malformed JSON-RPC frame (truncated, encoding-broken, mid-chunk split), the user sees an empty tool result and has no way to find out the wire-level cause. Today the only path is editing `stdio.ts` locally to add a `console.error`.

## Proposed change

Gate the catch on a debug-mode env var. When `REASONIX_DEBUG_MCP=1`, write the rejected line to `process.stderr` with a clear prefix so it is visible but separable from real MCP traffic:

```ts
} catch {
  if (process.env.REASONIX_DEBUG_MCP) {
    process.stderr.write(`[mcp-stdio] dropped malformed line: ${JSON.stringify(line)}\n`);
  }
}
```

Off by default — zero behavior change for normal users. On for anyone investigating "why did my MCP tool return nothing".

Mention the env var in `src/mcp/README.md` under a short "Debugging" section.

## Scope

- `src/mcp/stdio.ts` — one-line debug log behind env var.
- `src/mcp/README.md` — mention `REASONIX_DEBUG_MCP` in a debugging note.
- `tests/mcp-stdio-close.test.ts` (or a new small test) — case where a malformed line is emitted: with env unset, nothing on stderr; with env set, one stderr line containing the malformed content.

Out of scope:
- A structured logger / event sink for MCP frames.
- Persisting dropped lines.
- Buffering partial JSON across chunks (the bigger NDJSON-fragment problem — separate issue).

## Good first issue

One file change + one test case + a docs paragraph. Touches an isolated, well-trodden path. Comment-policy rules in `CLAUDE.md` apply.

MyPrototypeWhat:
## What

Gate a stderr trace behind `REASONIX_DEBUG_MCP=1` so the MCP stdio transport surfaces the actual bytes when it drops a malformed JSON-RPC line. Default behavior is unchanged — nothing is logged unless the var is set to exactly `"1"`. Documented in `src/mcp/README.md`.

## Why

`StdioTransport.onStdout` silently swallows anything that fails `JSON.parse`, which is correct for the common case (servers emit non-JSON startup banners) but hides real wire-level bugs. When a server ships truncated or corrupted frames, users see empty tool results with no clue why. An opt-in env var keeps the hot path quiet for everyone else and gives a developer one switch to flip when diagnosing an empty `callTool` response. Closes #630.

## How to verify

```
npm run verify
npx vitest run tests/mcp-stdio-debug.test.ts
```

The new test file covers four cases: env=`"1"` (logs), unset (silent), `"0"` (silent), `"true"` (silent). Tests wait on child exit via the `messages()` iterator instead of a fixed sleep — Node guarantees
stdout `data` fires before `close`, so once the iterator returns the synchronous catch has already run.

## Checklist

### DIFF
--- a/src/mcp/README.md
+++ b/src/mcp/README.md
@@ -145,3 +145,7 @@ about it.
   process hasn't exited.
 - **Malformed lines**: dropped silently. Some servers emit non-JSON
   during startup; logging every dropped line would be noise.
+- **Debugging dropped lines**: set `REASONIX_DEBUG_MCP=1` to print each
+  dropped malformed line to stderr, prefixed with
+  `[mcp-stdio] dropped malformed line:`. Useful when an MCP server
+  ships truncated or corrupted frames and tool calls come back empty.
--- a/src/mcp/stdio.ts
+++ b/src/mcp/stdio.ts
@@ -141,6 +141,9 @@ export class StdioTransport implements McpTransport {
         // Malformed lines are dropped — some servers emit startup banners
         // before the JSON-RPC loop begins. We surface the noise to stderr
         // via the inherited stderr stream, not our event queue.
+        if (process.env.REASONIX_DEBUG_MCP === "1") {
+          process.stderr.write(`[mcp-stdio] dropped malformed line: ${line}\n`);
+        }
       }
     }
   }
--- a/tests/mcp-stdio-debug.test.ts
+++ b/tests/mcp-stdio-debug.test.ts
@@ -1,0 +1,63 @@
+/** REASONIX_DEBUG_MCP=1 surfaces dropped malformed lines on stderr; otherwise silent. */
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { StdioTransport } from "../src/mcp/stdio.js";
+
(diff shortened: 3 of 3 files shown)


esengine:
Nice and tight — the strict `=== "1"` check is the right call (no truthy traps), and waiting on `messages()` instead of a fixed sleep means the test won't flake on slow Windows CI. Closes #630 as intended.

Merging.

PR #670, opened 2026-05-11
feat(tools,loop): warn when readOnlyCheck throws instead of swallowin…

### LINKED ISSUE #629: Log silently-caught readOnlyCheck / isMutating exceptions
## Problem

`isReadOnlyCall` in `src/tools.ts:278-285` catches all exceptions from a tool's `readOnlyCheck` function and falls through to `return false` (i.e. "not read-only", so the call is **allowed to mutate**):

```ts
function isReadOnlyCall(tool: InternalTool, args: Record<string, unknown>): boolean {
  if (tool.readOnlyCheck) {
    try {
      return Boolean(tool.readOnlyCheck(args as never));
    } catch {
      return false;
    }
  }
  return tool.readOnly === true;
}
```

Two problems:

1. **Silent failure** — if a `readOnlyCheck` throws because of a bug in the check function itself (bad arg shape, regex compile fail, etc.), nobody sees it. The tool just gets treated as mutating forever.
2. **Asymmetric with `loop.ts`** — `src/loop.ts:182-201` has a parallel `isMutating` closure that uses the opposite fall-through ("if it throws, defer to the static `readOnly` flag"). The two paths can disagree on the same call without anyone noticing.

Both catches should at least log so a future debugger sees the signal instead of having to read both files to understand the divergence.

## Proposed change

Add a single `console.warn` in each catch, naming the tool and the error:

```ts
} catch (err) {
  console.warn(
    `readOnlyCheck for ${tool.name} threw: ${(err as Error).message} - treating as not read-only`,
  );
  return false;
}
```

Same treatment in `src/loop.ts:197-199` (with a wording that reflects its different fall-through).

Keep existing semantics — we are not changing what happens, only making the failure visible.

## Scope

- `src/tools.ts` — log in the `isReadOnlyCall` catch.
- `src/loop.ts` — log in the `isMutating` catch.
- `tests/tools.test.ts` — one case: a tool with a throwing `readOnlyCheck` produces a warn-log line; behavior (returns false / falls through to readOnly flag) is unchanged.

Out of scope:
- Changing the default-on-throw semantic.
- Routing the warn through a structured event sink.

## Good first issue

Two small edits in adjacent files plus a single-case test. Follow the existing log style in the codebase (no new logger framework). Comment-policy rules in `CLAUDE.md` apply.

MyPrototypeWhat:
Closes #629.

Two catch blocks treat a throwing readOnlyCheck as 'not read-only' and move on:

- src/tools.ts isReadOnlyCall — used by plan-mode refusal
- src/loop.ts isMutating — used by the storm-breaker window

If the check itself has a bug, nobody sees it; the tool just silently behaves as if it were mutating. Log the exception on console.warn at each catch so the bug surfaces during development without changing default-on-throw semantics.

## What

Replace two silent `catch {}` blocks with `console.warn(...)` calls that surface buggy `readOnlyCheck` functions. Default-on-throw semantics are unchanged — `tools.ts isReadOnlyCall` still returns `false`, `loop.ts isMutating` still falls through to the static `readOnly` flag — but the bug in the check is no longer invisible. Closes #629.

## Why

`readOnlyCheck` lets a tool decide read-only-ness per-call (e.g. `run_command` is read-only iff its argv hits an allowlist). If that function throws because of a bug in the check itself, the tool silently behaves as if it might mutate: plan-mode quietly refuses it, or the storm-breaker reclassifies the call. Nobody sees the underlying exception. A bare `console.warn` is the minimum that surfaces it without dragging in a logging framework or changing observable behavior.

## How to verify

```
npm run verify
npx vitest run tests/tools.test.ts
```

The new test "warns when readOnlyCheck throws and treats the call as not read-only" registers a tool with a throwing `readOnlyCheck`, enables plan mode, dispatches it, and asserts both the plan-mode refusal AND the exact warn message string.

## Checklist

### DIFF
--- a/src/loop.ts
+++ b/src/loop.ts
@@ -194,8 +194,10 @@ export class CacheFirstLoop {
         }
         try {
           if (def.readOnlyCheck(args as never)) return false;
-        } catch {
-          /* ignore — fall through */
+        } catch (err) {
+          // Mirror tools.ts: surface buggy readOnlyCheck instead of silently
+          // falling through to the static flag.
+          console.warn(`readOnlyCheck for ${name} threw: ${(err as Error).message}`);
         }
       }
       return def.readOnly !== true;
--- a/src/tools.ts
+++ b/src/tools.ts
@@ -301,7 +301,10 @@ function isReadOnlyCall(tool: InternalTool, args: Record<string, unknown>): bool
   if (tool.readOnlyCheck) {
     try {
       return Boolean(tool.readOnlyCheck(args as never));
-    } catch {
+    } catch (err) {
+      // A buggy readOnlyCheck silently downgrades to "may mutate" — log it so
+      // the bug doesn't hide behind plan-mode refusals or storm-breaker noise.
+      console.warn(`readOnlyCheck for ${tool.name} threw: ${(err as Error).message}`);
       return false;
     }
   }
--- a/tests/tools.test.ts
+++ b/tests/tools.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 import { ToolRegistry } from "../src/tools.js";
 
 describe("ToolRegistry", () => {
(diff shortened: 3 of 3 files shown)


esengine:
Good catch on the silent downgrade — a buggy `readOnlyCheck` quietly reclassifying as "may mutate" is exactly the kind of bug that hides for weeks. Comments explain the *why* cleanly, the asymmetric handling between `tools.ts` (returns false) and `loop.ts` (falls through to static flag) is preserved correctly, and the test asserts both the plan-mode refusal AND the exact warn string.

Merging.

PR #671, opened 2026-05-11
feat(tools/web): attach actionable hints to web_search / web_fetch er…

### LINKED ISSUE #16: Make `web_search` / `web_fetch` errors actionable

MyPrototypeWhat:
Closes #16.

Tool errors round-trip to the model verbatim, so a bare 'web_search 429' prompts the model to retry the same call with the same args. Pick a status-specific i18n key with a hint clause for the four cases the model most commonly hits:

- 429 — wait 10s, or rephrase the query
- 403 — confirm public access / network reachability / User-Agent
- 5xx — try a browser; if it loads, retry in 30s
- other — confirm reachability and parseability (existing message, reused as the default)

Also surface the AbortController timeout as 'web_fetch timed out after {ms}ms' with a 'use a lighter URL' hint, instead of letting the raw AbortError bubble up. Caller-driven aborts still propagate as-is.

Both EN and zh-CN translations updated; the existing /web_search 429/ regex assertion is preserved as a prefix.

## What

Pick a status-code-specific i18n key with an actionable hint when `web_search` or `web_fetch` hits a non-2xx response, instead of returning the bare status. Four buckets: `429`, `403`, `5xx`, and a generic catch-all that's still actionable. Also surface `AbortController` timeouts in `web_fetch` as `"web_fetch timed out after {ms}ms"` instead of letting the raw `AbortError` bubble up; caller-driven aborts (Esc) still propagate unchanged. EN + zh-CN translations and `TranslationSchema` updated in step. Closes #16.

## Why

Tool errors go back to the model verbatim — they're how the agent decides what to do next. A bare `web_search 429` reads as a transient failure, so the model retries with identical args and burns tokens. `web_search 429 — try: wait 10s before retrying, or rephrase the query` tells the model both that a same-args retry won't help and what to do instead. Same shape for 403 (check public access / UA), 5xx (open in a browser, retry in 30s), timeout (lighter URL).

## How to verify

```
npm run verify
npx vitest run tests/web-tools.test.ts
```

The new `actionable error hints` describe block adds four cases: search 429, search 503, fetch 403, fetch timeout — each asserts on the hint substring, not just the status. Existing assertions on `/web_search 429/` and `/web_fetch 404/` still match (the new messages start with the same prefix).

## Checklist

### DIFF
--- a/src/i18n/EN.ts
+++ b/src/i18n/EN.ts
@@ -1234,7 +1234,14 @@ export const EN: TranslationSchema = {
     probeFailed: "probe failed \u2014 {message}",
   },
   webErrors: {
-    status: "web_search {status}",
+    status:
+      "web_search {status} \u2014 try: confirm the search engine is reachable and your query parses",
+    searchStatus429:
+      "web_search {status} \u2014 try: wait 10s before retrying, or rephrase the query (the engine is rate-limiting)",
+    searchStatus403:
+      "web_search {status} \u2014 try: confirm your network reaches the engine; switch with /search-engine if you're behind a proxy",
+    searchStatus5xx:
+      "web_search {status} \u2014 try: open the search URL in a browser; if it loads this is transient and a retry in 30s may help",
     mojeekBlocked: "web_search: Mojeek anti-bot page \u2014 rate-limited or blocked",
     mojeekNoResults:
       "web_search: 0 results but response doesn't look like a real empty page ({chars} chars, first 120: {preview})",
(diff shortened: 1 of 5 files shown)


esengine:
Hey, thanks for putting this together — the four-bucket strategy, the helper-function shape, and the timeout-vs-abort distinction are all the right calls.

Unfortunately #632 landed first and overlaps heavily with this PR — `rateLimit429` / `forbidden403` / `fetchRateLimit429` / `fetchForbidden403` / `fetchTimeout` are already on `main`, all closing the same #16, with the same general approach. Re-landing those under the `searchStatusNNN` naming would only churn the schema with no behavioral gain, so I'm going to close this one as superseded.

But the **5xx bucket** is a real gap — #632 stopped at 429/403, and surfacing transient 5xx with a "try again in 30s / check in a browser" hint is genuinely valuable. I'll open a small follow-up that adds just `serverError5xx` + `fetchServerError5xx` (sticking to #632's naming convention since that's what's on main), with credit to you in the PR body. Your 503 test case will land alongside.

Sorry for the timing — both PRs were in flight on the same evening and #632 happened to hit `main` first. The contribution isn't wasted; the part that was new sticks.

PR #676, opened 2026-05-11
feat(tools/web): bucket 5xx separately with a transient-retry hint

esengine:
## Summary

Extends the actionable-error work from #632 with a dedicated 5xx bucket. Before this, generic 5xx fell through to the catch-all `webErrors.status` / `webErrors.fetchStatus` message — which advises rephrasing the query or switching engines, neither of which helps when the upstream server is just temporarily broken.

After:

- `web_search 503 — try: open the search URL in a browser; if it loads this is transient and a retry in 30s may help`
- `web_fetch 502 for {url} — try: open the URL in a browser; if it loads this is transient and a retry in 30s may help`

`searchStatusError` / `fetchStatusError` get one extra branch each (`status >= 500 && status <= 599`); EN + zh-CN + `TranslationSchema` updated in step.

Credit to @MyPrototypeWhat — #671 surfaced this gap (their PR added a 5xx bucket alongside a parallel re-do of #632's 429/403 work). #671 is closed as superseded, but the 5xx insight + the 503 test case land here.

## Test plan

- [x] `npm run verify` — 170 files / 2628 tests passing
- [x] Existing `webSearch` 503 test tightened from `/web_search 503.*try:/` to assert on the new `retry in 30s` substring
- [x] New parallel `webFetch` 502 test in the same shape

### DIFF
--- a/src/i18n/EN.ts
+++ b/src/i18n/EN.ts
@@ -1247,6 +1247,8 @@ export const EN: TranslationSchema = {
       "web_search 429 \u2014 try: wait 10s before retrying, or rephrase the query; the search backend is rate-limiting this client",
     forbidden403:
       "web_search 403 \u2014 try: the search backend is blocking this client; switch engine with /search-engine mojeek|searxng, or wait and retry later",
+    serverError5xx:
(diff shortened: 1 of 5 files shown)


MyPrototypeWhat:
Thanks for picking this up cleanly 🙏

PR #682, opened 2026-05-11
feat(changes): checkpoint API + git changes panel with zero external deps

ForeverYoungPp:
What
新增 Checkpoint 完整 CRUD API（创建/列表/diff/还原/删除）及 Dashboard 界面，同时移植了 Git Changes 面板，diff 解析改用手写解析器。零新外部依赖。

Why
缺乏可视化操作，既然是有操作界面就把缺的完善。

How to verify
npm run build — 编译通过
npm run test — 有4个好像是旧失败，我就跳过了
启动 Dashboard，分别切换 Git changes / Checkpoint / Previous session 三个 tab 查看


esengine:
Thanks for the PR — build is green, but CodeQL is failing on a real one: `js/file-system-race` (high severity) at `src/server/api/file-read.ts:85`. Worth fixing before merge.

The pattern flagged is:

```ts
if (!existsSync(fullPath)) { ... }   // check
st = statSync(fullPath);              // race window — file can change
...
readFileSync(fullPath, "utf-8");      // race window again
```

Three separate filesystem touches against the same path. Between any two of them the path can be swapped (symlink flip, rename), so the `isFile()` / `size` guards you ran on `statSync` don't necessarily describe the bytes `readFileSync` ends up returning. CodeQL refuses to assume otherwise.

The fix is to open the file once and do all the checks against the same file descriptor:

```ts
import { openSync, fstatSync, readSync, closeSync } from "node:fs";

let fd: number;
try {
  fd = openSync(fullPath, "r");
} catch (err) {
  const code = (err as NodeJS.ErrnoException).code;
  if (code === "ENOENT") {
    return { status: 404, body: { error: `file not found: ${filePath}` } };
  }
  return { status: 500, body: { error: "cannot open file" } };
}
try {
  const st = fstatSync(fd);
  if (!st.isFile()) return { status: 400, body: { error: "not a file" } };
  if (st.size > MAX_FILE_SIZE) {
    return {
      status: 413,
      body: { error: `file too large (${st.size} bytes, max ${MAX_FILE_SIZE})` },
    };
  }
  const buf = Buffer.alloc(st.size);
  readSync(fd, buf, 0, st.size, 0);
  return { status: 200, body: { content: buf.toString("utf-8"), path: filePath, size: st.size } };
} finally {
  closeSync(fd);
}
```

`fstatSync` + `readSync` both operate on the fd, not the path, so the inode is pinned for the duration. Drop the leading `existsSync` — `openSync`'s `ENOENT` covers that case.

One more thing while you're in this file: `join(cwd, filePath)` does not block path traversal. `filePath = "../../etc/passwd"` resolves outside `cwd`. Normalize with `path.resolve(cwd, filePath)` and reject if the result doesn't start with `cwd + path.sep`. The dashboard binds localhost-only by default, but the API shouldn't trust the path either way.

Push the fix and CodeQL should clear.

ForeverYoungPp:
感谢你细致的评审

PR #712, opened 2026-05-12
feat(config): add statusBar toggles to hide individual status row fields (#626)

### LINKED ISSUE #626: Add statusBar config toggles to hide individual fields (balance, spent, cache hit, version, etc.)
## Problem

The bottom status row currently renders every available signal it has — mode pill, session id, branch, turn cost, cache-hit %, session spent, account balance, version, `/feedback` hint. That's the right default for someone who wants the full cockpit view, but it's not configurable. Anyone who finds the wallet section (`⛁ $X.XX spent / ¥Y.YY left`) noisy or doesn't want to see their balance over their shoulder during a screen share has no way to turn it off short of a code fork.

Everything in the row is already conditional on terminal width and on whether the underlying value exists (`hasBalance`, `hasTurn`, etc.), so the rendering paths are already segmented — we just don't expose them.

## Proposed change

Add an optional `statusBar` section to `ReasonixConfig` in `src/config.ts`:

```ts
statusBar?: {
  showBalance?: boolean;     // default true
  showSessionCost?: boolean; // default true
  showTurnCost?: boolean;    // default true
  showCacheHit?: boolean;    // default true
  showVersion?: boolean;     // default true
  showFeedbackHint?: boolean;// default true
};
```

Defaults preserve current behavior. `false` removes the corresponding field from `src/cli/ui/layout/StatusRow.tsx`. When both `showBalance` and `showSessionCost` are false, drop the whole wallet pill (the `⛁` glyph + separator).

## Scope

- `src/config.ts` — extend `ReasonixConfig`, no migration needed (all optional).
- `src/cli/ui/layout/StatusRow.tsx` — read config (via the existing state provider or a small `useAgentState` selector), gate each block on its flag.
- `tests/` — one test covering "all flags false → wallet pill not rendered" and one covering "default config still renders balance when present".

Out of scope: a `/config` slash UI for this, a per-project override, runtime toggle. Just the file-level switch — that's all this needs.

## Good first issue

Single file in the rendering layer, mechanical config plumbing, clear test surface. Reviewer guidance in `CLAUDE.md` (especially the comment rules) applies. Ping me on the PR if anything's unclear.

Evan-Pycraft:
## Summary

Adds per-field visibility toggles for the bottom status row, so users can hide noisy fields without a code fork.

- New `statusBar` config section in `ReasonixConfig` with 6 optional boolean flags (all default `true`)
- Each field gated independently:
  - `showTurnCost` — current turn cost
  - `showCacheHit` — prefix-cache hit %
  - `showBalance` / `showSessionCost` — wallet fields (wallet pill dropped entirely when both false)
  - `showVersion` — version string
  - `showFeedbackHint` — `/feedback` tip
- Config read once via `readConfig()` in `App`, passed as prop to `StatusRow`, with defaults for testability
- 8 new tests covering all toggle scenarios

## Usage

```json
{
  "statusBar": {
    "showBalance": false,
    "showVersion": false
  }
}
```

## Files

| File | Change |
|------|--------|
| `src/config.ts` | Add `statusBar` section to `ReasonixConfig` |
| `src/cli/ui/App.tsx` | Read config, define `StatusBarConfig` type, pass through `AppInner` |
| `src/cli/ui/layout/StatusRow.tsx` | Accept `statusBar` prop, gate each field |
| `tests/ui-status-row-balance.test.tsx` | 8 new tests for configuration toggles |

Closes #626

### DIFF
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -40,6 +40,7 @@ import {
   markEditModeHintShown,
   markMouseClipboardHintShown,
   mouseClipboardHintShown,
+  readConfig,
   resolveThemePreference,
   saveEditMode,
   savePreset,
@@ -149,6 +150,7 @@ import {
   UndoBanner,
 } from "./layout/LiveRows.js";
 import { StatusRow } from "./layout/StatusRow.js";
+import type { StatusBarConfig } from "./layout/StatusRow.js";
 import { ToastRail } from "./layout/ToastRail.js";
 import { PlanLiveRow } from "./layout/plan-live-row.js";
 import { ViewportBudgetProvider } from "./layout/viewport-budget.js";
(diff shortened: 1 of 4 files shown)


esengine:
Thanks @Evan-Pycraft — clean shape: `readConfig()` once in App via `useMemo`, exported `StatusBarConfig` with sane defaults at the `StatusRow` level (so the component still renders without a prop in standalone tests), 8 toggle-scenario tests. CI all green. Adding the missing `Closes #626` line and merging.

PR #713, opened 2026-05-12
fix(healing): stamp fallback id on bare tool_calls to prevent DeepSeek 400

ForeverYoungPp:
# [Bug] DeepSeek 400 Error: Missing `id` field in tool_calls during sessions

##  **What happened**
DeepSeek returns a **400 Bad Request** error after long sessions:  
`"failed to deserialize the json body into the target type: messages[n]: missing field id"`

In `src/types.ts`, `ToolCall.id` is defined as optional (`id?: string`), but the DeepSeek API strictly requires this field to be present.

##  **Expected**
Sessions should not error out due to missing `tool_call.id`. The application should ensure every tool call sent to the API contains a valid ID.

##  **Reproduction**
1. **Run a long coding session:** 50+ turns with multiple tool calls.
2. **Trigger context compaction:** Let the conversation grow large enough to trigger the `/compact` logic.
3. **Next request:** Wait for the next model interaction → the 400 error appears.

##  **Environment**
- **Reasonix Version:** 0.39.1
- **Node Version:** 24.14.1
- **OS:** Windows 11
- **Shell:** PowerShell
- **Terminal App:** Windows Terminal
- **DeepSeek Model:** deepseek-v4-flash

##  **Logs / Transcript**
- **Symptoms:** Error is displayed in the TUI status bar.
- **Persistence:** The error persists even across `/compact` commands.

## **Root Cause**
Located in `src/types.ts:24`:  
`ToolCall.id` is currently typed as `id?: string` (optional). During certain serialization paths or after context compaction, some `tool_call` entries are generated or restored without an `id` field. DeepSeek's JSON deserializer rejects these incomplete objects.

## **Fix Applied in PR**
Modified `src/loop/healing.ts`:
- Added a new function `stampMissingIds()`.
- This function scans for any `ToolCall` missing an `id` and assigns a fallback ID in the format:  
  `z-ext-{timestamp}-{sequence}`  
  This occurs immediately before the payload is sent to the API.

**Since the bug is hard to reproduce consistently, I'm not sure if this fix is effective. Please let me know what you think.**

---

Image

---

Image

---
image

### DIFF
--- a/src/loop/healing.ts
+++ b/src/loop/healing.ts
@@ -1,6 +1,13 @@
-import type { ChatMessage } from "../types.js";
+import type { ChatMessage, ToolCall } from "../types.js";
 import { shrinkOversizedToolResults, shrinkOversizedToolResultsByTokens } from "./shrink.js";
 import { isThinkingModeModel } from "./thinking.js";
+
+let _stampSeq = 0;
+
+/** DeepSeek 400s on tool_calls missing `id`. Give bare calls a fallback. */
+function stampMissingIds(calls: ToolCall[]): ToolCall[] {
+  return calls.map((c) => (c.id ? c : { ...c, id: `z-ext-${Date.now()}-${_stampSeq++}` }));
+}
 
 /** Drops both unpaired assistant.tool_calls and stray tool messages — DeepSeek 400s on either. */
 export function fixToolCallPairing(messages: ChatMessage[]): {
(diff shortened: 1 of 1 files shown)


esengine:
Merging — the stamp approach drops invalid pairs cleanly via the existing pairing logic (stamped ids never match a real tool response → falls through to the drop branch). Thanks for chasing this down. Will tighten it to an explicit "drop if any tool_call lacks id" in a follow-up; behaviorally identical.

esengine:
Looks like the PR is still marked as a draft — flip it to "Ready for review" when you're comfortable and I'll merge it. The behavior I described above is what I've verified against the existing pairing logic; no further changes needed from your side.

ForeverYoungPp:
Got it, thanks.

esengine:
Thanks — merging. As noted earlier, the stamped fallback id won't match any real tool result, so the existing pairing logic drops the broken assistant turn instead of sending it on to the API. Same end state, and the 400 path is gone.

PR #734, opened 2026-05-13
fix(plan): stabilize expanded approval modal layout

kabaka9527:
## What

This change stabilizes the expanded plan approval modal in the CLI/TUI. It keeps the approval card from being vertically compressed inside the app-style modal stack, tightens
  the expanded detail/options layout, and makes expanded-plan scrolling stable when the detail window encounters blank lines at the top of the viewport. It also documents the
  `Ctrl+P` summary/detail toggle in the CLI reference, updates the related i18n strings/types, and adds regression coverage for the clipping and scroll-jitter cases. A follow-up
  test fix restores runtime `React` imports in JSX-based Ink tests so the push-time verify suite passes.

## Why

The expanded `PlanConfirm` UI had two regressions: the detail header/content could be clipped in the real app layout, and the visible content could jump while scrolling long
  plans. Those issues made reviewing plans harder and caused confusing rendering in the approval flow. The test import fix was needed because the verify hook executes the
  JSX-based tests under Vitest, where those files still need `React` in value scope at runtime.

## How to verify

 1. Run `npm run verify`.
  2. Run `npx vitest run tests/plan-confirm.test.tsx tests/user-card-verbatim.test.tsx`.
  3. Start the CLI UI and open a submitted plan in the approval modal.
  4. Press `Ctrl+P` to expand the plan body.
  5. Confirm the expanded detail header is fully visible, the scroll hint sits directly above the options list, and scrolling through long plans does not cause the visible content
   to jump when blank lines enter the top of the detail window.

## Checklist

- [✅ ] `npm run verify` passes locally (lint + typecheck + tests + comment-policy gate)
- [✅ ] No `Co-Authored-By: Claude` trailer in commits
- [✅ ] Comments follow CONTRIBUTING.md (no module-essay headers, no incident history)
- [✅ ] No edits to `CHANGELOG.md` — release notes are maintainer-written at release time

### DIFF
--- a/docs/CLI-REFERENCE.md
+++ b/docs/CLI-REFERENCE.md
@@ -109,7 +109,7 @@ Type `/` mid-chat to open the picker. Aliases shown in parentheses. Code-mode-on
 | `/show [id]` | Dump a stored edit diff |
 | `/commit "msg"` | `git add -A && git commit -m ...` |
 | `/mode <review\|auto\|yolo>` | Edit-gate mode. Shift+Tab cycles |
-| `/plan [on\|off]` | Toggle read-only plan mode |
+| `/plan [on\|off]` | Toggle read-only plan mode. Submitted plans initially show a compact summary; press `Ctrl+P` in the plan confirmation modal to expand/collapse full details |
 | `/checkpoint [name\|list\|forget]` | Snapshot every file the session has touched |
 | `/restore <name\|id>` | Roll back to a named checkpoint |
 | `/cwd <path>` (`/sandbox`) | Switch the workspace root mid-session |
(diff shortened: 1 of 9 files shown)


esengine:
Thanks — nice fix. Read the diff: layout-stability + bounded expand + the blank-line jitter guard are all solid, and the regression tests pin down the exact repro cases you found. The arrow-keys-own-the-detail-window trade-off when expanded is fine, the test makes the contract explicit.

A couple of follow-up nits I'll handle separately, no need to redo this PR:

- The two magic numbers (`EXPANDED_MODAL_OVERHEAD_ROWS = 12`, `EXPANDED_DETAIL_CHROME_ROWS = 4`) are derived from the current card chrome — a one-line comment naming the components (header / divider / footer hints) would future-proof them.
- `stripAnsi()` in the test helper rolls its own; we already depend on `string-width` which ships a battle-tested `strip-ansi`. I'll swap that over.

Merging.

PR #760, opened 2026-05-13
fix(ui): capture keystrokes while scrolled up in history

rokyplay:
## What

When the user scrolls up to read chat history, `PromptInput` is unmounted and all keystrokes are silently dropped. This PR adds an App-level `useKeystroke` handler that captures printable characters, backspace, and Enter (returns to bottom) when `pinned=false`, so the user can type blind while reading history and see their input when they scroll back — Claude Code parity.

## Why

Reported by a user testing Reasonix for the first time. The current behavior (keystrokes silently lost) is confusing and breaks the flow of "read the agent output, start typing my next question, scroll back to see it."

## How

- Adds `const pinned = useChatScrollState((s) => s.pinned)` to `AppInner`
- Registers a new `useKeystroke` handler active when `!modalOpen && !pinned && !busy`
- Handler: printable chars → append to `input`, backspace → delete last char, Enter → `chatScroll.jumpToBottom()`
- Supports CJK IME input (multi-character `ev.input`)

## Test plan

- [x] `npm run verify` passes (same 2 pre-existing memory test failures as main)
- [x] Manual: scroll up → type letters, numbers, symbols, Chinese → scroll back → input preserved
- [x] Manual: scroll up → press Enter → returns to bottom, input visible

### DIFF
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -413,6 +413,7 @@ function AppInner({
   const isStreaming = useAgentState((s) => s.cards.some((c) => c.kind === "streaming" && !c.done));
   const activityLabel = useActivityLabel();
   const chatScroll = useChatScrollActions();
+  const pinned = useChatScrollState((s) => s.pinned);
   const [input, setInput] = useState("");
   const [busy, setBusy] = useState(false);
   const [slashUsage, setSlashUsage] = useState<Readonly<Record<string, number>>>(() =>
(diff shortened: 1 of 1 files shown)


esengine:
Nice catch — the blind-typing flow is a real papercut. Merging. Will follow up with a small refactor so the keystroke handler subscribes to `pinned` from a leaf component (the same pattern `InputAreaWithHistoryHint` uses), so AppInner stays untouched on scroll ticks; pure perf, no behaviour change.

Issue #765, opened 2026-05-13
acp: --transcript + --yolo flags for headless callers


ukint-vs:
ACP works end-to-end on `main` today (`v0.40.0` + the post-tag ACP commits). I drove a full session against a local build and watched `initialize` → `session/new` → `session/prompt` round-trip cleanly, with kernel events streaming as `agent_thought_chunk` / `agent_message_chunk` / `tool_call` / `tool_call_update`, and a real `write_file` tool call landing bytes on disk.

Two small gaps are blocking a downstream adapter that drives Reasonix from a non-interactive parent process (the use case here is a Paperclip cost-tracking layer; the same shape applies to any CI harness or headless caller).

### 1. `--transcript <path>` on `acp`

`chat` and `code` already accept `--transcript <path>` and write JSONL receipts via `openTranscriptFile()` + `recordFromLoopEvent()` in `src/transcript/log.ts`. The ACP path produces the same `LoopEvent` stream — with `usage` / `cost` / `prefixHash` populated on `assistant_final` — so reusing both helpers verbatim is a 5-line tee in the `session/prompt` handler:

```ts
// transcript needs raw LoopEvent (usage/cost/stats); kernel events lose those fields
if (stream) writeRecord(stream, recordFromLoopEvent(ev, { model: session.ctx.model, prefixHash: session.ctx.prefixHash }));
for (const kev of session.eventizer.consume(ev, session.ctx)) {
  dispatchKernelEvent(server, session.id, kev);
  ...
}
```

No new types. No new shape. The flag would be **single-session-per-process**, mirroring how `chat` / `code` already behave (one process = one transcript). I'd document that on the `--help` line.

### 2. `--yolo` on `acp`

`EditMode = "yolo"` already exists as a persisted config value, and the ACP `pauseGate` listener at `src/cli/commands/acp.ts:114` already honors it via `autoResolveVerdict(req, loadEditMode())`. Headless callers shouldn't have to mutate `~/.reasonix/config.json` to skip plan modals, so:

```ts
const editMode = opts.yolo ? "yolo" : loadEditMode();
const auto = autoResolveVerdict(req, editMode);
```

Scope intentionally narrow: **plan-checkpoint-only**, matching the existing `editMode = "yolo"` semantics. Shell / edit / future gates still bridge via `requestPermissionForGate()` → `session/request_permission` — unchanged. Not adding `--yolo` to `chat` / `code` (they have a TUI that surfaces the modal).

[...]

PR #766, opened 2026-05-13
feat(acp): --transcript flag for headless cost capture

ukint-vs:
Closes part 1 of #765.

## What

Adds `--transcript <path>` to `reasonix acp`, matching the flag already accepted by `chat` and `code`. Headless ACP callers (the immediate use case is a Paperclip cost-tracking adapter; same shape applies to any CI harness) can now parse `usage` / `cost` / `prefixHash` without driving the TUI.

## How

Five-line tee in the `session/prompt` handler, reusing existing helpers:

```ts
// transcript needs raw LoopEvent (usage/cost/stats); kernel events lose those fields
if (transcriptStream) {
  writeRecord(
    transcriptStream,
    recordFromLoopEvent(ev, {
      model: session.ctx.model,
      prefixHash: session.ctx.prefixHash,
    }),
  );
}
```

- Reuses `openTranscriptFile` / `writeRecord` / `recordFromLoopEvent` from `src/transcript/log.ts`. Zero changes to those helpers — same JSONL shape `chat` and `code` already emit.
- `_meta` header carries `source: \"reasonix acp\"`.
- Single transcript per process, mirroring `chat`/`code` precedent.

## Scope

| File | Lines | Why |
|---|---|---|
| `src/cli/commands/acp.ts` | +29 / -2 | `AcpOptions.transcript`, open/close stream, tee in `session/prompt` |
| `src/cli/index.ts` | +2 | Register `--transcript` on `.command(\"acp\")`, forward to `acpCommand` |
| `tests/acp-transcript.test.ts` | +128 (new) | Unit-level coverage matching `tests/acp.test.ts` precedent (mocked stdio, scripted `LoopEvent` generator); meta header shape, multi-prompt ordering, fixture validation |
| `tests/fixtures/acp-driver.ndjson` | +3 (new) | NDJSON driver fixture per @esengine's request — foundation for future end-to-end smoke |

Not included (split per \"one logical change per PR\"): `--yolo` flag — coming as a separate PR rebased on this one.

## Review notes

The stream lifecycle (no `'error'` handler, `.end()` not awaited) intentionally matches `src/cli/ui/App.tsx:775-799` precedent exactly. Adversarial pre-landing review flagged that an unwritable transcript path crashes the host process — but `chat` / `code` have the same behavior. Diverging on `acp` alone would create surface-area drift. Happy to revisit as a separate cross-cutting issue.

Multi-session-per-process: not enforced in code (the maintainer noted in #765 that documenting single-session is the right call rather than building multiplexing). Records carry per-session `model` / `prefixHash` already; only the `_meta.model` and the implicit single-writer assumption come from process startup.

## Verification

```sh
npm run verify     # lint + typecheck + 2,779 tests + build — green
node dist/cli/index.js acp --transcript /tmp/rx.jsonl   # EOF stdin writes _meta and exits clean
```

End-to-end against a real DeepSeek key not run here (no key in the dev environment); the unit tests cover the full `LoopEvent` → record path including `assistant_final` with `usage` / `cost` / `prefixHash`.

Refs: #765

### DIFF
--- a/src/cli/commands/acp.ts
+++ b/src/cli/commands/acp.ts
@@ -1,7 +1,7 @@
 /** ACP (Agent Client Protocol) agent — drives the cache-first loop over stdio NDJSON JSON-RPC. */
 
 import { AsyncLocalStorage } from "node:async_hooks";
-import { existsSync, statSync } from "node:fs";
+import { type WriteStream, existsSync, statSync } from "node:fs";
 import { resolve } from "node:path";
 import { dispatchKernelEvent } from "../../acp/dispatch.js";
 import { requestPermissionForGate } from "../../acp/gates.js";
@@ -31,6 +31,7 @@ import { autoResolveVerdict } from "../../core/pause-policy.js";
 import { loadDotenv } from "../../env.js";
 import { CacheFirstLoop, DeepSeekClient, ImmutablePrefix } from "../../index.js";
 import { timestampSuffix } from "../../memory/session.js";
+import { openTranscriptFile, recordFromLoopEvent, writeRecord } from "../../transcript/log.js";
 import { VERSION } from "../../version.js";
 import { canonicalPresetName, resolvePreset } from "../ui/presets.js";
 
@@ -38,6 +39,7 @@ export interface AcpOptions {
   model?: string;
   dir?: string;
   budgetUsd?: number;
+  transcript?: string;
 }
 
 interface Session {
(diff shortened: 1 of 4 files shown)


esengine:
Clean part-1 work — scope is tight, the helpers it reuses are exactly the right ones, the stream-lifecycle / single-session calls match the precedent we set in #765, and the NDJSON driver fixture lands as planned. CI green. Merging.

PR #767, opened 2026-05-13
feat(acp): --yolo flag to bypass plan checkpoints

ukint-vs:
Closes part 2 of #765 (ack from @esengine: https://github.com/esengine/DeepSeek-Reasonix/issues/765#issuecomment-4439399331).

**Stacked on #766** — that PR adds `--transcript`. This PR shows both commits until #766 merges, after which I'll rebase and only the `--yolo` commit will remain.

## What

Per-invocation override of `editMode` for plan-checkpoint gates only. Headless ACP callers can skip plan modals without mutating `~/.reasonix/config.json`.

## How

Three lines in the `pauseGate.on` listener:

```ts
const editMode = opts.yolo ? "yolo" : loadEditMode();
const auto = autoResolveVerdict(req, editMode);
```

- `autoResolveVerdict` already routes `editMode = "yolo"` to `{type: "continue"}` for `plan_checkpoint` — see `src/core/pause-policy.ts:12-17`. Zero changes there.
- **Plan-checkpoint-only scope.** `run_command`, `edit_file`, and any future gates still bridge to the client via `session/request_permission`. The test asserts this explicitly.
- No env var, no global flag mode, no surface beyond the one option.

## Scope

| File | Lines | Why |
|---|---|---|
| `src/cli/commands/acp.ts` | +3 / -1 | `AcpOptions.yolo` field; editMode source switch in pauseGate listener |
| `src/cli/index.ts` | +5 | Register `--yolo` on `.command(\"acp\")`, forward to `acpCommand` |
| `tests/acp-yolo.test.ts` | +75 (new) | Unit-level: yolo overrides config, fallback to config when undefined, `run_command` still bridges even with `--yolo` |

## Review notes

I considered extracting a small `resolveEditMode(opts)` helper to avoid duplicating the inline expression between production code and the test. Skipped: one call site, and CONTRIBUTING.md (\"no abstractions before two call sites\"). The test replicates the 1-line decision inline.

## Verification

```sh
npm run verify     # lint + typecheck + 2,783 tests + build — green
node dist/cli/index.js acp --help   # shows --yolo with the scoped description
```

Refs: #765

### DIFF
--- a/src/cli/commands/acp.ts
+++ b/src/cli/commands/acp.ts
@@ -40,6 +40,7 @@ export interface AcpOptions {
   dir?: string;
   budgetUsd?: number;
   transcript?: string;
+  yolo?: boolean;
 }
 
 interface Session {
@@ -124,7 +125,8 @@ export async function acpCommand(opts: AcpOptions): Promise<void> {
   }
 
   pauseGate.on((req) => {
-    const auto = autoResolveVerdict(req, loadEditMode());
+    const editMode = opts.yolo ? "yolo" : loadEditMode();
+    const auto = autoResolveVerdict(req, editMode);
     if (auto !== null) {
       pauseGate.resolve(req.id, auto);
       return;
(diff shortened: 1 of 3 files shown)


esengine:
Same shape as part-1 — minimal diff, scope kept to plan-checkpoint via the `editMode` source switch, and the test asserting `run_command` still bridges with `--yolo` is exactly the line of defense this needed. #766 just landed; go ahead and rebase to drop the stacked commits and I'll merge as soon as CI's green on the new base.

esengine:
Clean rebase — just the yolo commit now. CI's green across build + CodeQL. Merging.

PR #780, opened 2026-05-13
feat(acp): --mcp + --mcp-prefix for headless callers

### LINKED ISSUE #779: acp: --mcp + --mcp-prefix flags for headless callers
Follow-up to #765 — same headless-callers context (Paperclip adapter, CI harnesses).

`chat`, `run`, and `code` all accept `--mcp <spec>` (multi-value) and `--mcp-prefix <str>`. `acp` doesn't. With #766 (`--transcript`) and #767 (`--yolo`) merged, this is the last upstream gap before a fully-headless ACP session can call out to bundled MCP servers.

## What

```sh
reasonix acp --mcp paperclip=/abs/paperclip-mcp-server \
             --mcp paperclip-extras=/abs/another-server \
             --mcp-prefix paperclip_ \
             --transcript /tmp/run.jsonl --yolo
```

The agent gets `paperclip_<tool>` MCP tools wired into its `ToolRegistry`, same as `chat`/`run`/`code` do today.

## How

Lift the MCP loading loop from `src/cli/commands/run.ts:81-142` verbatim into `acpCommand` (after `buildCodeToolset` and before constructing `ImmutablePrefix`). The bridged tools land in the immutable prefix, picked up by the cache-first loop with no further wiring.

- `src/cli/commands/acp.ts` — extend `AcpOptions` with `mcpSpecs?: string[]`, `mcpPrefix?: string`; do the spec → preflight → transport → client → `bridgeMcpTools` dance in `buildSession`; call `McpClient.close()` for each registered server on shutdown so child MCP processes don't leak.
- `src/cli/index.ts` — `.option(\"--mcp <spec>\", ..., multi-value)` + `.option(\"--mcp-prefix <str>\", ...)` on the acp command block. Forward into `acpCommand({...})`.

Estimated footprint: ~75 LOC prod across two files; ~100 LOC test in a new `tests/acp-mcp.test.ts` (multi-spec parsing, malformed-spec error path, MCP child dispose on session end, MCP child crash propagation).

## Scope

Both flags are additive, off by default. Matches the existing `chat`/`run`/`code` parser surface byte-for-byte (same `parseMcpSpec`, same `bridgeMcpTools`, same `mcpEnvFor` config layering). Zero changes to `src/mcp/*` helpers.

No interaction with `--yolo` or `--transcript` — the three flags compose.

Happy to send the PR. Standard branch+commit conventions from #766/#767 (issue-first per CONTRIBUTING.md, no Co-Authored-By, no CHANGELOG.md touch).

Refs: #765 #766 #767

ukint-vs:
## Summary
- Adds `--mcp <spec>` (repeatable) and `--mcp-prefix <str>` to `reasonix acp`, lifting the spec/transport/bridge logic from `run.ts:81-142` into a shared `loadMcpServers` helper.
- MCP tools are bridged into the toolset registry **before** `ImmutablePrefix` is built, so they end up in the cache key alongside Reasonix's native tools.
- Spawned MCP children are disposed when the ACP stdin closes — no leaked stdio pipes when the headless caller exits.
- Same failure semantics as `run` / `chat`: a broken side-concern server logs to stderr and the session keeps going.

## Why
`reasonix acp` is the entrypoint headless ACP clients use to drive Reasonix. Without this flag, an embedder that wants domain-specific tools (Paperclip's `reasonix-local` adapter, custom MCP servers, etc.) has to mutate the user's saved config — which is unsafe for a process-per-task invocation. Mirrors what's already accepted on `run` / `chat`.

Closes #779.

## Test plan
- [x] `tests/acp-mcp.test.ts` — 7 cases: multi-spec parsing, single-anon `--mcp-prefix` honored, multi-spec prefix dropped, `mcpDisabled` skipped, initialize failure non-fatal (client closed), malformed-spec non-fatal, empty array no-op.
- [x] `node dist/cli/index.js acp --help` shows `--mcp` and `--mcp-prefix`.
- [x] `npm run verify` green (186 test files, 2814 tests, 0 failures).

## Lineage
Third in a small stack of upstream-able pieces of headless-ACP plumbing:
- #766 `--transcript <path>` (merged)
- #767 `--yolo` (merged)
- this PR — `--mcp` + `--mcp-prefix`

### DIFF
--- a/src/cli/commands/acp.ts
+++ b/src/cli/commands/acp.ts
@@ -23,16 +23,30 @@ import {
 import { AcpServer } from "../../acp/server.js";
 import { codeSystemPrompt } from "../../code/prompt.js";
 import { buildCodeToolset } from "../../code/setup.js";
-import { loadApiKey, loadBaseUrl, loadPreset, loadReasoningEffort } from "../../config.js";
+import {
+  loadApiKey,
+  loadBaseUrl,
+  loadPreset,
+  loadReasoningEffort,
+  mcpEnvFor,
+  readConfig,
+} from "../../config.js";
 import { loadEditMode } from "../../config.js";
 import { Eventizer } from "../../core/eventize.js";
 import { pauseGate } from "../../core/pause-gate.js";
 import { autoResolveVerdict } from "../../core/pause-policy.js";
 import { loadDotenv } from "../../env.js";
 import { CacheFirstLoop, DeepSeekClient, ImmutablePrefix } from "../../index.js";
+import { McpClient } from "../../mcp/client.js";
+import { preflightStdioSpec } from "../../mcp/preflight.js";
+import { bridgeMcpTools } from "../../mcp/registry.js";
+import { parseMcpSpec } from "../../mcp/spec.js";
+import { buildTransportFromSpec } from "../../mcp/transport-from-spec.js";
 import { timestampSuffix } from "../../memory/session.js";
 import { openTranscriptFile, recordFromLoopEvent, writeRecord } from "../../transcript/log.js";
 import { VERSION } from "../../version.js";
+import { formatMcpLifecycleEvent } from "../ui/mcp-lifecycle.js";
+import { formatMcpSlowToast } from "../ui/mcp-toast.js";
 import { canonicalPresetName, resolvePreset } from "../ui/presets.js";
 
 export interface AcpOptions {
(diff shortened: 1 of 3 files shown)


esengine:
Approach is right on all three points — bridging MCP tools before `ImmutablePrefix` is built keeps the prefix-cache key honest, lifting the spec/transport logic into a shared `loadMcpServers` is the cleanup I'd have asked for anyway, and the stdin-close → dispose hook closes the leaked-stdio-pipe failure mode that mattered most for embedder process-per-task. CI green, scope matches #766 / #767. Approved — thanks for keeping the stack tight.

PR #805, opened 2026-05-14
feat(slash): autocomplete skill names in /skill arg picker

douglarek:
Add `argCompleter: "skills"` to the `/skill\ command so users see a picker with available user-defined skills (project + global scope) when typing `/skill ` — excludes built-in skills to reduce noise.

## Changes

- **`src/cli/ui/slash/types.ts`** — Add `"skills"\ to the `argCompleter` type union in `SlashCommandSpec`
- **`src/cli/ui/slash/commands.ts`** — Register `argCompleter: "skills"` on the skill command spec
- **`src/cli/ui/useCompletionPickers.ts`** — Add `SkillStore`-based resolver that lists project + global skills (filters out `builtin` scope)

## Testing

- Typecheck: ✅ `tsc --noEmit`
- Build: ✅ `tsup`
- All existing tests pass (GPG-related failures in `server-dashboard.test.ts` are pre-existing)

image

### DIFF
--- a/src/cli/ui/slash/commands.ts
+++ b/src/cli/ui/slash/commands.ts
@@ -175,6 +175,7 @@ export const SLASH_COMMANDS: readonly SlashCommandSpec[] = [
     group: "extend",
     argsHint: "[list|show <name>|new <name>|<name> [args]]",
     summary: "list / run / scaffold user skills (<project>/.reasonix/skills + ~/.reasonix/skills)",
+    argCompleter: "skills",
   },
 
   {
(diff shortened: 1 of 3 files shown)


esengine:
Nice — small, focused, drops cleanly into the existing `argCompleter` pattern. Filtering out builtin scope is the right call (nothing the user can author or edit, so showing them would be noise).

Merging. Thanks for the screenshot — picker behavior is exactly what `/skill ` should do.

PR #851, opened 2026-05-14
增加模型在第一次会话会根据 memory.md 命中对应规则

ForeverYoungPp:
**解决 memory.md 索引命中问题**

本次提示词是通过模型写的，质量或许存在问题，希望作者能提供更好的解决方案。

# 问题

  目前现象是在 Reasonix ，首次会话时无法对指令进行遵守，常常无视直接执行，在 v4-flash 下会经常发生，而 v4-pro因为模型质量比较好会遵守执行，但还是有小部分概率忽略规则执行。

  **触发现象**
- 在当前会话让它保存相应的内容到目录中，开启新的会话时，会直接忽略我在 global/MEMORY 中写入的规则，执行搜索在读取，而不是直接加载。
- 使用 GitNexus 明确告诉 FTS 在 Win11 环境下有问题不要调用，但是模型还是会先进行调用

排查发现在 memory.ts 进对索引的描述只有 简短的一句 ‘One-line summary shown in MEMORY.md (under ~150 chars).’ 这就很有可能 v4-flash 即使读取了，也会再首次会话开始时直接忽略，需要再次提醒。

# 解决方案
 
**调整提示词，增加命中关键词，提供可供参考的示例，让模型自己完善 memory.md 索引**
```
One-line summary shown in MEMORY.md index (capped ~110-130 chars) — this is ALL the model sees when deciding if the memory is relevant.

MUST include trigger keywords — words the user would SAY or the model would THINK when this memory should activate.
Rule: if a word never appears in a user message, it's a topic name, not a trigger keyword.

Format: trigger:<word1/word2/word3> | <specific action to take>

GOOD — trigger keywords are user-language, action is concrete:
  trigger:deploy/release/ship | Blue-green deploy, 10% traffic first
  trigger:write-tests/add-tests/run-tests | Mock external services, never connect real databases
  trigger:reply/tone/how-to-answer | Keep replies direct, skip greetings and filler

BAD — with reason:
  Deployment rules           — pure concept name, no trigger keywords, no action
  Be careful with tests      — vague advice, no trigger scenario, no action
  FORMAT CORRECT, CONTENT WRONG:
    trigger:project/code/work | Follow best practices
      — trigger words match EVERY message (too broad),
        action is empty filler, says nothing concrete

```

# 效果

**修改前**
- memory.md
image

image

-----

**修改后**

- memory.md
image

image

### DIFF
--- a/src/tools/memory.ts
+++ b/src/tools/memory.ts
@@ -60,7 +60,26 @@ export function registerMemoryTools(
         },
         description: {
           type: "string",
-          description: "One-line summary shown in MEMORY.md (under ~150 chars).",
+          description:
+            "One-line summary shown in MEMORY.md index (capped ~110-130 chars) — this is ALL the model sees when deciding if the memory is relevant.\n" +
+            "\n" +
+            "MUST include trigger keywords — words the user would SAY or the model would THINK when this memory should activate.\n" +
+            "Rule: if a word never appears in a user message, it's a topic name, not a trigger keyword.\n" +
+            "\n" +
+            "Format: trigger:<word1/word2/word3> | <specific action to take>\n" +
+            "\n" +
+            "GOOD — trigger keywords are user-language, action is concrete:\n" +
+            "  trigger:deploy/release/ship | Blue-green deploy, 10% traffic first\n" +
+            "  trigger:write-tests/add-tests/run-tests | Mock external services, never connect real databases\n" +
+            "  trigger:reply/tone/how-to-answer | Keep replies direct, skip greetings and filler\n" +
+            "\n" +
+            "BAD — with reason:\n" +
+            "  Deployment rules           — pure concept name, no trigger keywords, no action\n" +
+            "  Be careful with tests      — vague advice, no trigger scenario, no action\n" +
+            "  FORMAT CORRECT, CONTENT WRONG:\n" +
+            "    trigger:project/code/work | Follow best practices\n" +
+            "      — trigger words match EVERY message (too broad),\n" +
+            "        action is empty filler, says nothing concrete",
         },
         content: {
           type: "string",


esengine:
Thanks for the writeup — and looking at the before/after screenshots, the symptom is more concrete than the diff alone showed. The "before" hook (`test-cognee1 技能：文件→cognee 的完整 MCP 方法链`, 92 chars) is an abstract noun phrase with no actionable cue; the "after" hook carries explicit trigger keywords + a concrete `read_file → cognee_remember → cognee_cognify → …` recipe. That kind of hook *would* survive recall better. The diagnosis is right.

Going to push back on the specific implementation shape though.

**The `trigger:` prefix is a new format competing with the existing `— hook` convention.** The actionable-recipe half of your "after" example is exactly the kind of hook the existing convention already asks for — what's new is the `trigger:` keyword block. We don't need a new schema for that; it's a *writing-guidance* change ("hooks should embed the user-words that should trigger them"), not a *format* change. Two formats coexisting splits the index and forces the model to learn both.

**164 chars exceeds the ~150 cap.** That cap exists because `MEMORY.md` gets loaded into every session prompt; breaking it on every entry adds up. The recipe part of your example is doing real work, but a cleaner version drops the explicit `trigger:` literal and folds the trigger words into natural prose.

**Tool-parameter descriptions are permanent cached-prompt cost.** This description sits in the system prompt every turn the memory tools are exposed — i.e. essentially always. 20 lines of guidance + examples is a permanent cache-prefix tax. Cheap is load-bearing here.

**The primary lever for "flash skipped MEMORY.md on first turn" is upstream.** Better hooks help at the margin, but if flash isn't being told *when* to consult memory, no hook saves it. That's a system-prompt fix, not a tool-schema fix.

Closing this PR — the right next step is an issue with a concrete repro ("flash ignored this rule, here's the MEMORY.md entry, here's the session-start prompt, here's what it did instead"). I'll dig in from the recall side. If a writing-guidance refresh turns out to be part of the right answer afterwards, it lands as a tightening of the existing one-line convention, not as a new `trigger:` schema. Appreciate the writeup either way — the screenshots are what made this useful.

Issue #681, opened 2026-05-11
模型一直在思考和读文件，没有任何输出


senweiliang:
**Reasonix**: 0.39.1 (latest)
**Platform**: win32 (10.0.26200)
**Terminal**: (unknown) (WT_SESSION=set)
**Size**: 131×35
**Node**: v22.17.1
**Locale**: zh-CN
**Theme**: default
**Model**: deepseek-v4-flash · effort=max
**Mode**: edit=review · plan=off
**MCP**: 0 server(s)
**Session**: code-code

使用auto模式，问了一个问题，感觉是没有限制的一直在推理，花费了¥3.41全程只有思考和读文件，是否是deepseek能力不够？

 ⌘ 上下文 · 71,603 / 1,000,000 (7.2%)
▎ 系统       ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 5,902 · 0.6%
▎ 工具       ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 8,932 · 0.9%
▎ 日志      ██ ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 56,769 · 5.7%
▎ 输入       ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 0 · 0.0%
▎ 常用工具 · 34 工具 · 142 日志消息
▎ read_file · 轮 3 · 2,856
▎ read_file · 轮 3 · 2,487
▎ read_file · 轮 3 · 1,877
▎ read_file · 轮 3 · 1,776
▎ read_file · 轮 3 · 1,694

<img width="1423" height="709" alt="Image" src="https://github.com/user-attachments/assets/6b77aae6-1680-476b-9b7d-a3cad05c5dcd" />

esengine:
Diagnosis: model stuck in a read-only loop — v4-flash repeatedly `read_file` + reason + `read_file` again, 34 tool calls / 142 log messages all skewed to reads, never produced an edit or final answer. Classic "task too heavy for flash + self-loop" shape.

**Existing levers for next time**

- `--budget <usd>` caps session spend; unlimited by default — worth setting on heavy tasks.
- `/preset pro` forces v4-pro, bypassing the auto judgement; usually one or two turns to converge on the same task.
- Esc aborts the current turn; Ctrl-C exits the session.

[...]

PR #860, opened 2026-05-14
chore(loop): remove read-only loop auto-escalation

ForeverYoungPp:
Remove the auto-escalation that bumped flash→pro after 8 consecutive read-only tool calls without a write.

## Why

The shape of this safety net was wrong. Escalating to pro doesn't break a read-only spiral — pro keeps reading files if that's the path the model is on. All the escalation did was make the spiral more expensive, which works against the cache-first / cheap positioning the project leans on. The originating bug (#681 — ¥3.41 burned reading files without producing output) needs a different fix: a stop-after-N-reads-without-progress gate that halts the turn and surfaces the situation, not a model swap. That replacement is on the roadmap separately.

## Changes

- `src/loop/read-only-loop-tracker.ts` — entire file deleted
- `src/loop.ts` — import, field, init, method, and call-site references removed
- `tests/loop.test.ts` — 3 related test cases removed
- `src/i18n/{EN,zh-CN,types}.ts` — `readOnlyLoopEscalation` strings and key removed

## Checklist


esengine:
Updated the body with the rationale myself — my earlier comment was off, sorry for the noise. Thanks for the cleanup; merging.

PR #864, opened 2026-05-14
Add QQ Channel Support to Chat and Code Sessions

wade19990814-hue:
## What

This PR adds QQ as a first-class communication channel for Reasonix, integrated into the existing `chat` and `code` modes rather than introducing a separate runtime mode.

It adds QQ connection management through slash commands (`/qq connect`, `/qq status`, `/qq disconnect`), persists QQ credentials/config in local config, and automatically starts the QQ channel in later `chat` / `code` sessions once enabled. It also bridges interactive pause/confirmation flows to QQ, so when a tool call, plan approval, checkpoint, or choice prompt needs user input, the prompt can be delivered to QQ and the reply can continue the same turn remotely. Follow-up assistant output from that turn is also routed back to QQ.

## Why

Reasonix already has strong interactive flows in the terminal UI, but it had no QQ support as a remote communication surface. That made QQ-based workflows awkward, especially when the user needed to monitor or steer a running session away from the terminal.

This change is designed to stay close to the existing Reasonix architecture: QQ is not treated as a third mode, but as an optional channel layered onto the current `chat` / `code` experience. Once connected, the same session can be continued through QQ for notifications, approvals, and replies.

## How to verify

1. Start an interactive session:
   - `npm run dev -- chat`
   - or `npm run dev -- code`

2. Configure QQ:
   - Run `/qq connect`
   - Provide `appId` / `appSecret` if they are not already configured

3. Confirm connection:
   - Run `/qq status`
   - Verify QQ is enabled and connected

4. Verify auto-start:
   - Exit the session
   - Start `chat` or `code` again
   - Verify QQ starts automatically once previously enabled

5. Verify normal QQ messaging:
   - Send a message from QQ
   - Verify Reasonix receives it and sends the reply back to QQ

6. Verify interactive remote approval flows:
   - Trigger a tool confirmation, plan approval, checkpoint decision, or choice prompt
   - Verify the prompt is sent to QQ
   - Reply from QQ with the requested option or follow-up text
   - Verify the turn continues without terminal-side confirmation
   - Verify the final assistant response for that turn is also delivered to QQ

7. Verify slash command handling from QQ:
   - Send a slash command from QQ
   - Verify it is routed through the existing slash-command dispatcher correctly

## Checklist

## Notes

Update: `npm run verify` now passes locally after syncing the QQ-related tests and fixing a few pre-existing test-environment issues surfaced during this work.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,6 +20,7 @@
         "react": "^19.2.6",
         "string-width": "^7.2.0",
         "undici": "^8.2.0",
+        "ws": "^8.20.1",
         "zod": "^4.4.1"
       },
       "bin": {
@@ -32,6 +33,7 @@
         "@types/node": "^22.9.0",
         "@types/picomatch": "^4.0.3",
         "@types/react": "^19.2.14",
+        "@types/ws": "^8.18.1",
         "@vitest/coverage-v8": "^2.1.5",
         "esbuild": "^0.21.5",
         "highlight.js": "^11.10.0",
@@ -2218,6 +2220,16 @@
       "license": "MIT",
       "dependencies": {
         "csstype": "^3.2.2"
+      }
+    },
+    "node_modules/@types/ws": {
+      "version": "8.18.1",
+      "resolved": "https://registry.npmmirror.com/@types/ws/-/ws-8.18.1.tgz",
+      "integrity": "sha512-ThVF6DCVhA8kUGy+aazFQ4kXQ7E1Ty7A3ypFOe0IcJV8O/M511G99AW24irKrW56Wt44yG9+ij8FaqoBGkuBXg==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@types/node": "*"
       }
     },
     "node_modules/@vitest/coverage-v8": {
@@ -6877,9 +6889,9 @@
       }
     },
     "node_modules/ws": {
-      "version": "8.20.0",
-      "resolved": "https://registry.npmjs.org/ws/-/ws-8.20.0.tgz",
-      "integrity": "sha512-sAt8BhgNbzCtgGbt2OxmpuryO63ZoDk/sqaB/znQm94T4fCEsy/yV+7CdC1kJhOU9lboAEU7R3kquuycDoibVA==",
+      "version": "8.20.1",
+      "resolved": "https://registry.npmmirror.com/ws/-/ws-8.20.1.tgz",
+      "integrity": "sha512-It4dO0K5v//JtTXuPkfEOaI3uUN87iYPnqo/ZzqCoG3g8uhA66QUMs/SrM0YK7/NAu+r4LMh/9dq2A7k+rHs+w==",
       "license": "MIT",
       "engines": {
         "node": ">=10.0.0"
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "react": "^19.2.6",
     "string-width": "^7.2.0",
     "undici": "^8.2.0",
+    "ws": "^8.20.1",
     "zod": "^4.4.1"
   },
   "devDependencies": {
@@ -88,6 +89,7 @@
     "@types/node": "^22.9.0",
     "@types/picomatch": "^4.0.3",
     "@types/react": "^19.2.14",
+    "@types/ws": "^8.18.1",
     "@vitest/coverage-v8": "^2.1.5",
     "esbuild": "^0.21.5",
     "highlight.js": "^11.10.0",
--- a/src/cli/commands/chat.tsx
+++ b/src/cli/commands/chat.tsx
@@ -16,6 +16,7 @@ import {
   renameSession,
   resolveSession,
 } from "../../memory/session.js";
+import { QQChannel } from "../../qq/channel.js";
 import { ToolRegistry } from "../../tools.js";
 import { registerChoiceTool } from "../../tools/choice.js";
 import { registerMemoryTools } from "../../tools/memory.js";
@@ -128,6 +129,12 @@ interface RootProps extends ChatOptions {
   mcpRuntime: McpRuntime;
   /** One-time startup info rows shown after App mounts. */
   startupInfoHints: string[];
+  /** Pre-created QQ channel (started before TUI mounts). */
+  qqChannel?: QQChannel;
+  /** App fills this ref on mount so QQ messages flow into the TUI input queue. */
+  qqSubmitRef: { current: ((text: string) => void) | null };
+  /** App fills this ref on mount so QQ errors appear in the TUI log. */
+  qqErrorRef: { current: ((msg: string) => void) | null };
 }
 
 function Root({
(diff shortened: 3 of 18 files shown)


wade19990814-hue:
I pushed follow-up fixes for the PR:

- validated QQ gateway input more defensively
- bounded heartbeat interval handling
- synced QQ-related CLI / slash tests
- fixed the remaining local `npm run verify` failures that were blocking CI review

`npm run verify` now passes locally.

esengine:
Quick consolidated review on where this stands — splitting the CI failure from the CodeQL alerts so it's clear what's yours to fix and what's mine.

## CI — i18n keys missing

Same failure as the previous push: `tests/slash-help-i18n.test.ts` is red with `expected ['qq'] to deeply equal []`. The last two pushes were merge-from-main commits — the i18n files weren't touched. Two paste-ready diffs:

**`src/i18n/EN.ts`** — next to `skill` (around line 283):

```ts
    qq: {
      description: "connect / inspect / disconnect the QQ channel",
      argsHint: "<connect|status|disconnect>",
    },
```

**`src/i18n/zh-CN.ts`** — matching entry next to `skill` (around line 276):

```ts
    qq: {
      description: "连接 / 查看 / 断开 QQ 通道",
      argsHint: "<连接|状态|断开>",
    },
```

Run `npm run verify` locally after adding both — the rest of the suite is already green.

If you're iterating with an agent, please feed the failing test output back to it directly — the assertion (`expected ['qq'] to deeply equal []`) literally names the missing key, so it's a mechanical fix, not a guessing game.

## CodeQL — 3 alerts remaining (2 will be dismissed, 1 is on you)

**First, the inline `// codeql[...]` comments don't work.** CodeQL doesn't honor inline magic comments the way ESLint does — suppression has to go through `.github/codeql-config.yml` (paths-ignore / query-filters) or a manual dismissal in the Security tab. Please drop those `// codeql[...]` lines, they're noise.

**#70 / #71 — `js/file-access-to-http` (lines 82, 267):** these are *correct* alerts but *intended* behavior. A QQ bot has to read `appid` / `secret` from disk and send them to the QQ OAuth endpoint — that's how the protocol works. There's no source-level fix. I'll dismiss these as "used in intended way" on the Security tab once the PR is ready to land; you don't need to change anything in code for these two.

**#72 — `js/resource-exhaustion` (line 137):** this one *is* fixable in source. You already clamp via `sanitizeHeartbeatInterval`, but CodeQL's dataflow tracker can't see through `Math.max(Math.min(...))` on an `unknown` cast. Restructure so each return path is a clean constant or a guarded narrow:

```ts
private sanitizeHeartbeatInterval(interval: unknown): number | null {
  if (typeof interval !== "number" || !Number.isFinite(interval)) return null;
  if (interval < MIN_HEARTBEAT_INTERVAL_MS) return MIN_HEARTBEAT_INTERVAL_MS;
  if (interval > MAX_HEARTBEAT_INTERVAL_MS) return MAX_HEARTBEAT_INTERVAL_MS;
  return Math.trunc(interval);
}
```

Same end result, but every return value is either a literal constant or a number that's already been range-checked, so the taint tracker resolves and the alert closes. (Alternatively: hard-code `30_000` and ignore the server-provided interval — QQ's heartbeat is a known constant. Either is fine.)

Nice work on the SSRF fix btw — the `validateGatewayUrl` with host allowlist + `wss:` + rejecting user/pass/query/hash is exactly the right shape.

wade19990814-hue:
Force-pushed this PR onto the latest main and collapsed it to a single commit so the diff stays focused on QQ channel support. The branch now removes the legacy standalone QQ CLI surface, keeps QQ integrated under chat / code via /qq connect|status|disconnect, and passes 
pm run verify locally. @esengine could you please take a look when convenient?

esengine:
Nice progress — force-pushing onto current main, collapsing to a single commit, and dropping the standalone `--qq` CLI surface all made the PR much easier to read. CI is green. Three things left before this lands.

## 1. Drop the inline `// codeql[js/...]` comments

Still present at `src/qq/bot.ts:78`, `:134`, `:259`. They're dead text — CodeQL doesn't honor inline magic comments, so they neither suppress anything nor convey signal to readers. Please remove them.

## 2. Restructure `sanitizeHeartbeatInterval` so #72 closes

The `Math.max(MIN, Math.min(MAX, ...))` shape still trips CodeQL's dataflow tracker on `js/resource-exhaustion`. Same end result with explicit branches:

```ts
private sanitizeHeartbeatInterval(interval: unknown): number | null {
  if (typeof interval !== "number" || !Number.isFinite(interval)) return null;
  if (interval < MIN_HEARTBEAT_INTERVAL_MS) return MIN_HEARTBEAT_INTERVAL_MS;
  if (interval > MAX_HEARTBEAT_INTERVAL_MS) return MAX_HEARTBEAT_INTERVAL_MS;
  return Math.trunc(interval);
}
```

Once this is in, #72 should close on the next scan. I'll dismiss #70 / #71 as "used in intended way" on the Security tab — they're inherent to OAuth bot auth and there's no source-level fix.

## 3. Extract a `useQQChannel` hook out of App.tsx

The PR adds **+728 lines to `src/cli/ui/App.tsx`** — `promptLine`, `sendQQInfo`, `qqConnect`, `qqDisconnect`, `qqStatus`, plus the `fromQQ` routing branches inside `handleSubmit` and the slash-interaction handlers. App.tsx is already past the 800-line hard ceiling on main (it's debt we're carrying), and inlining a whole feature's state machine into it makes the debt strictly worse and the feature impossible to test in isolation.

Concretely:

- Move `promptLine` / `sendQQInfo` / `qqConnect` / `qqDisconnect` / `qqStatus` plus the `qqChannelRef` and `qqSlashInteractionRef` state into a new `src/qq/use-qq-channel.ts` exporting a `useQQChannel({ log, codeMode, setQueuedSubmit, ... })` hook that returns `{ connect, disconnect, status, sendInfo, channelRef }`.
- Keep only the wiring in App.tsx: one `const qq = useQQChannel({ ... })` call, plus the minimal `fromQQ` branches in `handleSubmit` that *call* `qq.*` rather than inline the logic.

Target: net diff to App.tsx ≤ ~80 lines (channel wiring + the `[QQ] ` prefix detection). Everything else goes into the new file.

Take your time on this one — the rest of the PR shape is good, this is the load-bearing piece.

wade19990814-hue:
Addressed the remaining review items in `43c60c6`:

- removed the inline `// codeql[...]` comments from `src/qq/bot.ts`
- rewrote `sanitizeHeartbeatInterval()` to the explicit branch form
- extracted the QQ remote interaction/state machine into `src/qq/use-qq-channel.ts` and reduced `App.tsx` to wiring + dispatch branches

`npm run verify` passes locally after rebasing onto the current branch head.

wade19990814-hue:
@esengine I pushed a follow-up fix in e8bc31 for the CI failure after the rebase. The issue was that package-lock.json was out of sync with package.json, which caused 
pm ci to fail (ws / @types/ws).

I re-synced the lockfile and re-ran 
pm run verify locally.

If the fork PR workflows are gated on your side, could you also approve / re-run the pending Actions checks?

esengine:
Thanks for sticking with this through all the rounds — looks good now. CI's green, scope is clean. Merging.

PR #897, opened 2026-05-15
feat(web): add Metaso web search engine.

mrluanma:
## What

Add [Metaso](https://metaso.cn/) as a web search engine backend (alongside Mojeek and SearXNG).

## Why

Provides a zero-setup, free-quota (100 searches/day) alternative to self-hosted SearXNG.

## How to verify

`npm run verify` passes. Test switching engines with /search-engine metaso, and verify web_search tool works via the Metaso API.

## Checklist

### DIFF
--- a/src/cli/ui/slash/handlers/web-search-engine.ts
+++ b/src/cli/ui/slash/handlers/web-search-engine.ts
@@ -5,7 +5,7 @@ import type { SlashHandler } from "../dispatch.js";
 export const handlers: Record<string, SlashHandler> = {
   "search-engine": (args, _loop, ctx) => {
     const engine = args[0];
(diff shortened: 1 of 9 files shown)


esengine:
Thanks — clean PR, the error-mode tests cover the daily-limit / unauthorized / rate-limit / parse paths properly. Merged.

One awareness note, not blocking: the bundled default key pools the 100/day quota across every user who doesn't set their own `METASO_API_KEY`, so under any real usage we'll see `metasoDailyLimit` fire quickly. The error message already points people at the playground to get their own key, so the failure mode is graceful — flagging this so it's not a surprise. If quota churn gets painful we can drop the bundled fallback (require BYOK) or fold the obtain-a-key step into the setup wizard in a follow-up.

Issue #259, opened 2026-05-05
Sensitive-path pattern guard for run_command arguments


esengine:
## Background

The shell tool's allowlist + per-prefix risky-arg demotion (#257, #258) blocks destructive operations, but read-only commands like `cat`, `grep`, `find`, `head`, `tail` are still on the fast path with *any* path argument. The agent can `cat ~/.ssh/id_rsa` or `grep -r AWS_SECRET /etc` and pipe the contents straight into the LLM provider's chat context — a real exfiltration vector even on a local CLI.

OWASP AI Agent Cheat Sheet § 8 (Data Protection / Path traversal) and § 1's example config (`"blocked_patterns": ["*.env", "*.key", "*.pem", "*secret*"]`) frame this as the complement to per-tool least-privilege: not just *which* command, but *against which resources*.

## Proposal

Path-argument scanner on top of the existing allowlist gate. When a `run_command` call matches the allowlist **and** one of its argv tokens resolves into a configured sensitive prefix or matches a sensitive filename pattern, demote it to the confirm gate (same code path as `RISKY_ARGS`).

Default blocklist (suggested):

- **Path prefixes**: `~/.ssh`, `~/.aws`, `~/.gnupg`, `~/.kube`, `/etc/shadow`, `/etc/sudoers`
- **Filename patterns**: `*.env`, `*.env.*`, `*.key`, `*.pem`, `id_rsa*`, `id_ed25519*`, `*credentials*`, `*secret*`

User-configurable. Case-insensitive on Windows. Tilde expansion + cwd-relative resolution required.

## Acceptance criteria

- [ ] New `SENSITIVE_PATHS` config layer (default list + user override)
- [ ] Path-arg scanner in `src/tools/shell.ts` that flags any argv token resolving into the blocklist
- [ ] Demotion fires in the same code path as `RISKY_ARGS` (chains + confirm gate work uniformly)
- [ ] Tests: absolute paths, `~/`-relative paths, cwd-relative paths, glob patterns, case-insensitive on win32
- [ ] No false positives on legit project paths; `./src/.env.example` *should* still trigger (we do not want the agent reading `.env`-shaped files at all)

## Out of scope

- Output-side scanning (post-execution PII filter on stdout) — separate concern, much larger
- Sandboxing / seccomp / chroot — the local-CLI threat model assumes user trust of the binary, just not of the model

## References

- OWASP AI Agent Cheat Sheet § 1, § 8
- Sibling work: #257, #258

[...]

PR #911, opened 2026-05-15
feat(shell): sensitive-path pattern guard for run_command args (#259)

KuaaMU:
## What

Path-argument scanner on top of the existing allowlist gate. When a `run_command` or `run_background` call matches the allowlist but one of its argv tokens resolves into a sensitive prefix or matches a sensitive filename pattern, the call is demoted to the confirm gate — same code path as `RISKY_ARGS`.

## Why

Read-only commands like `cat`, `grep`, `find`, `head`, `tail` are on the fast path with *any* path argument. The agent can `cat ~/.ssh/id_rsa` or `grep -r AWS_SECRET /etc` and pipe the contents straight into the LLM context — a real exfiltration vector even on a local CLI.

OWASP AI Agent Cheat Sheet § 8 (Data Protection / Path traversal) and § 1 frame this as the complement to per-tool least-privilege: not just *which* command, but *against which resources*.

## Default blocklist

**Path prefixes** (tilde-relative, resolved at runtime):
- `~/.ssh`, `~/.aws`, `~/.gnupg`, `~/.kube`
- `/etc/shadow`, `/etc/sudoers`

**Filename patterns** (case-insensitive basename match):
- `*.env`, `*.env.*`, `*.key`, `*.pem`
- `id_rsa*`, `id_ed25519*`
- `*credentials*`, `*secret*`

## User configuration

```yaml
# ~/.reasonix/config.yaml
sensitivePaths:
  prefixes:
    - /opt/secrets
    - ~/work/vault
  patterns:
    - "*.p12"
    - "*token*"
```

## Changes

| File | What |
|------|------|
| `src/tools/shell/parse.ts` | Default blocklist, `hasSensitivePathArgs()`, `isAllowed()`/`isCommandAllowed()` accept optional `projectRoot` + `sensitivePathConfig` |
| `src/tools/shell.ts` | `ShellToolsOptions.sensitivePaths`, pass `rootDir` + config through to dispatch |
| `src/config.ts` | `ReasonixConfig.sensitivePaths` |
| `tests/shell-tools.test.ts` | 12 new tests: tilde-prefix, glob patterns, relative paths, case-insensitivity, custom config, chain commands, backward compat |

## Design notes

- **Backward compatible**: `isAllowed(cmd)` (2-arg form) still works — sensitive-path check only fires when `projectRoot` is provided.
- **Non-path tokens skipped**: flags (`-d`), URLs (`http://...`), env vars (`$HOME`) are not resolved as paths.
- **Prefix boundary matching**: `~/.ssh` matches `~/.ssh/id_rsa` but NOT `~/.ssh_extra/config`.
- **Runs alongside RISKY_ARGS**: both checks are applied; either can demote.

## Acceptance criteria (from #259)

- [x] `SENSITIVE_PATHS` config layer (default list + user override)
- [x] Path-arg scanner in `src/tools/shell.ts` / `parse.ts` that flags any argv token resolving into the blocklist
- [x] Demotion fires in the same code path as `RISKY_ARGS` (chains + confirm gate work uniformly)
- [x] Tests: absolute paths, `~/`-relative paths, cwd-relative paths, glob patterns, case-insensitive on win32
- [x] No false positives on legit project paths; `./src/.env.example` *should* still trigger


esengine:
Solid piece of work — OWASP § 1 / § 8 is the right framing, and demoting through the existing `RISKY_ARGS` code path keeps the gate uniform. Edge coverage (tilde expansion, glob, relative paths, chains, case folding, custom config) is thorough.

Two real fixes before merge:

1. **The user-config layer isn't actually wired up.** `ShellToolsOptions.sensitivePaths` is plumbed, but `src/code/setup.ts:44`'s `registerShellTools` call never reads `config.sensitivePaths` from `readConfig()` and forwards it — so the YAML override the PR documents is silently ignored and only the default list takes effect. Add the read in `registerRooted` and pass it through.

2. **Windows path separator.** `pathStartsWithPrefix` does `normalized.startsWith(`${prefix}/`)` with a literal forward slash. After `pathMod.normalize()` on Windows the result uses backslashes, so `C:\Users\bob\.ssh` followed by `\id_rsa` won't match (Linux CI passes; real Windows misses). Use `pathMod.sep` or accept both separators.

After those two, happy to merge — the defaults, chain handling, and boundary-correct prefix match are otherwise exactly right.

KuaaMU:
Fixed both issues in the latest push:

1. **User-config wiring**: Added `readConfig()` call in `registerRooted` and forwarded `cfg.sensitivePaths` to `registerShellTools`. The YAML override now takes effect.

2. **Windows path separator**: Changed `pathStartsWithPrefix` to use `pathMod.sep` instead of hardcoded `/`, so `C:\Users\bob\.ssh\id_rsa` correctly matches prefix `C:\Users\bob\.ssh` on Windows.

esengine:
Thanks for this — and apologies for the delay; while it was sitting open, the `mcpServers` config (#1037) landed and the `onJobsChanged` callback got added to `ShellToolsOptions`, so I rebased it against `main` and merged the two `ShellToolsOptions`/`registerShellTools` call-sites by hand.

Caught one Windows path issue on the way through: `pathMod.resolve("/etc/shadow")` on Windows turns into `<drive>:\etc\shadow`, but `expandPrefix` was leaving the configured prefix as the literal POSIX string, so the comparison missed and the `/etc/shadow` / `/etc/sudoers` defaults (plus any user-supplied POSIX-absolute prefix like `/opt/secrets`) silently never fired on Windows. One-line fix in `expandPrefix` — run the prefix through `pathMod.resolve` too. Tilde-relative prefixes keep their existing `homedir()` join.

`npm run verify` green on Windows (3080 tests). Merging.

PR #922, opened 2026-05-15
feat(ui): add /cwd path tab-completion with async directory picker

douglarek:
The /cwd slash command now exposes \`argCompleter: "path"\`, a new completer type that async-lists the filesystem via \`listDirectory()\`. Tab drills into directories (trailing '/'), Enter commits. Matches are capped at SEARCH_RESULT_CAP (200) to match @-mention search.

- types.ts: add "path" to argCompleter union
- commands.ts: wire /cwd to argCompleter: "path"
- useCompletionPickers.ts: new usePathCandidates hook; update pickSlashArg for dir drill-down; expose slashArgPathCandidates
- App.tsx: Tab handler checks candidate.isDir for drill vs. commit
- SlashArgPicker.tsx: optional pathCandidates prop, renders trailing / on directory rows

## What

https://github.com/user-attachments/assets/c2116781-041e-4864-b5c4-65389c8127e2

## Checklist

### DIFF
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -1260,6 +1260,7 @@ function AppInner({
     slashArgMatches,
     slashArgSelected,
     setSlashArgSelected,
+    slashArgPathCandidates,
     pickSlashArg,
   } = useCompletionPickers({
     input,
@@ -1637,9 +1638,10 @@ function AppInner({
     }
 
     // Slash-argument picker. Fires inside `/<cmd> <partial>` — either
-    // a file picker (for /edit), enum picker (for /preset, /model,
+    // a path picker (for /cwd), enum picker (for /preset, /model,
     // /plan, /branch, /harvest), or hint-only row. Navigation + Tab
-    // substitute the highlighted value at the arg's offset.
+    // substitute the highlighted value at the arg's offset. For path
+    // completers, directories drill (trailing `/`), files commit.
     if (slashArgMatches && slashArgMatches.length > 0) {
       if (key.upArrow) {
         setSlashArgSelected((i) => Math.max(0, i - 1));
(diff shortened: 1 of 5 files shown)


esengine:
Nice extension of the `argCompleter` abstraction — drill-on-`/` vs commit-on-space is the right shape, and capping at `SEARCH_RESULT_CAP` keeps it consistent with `@`-mention.

Two things before merge:

1. **Windows absolute paths are missed.** Detection at `usePathCandidates` uses `partial.startsWith("/")`, which is POSIX-only. `C:\Users\bob` or `C:/Users/bob` skips the absolute branch, gets resolved against `rootDir`, then immediately falls through the `..`-escape path and tries to list from `/` — wrong twice (root `/` isn't a real path on Windows, and the canonical separator is `\`). Use `node:path.isAbsolute` for detection and `path.parse(p).root` for the listing root. Windows is the project's primary platform; worth handling properly.

2. **`/cwd` lists files too.** The completer offers everything from `listDirectory`. A user tabbing a `.ts` row gets it committed with trailing space; Enter then sends `/cwd path/to/file.ts` which `/cwd` rejects — the picker shouldn't be offering invalid options in the first place. Easiest fix: filter to `isDir` only inside the `"path"` branch (single consumer today, safe). If we later need a file-completer for some other command, introduce `argCompleterMode: "dirs" | "files" | "any"` then.

A couple of unit tests around drill-down and the escape-to-absolute branch would be nice but not blocking — happy to merge once the two above are in.

esengine:
Both addressed cleanly — `isAbsolute(partial)` + `parse(partial).root` is exactly the shape I wanted, and the backslash→forward-slash normalization keeps `parseAtQuery` happy on Windows. Dir-only filter on the listDirectory result is the right place to do it.

Heads-up: the matrix CI change landed in #926 about an hour ago, so on your next contribution the build job will run on `windows-latest` too — the path-handling code in this PR would have been verified end-to-end there. This one's already CI-green on Linux; merging.

PR #931, opened 2026-05-15
fix(repair): prevent silent execution with empty args on truncation failure, fix POSIX process-group kill

zj-rrissh:
## Summary

- **Truncation repair fallback**: When JSON truncation repair completely fails, keep the original truncated args instead of replacing with `{}`. This lets `tools.ts` reject the call with an informative "invalid JSON" error rather than silently running with empty args that miss required parameters.
- **POSIX process-group fix**: Set `detached: true` on spawn options (POSIX only) so child processes become their own process-group leaders. This allows `killProcessTree` to terminate the entire subtree (e.g., `npm → node → esbuild`) instead of only the direct child.

## Test plan

- [x] `npm run lint` — 562 files, no issues
- [x] `npm run typecheck` — passes
- [x] `npm run test:coverage` — 191 files / 2905 tests passed
- [x] `npm run build` — successful

### DIFF
--- a/src/repair/index.ts
+++ b/src/repair/index.ts
@@ -89,9 +89,22 @@ export class ToolCallRepair {
       const args = call.function?.arguments ?? "";
       const r = repairTruncatedJson(args);
       if (r.changed) {
-        call.function.arguments = r.repaired;
-        report.truncationsFixed++;
-        report.notes.push(...r.notes.map((n) => `[${call.function.name}] ${n}`));
+        if (r.fallback) {
+          // Hard fallback — all repair attempts failed. Leave the
+          // original truncated args untouched so tools.ts dispatch
+          // rejects them with "invalid JSON" rather than silently
+          // running with {} (which would miss required params or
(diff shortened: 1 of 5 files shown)


esengine:
Both fixes look good.

**Truncation fallback**: keeping the original truncated args and letting `tools.ts` surface "invalid JSON" beats silently dispatching with `{}` and forcing the model to chase a phantom "missing required parameter" downstream. The `unrecoverable truncation — original args preview: …` note will help when this fires in the wild. The `changed: s !== input` tightening is a quiet improvement.

**POSIX detach**: matches what `JobRegistry` already does for `run_background` (`src/tools/jobs.ts:139`). Bringing `run_command` and the pipe-group path to parity is the right move — `killProcessTree`'s `process.kill(-pid, …)` was a no-op without it on every chain we ran synchronously. Comment placement and the Windows-skip carve-out are correct.

CI green on both Linux and Windows. Merging — thanks for a clean first contribution.

PR #941, opened 2026-05-15
docs: document Metaso as a third web search backend

mrluanma:
## What

Document Metaso as a third web search backend.


esengine:
Thanks — clean follow-up to your own #897. Merging.

PR #965, opened 2026-05-15
fix(desktop): add sticky-to-bottom auto-scroll for chat area

zarttic:
Refs #955 (item 1: sticky-to-bottom auto-scroll — #955 is an umbrella with 10 items, keeping it open)

## Summary

Implements sticky-to-bottom auto-scroll for the desktop chat area, addressing the first suggestion in #955.

## Problem

During long tasks with streaming output, the chat area's auto-scroll only triggered on message count changes (`state.messages.length`). As content grew within existing messages (streaming text, tool results, reasoning), the scroll position would drift — leaving users stranded mid-history, unable to see the latest output without manual scrolling.

## What changed

- **New `useAutoScroll` hook** (`desktop/src/ui/useAutoScroll.ts`) that:
  - Uses `ResizeObserver` on `.thread-inner` to detect **content growth during streaming** (not just message count changes)
  - Uses `requestAnimationFrame` batching to throttle rapid resize events
  - Tracks user scroll position via `scroll` event listener (pinned threshold: 80px)
  - **Auto-scrolls only when the user is pinned to the bottom** — respects user's intent to review history
  - **Auto-scrolls to final answer on turn completion** (busy -> idle transition)
  - Exposes `showJumpButton` + `scrollToBottom` for UI binding

- **`App.tsx`**: Replaced the old `useEffect` scroll hook with `useAutoScroll`; added a floating "jump to bottom" button with chevron icon, positioned after `.thread-inner` inside the scroll container

- **`styles.css`**: Added `.thread-jump-bottom` styles — circular sticky button with shadow, hover scale effect, light/dark theme support

- **i18n**: Added `app.jumpToBottom` key for EN ("Jump to bottom") and zh-CN ("回到底部")

## Behavior

| State | Behavior |
|-------|----------|
| User at bottom + streaming | Auto-scroll follows new content (instant) |
| User scrolled up + streaming | Auto-scroll pauses; jump button appears |
| Task completes (busy->idle) | Always scroll to final answer (smooth) |
| Click jump button | Smooth scroll to bottom, re-pin |

### DIFF
--- a/desktop/src/App.tsx
+++ b/desktop/src/App.tsx
@@ -56,6 +56,7 @@ import {
   UserMsg,
 } from "./ui/thread";
 import { WorkdirPop } from "./ui/workdir-pop";
+import { useAutoScroll } from "./ui/useAutoScroll";
 
 export type AssistantSegment =
   | { kind: "text"; text: string }
@@ -957,6 +958,7 @@ function TabRuntime({
   >(undefined);
   const composerRef = useRef<HTMLTextAreaElement>(null);
   const threadRef = useRef<HTMLDivElement>(null);
+  const threadInnerRef = useRef<HTMLDivElement>(null);
   const [settingsOpen, setSettingsOpen] = useState(false);
   const [settingsPage, setSettingsPage] = useState<SettingsPageId>("general");
   const [jobsOpen, setJobsOpen] = useState(false);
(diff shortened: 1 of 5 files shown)


esengine:
LGTM — ResizeObserver on `.thread-inner` + rAF batching is the right shape for streaming, pin threshold respects user scrollback intent. CI green across all platforms. (Re-pointed the closing keyword to a non-closing ref since #955 is an umbrella with 10 items.) Merging.

Issue #868, opened 2026-05-15
RFC: rewrite the TUI renderer in Rust, migrate the agent core in stages


esengine:
# Summary

Rewrite the TUI renderer in Rust, and migrate the agent core to Rust in staged follow-ups. The Node/TypeScript surface shrinks to a thin launcher + slash-command authoring layer; everything that renders, parses streams, drives the event log, or talks to providers moves to a native binary.

This is an architectural bet, not a perf chase. The justification is robustness on the long tail of terminal hosts and a smaller, sharper internal surface — not "Rust is faster."

# Motivation

We have shipped credible workarounds for every individual rendering bug, but the rate of new bugs in the same family is not slowing. Recent representative cases:

- **#412** — xterm.js hosts (VSCode, Cursor, Hyper) drop our `DECSET 2026` synchronized-output envelopes on older builds. We can detect the host and degrade frame rate, but we cannot fix the host.
- **#663** — Ink + Yoga undercount visual rows on CJK terminals in alt-screen mode. The fix is a one-line patch to Ink's render path. We had to fork Ink and publish it under our own name because `patch-package` breaks for downstream consumers.
- **#565** — `App.tsx` is 3782 lines because Ink/React forces every concern into the render tree to participate in reconciliation. The staged split is in flight, but the ceiling on what we can decompose is set by Ink's component model.
- **#867** — mouse-wheel scroll injects raw SGR escape sequences into the composer because Ink does not handle mouse-tracking modes, and we have no good seam to intercept them ahead of the line editor.
- **#665** — flicker on Tabby. Same root family.

The pattern: React + Yoga + Ink is a layout engine designed for the DOM, retargeted at a cell grid. The mismatches it papers over (character width vs cells, ZWJ clusters, alt-screen vs main buffer, synchronized output, mouse reports, ConPTY input quirks) are exactly where our bugs live. Every fix is a workaround against a layer that does not model the terminal.

Secondary motivations:

- **Startup cost.** ~250–400ms of Node + npm graph load on cold start. We have not chased it because shaving npm deps is whack-a-mole; the realistic win is removing the Node dependency from the hot path entirely.
- **Install pain on Windows / Termux.** native-module rebuild prompts, npm registry hiccups, postinstall scripts. A prebuilt single binary replaces all of it.
- **Test surface.** Ink renderers are hard to assert against without spinning up the whole app. A Rust render layer fed by a serialized scene graph is trivially snapshot-testable.

[...]

PR #1007, opened 2026-05-16
feat(scene): pull more design-mock content into sidebar, ctx pane, composer

esengine:
Closes the \"left/right/composer have different things from the mock\"
gap. Builds on the layout + palette + sessions list work so far
(#1003, #1005, #1006) — this PR is purely producer-side content
additions, no schema or Rust render changes.

## Sidebar (\`SESSIONS\` → \`new chat / HISTORY\` + footer)

The mock's sidebar has, top to bottom: a prominent \`+ new chat ⌘N\`
button, a search row, sessions under \`PINNED\` / \`HISTORY\` headers,
and a footer with rules / mcp / settings rows. We can't do a real
button or a persistent text input in a scene frame, but the
structure carries over:

- Top: \`＋ new chat   ⌘N\` — accent-tinted bg, bold label, dimmed
  keybinding hint. Reads as the call-to-action even though clicking
  isn't a thing in TUI.
- \`HISTORY\` header + sessions list. \`●\` marks the active session,
  \`○\` the others. Branch meta tucked dim after the title.
- \`height: \"fill\"\` spacer pushes the footer to the bottom.
- Footer: \`⚙ settings ⌘,\` and \`◈ mcp servers  N\` rows. mcp count
  comes from \`liveMcpServers.length\` (a wired number) — rules is
  skipped this PR since we don't surface that count to scene yet.

Sessions cap is now 6 (was 8) so the new-chat row + history section
+ footer all fit in a 24-row terminal.

## Context pane (tabs above the existing kv list)

Adds a tab strip \`files  tools  memory  rules\` with \`files\` as the
active tab in bold accent. Switching between tabs isn't wired — this
is the visual hook for that follow-up. Adds an \`mcp\` row to the
existing model / cards / wallet kv list when the count is non-zero.

## Composer (mode selector + hint row)

Mock has: \`hint with kbd shortcuts → chips → textarea → mode + send
footer\`. We do:

- \`composerHintRow\`: \`⏎ send  ⇧⏎ newline  / commands  @ mention\`,
  accent glyph + muted label per pair.
- Existing composer \`❯\` row.
- \`modeSelectorRow\`: \`mode  [review] auto yolo\` with the active mode
  highlighted in a colored chip (accent / warning / danger for
  review / auto / yolo respectively — matches the trust-dial
  semantics).

\`editMode\` flows from \`useEditGate\` through \`useSceneTrace\`.

## Tests

Existing 67 scene-trace tests rewritten to query by content (helper
funcs \`composerOf\`, \`approvalRowOf\`, \`pickerSessionRows\`,
\`slashRowsOf\`) rather than fixed indices — the main pane structure
changed (added 2 rows: hint + mode), and queries by position were
the load-bearing fragile bit.

\`npm run verify\` green via prepush gate.

## What still falls short

- Real tab switching in ctx pane — needs focused tab state +
  per-tab content branches.
- Token meter (cache-hit %, used/cached/reserved gauges) — needs
  telemetry wiring.
- Chips (\`@reviewer\`, \`src/foo/bar.ts\`) above composer — needs
  separate state for parsed @-mentions vs raw buffer text.
- Real search input — can't do persistent input boxes in a single
  scene frame; needs a dedicated overlay.
- Card head/body collapse — biggest remaining gap on main pane.

Refs #868

### DIFF
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -1370,6 +1370,8 @@ function AppInner({
     walletCurrency: balance?.currency,
     sidebarSessionsJson,
     sidebarActiveSession: session ?? undefined,
+    mcpServerCount: liveMcpServers.length,
+    editMode,
   });
 
   // Ctrl+P / Ctrl+N from PromptInput route here. When any input-prefix
--- a/src/cli/ui/hooks/useSceneTrace.ts
+++ b/src/cli/ui/hooks/useSceneTrace.ts
@@ -36,6 +36,10 @@ export type SceneTraceInput = {
   sidebarSessionsJson?: string;
   /** Name of the currently-active session — highlighted in the sidebar list. */
   sidebarActiveSession?: string;
+  /** Number of live MCP servers, surfaced in the sidebar footer. */
+  mcpServerCount?: number;
+  /** Edit mode (`review` / `auto` / `yolo`) — drives the composer mode segment. */
+  editMode?: "review" | "auto" | "yolo";
 };
 
 type BuildInput = {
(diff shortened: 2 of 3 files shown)


esengine:
Closing — the design mock changed direction (single-column terminal stream per reasonix-v1, not three-pane). Restructuring under a fresh PR; reusable wiring (editMode / mcpServerCount / sidebarSessions) will be re-added there with the new layout shape.

PR #1009, opened 2026-05-16
feat(slash): add colonCompleter — /cmd:arg triggers a second, independent arg-picker

### LINKED ISSUE #1008: /skill: colon-syntax for arg completion (subcommands vs skill names)
## Problem

`/skill` currently has `argCompleter: "skills"`, which means typing `/skill ` always pops up a picker listing every installed skill name. This has two issues:

1. **Subcommands trigger the wrong picker.** Typing `/skill list`, `/skill show`, `/skill new`, or `/skill paths` still fires the skills-name picker, even though these are subcommands that should just execute directly — they don't need completion at all.
2. **No way to complete skill names intentionally.** If the user actually wants the skill-name picker (to run a skill), they have to type `/skill ` and scroll through every installed skill — there's no clean way to signal "I want to run a skill" vs "I want to run a subcommand."

## Proposed solution

Introduce a **colon separator** to disambiguate:

- `/skill ` → completes **subcommands**: `list`, `show`, `new`, `paths`
- `/skill:` → completes **skill names** (the existing skills-store list)

This is implemented via a general-purpose mechanism:

1. `SlashCommandSpec` gains an optional `colonCompleter` field — same type as `argCompleter`, but triggered by `:<arg>` after the command instead of ` <arg>`.
2. `SlashArgContext` gains a `separator` field (`" "` or `":"`) so the completion hooks can choose the right completer.
3. `detectSlashArgContext` matches both `/cmd arg` and `/cmd:arg` patterns, recording the separator.
4. `parseSlash` splits colon-separated first tokens so `/skill:weather 北京` correctly resolves to `cmd: "skill"`, `args: ["weather", "北京"]`.
5. `useCompletionPickers` stops slash-name matching on `:` (same as space), and picks `colonCompleter` vs `argCompleter` based on the separator.
6. `App.tsx` Enter-handler skips command-name substitution when `:` is present.

The `/skill` spec is updated to:
```ts
argCompleter: ["list", "show", "new", "paths"],
colonCompleter: "skills",
```

`colonCompleter` is a general mechanism — any future command can use `/cmd:arg` to trigger a different completer.

## Files touched

| File | Delta |
|------|-------|
| `src/cli/ui/slash/types.ts` | +4 (new fields) |
| `src/cli/ui/slash/commands.ts` | +38/-7 (detection, parsing, spec) |
| `src/cli/ui/useCompletionPickers.ts` | +17/-5 (separator-aware completion) |
| `src/cli/ui/App.tsx` | +2/-1 (Enter guard) |

All 3088 existing tests pass with no changes needed.

douglarek:
## Problem

`/skill` currently has `argCompleter: "skills"`, which means typing `/skill ` always pops up a picker listing every installed skill name. This has two issues:

1. **Subcommands trigger the wrong picker.** Typing `/skill list`, `/skill show`, `/skill new`, or `/skill paths` still fires the skills-name picker, even though these are subcommands that should just execute directly.
2. **No way to complete skill names intentionally.** If the user wants the skill-name picker (to run a skill), they have to type `/skill ` and scroll through every installed skill — there's no clean way to signal "I want to run a skill" vs "I want to run a subcommand."

## Solution

Introduce a **colon separator** to disambiguate:

- `/skill ` → completes **subcommands**: `list`, `show`, `new`, `paths`
- `/skill:` → completes **skill names** (the existing skills-store list)

This is implemented via a general-purpose mechanism:

1. `SlashCommandSpec` gains an optional `colonCompleter` field — same type as `argCompleter`, triggered by `:<arg>` after the command.
2. `SlashArgContext` gains a `separator` field (`" "` or `":"`) so completion hooks can choose the right completer.
3. `detectSlashArgContext` matches both `/cmd arg` and `/cmd:arg` patterns, recording the separator.
4. `parseSlash` splits colon-separated first tokens so `/skill:weather Beijing` correctly resolves to `cmd: "skill"`, `args: ["weather", "Beijing"]`.
5. `useCompletionPickers` stops slash-name matching on `:` (same as space), and picks `colonCompleter` vs `argCompleter` based on the separator.
6. `App.tsx` Enter-handler skips command-name substitution when `:` is present.

`colonCompleter` is a general mechanism — any future command can use `/cmd:arg` to trigger a different completer.

## Files

| File | Delta |
|------|-------|
| `src/cli/ui/slash/types.ts` | +4 |
| `src/cli/ui/slash/commands.ts` | +38/-7 |
| `src/cli/ui/useCompletionPickers.ts` | +17/-5 |
| `src/cli/ui/App.tsx` | +2/-1 |

All 3092 existing tests pass. Closes #1008.

### DIFF
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -2484,7 +2484,7 @@ function AppInner({
       // + Enter, one keystroke less. Skip substitution if the user
       // already typed a full, exact command name (respect verbatim
       // input when they know what they want).
-      if (text.startsWith("/") && !text.includes(" ")) {
+      if (text.startsWith("/") && !text.includes(" ") && !text.includes(":")) {
         const typed = text.slice(1).toLowerCase();
         const matches = suggestSlashCommands(typed, !!codeMode, slashUsage);
         const exact = matches.find((m) => m.cmd === typed);
--- a/src/cli/ui/slash/commands.ts
+++ b/src/cli/ui/slash/commands.ts
@@ -176,7 +176,8 @@ export const SLASH_COMMANDS: readonly SlashCommandSpec[] = [
     argsHint:
       "[list|paths|paths add <path>|paths remove <path|N>|show <name>|new <name>|<name> [args]]",
     summary: "list / run / scaffold skills (project + custom + global + builtin)",
-    argCompleter: "skills",
+    argCompleter: ["list", "show", "new", "paths"],
+    colonCompleter: "skills",
   },
   {
     cmd: "qq",
@@ -421,9 +422,16 @@ export function resolveSlashAlias(name: string): string {
   return ALIAS_TO_CMD[name] ?? name;
 }
 
-/** Picker fires only when arg tail has no internal whitespace; past that it's a usage hint. */
+/** Picker fires only when arg tail has no internal whitespace; past that it's a usage hint.
+ *  Space-separated (`/cmd arg`) vs colon-separated (`/cmd:arg`) each drive their own completer. */
 export function detectSlashArgContext(input: string, codeMode = false): SlashArgContext | null {
-  const m = /^\/(\S+) ([\s\S]*)$/.exec(input);
+  // Try space-separated first, then colon-separated.
+  let m = /^\/(\S+) ([\s\S]*)$/.exec(input);
+  let separator: " " | ":" = " ";
+  if (!m) {
+    m = /^\/(\S+):([\s\S]*)$/.exec(input);
+    separator = ":";
+  }
   if (!m) return null;
   const cmdName = resolveSlashAlias(m[1]!.toLowerCase());
   const tail = m[2] ?? "";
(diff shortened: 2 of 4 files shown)


esengine:
Thanks for tackling this — the underlying UX problem is real. But I don't want to land `colonCompleter` as-is. Two reasons:

1. **`:` isn't discoverable.** `argsHint` doesn't mention it, no other slash command uses it, users won't intuit `/skill:weather`.
2. **One-caller abstraction.** The new spec field, `SlashArgContext.separator`, dual regex in `detectSlashArgContext`, second split path in `parseSlash`, Enter-guard in `App.tsx` — all exist for `/skill` alone. Per `CLAUDE.md` we don't add infrastructure for hypothetical future callers.

**What I'd merge instead** — extend `argCompleter` with a function variant:

```ts
argCompleter?:
  | "models" | "mcp-resources" | "mcp-prompts" | "skills" | "path"
  | readonly string[]
  | ((partial: string) => readonly string[] | Promise<readonly string[]>);
```

`/skill`'s spec becomes:

```ts
argCompleter: (partial) => [
  ...["list", "show", "new", "paths"].filter(s => s.startsWith(partial)),
  ...listInstalledSkillNames().filter(n => n.startsWith(partial)),
]
```

Single picker, subcommands on top, zero new syntax, ~20 LoC net. The function form is genuinely general — any command with dynamic sources can use it.

Also: parser/completer behavior needs unit tests — this PR adds none even though it touches `parseSlash` and `detectSlashArgContext`.

Happy to merge a revised version along those lines (with tests). Otherwise I'll close and pick it up — let me know.

douglarek:
> I'll close and pick it up

yes, agree with this

PR #1015, opened 2026-05-16
refactor(scene): move the entire scene producer from TS to Rust

esengine:
Stage A of the architectural pivot the user asked for —
\"先看一下这些底层架构机制，这个一定要先搭建好\".

The JS side no longer builds a \`SceneFrame\` layout tree; it just
serializes the latest input state. Rust deserializes the state and
runs the full layout / palette / row-shape logic against the **real**
terminal area.

## Why

The TS scene producer was a recurring source of \"things don't look
right\" bugs:

- JS reads its idea of terminal size from Ink's \`useStdout()\`, which
  is the **null stream** under \`REASONIX_RENDERER=rust\` mode. Cols
  and rows fell back to 80 × 24 every render, so layout decisions
  (boot block size, card cap, dock thickness) were divorced from the
  actual terminal — even though the renderer rendered against the
  real area. Producer + renderer disagreed on dimensions.
- Two languages held the same constants (palette hexes, row counts,
  glyphs). Every visual tweak required keeping both sides in sync.
- Behavior changes required JS rebuild + Rust rebuild and both had
  to land together.

Moving the producer to Rust makes the renderer authoritative for
layout. Single source of truth. Single rebuild target.

## What

New Rust modules under \`crates/reasonix-render/src/\`:

- \`state.rs\` — \`SceneState\` / \`SceneCard\` / \`SlashMatch\` /
  \`SessionItem\` / \`SetupState\` / \`Message\` (tagged enum
  \`{ type: \"trace\" | \"setup\", ... }\`).
- \`theme.rs\` — \`palette::{bg, bg2, fg, fg1..3, ds, ds_bright,
  ds_purple, ok, warn, err}\` matching the v1 mock's oklch tokens.
- \`producer.rs\` — \`build_trace_frame(&state, cols, rows)\` +
  \`build_setup_frame(&state, cols, rows)\`. Ports the boot block
  (REASONIX ASCII banner), card kinds (\`user\` / \`reasoning\` /
  \`streaming\` head+body, \`tool\` rich row, generic single-line for
  others), composer / meta / status dock, slash overlay, sessions
  picker, approval modal — all from the old TS producer.

\`main.rs\` reads messages from stdin (tries \`Message\`, then bare
\`SceneState\`, then \`SetupState\`, then legacy \`SceneFrame\` for
back-compat). Each frame calls \`terminal.size()\` to get the real
cols/rows and feeds them into the producer.

## JS side

\`useSceneTrace.ts\` shrunk from ~700 LOC to ~270 LOC — purely state
shaping now. The hook builds a plain object
(\`{ type: \"trace\", ...state }\`) and emits via the renamed
\`emitSceneMessage(message: unknown)\`. Same for \`useSetupSceneTrace\`.

Helpers that survived (still used to serialize wire-format payloads):
\`toSceneCard\`, \`parseRecentCards\`, \`parseSlashMatches\`,
\`parseSessions\`, \`summarizeCard\`.

Gone (moved to Rust): \`buildTraceFrame\`, \`buildSetupFrame\`, all the
row builders, the kind-glyph / color / label tables, \`PALETTE\`,
\`listWindow\`, \`slashWindow\`, \`cardsForHeight\`.

## Tests

- Rust (\`crates/reasonix-render/tests/producer.rs\`) — 14 new cases
  covering boot block, card kinds, composer cursor, dock rows,
  approval / slash / sessions overlays, status bar segments, edit
  mode color, rich tool format, setup frame, body line cap.
- TS (\`tests/scene-trace-frame.test.ts\`) — pared down to the 20
  wire-format tests that still apply (\`summarizeCard\`,
  \`toSceneCard\`, \`parseRecentCards\`, \`parseSlashMatches\`,
  \`parseSessions\`).

\`cargo test -p reasonix-render\` — 51 / 51 green.
\`npm run verify\` — 3071 / 3074 green via prepush gate.

## Migration notes

- Rust binary **must** be rebuilt for users running off the prebuilt
  \`target/release/reasonix-render.exe\` — the new producer is the
  binary's job now. \`cargo build --release --bin reasonix-render\`.
- The legacy \`SceneFrame\` JSON shape is still accepted by Rust as a
  fourth-tier fallback, so a stale JS bundle paired with a fresh
  Rust binary keeps working until both roll forward.

Refs #868

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,16 +9,37 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
+name = "android_system_properties"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "anyhow"
 version = "1.0.102"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
+name = "autocfg"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
+
+[[package]]
 name = "bitflags"
 version = "2.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c4512299f36f043ab09a583e57bceb5a5aab7a73db1805848e8fef3c9e8c78b3"
+
+[[package]]
+name = "bumpalo"
+version = "3.20.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5d20789868f4b01b2f2caec9f5c4e0213b41e3e5702a50157d699ae31ced2fcb"
 
 [[package]]
 name = "cassowary"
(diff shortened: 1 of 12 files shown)


esengine:
Holding per user request — found a likely root cause (rust renderer missing enable_raw_mode, which can leave the bottom row unfilled). Investigating locally; will reopen after verification.

PR #1017, opened 2026-05-16
feat: transparent rtk integration for token-optimized command output

wul7chaos:
Integrate rtk (Rust Token Killer) as a transparent output filter layer for `run_command`. When rtk is available, every shell command is automatically routed through `rtk <original-command>`, applying built-in output filters (git, cargo, npm, pytest, ls, etc.) to reduce LLM context consumption by 60-90% on CLI output.

## How it works
- `prepareSpawn` accepts a new `wrapCommand` option
- When set, the original argv is transparently wrapped: `["git","status"]` → `["rtk","git","status"]`
- The allowlist and confirmation gates still check the **original** command — rtk is an invisible execution layer
- Works for both simple and chain commands (`|`, `&&`, `||`)
- Auto-detects rtk binary: `bin/rtk` → `$PATH` → project-local build
- Disabled for background jobs (`run_background`) where filtering would interfere
- Zero behavioral change when rtk is not installed

## Benchmark
`git status` benchmark shows **45.7% character reduction** (1419 → 770 chars). Estimated savings across a typical session: 60-80%.

## Files changed
- `src/tools/shell/exec.ts` — core: `prepareSpawn` wrapCommand
- `src/tools/shell-chain.ts` — chain command passthrough
- `src/tools/shell.ts` — rtk auto-detection + wiring

### DIFF
--- a/src/tools/shell-chain.ts
+++ b/src/tools/shell-chain.ts
@@ -293,6 +293,8 @@ export interface RunChainOptions {
   timeoutSec: number;
   maxOutputChars: number;
   signal?: AbortSignal;
+  /** Prepend this wrapper to each segment (e.g. "rtk" for token filtering). */
+  wrapCommand?: string;
 }
 
 export async function runChain(chain: CommandChain, opts: RunChainOptions): Promise<ChainResult> {
@@ -314,6 +316,7 @@ export async function runChain(chain: CommandChain, opts: RunChainOptions): Prom
       timeoutMs: remainingMs,
       buf,
       signal: opts.signal,
+      wrapCommand: opts.wrapCommand,
     });
     lastExit = result.exitCode;
     if (result.timedOut) {
@@ -340,6 +343,7 @@ interface PipeGroupOptions {
   timeoutMs: number;
   buf: OutputBuffer;
   signal?: AbortSignal;
+  wrapCommand?: string;
 }
 
 interface SegmentStdio {
@@ -419,7 +423,9 @@ async function runPipeGroup(
       const seg = segments[i]!;
       const io = openRedirects(seg.redirects, opts.cwd);
       allFds.push(...io.toClose);
-      const { bin, args, spawnOverrides } = prepareSpawn(seg.argv);
+      const { bin, args, spawnOverrides } = prepareSpawn(seg.argv, {
+        wrapCommand: opts.wrapCommand,
+      });
       const stdoutSpec = io.stdoutFd !== null ? io.stdoutFd : "pipe";
       const stderrSpec =
         io.stderrFd !== null ? io.stderrFd : io.mergeStderrToStdout ? stdoutSpec : "pipe";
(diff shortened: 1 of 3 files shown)


wul7chaos:
## 关于 rtk 二进制分发的补充说明

当前 PR 仅包含代码层面的集成（3 个文件），rtk 本身的二进制需要用户自行安装：

### 各平台安装方式

| 平台 | 推荐安装方式 |
|------|------------|
| **macOS** | `brew install rtk` |
| **Linux** | `cargo install rtk` 或下载 GitHub release |
| **Windows** | `cargo install rtk` 或下载 release 二进制放 PATH |

### 安装后如何使用

只要 rtk 在系统 PATH 上，Reasonix 会自动检测并透明启用——所有 `run_command` 调用自动路由通过 rtk 过滤输出，无需额外配置。

### 如果未安装 rtk

自动检测返回 null，行为与集成前完全一致，无任何副作用和性能开销。

### 后续可以改进的方向

1. **npm optionalDependencies**：在 `package.json` 中添加各平台 rtk 二进制作为可选的 npm 包，`postinstall` 自动部署到 `bin/` 目录
2. **`reasonix setup-rtk` 子命令**：跨平台自动下载对应 release 二进制
3. **与 rtk 项目合作**：让 rtk 发布 npm 包以便 `npx rtk` 直接可用

这些改进可以作为后续 PR 跟进，当前纯代码改动先合并，让社区知道这个功能的存在。

esengine:
Thanks for the writeup, but I'm going to close this. Three blockers, in priority order:

**1. Wrong cost lever for this project.** Reasonix's "cheap" positioning runs on cache hits, not output truncation. After #978 v4-flash cache-hit input is `$0.0028/1M` — cached input is effectively free. A 45% cut on tool output saves cents per session against a real risk of dropping lines the agent actually needs (`modified: foo.ts`, a specific error message, untracked files). We've already won the token budget via caching; filtering tool output is premature optimization for a budget we don't need.

**2. Security footgun.** Auto-resolving `rtk` from `$PATH` and silently prepending it to every `run_command` means any other binary named `rtk` (legitimate or planted by a malicious npm package) becomes an invisible execution wrapper. The allowlist and confirm gate still display the *original* command, so the user has no signal the call was intercepted. The `../rtk/rtk-install/bin` fallback assuming a sibling clone is also too presumptuous to ship.

**3. Speculative infrastructure for one unproven external tool.** `wrapCommand` becomes a permanent generic "wrap arbitrary command with arbitrary binary" hook in `prepareSpawn`, threaded through three files, for a Rust binary that isn't on your GitHub profile. CLAUDE.md is explicit about not adding infrastructure for hypothetical future callers.

Minor code smells: duplicated "Apply command wrapper..." comment in `exec.ts` (written twice in a row), mixed Chinese/English inline comments, oversized comment block in the rtk resolver.

If you want to ship rtk as an opt-in user-side wrapper (alias / shell function / custom skill), that's a fine experiment — but it shouldn't live inside Reasonix as a transparent execution layer.

Issue #949, opened 2026-05-15
计算花费似乎是错误的，大了一个量级


futianhua:
**Reasonix**: 0.43.0 (latest)
**Platform**: win32 (10.0.19045)
**Terminal**: (unknown)
**Size**: 237×63
**Node**: v24.15.0
**Locale**: zh-CN
**Theme**: default
**Model**: deepseek-v4-flash · effort=max
**Mode**: edit=auto · plan=off
**MCP**: 0 server(s)
**Session**: code-Vibe_Coding

显示的花费是实际使用的10倍左右

kabaka9527:
Because under zh-cn, only the currency symbol was converted, not the amount

esengine:
Confirmed — real bug, already fixed on `main`, just not released yet.

**Root cause:** `src/telemetry/stats.ts` shipped v4-flash's `inputCacheHit` constant as `0.028` USD/1M when it should be `0.0028` — a misplaced decimal, 10× too high. Since most tokens in a real session go through the cache-hit path, displayed cost ends up ~10× the actual bill, which matches what you observed.

**Fix:** #978 corrected the constants to match DeepSeek's official pricing page exactly — v4-flash `0.0028 / 0.14 / 0.28` USD/1M, v4-pro `0.003625 / 0.435 / 0.87` (current 75%-off promo).

**Release status:** 0.43.1 patch hasn't been cut yet. Users on `npm i reasonix@0.43.0` still see the inflated number. Closing this once 0.43.1 ships; tracking via #978 in the meantime.

Thanks for the report — wide-impact bug.

PR #1018, opened 2026-05-16
Fix CNY symbol from ¥ to $ in tests and constants

kabaka9527:
## What

This change updates the CNY display symbol in the CLI UI from ¥ to $ by adjusting the shared currency formatting used by balance and cost rendering. It also updates the affected UI tests so their expectations match the new CNY output in 

## Why

The requested behavior change was to show $ instead of ￥/¥ in the zh-cn experience. After tracing the rendering path, the symbol source was confirmed to be formatting logic in [SYMBOL] not translation text in src/i18n/zh-CN.ts. Because of that, the correct fix was to update the currency-symbol mapping and then align the directly affected tests with the new behavior.

## How to verify

Run the affected tests, including tests/ui-usage-card-balance.test.tsx
Confirm CNY balance/cost output now renders with $ in the CLI UI.
Confirm no remaining assertions expect ¥ for CNY in the updated UI formatting tests.

## Checklist

### DIFF
--- a/src/cli/ui/theme/tokens.ts
+++ b/src/cli/ui/theme/tokens.ts
@@ -360,7 +360,7 @@ export type CardTone = keyof ThemeTokens["card"];
 /** DeepSeek prices in CNY; our internal table is USD divided by 7.2. Multiply back for display. */
 export const USD_TO_CNY = 7.2;
 
-const SYMBOL: Record<string, string> = { USD: "$", CNY: "¥" };
+const SYMBOL: Record<string, string> = { USD: "$", CNY: "$" };
 
 /** Format an amount already in `currency`. Undefined currency → CNY (matches pre-fix behavior). */
 export function formatBalance(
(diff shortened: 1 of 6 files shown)


kabaka9527:
#949 issues

esengine:
Thanks for taking a look at #949, but I'm going to close this — the diagnosis here isn't quite right and the fix would make the bug worse.

**Real root cause:** the v4-flash `inputCacheHit` constant in `src/telemetry/stats.ts` shipped as `0.028` USD/1M when it should be `0.0028` — a misplaced decimal that inflates cost ~10× for users with high cache-hit ratios (which is most real sessions). Already fixed on `main` in #978: constants now match DeepSeek's official USD pricing page exactly, and `formatCost` correctly converts USD → CNY via the 7.2× factor at the UI boundary.

**Why this PR isn't the fix:** `formatCost` already converts the amount (`costUsd * USD_TO_CNY`). Changing the CNY symbol to `$` doesn't undo a missing conversion — there is no missing conversion. It only relabels a CNY value as USD, so a user billed `¥0.22` would now read `$0.22` on screen, which is ~7× higher than the real USD figure. That's a worse misread than the original bug.

Appreciate the effort tracing the symbol path. The pricing fix is on `main` already and will go out in the next release.

PR #1035, opened 2026-05-16
[GTC] fix(tokenizer): bound pathological token counting

### LINKED ISSUE #558: Audit `countTokens` callers — add upper-char bound to defend non-MCP paths from pathological BPE input
## Summary

A user report flagged the pure-TS BPE tokenizer as O(n²) on
pathological repetitive input (`AAAA…`) and claimed multiple call
paths could hang for 30s+. Most of the report misread our code —
specifically, the `Never tokenizes full input — pathological
repetitive text (AAAA…) costs 30s+...` comment at
`src/mcp/registry.ts:152` is **defense documentation**, not a bug
admission. `truncateForModelByTokens` is deliberately designed to
never feed full input into BPE; `sizePrefixToTokens` only tokenizes
budget-bounded slices with a 6-iter cap. That whole MCP truncation
path is already safe.

But there's a narrower real surface that the report stumbled into,
worth fixing.

## The actual exposure

Three call sites tokenize **full message content** with only a
lower-bound short-circuit (`length <= maxTokens` skips), and no upper
char cap:

1. **`src/loop/shrink.ts:52`** — `shrinkOversizedToolResultsByTokens`:
   ```ts
   if (content.length <= maxTokens) return msg;
   const beforeTokens = countTokens(content);  // full content
   ```
2. **`src/loop/shrink.ts:83`** — same pattern for `tool_calls.arguments`.
3. **`src/tokenizer.ts:258` `estimateConversationTokens`** — preflight
   sums `countTokens(m.content)` over every message in the
   conversation; no per-message char cap.

Most of the time these are safe because upstream caps (MCP default
`maxResultChars: 8000`, subagent default 8000) keep individual
strings small. But:

- Non-MCP tools (`read_file`, shell stdout) don't share the MCP cap —
  a `read_file` of a 200KB log file lands in the message log at full
  size, and the next preflight tokenizes the whole thing.
- A repetitive payload (CSV columns, log timestamps, base64 chunks)
  hits the BPE inner loop's worst case where merges keep finding
  matches across the full string.

Worst case isn't 30s on realistic inputs — that number was
speculation in the report — but tokenizing 200KB of moderately
repetitive content on the pure-TS port does take seconds, and it
runs synchronously on the main thread. That's enough to:

- Stall the loop's preflight
- Make `decidePreflight` (`src/context-manager.ts:107`) feel like a
  freeze on long sessions
- Compound when shrink + preflight + estimate all touch the same
  oversized message in one turn

## Proposed fix

Add a bounded-tokenize helper alongside `countTokens`:

```ts
// Returns an exact count when feasible, an estimate for oversized
// inputs. Never tokenizes more than `maxChars` of the input.
export function countTokensBounded(text: string, maxChars: number): number;
```

Implementation: when `text.length <= maxChars`, just delegate to
`countTokens`. Otherwise tokenize a head + tail sample (same shape as
`registry.ts:170-182` already uses for size estimation) and scale by
char/token ratio. The math is already in-tree at `registry.ts:180`:

```ts
const ratio = sampleChars > 0 ? sampleTokens / sampleChars : 0.3;
const estTotalTokens = Math.ceil(s.length * ratio);
```

Then swap the three sites above to call `countTokensBounded` with a
sane cap (e.g. 32KB — well above any realistic single message that
matters for a token-budget decision, well below the slow zone).

For `shrinkOversizedToolResultsByTokens` the cap is fine to be loose
because the function's job is just deciding "is this over budget";
an estimate within ±10% is enough to gate the truncation pass.

## Out of scope

- Replacing the pure-TS BPE port with a native binding. The TS port
  is "good enough for budgeting decisions" by design; native is a
  separate, much larger change.
- Caching `countTokens` results. Useful but orthogonal.
- The MCP truncation path. Already defended; do not touch.

## Credit

Original report flagged the right family of risk even though the
specific claims were wrong. The 30s number isn't real on observed
inputs and the cited comment was misread, but the underlying "we
sometimes tokenize unbounded user input on the main thread" exposure
is genuine and worth closing.

GTC2080:
## What

Adds bounded token counting for oversized text and routes the hot token-estimation paths through it: request estimation, context shrink ordering, `/context` breakdowns, desktop context telemetry, `/cost`, streaming token-rate estimates, and subagent distillation metrics. This keeps exact counting for normal-sized inputs while sampling pathological large strings.

## Why

Fixes #558. A reproduction is a very large repetitive payload such as `"A".repeat(100_000)` flowing into token-counting paths; exact BPE counting can spend multiple seconds there before shrink/healing gets a chance to reduce the content. The bounded counter avoids that stall while preserving a proportional estimate for context budgeting and UI telemetry.

## How to verify

`npm run verify`

Regression coverage added for `countTokensBounded("A".repeat(100_000))` and for `/context` breakdown over a 100KB tool result, both asserting the pathological path stays under 1 second.

## Checklist

### DIFF
--- a/src/cli/commands/desktop.ts
+++ b/src/cli/commands/desktop.ts
@@ -61,7 +61,7 @@ import {
 } from "../../memory/session.js";
 import { MemoryStore } from "../../memory/user.js";
 import { SkillStore } from "../../skills.js";
-import { countTokens } from "../../tokenizer.js";
+import { countTokensBounded } from "../../tokenizer.js";
 import type { ChoiceOption } from "../../tools/choice.js";
 import type { ChatMessage } from "../../types.js";
 import { VERSION } from "../../version.js";
@@ -548,8 +548,8 @@ function emitMemory(tab: Tab): void {
 function emitCtxBreakdown(tab: Tab): void {
   if (!tab.runtime) return;
   try {
-    const sys = countTokens(tab.runtime.loop.prefix.system);
-    const tools = countTokens(JSON.stringify(tab.runtime.loop.prefix.toolSpecs));
+    const sys = countTokensBounded(tab.runtime.loop.prefix.system);
+    const tools = countTokensBounded(JSON.stringify(tab.runtime.loop.prefix.toolSpecs));
     emit({ type: "$ctx_breakdown", reservedTokens: sys + tools }, tab.id);
   } catch {
     // tokenizer warmup can throw on first call before the data file loads
--- a/src/cli/ui/cards/StreamingCard.tsx
+++ b/src/cli/ui/cards/StreamingCard.tsx
@@ -2,7 +2,7 @@ import { Box, Text, useStdout } from "ink";
 import React, { useContext } from "react";
 import { clipToCells } from "../../../frame/width.js";
 import { t } from "../../../i18n/index.js";
-import { countTokens } from "../../../tokenizer.js";
+import { countTokensBounded } from "../../../tokenizer.js";
 import { LiveExpandContext } from "../layout/LiveExpandContext.js";
 import { useReserveRows } from "../layout/viewport-budget.js";
 import { Markdown } from "../markdown.js";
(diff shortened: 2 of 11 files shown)


GTC2080:
Withdrawing this PR for now.

PR #1040, opened 2026-05-16
fix: MCP bridge state — record tools immediately, use live getter for…

Oush1N:
## 问题

MCP 管理面板（`/mcp` 和 Dashboard）中，已成功启动的 MCP 服务器（工具已注册、可正常调用）仍显示为"未桥接"。

### 第一层：`addSpec` 过早判定失败（`mcp-runtime.ts`）

`bridgeMcpTools` 成功后工具已注册到 ToolRegistry，但 `records.set()` 在 `inspectMcpServer` 和 `loop.prefix.addTool` 之后才执行。若后两步中任一步抛出异常，整个 spec 被标记为失败，即使工具已可用。`rejectReady` 也会被错误调用。

### 第二层：Dashboard API 读取冻结引用（`context.ts` → `App.tsx`）

`DashboardContext.mcpServers` 是静态数组引用，在上下文创建时冻结。其他实时状态（`planMode`、`editMode` 等）已使用 getter 函数模式，唯独 MCP servers 没有。Dashboard 和 `/api/mcp` 读到的一直是启动时的空数组。

### `src/cli/commands/mcp-runtime.ts`

- `bridgeMcpTools` 成功后**立即**创建 provisional record（`records.set` + `insertionOrder.push` + `resolveReady`），不等 inspect 完成
- `inspectMcpServer` 降级为非关键步骤，失败用 fallback 不影响桥接状态
- `loop.prefix.addTool` 用 `try/catch` 包住，失败不回滚
- catch 块中检查 `records.has(raw)`：若已有 provisional record 则保留，不关闭 MCP client

### `src/server/context.ts`

- `mcpServers?: McpServerSummary[]` → `getMcpServers?: () => McpServerSummary[]`

### `src/server/api/mcp.ts` / `src/server/api/overview.ts`

- `ctx.mcpServers` → `ctx.getMcpServers?.()`

### `src/cli/ui/App.tsx`

- 添加 `liveMcpServersRef`，赋值改为 `getMcpServers: () => liveMcpServersRef.current`

## 测试

- 使用带有 2 个 MCP 服务器的 config.json 启动
- 终端 lifecycle 日志显示 `✓ 已连接`
- `/mcp` 面板和 Dashboard 均正确显示"已桥接"状态

## 已知限制

`server-filesystem` 传入多个路径（`C:\ D:\ E:\`）时，子进程无法启动，面板仍显示"未桥接"。此为 config.json 中 MCP 规范格式问题（`server-filesystem` 只接受单个目录参数），与本次修复无关。不影响功能——Reasonix 内置文件系统工具已覆盖所有文件操作。

### DIFF
--- a/src/cli/commands/mcp-runtime.ts
+++ b/src/cli/commands/mcp-runtime.ts
@@ -10,7 +10,7 @@ import { buildMcpServerSummary } from "../../mcp/summary.js";
 import { buildTransportFromSpec } from "../../mcp/transport-from-spec.js";
 import type { ToolRegistry } from "../../tools.js";
 import type { ToolSpec } from "../../types.js";
-import { formatMcpLifecycleEvent } from "../ui/mcp-lifecycle.js";
+import { type McpLifecycleEvent, formatMcpLifecycleEvent } from "../ui/mcp-lifecycle.js";
 import { formatMcpSlowToast } from "../ui/mcp-toast.js";
 import type { McpServerSummary } from "../ui/slash.js";
 
@@ -50,7 +50,9 @@ export type McpLifecycleNotice =
     }
   | { kind: "disabled"; name: string }
   | { kind: "failed"; name: string; reason: string }
-  | { kind: "slow"; serverName: string; p95Ms: number; sampleSize: number };
+  | { kind: "slow"; serverName: string; p95Ms: number; sampleSize: number }
+  | { kind: "tools-ready"; name: string; tools: number; ms: number }
+  | { kind: "warn"; name: string; reason: string };
 
 export type McpLifecycleSink = (notice: McpLifecycleNotice) => void;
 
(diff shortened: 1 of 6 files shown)


esengine (changes requested):
Thanks @Oush1N — nice diagnosis. The two-layer root cause is real: `records.set` ordering in `mcp-runtime.ts` and the frozen `mcpServers` array in `DashboardContext` were both wrong, and the second-layer fix (`getMcpServers: () => liveMcpServersRef.current`) matches the existing `getPlanMode` / `getEditMode` getter pattern exactly. That half I'd take as-is.

The runtime-side fix needs three adjustments before I can merge:

1. **Don't emit `failed` after a provisional record exists.** The new catch block still calls `sink({ kind: "failed", name: label, reason })` even when `records.has(raw)` is true, so the lifecycle log shows `✓ connected` followed by `✗ failed` for the same server while `/mcp` and the Dashboard show it as bridged. Either skip the emit when `records.has(raw)`, or introduce a `kind: "warn"` for "bridged but inspection degraded".

2. **Don't silently swallow `loop.prefix.addTool` failures.** The new

   ```ts
   try { loop.prefix.addTool(s); } catch { /* non-critical */ }
   ```

   violates CLAUDE.md's "log + crash > silent wrong output". If `addTool` fails the tool won't reach the next turn even though we just reported "bridged" — that's the silent-wrong-output case. Surface it through the same `sink` channel (e.g. a `warn` event) or let it propagate.

3. **Re-emit the `connected` event after inspection** (or update it). The provisional emit hard-codes `resources: 0, prompts: 0`, so the lifecycle log now under-reports counts even when inspection later succeeds. A second `sink({ kind: "connected", ..., resources: resourceCount, prompts: promptCount })` after `inspectMcpServer` returns would close the gap — or change the provisional event to a different `kind` so the post-inspect one is the canonical "connected".

Unrelated nit for next time: this PR was pushed from your own `main` branch. Topic branches (e.g. `fix/mcp-bridge-state`) make it easier on your end if `upstream/main` moves while we're iterating — no action needed here.

Once those three are in, I'll merge.

Oush1N (commented):
三个问题已按你的要求改完：

1.catch 块会先判断 records.has(raw)，已有 provisional record 时发 warn 而非 failed
2.addTool 失败不再静默吞掉，通过 sink({ kind: "warn", ... }) 输出工具名和错误原因
3.临时事件改成 kind: "tools-ready"，inspect 成功后再发正式的 connected（带真实 resources/prompts 数量）

  新加了 tools-ready 和 warn 两种 lifecycle 状态，McpLifecycleNotice、McpLifecycleEvent、stderrLifecycleSink 和 App.tsx 的 TUI sink 都已同步处理。

PR #1060, opened 2026-05-17
fix(loop): avoid stalled context fold and unsafe preflight compaction

### LINKED ISSUE #561: Add timeout to context fold + replace emergency preflight fold with mechanical truncation
## Summary

Context folding (`src/context-manager.ts:122` `fold()` →
`summarizeForFold` at line 186) makes a synchronous LLM call to
`deepseek-v4-flash` inside the main turn loop. Two real problems:

1. **No timeout.** `summarizeForFold` only wires
   `signal: getAbortSignal()` (the user's Esc). If the API hangs
   network-level (TCP stuck, slow socket), the turn is stuck until
   the user manually aborts. The `try/catch` at line 210 handles
   error-on-return cases gracefully, but not no-response cases.

2. **Emergency preflight path adds latency on top of a request the
   user is already waiting for.** `loop.ts:610-633` runs
   `decidePreflight`; if `> 95%` it `await fold()` *before* sending
   the user's request. Worst-case UX: user sees nothing for the 2-5s
   the summary call takes, then the request finally goes out.

The post-turn fold path (the more common one, triggered by
`decideAfterUsage` at >50% / >70% / >80%) is correctly placed —
there's no user waiting on a response when it runs, and the
semantic compaction is worth its cost. Keep that path as-is.

## Verified

- `fold()` at `context-manager.ts:122-167` calls
  `summarizeForFold` at line 151.
- `summarizeForFold:186-213` chats with `deepseek-v4-flash`,
  reasoning effort `high`, no explicit timeout.
- `loop.ts:618` `await this.context.fold(this.model)` — emergency
  preflight path.
- `loop.ts:632-644` already handles `result.folded === false`
  gracefully (warns + lets the request go).

### A. Add a timeout to `summarizeForFold` (universal)

Wrap the chat call in `Promise.race` with a 15s deadline:

```ts
const TIMEOUT_MS = 15_000;
const resp = await Promise.race([
  this.deps.client.chat({ ... }),
  new Promise<never>((_, rej) =>
    setTimeout(() => rej(new Error("fold-timeout")), TIMEOUT_MS),
  ),
]);
```

The existing catch at line 210 already returns `""` on any failure,
which makes `fold()` return a `noop` `FoldResult` (`folded: false`).
Callers — both `loop.ts:618` and the post-turn path — already
handle `folded: false` correctly. Fail-open by construction.

The 15s number is a starting point; flash on a long-history summary
is typically 1-3s, so 15s gives ample headroom while still bounding
the worst case. Tune later if needed.

### B. Mechanical truncation for the preflight emergency path

Add a new method `ContextManager.mechanicalTruncate(model)`:

- Drop oldest messages (keep `system` + tail) until the estimated
  token count is under a target (e.g. 70% `ctxMax`).
- Pure local operation: no LLM call.
- Returns the same `FoldResult` shape so `loop.ts:619-644` can
  consume it without branching.

Swap `loop.ts:618` from `await this.context.fold(...)` to
`this.context.mechanicalTruncate(...)`. The user no longer waits
2-5s before their request goes out; they trade semantic continuity
of older history for latency. That's the right tradeoff at >95% —
the user is already in an emergency context state.

The post-turn fold (`decideAfterUsage` → `kind: "fold"`) is
unaffected. It still runs the LLM summary, because:
- No user waiting on the call (it runs after the response, before
  the next prompt).
- Semantic compaction is worth the wait when not blocking.
- Keeping it preserves the cache-friendly long-session shape that
  the project relies on.

## Out of scope

- **True async / background fold.** Considered. Rejected: `fold()`
  mutates `log` and persists via `rewriteSession`. Concurrent reads
  from the next turn would race; the next turn could read a
  half-folded state. Solving that needs locks or double-buffering,
  which is a much larger change. Revisit if A + B prove
  insufficient.
- **Removing fold entirely.** Not an option — long sessions depend
  on it for cache + cost discipline.
- **Replacing flash with an even cheaper local summarizer.** Out of
  scope for this issue; orthogonal to latency.

## Acceptance

- `summarizeForFold` enforces a 15s timeout; on timeout, fold
  returns `folded: false` and the existing warning path fires.
- Preflight emergency (`>95%`) no longer makes an LLM call — it
  takes the mechanical-truncation branch and returns within ms.
- Post-turn fold path behavior unchanged.
- Existing tests for `decidePreflight` / `fold` still pass; new
  tests for timeout fail-open and mechanical truncation added.

## Why now

Self-noticed via reviewing a user report. The hang risk (no
timeout) is the most concerning of the two — it's a foot-gun that
only surfaces under network anomalies, which are exactly the cases
where users least want a frozen UI. The emergency-path latency is
secondary but cheap to fix in the same PR.

GTC2080:
## What

Adds a bounded context-fold summary path and switches the emergency preflight path to local mechanical truncation. The truncation now accounts for rendered message cost, including tool-call arguments, reruns the preflight estimate after compaction, and preserves the active tool turn when a large tool result cannot safely fit under the target.

## Why

Fixes #561. A semantic fold during preflight could stall the user's turn, and a naive mechanical truncation could either miss large tool-call arguments or drop the active user/tool context before the follow-up request.

## How to verify

- `npm run verify`

## Checklist

### DIFF
--- a/src/context-manager.ts
+++ b/src/context-manager.ts
@@ -12,7 +12,11 @@ import {
   DEFAULT_CONTEXT_TOKENS,
   type SessionStats,
 } from "./telemetry/stats.js";
-import { countTokensBounded, estimateRequestTokens } from "./tokenizer.js";
+import {
+  countTokensBounded,
+  estimateConversationTokens,
+  estimateRequestTokens,
+} from "./tokenizer.js";
 import type { ChatMessage } from "./types.js";
 
 /** Auto-fold when a turn's response shows promptTokens above this fraction of ctxMax. */
@@ -29,6 +33,10 @@ export const HISTORY_FOLD_MIN_SAVINGS_FRACTION = 0.3;
 export const FORCE_SUMMARY_THRESHOLD = 0.8;
 /** Local preflight estimate above this fraction trips the emergency in-place compact path. */
 export const PREFLIGHT_EMERGENCY_THRESHOLD = 0.95;
+/** Emergency preflight target after local truncation, as a fraction of ctxMax. */
+export const PREFLIGHT_MECHANICAL_TARGET_FRACTION = 0.7;
+/** Hard deadline for semantic fold summaries so a hung request cannot stall the turn loop. */
+export const HISTORY_FOLD_SUMMARY_TIMEOUT_MS = 15_000;
 /** Prepended to fold summary content so the model knows it's a synthesized recap. */
 export const HISTORY_FOLD_MARKER =
   "[CONVERSATION HISTORY SUMMARY — earlier turns folded for context efficiency]\n\n";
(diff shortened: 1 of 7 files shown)


esengine:
This is exactly the surgery #561 asked for, and the implementation is tighter than I'd have written it.

Two things I want to call out as load-bearing:

- **Fold timeout uses a real abort, not just a timer.** `foldCtrl.abort()` is wired into the `client.chat({ signal: foldCtrl.signal })` call, so when the 15s deadline (or turn-Esc) fires the in-flight fetch actually unwinds — no orphaned promise sitting on a hung socket waiting for GC. The `finally` cleans up the timeout and the abort listener regardless of which leg of the `Promise.race` won, so we're not leaking either side on the happy path.
- **`mechanicalTruncate` accounts for tool-call argument tokens.** Using `estimateConversationTokens` per message (the same path the actual request uses) is the right move — a naive char-count would have missed the 600-token JSON arg case the second new test covers, which is exactly the scenario where preflight has to fire in the first place. And re-running `decidePreflight` after truncation, plus the `preflightTruncatedStillFull` warning when the tail still won't fit, means the user gets honest feedback instead of a silently-still-too-big request hitting a 400.

The `allowEmpty: pendingUser !== null` / latest-user-boundary fallback also correctly distinguishes "fresh user turn coming, safe to nuke the tail entirely" from "we're mid tool-loop, must preserve the active user+tool messages so the follow-up request isn't malformed" — the fourth new test pins that invariant against regressions.

Will pull locally, `npm run verify`, and merge.

Issue #947, opened 2026-05-15
Tracking: Rust rewrite (TUI renderer → agent core), staged


esengine:
## Why

#868 has the full design and motivation. Short version: React + Yoga + Ink is the wrong model for a terminal renderer, and every workaround we ship costs more than it earns. Migration is staged, reversible, kill-switched — if any stage misses its acceptance gate, we stop and reassess.

This issue is the execution log. Design discussion lives on #868.

### Stage 0 — landing pad (no Rust)

- [x] Scene-graph contract (`src/cli/ui/scene/types.ts`) — #946
- [ ] Finish #565 — App.tsx ≤ 500 lines (Phase 1 done; 2–4 in flight)
- [ ] Lowering pass: Ink render tree → `SceneFrame`, no behavior change, snapshot-tested
- [ ] Frame stdio harness — JS emits frames on a flag, no consumer yet

**Gate:** scene-graph contract + lowering exist; App.tsx ≤ 500 lines; snapshot tests cover the card-stream view.

### Stage 1 — Rust renderer behind a flag

- [ ] `reasonix-render` Rust crate skeleton (cargo workspace at repo root)
- [ ] ratatui consumer for `SceneFrame` over stdio
- [ ] `REASONIX_RENDERER=rust` wires JS lowering → Rust process
- [ ] Card-stream parity on Windows Terminal / iTerm2 / VSCode / Linux ssh
- [ ] #867 (mouse-wheel SGR leak) fixed as a side effect

**Gate:** parity with Ink on the card-stream view across the four host configurations; #867 closed.

### Stage 2 — Rust input layer

- [ ] Composer input → Rust
- [ ] Key chords + paste bracketing + mouse → Rust
- [ ] JS receives semantic events (`submit`, `history-prev`, `approve`), not raw bytes
- [ ] `Home` / `End` / `Ctrl+U` regressions fixed (#20)

**Gate:** every existing key chord works; no UX regressions; #20 closed.

### Stage 3 — flip default, deprecate Ink

- [ ] `REASONIX_RENDERER=ink` opt-in for one minor release
- [ ] Retire Ink fork (#663)
- [ ] Delete Ink renderer path

**Gate:** two weeks on `main` Rust-default, no severity-1 render regressions.

### Stage 4 — migrate agent core

- [ ] Event log → Rust
- [ ] Streaming parser → Rust
- [ ] Tool dispatcher → Rust
- [ ] MCP client → Rust
[...]

PR #1070, opened 2026-05-17
fix(rust): zero-width char handling + tab expansion + Unicode rendering tests

Bernardxu123:
## Problem

\`paint.rs\` used \`UnicodeWidthChar::width(ch).unwrap_or(1)\` — control characters (U+0000-U+001F, U+007F-U+009F) return \`None\` from unicode-width, so they were given a spurious width-1 cell. This caused zero-width characters to overwrite the next character's cell, producing garbled output for text containing combining marks, ZWJ/ZWNJ, variation selectors, or control codes. Tabs were also rendered as single-width unknown glyphs.

## Fix

1. **\`char_width()\`** — new public helper returning 0 for \`None\` (control chars) and combining marks, instead of the old \`unwrap_or(1)\`.
2. **\`paint()\`** — bails early on w==0 characters (combining marks, ZWJ, control codes).
3. **\`paint_str_to()\`** — expands \`\t\` to 2 spaces, skips all w==0 chars, uses \`char_width()\`.
4. **\`overlay_at.rs\`** — switched from \`unwrap_or(1)\` to \`char_width()\`.

## Tests (6 new)

| Test | Covers |
|------|--------|
| \`emoji_characters_consume_two_cells\` | 🤖🚀 → 4 cells, no overwrite |
| \`zero_width_chars_are_skipped_not_overwritten\` | ZWJ (U+200D) skipped, 'ab' preserved |
| \`combining_marks_are_skipped\` | Combining acute (U+0301) skipped, 'e' and 'x' neighbors |
| \`tab_char_expands_to_spaces\` | \`\t\` → 2 spaces |
| \`control_characters_are_filtered\` | \`\x00\x01\` filtered, 'abc' rendered |
| \`cjk_and_emoji_mixed_preserve_each_other\` | 你好🤖世界 — CJK + emoji side by side |

## Aligns with #947

\`\`\`
Stage 1 — Rust renderer behind a flag
  Card-stream parity on Windows Terminal / iTerm2 / VSCode / Linux ssh
\`\`\`

Unicode correctness is prerequisite for card-stream parity — CJK and emoji are fundamental to DeepSeek's user base. Every card kind (user messages, reasoning, streaming, tool output, file views, errors) passes through \`paint_str_to\` → \`paint\`.

### DIFF
--- a/crates/reasonix-render/src/whole_screen/overlay_at.rs
+++ b/crates/reasonix-render/src/whole_screen/overlay_at.rs
@@ -326,7 +326,7 @@ fn paint_clipped(
     let mut w = 0u16;
     let mut clipped = String::new();
     for ch in s.chars() {
-        let cw = unicode_width::UnicodeWidthChar::width(ch).unwrap_or(1) as u16;
+        let cw = crate::whole_screen::paint::char_width(ch);
         if w + cw > *budget {
             break;
         }
--- a/crates/reasonix-render/src/whole_screen/paint.rs
+++ b/crates/reasonix-render/src/whole_screen/paint.rs
@@ -4,15 +4,26 @@ use ratatui::layout::Rect;
 use ratatui::style::{Color, Modifier, Style};
 use unicode_width::UnicodeWidthChar;
 
+/// Display width of a character.  Returns 0 for control characters
+/// (where unicode-width returns None) and combining marks, so the
+/// layout engine skips them instead of giving them a spurious width-1
+/// cell that overwrites the next character.
+pub fn char_width(ch: char) -> u16 {
+    UnicodeWidthChar::width(ch).unwrap_or(0) as u16
+}
+
 pub fn paint(buf: &mut Buffer, x: u16, y: u16, ch: char, fg: Color, bg: Color, modifier: Modifier) {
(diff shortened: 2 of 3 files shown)


esengine:
Solid catch and a clean fix. `unwrap_or(1)` on `UnicodeWidthChar::width` was exactly the wrong default — control chars and combining marks return `None`, so they were silently claiming a cell and stomping the next column. Centralizing the policy in `char_width()` and using it from both `paint()` / `paint_str_to()` / `paint_clipped()` is the right shape (one definition of "what's the cost of this char" rather than three).

Tests cover the failure modes I'd want pinned against regression — ZWJ, combining acute, control bytes, tab, CJK+emoji adjacency.

One known limitation worth flagging now so it doesn't surprise us later: dropping zero-width combining marks means **NFD-decomposed** accented text (`e` + U+0301 instead of pre-composed `é`) will lose its accent. In practice DeepSeek streams and most file content is NFC and renders fine, but if we ever hit a corpus with NFD (some macOS filenames, certain APIs), the right move is grapheme-cluster-aware painting that attaches the mark to the previous cell's symbol rather than dropping it. Out of scope for this PR — file as a follow-up if it ever bites.

Merging.

PR #1077, opened 2026-05-17
fix(desktop): fix ModeSwitch per-segment tooltips + i18n for all composer strings

Orchardxyz:
## What

- Fixed the ModeSwitch tooltip: the `title` was on the parent `<div>`, so hovering any segment showed the current mode's hint rather than the segment's own.
- Internationalized all hardcoded Chinese in the composer component: hints, placeholders, keyboard shortcuts, popup headers, button titles, and the image filter label.

## Why

The tooltip was a cosmetic bug. The hardcoded Chinese was inconsistent with the rest of the desktop UI, where settings and status bar labels already go through `t()`.

## How to verify

1. `npm run tauri dev` — hover each mode segment (Review / Auto / YOLO), confirm the tooltip matches the segment, not the current mode
2. Switch language to English in settings, confirm all composer text switches
3. Open the model picker, confirm preset descriptions render translated

### Before

image

### After

image

## Checklist


esengine:
Thanks — nice catch on the tooltip (the parent-div `title` bug is the kind of thing that's easy to miss in a quick visual pass). Bundling the composer-string i18n cleanup with it makes sense since you're already in the same component. Typed `TKey` on `PRESET_INFO` / `MODE_INFO` is a nicer shape than the prior inline strings too. Both i18n dicts are in sync, CI's green. Merging.

PR #1101, opened 2026-05-17
fix(session): make session rewrites recoverable

GTC2080:
## Summary
- make session rewrites crash-safer by writing a sibling temp file before replacing the live JSONL
- keep a `.jsonl.bak` recovery copy and fall back to it when the live transcript has no valid entries
- carry the recovery sidecar through rename, delete, and archive session flows

## Test Plan
- [x] npm run test -- tests/session.test.ts tests/loop-error.test.ts
- [x] npm run verify

Refs #1079

### DIFF
--- a/src/memory/session.ts
+++ b/src/memory/session.ts
@@ -4,6 +4,7 @@ import { execFileSync } from "node:child_process";
 import {
   appendFileSync,
   chmodSync,
+  copyFileSync,
   existsSync,
   mkdirSync,
   readFileSync,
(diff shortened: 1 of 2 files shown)


esengine:
Nice — this is exactly the kind of stability fix #1079 was asking for. The pattern is right: tmp-sibling write + atomic rename + `.bak` snapshot, with the recovery sidecar carried through rename/delete/archive so users don't get orphaned backups. The `hadContent` distinction in `readSessionMessages` is the subtle bit — it correctly separates "user cleared the session" from "file got corrupted," which is what makes the fallback safe.

Merging.

PR #1102, opened 2026-05-17
fix(loop): persist send-time session healing

GTC2080:
## Summary
- persist the existing send-time session healing back into the active log before each model request
- rewrite the session JSONL when a dangling `assistant.tool_calls` tail or stray tool result is removed during send preparation
- add a regression that simulates a session poisoned in-process after sleep/network interruption and verifies both memory and disk are healed

## Test Plan
- [x] npm run test -- tests/loop-user-persist.test.ts tests/loop-error.test.ts tests/loop.test.ts
- [x] npm run verify

Refs #1079

### DIFF
--- a/src/loop.ts
+++ b/src/loop.ts
@@ -502,11 +502,25 @@ export class CacheFirstLoop {
   private _inflightCounter = 0;
 
   private buildMessages(pendingUser: string | null): ChatMessage[] {
-    // DeepSeek 400s on either unpaired tool_calls or stray tool entries — heal before sending.
-    const healed = healLoadedMessages(this.log.toMessages(), DEFAULT_MAX_RESULT_CHARS);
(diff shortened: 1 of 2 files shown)


GTC2080:
Withdrawing this PR for now because CI is blocked by the stale package-lock renderer optional dependency entries on the current base. I will fix that separately, then resubmit this session-healing change on a clean updated base.

GTC2080:
Update: opened #1105 to fix the stale package-lock renderer optionalDependency entries that made npm ci fail before tests. Once that lands, I will rebase/resubmit the session-healing change on a clean base.

GTC2080:
Follow-up: #1105 has been updated to the latest main base and all checks are green. After #1105 lands, I will resubmit this session-healing change on top of that clean base.

PR #1106, opened 2026-05-17
fix(loop): persist send-time session healing

GTC2080:
## Summary
- persist send-time healing before requests by compacting the active log after dropping invalid tool-call tails
- rewrite the active session JSONL after send-time healing so a poisoned session does not come back on the next run
- add a regression test for dangling assistant tool_calls being removed from the request, in-memory log, and persisted session
- sync package-lock renderer optionalDependencies with current package.json so CI npm ci can install cleanly

## Test Plan
- [x] npm ci --dry-run --ignore-scripts
- [x] npm run test -- tests/loop-user-persist.test.ts tests/loop-error.test.ts tests/session.test.ts
- [x] npm run verify

Resubmits the withdrawn #1102 on the latest main base and folds in the lockfile sync to keep this as one PR for #1079 second-batch work.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "reasonix",
-  "version": "0.43.0",
+  "version": "0.44.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "reasonix",
-      "version": "0.43.0",
+      "version": "0.44.1",
       "license": "MIT",
       "dependencies": {
         "cli-highlight": "^2.1.11",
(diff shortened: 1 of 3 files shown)


esengine:
Nice — this is the right fix for the second symptom in #1079. Before this, send-time `healLoadedMessages` would scrub dangling `tool_calls` from the outgoing request but the JSONL on disk stayed poisoned, so every restart reloaded the bad state and the heal-on-send dance repeated. Persisting the healed log via `rewriteSession` (now crash-safe after #1101) and `compactInPlace`-ing the in-memory log breaks that cycle. The try/catch around the disk write is correctly scoped — a FS failure shouldn't block the in-memory heal.

The lockfile rebump to `0.44.2-rc.4` will get overwritten by the upcoming 0.45.1 lockfile sync — fine to merge as-is.

Merging.

PR #1114, opened 2026-05-17
fix(desktop): keep auto-scroll pinned during streaming

### LINKED ISSUE #1103: 桌面版的对话窗口的换行问题
桌面端的对话窗口在进行几段对话后就会超出可视区域，需要自行滑动至底部 

Image

legeling:
## What

`useAutoScroll` froze the desktop chat view mid-stream once a turn produced
more than ~80 px of content. The thread stayed at the scroll position from
the previous turn while `scrollHeight` kept growing; the bottom-following
behaviour only resumed when the next `busy → idle` transition fired
`scrollToBottom(true)`. From the user's side it looked like the window
silently stopped following streaming output.

## Why

The hook computed `isPinnedRef` inside a `scroll`-event listener:

```ts
el.addEventListener("scroll", updateJumpButton, { passive: true });
```

Scroll events fire for both user gestures **and** our own `scrollTo` calls.
The mount effect and the busy → idle effect both call
`el.scrollTo({ top: el.scrollHeight, behavior: "smooth" })`. A smooth
scroll keeps dispatching `scroll` events for ~200–500 ms while the
animation runs. Each of those events runs `updateJumpButton`, which reads:

```
el.scrollTop + el.clientHeight >= el.scrollHeight - PIN_THRESHOLD
```

Mid-animation `scrollTop` is still climbing toward the (already grown)
`scrollHeight`, so `isPinned` evaluates to `false`. The handler writes
that into `isPinnedRef`. After that, every `ResizeObserver` callback
takes the `else` branch and stops scrolling.

Confirmed against the live Tauri app with a console probe:

```
[04:30.769] scrollTop=10194 scrollHeight=10712 distFromBottom=72
[04:32.531] scrollTop=10194 scrollHeight=10822 distFromBottom=182 UNPINNED
[04:32.732] scrollTop=10194 scrollHeight=10844 distFromBottom=204 UNPINNED
...
[04:36.904] scrollTop=10194 scrollHeight=11521 distFromBottom=881 UNPINNED
[04:47.309] scrollTop=12518 scrollHeight=12965    <-- busy->idle finally re-pins
```

`scrollTop` sat unchanged for ~14s while `scrollHeight` rose by ~900px.

## How

Drop the `scroll` listener entirely. Treat un-pinning as a user-intent
signal instead, and listen to the four input gestures that can move a
scrollable container in WebKit:

- `wheel` — mouse wheel / trackpad two-finger scroll
- `touchmove` — Force Touch / iPad
- `keydown` — arrow keys / PgUp / PgDn / Home / End
- `pointerdown` — scrollbar gutter drag

Programmatic `scrollTo` (mount, busy → idle, ResizeObserver follow-up)
no longer dispatches any of these, so it can't accidentally flip
`isPinnedRef`. ResizeObserver still pins-and-follows while the user
hasn't scrolled away from the bottom; it falls back to surfacing the
jump-to-bottom button only when the user has actually moved up.

## How to verify

Hand-tested in `npm run tauri dev` against `main` at `1c9bbae`:

1. Open the desktop app, send a long-output prompt (e.g. *"don't use any
   tools, just print 1 to 200 each on its own line"*).
2. **Before this fix:** the thread sticks at the prior scroll position
   for the entire stream and only jumps once when the turn completes.
3. **After this fix:** the thread follows the bottom continuously
   throughout the stream. Scrolling up by hand pauses the auto-follow
   and surfaces the jump-to-bottom button as before; clicking the button
   resumes following.

Also re-ran `npm run lint` (clean) and `npx tsc --noEmit` against both
root and `desktop/tsconfig.json` (clean).

## A question for maintainers — UI tests for `desktop/`

While working on this I noticed `desktop/` has no test setup
(`vitest.config.ts` is `environment: "node"`; no `@testing-library/react`,
`happy-dom`, `jsdom`, or hook tests anywhere in the tree). I sketched a
local repro test for this hook against `happy-dom` and used it for an
honest negative-control before-and-after comparison, but didn't include
it in the PR — adding a new dev dep and a new test environment feels
like it deserves its own discussion rather than slipping in via a bug
fix.

Would you be open to a follow-up PR that adds a minimal happy-dom test
slice for the desktop hooks? Happy to scope it however you'd prefer
(or skip it entirely if `desktop/` is intentionally untested).

## Checklist

- [x] `npm run lint` passes locally
- [x] `npx tsc --noEmit` passes for root and `desktop/`
- [x] No `Co-Authored-By: Claude` trailer
- [x] Comments explain *why* (smooth-scroll dispatches scroll events
  while animating), not *what*
- [x] `npm run verify` notes: 4 pre-existing test failures unrelated
  to this change reproduce on `main` at `1c9bbae`
  (`tests/cli-bare-routing.test.ts` macOS `/var` symlink mismatch,
  `tests/code-prompt.test.ts` `.gitignore` length threshold). Confirmed
  by stashing the change and re-running.

Closes #1103


esengine:
Merged — thanks for the excellent root-cause writeup. The smooth-scroll-dispatches-scroll-events trap is the kind of bug that's near-impossible to spot without exactly this kind of probe log, and the input-gesture-as-unpin-signal fix is much cleaner than the original listener. The failing CI was the unrelated stale-lockfile issue from the Rust drop on `main`; updating the branch cleared it.

On the desktop test infra question — yes, please do open a follow-up. A minimal happy-dom slice for `desktop/` hooks would be very welcome; `desktop/` being untested isn't intentional, just unfinished. Scope it however reads cleanest to you.

PR #1108, opened 2026-05-17
fix(session): improve session discovery

GTC2080:
## Summary
- add `/` search to the TUI session picker, filtering by session name, summary, branch, workspace, and currency
- show summary/workspace/branch detail in `reasonix sessions` so saved sessions are easier to identify before resuming
- sync the renderer optional dependency entries in `package-lock.json` so `npm ci` passes on the current base

Refs #1079.

## Test Plan
- [x] `npm ci --dry-run --ignore-scripts`
- [x] `npm run test -- tests/ui-session-picker-search.test.tsx tests/ui-session-picker-broadcast.test.tsx tests/ui-session-picker-currency.test.tsx tests/sessions-command.test.ts`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm run verify`

PR #1119, opened 2026-05-17
fix(workspace): add workspace picker

GTC2080:
## Summary
- Add known workspace discovery from the current root, saved recent workspaces, and session metadata.
- Make bare `/cwd` open a workspace picker; choosing a workspace re-registers root-bound tools, reloads hooks/semantic search, and then opens the session picker scoped to that workspace.
- Sync the renderer optional dependency lockfile entries to `0.44.2-rc.4` so `npm ci` stays green on current `main`.

## Test Plan
- `npm ci --dry-run --ignore-scripts`
- `npm run test -- tests/workspaces.test.ts tests/ui-workspace-picker.test.tsx tests/slash.test.ts`
- `npm run lint`
- `npm run typecheck`
- `npm run verify`


GTC2080:
Added a critical follow-up commit: `1ad0e05 fix(workspace): allow new session in empty integrated picker`.

Root cause: #1119 adds the `/cwd` workspace picker and, after a successful workspace switch, opens the session picker for the selected workspace. In `REASONIX_RENDERER_INTEGRATED=1`, however, the session picker is bridged through the scene `requestListPicker` path. That integrated path previously returned early when `sessionsPickerList.length === 0`.

Impact: switching to a brand-new or otherwise empty workspace would silently fail to show any session picker. The Ink `SessionPicker` already supports creating a new session from an empty list, but the integrated renderer path did not expose that fallback. Practically, a user could switch roots successfully and then have no visible way to create the first session for that workspace, which breaks the most common new-project workflow for multi-workspace switching.

Fix:

- Extracted integrated session picker option construction into `buildIntegratedSessionPickerOptions`.
- Always append an internal-sentinel `New session` option, even when there are no saved sessions.
- Selecting that sentinel now uses `freshSessionName(session)`, matching the Ink `SessionPicker` behavior.
- The sentinel key uses characters that `sanitizeName` cannot produce, so it cannot collide with a real session name.
- Added `tests/integrated-session-picker-options.test.ts` coverage for both saved-session lists and empty-workspace lists.

Validation on the #1119 branch:

- `npm run test -- tests/integrated-session-picker-options.test.ts tests/workspaces.test.ts tests/ui-workspace-picker.test.tsx tests/ui-session-picker-search.test.tsx tests/slash.test.ts`
- `npm ci --dry-run --ignore-scripts`
- `npm run lint`
- `npm run typecheck`
- `npm run verify`

The pre-push hook also reran `npm run verify` successfully: 217 test files passed, 3179 tests passed, 9 skipped.

esengine:
The feature itself looks good — workspace discovery + bare `/cwd` opening a picker is the natural next step after the session-picker work in #1108.

It can't merge in its current state though: status is `CONFLICTING` against `main`. Two sources of conflict:

1. The `package-lock.json` block re-adds the `@reasonix/render-*` optional-dep entries (at `0.44.2-rc.4`), but those packages were removed in the Rust renderer drop, so `main`'s lockfile no longer carries them.
2. `src/cli/ui/App.tsx` was substantially reworked by the pure-Ink rebuild on `main`, and this PR also rewrites parts of `App.tsx` — there will be real content conflicts to resolve, not just a mechanical rebase.

Could you rebase onto current `main`, drop the `package-lock.json` changes entirely (no longer needed), and re-apply the `App.tsx` workspace-picker wiring on top of the post-rebuild structure? Happy to review again once it's clean.

Thanks for the work — this and #1108 together are a real discoverability win.

GTC2080:
Thanks for the detailed guidance. I rebased this PR onto current `main` and force-pushed the refreshed branch (`b43add3`).

- Dropped the `package-lock.json` changes entirely.
- Re-applied the workspace picker wiring on top of the current pure-Ink `App.tsx` structure.
- Avoided reintroducing the old integrated-renderer scene picker code, since that path has been removed from `main`.
- Added a small formatting commit for the pricing override files so Biome/pre-push verification passes on the rebased branch.
- Local pre-push `npm run verify` passed, and GitHub CI/CodeQL are green.
- Current state: `CLEAN` / `MERGEABLE`.

Please take another look when you have a chance.

esengine:
Merged — thanks for the rebase, including the `App.tsx` re-apply on top of the pure-Ink rebuild. Workspace discovery + bare `/cwd` opening the picker is exactly the right shape, and it pairs cleanly with the session picker work from #1108 that just landed. Nice job.

PR #1127, opened 2026-05-17
chore(lockfile): sync package-lock.json to 0.45.0

esengine:
\`sync-render-versions.mjs\` updates \`package.json\` + subpackages but not \`package-lock.json\`. The 0.45.0 release commit landed without regenerating the lockfile, so \`npm ci\` on main rejects the install:

\`\`\`
npm error Invalid: lock file's @reasonix/render-darwin-arm64@0.44.2-rc.1
                   does not satisfy @reasonix/render-darwin-arm64@0.45.0
\`\`\`

This unblocks main. The lockfile-sync gap in the release script is a separate follow-up — sync-render-versions.mjs should run \`npm install --package-lock-only\` (or equivalent) so this stays in sync without a manual step.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "reasonix",
-  "version": "0.44.2-rc.1",
+  "version": "0.45.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "reasonix",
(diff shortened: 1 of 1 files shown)


esengine:
Superseded by lockfile sync inside #1129 (which also carries the 0.45.1 release bump).

PR #1131, opened 2026-05-17
fix(qq): add msg_seq for chunked C2C replies

wade19990814-hue:
## Summary
- add explicit `msg_seq` support when sending QQ C2C messages
- split long QQ replies by UTF-8 byte budget instead of naive string length
- stop after the first failed chunk so long replies do not silently lose later segments
- add regression coverage for chunk splitting and outbound `msg_seq` sequencing

## Why
Long QQ replies were being delivered partially: the first chunk reached QQ, but later chunks could fail with `40054005` (`消息被去重，请检查请求msgseq`).

The current implementation already chunks long replies, but it did not attach an explicit `msg_seq` to each outbound chunk. That makes multi-part replies look like duplicate sends on the QQ side.

## Changes
- extend `QQBot.sendPrivateMessage(...)` to accept and send `msg_seq`
- track a monotonic outbound sequence inside `QQChannel`
- assign incrementing `msg_seq` values per chunk in `sendResponse(...)`
- replace fixed `string.length` chunking with UTF-8 byte-aware splitting and natural break preference
- stop sending later chunks after the first failure and surface which chunk failed

## Verification
- `npm run verify`
- added `tests/qq-channel.test.ts` to cover:
  - UTF-8 byte-bounded chunking
  - incrementing `msg_seq` assignment across chunks
  - stopping after the first failed chunk

## Scope
This PR stays narrow: QQ transport/channel fixes only. It does not change the chat/code session loop or desktop-side architecture.

### DIFF
--- a/src/qq/bot.ts
+++ b/src/qq/bot.ts
@@ -249,13 +249,19 @@ export class QQBot extends EventEmitter {
     this.cleanup();
   }
 
-  async sendPrivateMessage(openid: string, content: string, msgId?: string): Promise<void> {
+  async sendPrivateMessage(
+    openid: string,
+    content: string,
+    msgId?: string,
+    msgSeq?: number,
+  ): Promise<void> {
     const token = await this.ensureToken();
     const body: Record<string, unknown> = {
       content,
       msg_type: 0,
     };
     if (msgId) body.msg_id = msgId;
+    if (typeof msgSeq === "number" && Number.isFinite(msgSeq)) body.msg_seq = Math.trunc(msgSeq);
     const res = await fetch(`${this.baseUrl}/v2/users/${encodeURIComponent(openid)}/messages`, {
       method: "POST",
       headers: {
--- a/src/qq/channel.ts
+++ b/src/qq/channel.ts
@@ -6,18 +6,45 @@ import { loadDotenv } from "../env.js";
 import { type C2CMessage, QQBot } from "./bot.js";
 
 const QQ_LOCK_FILE = join(homedir(), ".reasonix", "qq-channel.pid");
+const QQ_MAX_CHUNK_BYTES = 1500;
+const NATURAL_SPLIT_MIN_FRACTION = 0.6;
 
-function chunkMessage(text: string, maxLen = 1500): string[] {
+function fitUtf8Slice(text: string, maxBytes: number): string {
+  let end = 0;
+  let bytes = 0;
+  for (const char of text) {
+    const nextBytes = Buffer.byteLength(char, "utf8");
+    if (bytes > 0 && bytes + nextBytes > maxBytes) break;
+    end += char.length;
+    bytes += nextBytes;
+  }
+  return end > 0 ? text.slice(0, end) : text.slice(0, 1);
+}
+
+function pickNaturalSplit(candidate: string): number {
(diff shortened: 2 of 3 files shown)


wade19990814-hue:
After this bugfix, I'm considering a small QQ follow-up for access hardening (`ownerOpenId` / allowlist) and wanted to ask about one possible next step before I go further.

Would you be open to QQ channel coverage for the desktop host *if* it is done by extracting the current QQ interaction routing out of `src/qq/use-qq-channel.ts` into a host-agnostic session bridge, rather than adding a separate desktop-only QQ state machine?

The idea would be:
- keep QQ as the same channel surface over existing chat/code sessions
- move turn routing, pause/choice/checkpoint interactions, and final-reply routing into one shared bridge
- let the current CLI/TUI host and `desktopCommand` each adapt to that bridge

That seems like the smallest way to make QQ compatible with both hosts while staying inside Reasonix's current session model and avoiding duplicated maintenance as the CLI and desktop paths evolve.

If that is not a direction you want, I'll keep follow-ups limited to QQ hardening only.

wade19990814-hue:
@esengine One clarification on the possible desktop-host follow-up, since my earlier note was too brief.

I am **not** proposing a desktop-specific QQ implementation or a broader bridge product. If I do anything beyond QQ hardening, the only direction I would pursue is:

- keep QQ as the same channel over existing chat/code sessions
- extract the current QQ interaction routing from `src/qq/use-qq-channel.ts` into one host-agnostic session bridge
- let the current CLI/TUI host and `desktopCommand` each adapt to that same bridge

Why this seems like the best fit for Reasonix:
- it stays inside the current session model instead of adding a third mode
- it avoids maintaining separate QQ state machines for CLI and desktop
- it keeps future QQ follow-ups small even as the two hosts evolve independently

If that still feels out of scope, I will keep follow-ups limited to QQ hardening only (for example access binding / allowlist) and stop there.

esengine:
The QQ side of this PR looks solid — explicit `msg_seq` per chunk, UTF-8 byte budget instead of `.length`, fail-fast on the first chunk failure, and the regression tests cover all three cases. That's exactly the right fix for `40054005`.

It can't merge as-is though: status is `CONFLICTING`. The conflict is the `package-lock.json` commit — that's based on a stale snapshot of `main` from before the Rust renderer was removed, so it re-adds the `@reasonix/render-*` optional-dep entries that `main`'s lockfile no longer has. (CI on current `main` is green; the lockfile fix you saw was solving the inverse problem on the older base.)

Could you rebase onto current `main` and drop the `package-lock.json` commit entirely? The QQ files (`src/qq/bot.ts`, `src/qq/channel.ts`, `tests/qq-channel.test.ts`) shouldn't have any conflicts. Once that's clean I'll merge.

Thanks for the careful root-cause writeup on the dedup error.

wade19990814-hue:
@esengine Rebased onto current `main` and force-pushed. The `package-lock.json` commit is gone now; this PR should be QQ-only.

esengine:
Merged — thanks for the clean rebase. Per-chunk `msg_seq`, UTF-8 byte-bounded splitting, and fail-fast on the first failed chunk are exactly the right shape of fix for `40054005`, and the regression coverage on all three (split, sequencing, stop-on-fail) locks the behavior in. Nice work.

PR #956, opened 2026-05-15
docs: document QQ channel setup and usage

wade19990814-hue:
## What

Add user-facing QQ channel documentation:

- a short QQ channel overview in `README.md`
- a matching short overview in `README.zh-CN.md`
- a dedicated English guide at `docs/qq-connect.md`
- a dedicated Chinese guide at `docs/qq-connect.zh-CN.md`

The guide covers what the QQ channel is, how it fits into `chat` / `code`, the available slash commands, the basic setup flow, and how to obtain QQ Open Platform bot credentials.

## Why

QQ channel support has already landed in the runtime, but there is currently no user-facing documentation explaining how to discover it, connect it, or use it.

More broadly, there are now several features in Reasonix whose runtime support has moved faster than the top-level introduction pages. This PR keeps scope narrow and only documents QQ, but the underlying issue is that some newer capabilities are still under-documented from the README / quick-start surface.

## How to verify

1. Read the new QQ section in `README.md` and `README.zh-CN.md`
2. Open `docs/qq-connect.md` and `docs/qq-connect.zh-CN.md`
3. Confirm the docs consistently describe:
   - QQ as a channel for existing `chat` / `code` sessions rather than a third mode
   - available commands: `/qq connect`, `/qq status`, `/qq disconnect`
   - the setup flow and expected behavior after enablement
   - the QQ Open Platform credential flow at a user-facing level

## Checklist

- [x] `npm run verify` passes locally
- [x] No `Co-Authored-By: Claude` trailer in commits
- [x] Comments follow CONTRIBUTING.md (no module-essay headers, no incident history)
- [x] No edits to `CHANGELOG.md` — release notes are maintainer-written at release time

PR #1141, opened 2026-05-17
fix(qq): harden private channel ownership binding

wade19990814-hue:
## Summary
- add QQ access helpers for owner binding, allowlist checks, and redacted status text
- bind unconfigured QQ sessions to the first private sender for the current run
- support persistent `qq.ownerOpenId` and `qq.allowlist` config values without expanding the slash surface
- surface current QQ access state from `/qq status`
- add regression coverage for access decisions, config loading, and the compact `/qq` handler surface

## Why
Right now, once the QQ channel is connected, any private sender who reaches the bot can become the active input source for the current Reasonix session.

That is fine for a single-user bot in a tightly controlled environment, but it is a weak default for the native QQ channel because it leaves the session open to accidental takeover by another QQ account.

This PR closes that gap without changing QQ's product shape:
- QQ stays a channel over existing `chat` / `code` sessions
- `/qq` stays compact (`connect | status | disconnect`)
- persistent access control stays in config instead of becoming a new slash-managed subsystem

## Changes
- add `src/qq/access.ts` to centralize owner binding, allowlist decisions, and redacted access descriptions
- extend QQ config loading/saving to support:
  - `qq.ownerOpenId`
  - `qq.allowlist`
  - `QQ_OWNER_OPENID`
  - `QQ_ALLOWLIST`
- update `QQChannel` so that:
  - configured `ownerOpenId` / `allowlist` is enforced for private messages
  - otherwise the first private sender in the current run gets a temporary owner binding
  - unauthorized senders are ignored with a terminal warning
- update `/qq status` to show the current access mode, including a redacted owner id when applicable
- keep `/qq` command surface unchanged at `connect | status | disconnect`

## Verification
- `npm run verify`
- added tests for:
  - access decision and redaction logic
  - config normalization / env loading for `ownerOpenId` and `allowlist`
  - `/qq` slash behavior staying compact while still exposing access state through `status`

## Scope
This PR is intentionally limited to QQ access hardening and status visibility. It does not change the session loop, add a new QQ mode, or introduce desktop-host wiring.


wade19990814-hue:
@esengine Small QQ follow-up after #1131. This one is intentionally narrow: it only hardens private-session ownership/access and keeps the slash surface at `connect | status | disconnect`.

esengine:
This is a really solid hardening pass — closing the accidental-takeover gap with a temp owner-bind on first private sender (when unconfigured) is exactly the right default, and keeping `/qq` at `connect | status | disconnect` while putting the access controls in config keeps the slash surface from blowing up. Test coverage on access decisions, config normalization, and the redacted status output is the right scope.

Same blocker as #1131 though: status is `CONFLICTING` because of the `package-lock.json` commit. That commit re-adds the `@reasonix/render-*` optional-dep entries, but those packages were removed in the recent Rust renderer drop, so `main`'s lockfile no longer has them. CI on current `main` is green — the lockfile drift you saw was on a snapshot from before the drop.

Same ask as on #1131: rebase onto current `main` and drop the `package-lock.json` commit entirely. The QQ files (`access.ts`, `channel.ts`, `config.ts`, `use-qq-channel.ts`, slash handler, tests) shouldn't conflict.

Really appreciate the QQ track here — between #864, #956, #1131, and this one, the QQ channel is starting to look like the most robust transport in the project.

wade19990814-hue:
@esengine Rebased onto current `main` and force-pushed. The `package-lock.json` commit is gone now; this PR should be QQ-only hardening.

esengine:
Merged — thanks for the clean rebase. Default-to-first-private-sender as temp owner + persistent `qq.ownerOpenId` / `qq.allowlist` config is the right shape for closing the accidental-takeover gap without bloating the `/qq` slash surface, and the redacted owner id in `/qq status` is a nice touch. Test coverage on access decisions, config normalization, and the compact slash surface locks it all in. Great work on the QQ track.

PR #1142, opened 2026-05-17
fix: TUI stale-symbol persistence in ratatui renderer

AirTech-Full:
- Clear sidebar cells before painting to prevent stale '│' from lingering after a resize or when the sidebar toggles.
- Clear the gap between boot content and the card view so stray pixels from a smaller render don't ghost above the scroll area.
- Expose build_composer_visual_lines as pub(super) and add composer_cursor_position() so the renderer can set the hardware cursor at the composer input area in raw mode.
- Wire composer_cursor_position into the draw loop so typed characters appear at the correct screen location.

## Checklist

### DIFF
--- a/crates/reasonix-render/src/main.rs
+++ b/crates/reasonix-render/src/main.rs
@@ -18,7 +18,7 @@ use reasonix_render::frame_cache::FrameCache;
 use reasonix_render::input::{is_quit, paste_event, translate_key, translate_mouse};
 use reasonix_render::state::{decode_message, Payload};
 use reasonix_render::view::render_setup;
-use reasonix_render::whole_screen::{
+use reasonix_render::whole_screen::{composer_cursor_position, 
     at_completion, at_match_count, cards_layout, demo_state, extract_text, slash_completion,
     slash_is_exact, slash_match_count, Selection, WholeScreen,
 };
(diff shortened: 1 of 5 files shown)


esengine:
Thanks for digging into these — the stale-cell clear, the boot-to-card gap clear, and the `composer_cursor_position()` wiring all look like real, correctly-scoped fixes for what the ratatui renderer was doing.

Unfortunately the timing isn't on our side: the Rust renderer (`crates/reasonix-render/`) was removed entirely in the latest release. We've gone back to a pure Ink TUI on `main` and aren't planning another Rust attempt — background in #1148. So the path this PR targets no longer exists.

Closing this for that reason, not because of anything wrong with the change itself. If you hit any cursor / stale-render issues on the pure-Ink path now, please open a new issue and I'll take a look.

Thanks again for contributing.

AirTech-Full:
Thanks for the clarification, that makes sense regarding the architecture change and the removal of the Rust renderer.

From my perspective, it might be more valuable at this stage to focus engineering effort on stabilizing and improving the CLI core functionality. Web and desktop layers, while useful in the long term, may not bring as much immediate value during such rapid iteration cycles.

Once the core system becomes more stable, the rendering layer and other UX-related components can be revisited and optimized by contributors in a more structured way. Since dsv4 is still not a final stable release, there’s still quite a bit of core work to be done.

Overall, I still think the principle of prioritizing cost-performance and core stability first is the right direction

PR #1152, opened 2026-05-18
Harden TUI turn interrupt handling

### LINKED ISSUE #1147: In TUI, ESC key (also ctrl + C) doesn't stop model's task

**What happened**
When model is running, pressing the ESC key (or Ctrl+C) fails to stop the model as expected, and it keeps outputting content continuously.

**Expected**
Interrupt the current task and stop outputting.

**Reproduction**
After sending any session, press the ESC key (or Ctrl+C) and check if the operation stops immediately.

**Environment**
- OS: Windows 11
- Shell: PowerShell 7.6.1

```markdown
**Reasonix**: 0.45.1 (latest)
**Platform**: win32 (10.0.22631)
**Terminal**: (Windows Terminal 1.24.11321.0) (WT_SESSION=set)
**Size**: 156×43
**Node**: v24.14.0
**Locale**: zh-CN
**Theme**: default
**Model**: deepseek-v4-flash · effort=max
**Mode**: edit=review · plan=off
**MCP**: 1 server(s)
**Session**: code-DeepSeek-Reasonix__archive_202605171952__archive_2026051720
```

**Logs / transcript**

```powershell
PS D:\github-clone\DeepSeek-Reasonix> npx reasonix@latest doctor
reasonix 0.45.1  ·  doctor  (cwd: D:\github-clone\DeepSeek-Reasonix)
  home: C:\Users\10078

  ✓  api key        from C:\Users\10078\.reasonix\config.json (…4785)
  ✓  config         C:\Users\10078\.reasonix\config.json (preset=flash, editMode=review, mcp=1)
  ✓  api reach      /user/balance ok — 19.32 CNY
  ✓  tokenizer      C:\Users\10078\AppData\Local\npm-cache\_npx\8b87d63910bcac80\node_modules\reasonix\data\deepseek-tokenizer.json.gz (1.7 MB)
  ✓  sessions       303 saved · 2.2 MB · oldest 9d
  ✓  semantic       ollama daemon up · model nomic-embed-text ready
  ✓  project        D:\github-clone\DeepSeek-Reasonix (.git, REASONIX.md, package.json, Cargo.toml)

9 ok · 0 warn · 0 fail
```

GTC2080:
## Summary
- update the PR on top of the current pure-Ink TUI after upstream removed the Rust renderer in #1151
- centralize `Esc` interrupt handling so an active model turn aborts once, pending gates are cleared, and an active loop is stopped
- preserve idle loop cancellation while preventing non-model busy states (`!cmd`, `/btw`, dashboard work) from poisoning the next turn's abort controller

## Root Cause
The original Windows repro was tied to the now-removed integrated Rust renderer path, where terminal interrupt events did not reach the active loop abort path. With the current pure-Ink TUI, the remaining risk is using generic `busy` as the abort signal: unrelated async UI work can set `abortedThisTurn`/abort state even when no model turn is running. This PR keys interrupts to active model turns via `submittingRef` and keeps idle loop cancellation separate.

Fixes #1147

## Validation
- npm run verify
- npx vitest run tests/turn-interrupt.test.ts
- npm run typecheck


esengine:
Thanks a lot for the fix and the careful root-cause writeup — really appreciate the effort here.

Unfortunately the timing isn't on our side: the Rust renderer was removed entirely in the latest release and we've gone back to pure Ink, so the integrated renderer / interrupt-bridge path this PR targets no longer exists on `main`. Some background in #1148.

Closing this for that reason, not because of anything wrong with the change itself. If you hit interrupt issues on the pure-Ink path, please open a new issue and I'll take a look — and thanks again for contributing.

PR #1154, opened 2026-05-18
Fix pure-Ink TUI turn interrupts

### LINKED ISSUE #1147: In TUI, ESC key (also ctrl + C) doesn't stop model's task

GTC2080:
## Summary
- handle Ctrl+C as an active-turn interrupt in the pure-Ink TUI instead of always quitting the process
- gate Esc and dashboard aborts on an active model turn (`submittingRef`) instead of generic UI `busy`
- keep idle Esc auto-loop cancellation while avoiding stale `loop.abort()` state poisoning the next turn
- add focused coverage for active Ctrl+C abort, idle Ctrl+C quit, idle Esc loop stop, and unrelated busy no-op behavior

## Root Cause
On the current pure-Ink path, Ctrl+C was routed directly to `quitProcess()` before the active `CacheFirstLoop` could receive `loop.abort()`. Esc used the broader `busy` state, so unrelated async UI work such as `!cmd` or `/btw` could call `loop.abort()` with no model turn active. `CacheFirstLoop.step()` intentionally carries a previously-aborted controller into the next turn, so that stale abort can make the next real model turn stop immediately.

This supersedes the closed Rust-renderer PR #1152 and targets only the current pure-Ink TUI.

Fixes #1147

## Validation
- npm run verify
- npx vitest run tests/turn-interrupt.test.ts tests/stdin-reader.test.ts tests/loop.test.ts tests/loop-inflight.test.ts tests/dashboard-loop-control.test.ts tests/loop-to-dashboard.test.ts
- npm run typecheck
- npx biome check src/cli/ui/turn-interrupt.ts src/cli/ui/App.tsx tests/turn-interrupt.test.ts


esengine:
Merged — thanks for porting this to the pure-Ink path so quickly after #1152 closed.

This is a really solid fix. Gating on `submittingRef` instead of generic `busy` is the right call — it surgically excludes `!cmd` and `/btw` from the abort path, which closes the stale-controller-poisons-next-turn case. Extracting the state machine into a pure `handleTurnInterrupt` with explicit outcomes (`aborted | already-aborted | stopped-loop | idle | quit`) and the four-case test coverage is exactly how this should be structured. Ctrl+C now matching Claude/standard-CLI semantics (interrupt during turn, quit when idle) is a much nicer UX too. The `busyRef`/`submittingRef` sync alongside `setBusy` was a real latent bug.

PR #1159, opened 2026-05-18
fix: add metaso to /search-engine argCompleter, argsHint, and summary

mrluanma:
## What

fix: add metaso to /search-engine argCompleter, argsHint, and summary)


esengine:
Merged — thanks. Clean three-spot patch (argCompleter / argsHint / summary) so the metaso backend now shows up everywhere the other two engines do in the `/search-engine` slash UX.

PR #1175, opened 2026-05-18
feat(embedding): batch OpenAI-compat embedding requests to respect provider batch limits

LeekJay:
## Problem

`embedAllOpenAICompat` sends ALL texts in a single API call. Providers like Aliyun Bailian's `text-embedding-v4` limit batch size to **10 items/request**. When a file has more chunks than the limit, the API returns 400.

## Changes

- **`src/index/semantic/embedding.ts`** — core fix: split input into batches of `batchSize` (default 10)
- **`src/config.ts`** — thread `batchSize` through `OpenAICompatEmbeddingUserConfig`, `ResolvedOpenAICompatEmbeddingConfig`, `SemanticEmbeddingConfigView`, and resolve/normalize/view functions
- **`src/index/semantic/builder.ts`** — pass `batchSize` through `BuildOptions`, `QueryOptions`, `resolveBuildEmbeddingConfig`

## Configuration

```json
{
  "semantic": {
    "provider": "openai-compat",
    "openaiCompat": {
      "baseUrl": "https://dashscope.aliyuncs.com/compatible-mode/v1/embeddings",
      "apiKey": "sk-xxx",
      "model": "text-embedding-v4",
      "extraBody": { "dimensions": 1024 }
    }
  }
}
```

`batchSize` defaults to 10 — matches Bailian's limit, no explicit config needed.

### DIFF
--- a/dashboard/src/i18n/en.ts
+++ b/dashboard/src/i18n/en.ts
@@ -614,6 +614,7 @@ export const en = {
     customRequestBodyMustBeObject: "Custom request body must be a JSON object.",
     saveBeforeIndex: "Save semantic settings before starting an index.",
     extraBody: "extra body",
+    batchSize: "batch size",
     keepExistingKey: "leave blank to keep existing key",
     remoteProvider: "Remote embedding provider",
     remoteProviderDesc:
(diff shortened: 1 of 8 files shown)


esengine:
Merged — thanks for sticking with this across the iterations. The 10-item cap on Bailian was a real wall for anyone indexing a non-tiny tree, glad to have it batched.

Issue #896, opened 2026-05-15
windows桌面版，那个思考的小框Reasoning xxs ，位置太碍眼了


zangzy2023:
如图自己看吧：
就是那个Reasoning  10s 那个
<img width="900" height="651" alt="Image" src="https://github.com/user-attachments/assets/12578921-7a63-49b2-9e7b-2bc1b158f72d" />

esengine:
Good catch — the floating pill was centered 88px above the bottom, sitting right on top of the input's top edge, and it duplicated the stop button the composer already shows (the send button flips red `中断` while busy). Two stop buttons, one of them in your typing path.

#898 collapses it into the existing hint row above the textarea: animated pip · `Reasoning` (or `Skill · <name>`) · elapsed timer on the left, `⏎/esc 中断` hint on the right. No floating overlay, no overlap, one canonical stop control. Same vertical height — the hint row just morphs in place when generation kicks off.

Shipping in the next release.

PR #1195, opened 2026-05-18
fix(desktop): tighten slash picker labels

wade19990814-hue:
## Summary
- shorten the desktop preset labels to `auto` / `v4-flash` / `v4-pro`
- use the skill's own slash-picker description when available, with a localized generic fallback when it is missing
- generically truncate long slash-picker descriptions so they do not dominate the picker UI

## What changed
- update the desktop preset labels in `desktop/src/ui/composer.tsx`
- replace the desktop-side curated skill summary path with a localized generic fallback in `desktop/src/App.tsx`
- add the generic fallback strings in both desktop i18n dictionaries
- keep slash-picker descriptions to a single line in desktop CSS

## Verification
- `npm run verify`

### DIFF
--- a/desktop/src/App.tsx
+++ b/desktop/src/App.tsx
@@ -264,6 +264,20 @@ type Action =
   | { t: "dequeue_send"; index: number }
   | { t: "shift_queued_send" };
 
+function fallbackSkillDesc(skill: SkillInfo): string {
+  const scope =
+    skill.scope === "builtin"
+      ? t("app.skill.scope.builtin")
+      : skill.scope === "global"
+        ? t("app.skill.scope.global")
+        : t("app.skill.scope.project");
+  const runAs =
+    skill.runAs === "subagent"
+      ? t("app.skill.runAs.subagent")
+      : t("app.skill.runAs.inline");
+  return t("app.skill.generic", { scope, runAs });
+}
+
 function reduce(state: State, action: Action): State {
   switch (action.t) {
     case "send_user": {
@@ -1353,7 +1367,7 @@ function TabRuntime({
     },
     ...state.skills.map((s) => ({
       cmd: `/${s.name}`,
-      desc: s.description || `skill · ${s.scope}`,
+      desc: s.description?.trim() || fallbackSkillDesc(s),
       run: () => {
         dispatch({
           t: "start_skill",
--- a/desktop/src/i18n/en.ts
+++ b/desktop/src/i18n/en.ts
@@ -308,6 +308,18 @@ export const en = {
       exportMd: "Copy session as Markdown",
       help: "Show all commands",
     },
+    skill: {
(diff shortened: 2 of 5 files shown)


esengine:
Thanks for putting this together. Holding off on merging as-is for a few reasons — I want to be upfront about them so you can decide whether to revise or split:

1. **Merge conflict.** #1194 (the desktop i18n completion pass) landed a few minutes ago and touches `composer.tsx` + both i18n dictionaries. A rebase against `main` will be needed before any merge.

2. **No backing issue + bundled scope.** The PR mixes three things: a composer footer redesign (`+` quick-actions menu + edit-mode menu + bottom-right preset), a hand-curated skill-summary table, and shorter preset labels. The composer redesign is a real UX call I'd want to discuss in an issue first — the current layout is the result of #877, #896, #898, etc., and the bottom bar isn't load-bearing in the way the redesign assumes. Mind opening an issue with a screenshot of the proposed footer so we can align on the direction before another iteration?

3. **`KNOWN_SKILL_SUMMARY` is a maintenance trap.** Hard-coding seven skill names + EN/ZH summaries inside `desktop/src/skill-display.ts` means every new skill needs a desktop-side translation diff. Skill descriptions belong with the skill (frontmatter / `description:`); the desktop should render whatever the skill ships, not curate it. If the worry is that some descriptions are too long for the slash picker, a generic truncate is fine — but the curated list shouldn't ship.

4. **Localization-only piece is mergeable on its own.** The shorter `auto` / `v4-flash` / `v4-pro` preset labels, and routing the slash-picker description through i18n with a generic fallback, are small and uncontroversial. Happy to take those as a focused PR while the composer redesign discussion plays out.

To be clear — the work itself is thoughtful and the i18n cleanup is appreciated. I just don't want to land a composer rewrite + a hard-coded skill table in a single autonomous merge without a tracked design discussion. Let me know how you want to proceed.

wade19990814-hue:
@esengine Rebased onto current `main` and split the PR down to the focused piece you called out as mergeable.

What's left in this revision:
- shorter desktop preset labels: `auto` / `v4-flash` / `v4-pro`
- slash-picker skill descriptions now use the skill's own `description` when present, with a localized generic fallback for missing descriptions
- generic single-line truncation for long slash-picker descriptions

What's removed from this revision:
- the composer footer redesign
- `KNOWN_SKILL_SUMMARY` / desktop-side curated skill summaries
- the extra desktop-only quick-action strings that were only needed by the redesign

I also reran full `npm run verify` on the rebased branch before force-pushing. If you'd still prefer this split even further, I can peel the truncation rule into a separate follow-up.

esengine:
Merged — thanks for trimming this down to the focused slice. Three small, uncontroversial wins (compact preset labels, localized fallback without the curated table, ellipsis on the picker description) is exactly the shape that lands cleanly. Appreciated.

PR #1198, opened 2026-05-18
fix(mcp): add runtime schema validation for MCP server responses

T1anjiu:
## Summary

Add runtime schema validation for MCP tool-call results in `flattenMcpResult()`. TypeScript types provide no runtime protection, and MCP responses arrive over a network boundary — particularly over SSE / Streamable HTTP transports.

## Problem

A malicious or buggy MCP server can return malformed `CallToolResult` shapes that flow through unresisted:

- `content` is not an array
- content entries are not objects
- entries have an unrecognized `type` (e.g. `"resource"`)
- `text` blocks with a non-string `text` field
- `image` blocks with a non-string `data` or `mimeType` field

These pass through to `blockToString()` and ultimately to the LLM, causing unexpected behavior or garbled output. The MCP client's `dispatch()` resolves `resp.result` as-is [client.ts:306](src/mcp/client.ts:306), and `flattenMcpResult()` previously did only format conversion — neither layer validated structure.

## Fix

Rather than validating at the dispatch boundary (which would require changing the wire-protocol types), the check lives in `flattenMcpResult()` — the single funnel every MCP result passes through before reaching the model.

- **`src/mcp/registry.ts`** — Added `validateResultShape()`, called at the top of `flattenMcpResult()`. It validates step by step:
  - `result` is a non-null object
  - `content` is an array
  - each content entry is a non-null object with `type` of `"text"` or `"image"`
  - text blocks have a string `text` field
  - image blocks have string `data` and `mimeType` fields
  - Any mismatch → `throw Error` (follows the boundary-validation rule: crash loud > silent wrong output).

- **`tests/mcp.test.ts`** — 8 new test cases covering every rejection path plus one valid-result case (non-object result, missing/non-array content, non-object block, unknown type, non-string text, non-string data, non-string mimeType).

## Verification

| Check | Result |
|---|---|
| `npm run lint` | ✅ 607 files, no fixes needed |
| `npm run test` | ✅ all pass (including new cases) |

**Note:** `acp-mcp.test.ts` and `chat-mcp-startup-summary.test.ts` timed out during the full `npm run test` run, but both pass when targeted individually: `npx vitest run tests/acp-mcp.test.ts tests/chat-mcp-startup-summary.test.ts`. The timeouts are unrelated to this change.


esengine:
Merged — thanks for the careful test coverage (every rejection path + a passthrough is exactly the right shape for this kind of boundary check). The single-funnel placement at the top of `flattenMcpResult` is the right call too; that's where all MCP results converge before reaching the model, so any future caller picks the validation up for free.

T1anjiu:
Hey @esengine, have you ever considered switching the repository's homepage link to a custom domain? I'd be glad to offer a small token of support to fund it if you have any interest!

esengine:
Really appreciate the offer — thank you. The cost is small enough that we'd rather cover it ourselves, but you're right that a custom domain would be a nicer landing surface than the GitHub URL. I'll register one in the next few days and point the repo homepage at it once it's wired up.

PR #1225, opened 2026-05-18
fix(cli): sync system cursor to visual ▮ after Ink render for Windows…

AirTech-Full:
… IME

After each PromptInput render, reposition the system cursor to the visual input-cursor position by emitting \x1b[upA\x1b[rightC.  Only fires on Windows (process.platform === 'win32').

The problem: Ink renders the full screen via ANSI sequences, but the actual system cursor (which Windows IME queries via GetConsoleScreenBuffer) ends up at the last output position — not at the visual ▮.  This causes the IME candidate window to appear at the bottom of the terminal instead of tracking the real input cursor.

The fix runs in useEffect after every render; the up/right offsets are calculated from the known border + padding + hint-row layout.

## Checklist

### DIFF
--- a/src/cli/ui/PromptInput.tsx
+++ b/src/cli/ui/PromptInput.tsx
@@ -203,11 +203,11 @@ export function PromptInput({
   // Sync system cursor to visual cursor on Windows for IME positioning.
   // Ink renders the full screen via ANSI sequences; the actual system
   // cursor (queried by Windows IME via Console API) ends up at the last
-  // output position — not at the visual "▌".  Repositioning it after each
-  // render lets the IME candidate window track the real input position.
+  // output position — not at the visual "▌".  Repositioning it after
+  // each render lets the IME candidate window track the real input.
   useEffect(() => {
-    if (!stdout || process.platform !== 'win32') return;
-    // Offset from the last output position (bottom border) up to the
+    if (!stdout || process.platform !== "win32") return;
+    // Offset from the last output line (bottom border) up to the
     // cursor line, then right to the cursor column.  Values are tuned
     // to the border + padding + hint-row layout of the render below.
     const bottomBorder = 1;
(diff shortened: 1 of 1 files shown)


esengine:
Thanks for tackling this — Windows IME cursor placement is a real pain point and the diagnosis (Ink leaves the system cursor at the last output position, Windows IME queries Console API and ends up at the bottom) is correct. Holding off on merging for a few reasons, want to lay them out so you can decide whether to revise:

**1. CI is red.** Biome's flagging two issues in `PromptInput.tsx`:

- `useExhaustiveDependencies` warnings on `lines.length` and `showHugeBufferHints` (these are derived locally, not stable refs — passing them as `useEffect` deps doesn't do what you want).
- Formatter wants `"win32"` not `'win32'` and a single space before the trailing comment.

`npx biome check --write src/cli/ui/PromptInput.tsx` fixes both. Worth running `npm run verify` end-to-end before pushing — the project's pre-push gate runs the same checks.

**2. The PR title says "sync system cursor", but it also lands three other things that aren't called out:**

- **`IME_GUARD_MS = 50`** + `lastNonAsciiInputAtRef` — swallows `action.submit` for 50ms after non-ASCII input to drop CJK-commit-then-Enter. Plausible fix, but it has its own subtle failure mode (50ms is also the window where a real human Enter after a typed CJK character gets eaten). This deserves its own PR with its own test fixture (replay-style: simulated keystroke timing + assert which `action.submit`s land).
- **`useCursorBlink()` → `cursorVisible = true`** — removes the blinking cursor entirely. That's a deliberate UX regression with no explanation. If it's needed to make the system cursor sync land cleanly, say so; if it's incidental, it should be reverted.
- **`<Box paddingX={1}>` → `<Box borderStyle="round" borderColor={accentColor} paddingX={1}>`** — adds a visible round border around the prompt input. This isn't a bug fix, it's a visual change. And it looks like it was added to make your `right = 1 + 1 + 2 + cursorCol  // border + padding + prefix/indent` offset math line up — i.e., the visual change is load-bearing for the IME fix, which makes them inseparable.

**3. The cursor-sync math itself is fragile.**

`bottomBorder + hintMargin + hintLine + largeHints + linesBelow` and `1 + 1 + 2 + cursorCol` are magic constants derived by hand from the current layout. Any future change to the hint row, the border style, the padding, or the prefix indent silently misaligns the system cursor — and Windows IME breakage is hard to notice in tests. If we have to go down this road, the offsets should come from the same primitives the layout uses (e.g., a single `BORDER_WIDTH` / `HINT_HEIGHT` constant referenced both by the JSX and the cursor sync), not hand-tuned numbers.

**Suggested path forward:**

1. Split into three PRs:
   - **A.** Just the system-cursor write after Ink render, with computed (not hardcoded) offsets, no border change, no blink change. Title and description match the diff.
   - **B.** CJK IME commit-Enter guard (with a replay test for the timing window).
   - **C.** If the border was a separate UX intention — own PR, own justification.
2. Run `npm run verify` locally before each push.
3. Keep the existing `useCursorBlink` unless there's a specific reason it conflicts.

I'd happily merge **A** as a small, focused diff once the offsets stop being hand-tuned and the border / blink changes are out of it. The IME-guard piece in **B** I'd want to look at separately too — it's interesting but the 50ms window has tradeoffs that deserve their own discussion. Let me know how you want to split it.

PR #1241, opened 2026-05-18
feat(desktop): add QQ settings entry

### LINKED ISSUE #1218: desktop: discuss a minimal settings entry for the native QQ channel
QQ channel is already merged into Reasonix as a native remote channel for existing sessions. At the moment, the native QQ path is available through the CLI/TUI flow, but desktop has no first-class UI for configuring or checking it.

I would like to discuss a very small desktop addition: a settings entry for the existing QQ channel, without expanding the product scope.

## Why this seems worth doing

- **Keeps the desktop surface aligned with an already-merged native capability**
  QQ is already part of Reasonix as a session-attached remote channel. Desktop currently has no native place to configure or inspect that capability.

- **Keeps the main chat UI clean**
  QQ setup is low-frequency configuration, not a high-frequency composer action. A settings entry seems like the least intrusive desktop surface.

- **Gives QQ a stable maintenance point on desktop**
  Even if future QQ work stays narrow, having one consistent desktop entry point should make follow-up maintenance easier than scattering status/buttons into the main session UI.

## Proposed scope

For a first desktop pass, I think the surface should stay strictly aligned with the current CLI scope:

- connect
- status
- disconnect

Concretely:

- add a compact `QQ Channel` row at the bottom of `Settings -> General`
- show only current status there:
  - connected / disconnected
  - one `Configure…` action
- move `App ID`, `App Secret`, and `Sandbox / Production` into a second-level config view or modal
- keep the main session surface unchanged
- do not introduce a new mode
- do not introduce a desktop-only QQ product surface

### 1. General settings entry

QQ settings overview

### 2. Second-level config view

QQ settings config

## Design intent

The design goal here is:

- **simple**: one small status row in settings, not another always-visible desktop control
- **low-maintenance**: desktop adds only a thin settings/status UI for the existing QQ capability
- **compatible with future adjustments**: if the shared QQ behavior evolves later, this desktop entry point should not need to be redesigned

In other words, I am *not* proposing:

- a standalone `qq` mode
- a large desktop QQ session surface
- a generic chat-platform bridge in core
- a broad remote-channel redesign in this issue

## Implementation direction if this is in scope

My expectation would be:

- desktop owns the configuration/status UI
- the underlying QQ capability remains shared with the existing implementation
- desktop’s first step is only to expose connect / status / disconnect cleanly
- any deeper host-agnostic session-bridge work would be a separate discussion, not bundled into this initial UI entry

If this direction seems acceptable, I can follow up with a narrowly scoped PR for the desktop settings/status surface only.

wade19990814-hue:
Closes #1218.

## Summary
- add a `QQ Channel` row at the bottom of `Settings -> General`
- add a second-level desktop configure view for `App ID`, `App Secret`, and `Sandbox`
- keep desktop aligned with the existing QQ control surface: connect, status, disconnect
- persist through the existing `loadQQConfig` / `saveQQConfig` path with no new desktop secret store

## Notes
- this is a desktop settings front-end only
- it does not add a desktop QQ session mode or a chat-surface QQ panel
- it does not change the underlying QQ channel behavior

## Verification
- `npm test -- --run tests/qq-config.test.ts tests/desktop-qq-config.test.ts tests/desktop-qq-settings-ui.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm --prefix desktop run build`

## Local verify note
- `npm run verify` still hits the existing `tests/code-prompt.test.ts` failure in this environment (`expected 30744 to be less than 29387`)
- the QQ desktop changes themselves passed the targeted checks above


esengine:
Thanks @wade19990814-hue — this stays exactly inside the scope we agreed on in #1218: a thin desktop front-end over the existing `loadQQConfig` / `saveQQConfig` path, no new secret store, no chat-surface bleed, and the underlying channel implementation untouched. The `setDesktopQQEnabled` guard refusing to flip the flag without credentials is a nice safety net. Tests cover the right boundaries. Merged. Closes #1218.

One small thing for a follow-up only if you feel like it (not blocking): in the desktop context `qq_connect` flips the persisted `enabled` flag rather than driving a live websocket (the desktop process doesn't host a QQ runtime today). The "Connected" / "Disconnected" labels mean "enabled in config" — which is the right behaviour for the scope we picked, but if it turns out to confuse users the labels could shift to "Enabled" / "Disabled" later.

Issue #1227, opened 2026-05-18
桌面版本的技能使用无法将技能正常加入到输入栏中，还有技能栏小键盘上下选择无法滑动滚动条，只能鼠标滚轮


niszhum:
我使用/输入选择技能会直接发送一次，然后技能才能进入输入框，这时候才能将技能与文本一起发送。
这时候取消运行再次发送会报错

错误
aborted at iter 0 — stopped without producing a summary (press ↑ + Enter or /retry to resume)

其他命令也是同样，只要选择了立刻就触发，无法正常使用一些需要自己补充信息的操作

ADX15xs:
I have also reproduced this issue. It is indeed quite tricky when invoking skills that require matching text information or options, and I am currently trying to pinpoint the cause.

ADX15xs:
I submitted PR #1247 to fix the issue where skills cannot be added to the input box normally, and please verify it after the merge. 
The problem that the up and down keyboard keys fail to automatically move selected items to the /command list window will be addressed in subsequent PR.

esengine:
Both halves of this fixed:

- Slash-typed skill execution → #1247
- Keyboard ArrowUp/Down scrolling the slash + @ popup → #1255

Closing.

PR #1247, opened 2026-05-18
fix(desktop): enable skill execution via slash command syntax

ADX15xs:
- Implement regex-based slash command detection in `TabRuntime` to trigger skills
- Add `insertOnly` property to `SlashCmd` to support command insertion without immediate execution
- Update `Composer` to handle conditional execution based on `insertOnly` flag
- Include `state.skills` in `TabRuntime` dependency array for correct command matching

## What

Related #1227 
When a user types `/skillname [args]` in the composer and presses Enter, the `send()` handler now matches the input against registered skills and dispatches a `skill_run` RPC instead of sending the raw text as a user message.

## Why

Typing `/skillname` manually and hitting Enter sent the command as plain user text, bypassing skill execution. The skill was only invocable through the slash picker dropdown.

## How to verify

Type `/<skillname> <args>` in the composer and press Enter. Verify the `skill_run` RPC is sent instead of `user_input`.

### DIFF
--- a/desktop/src/App.tsx
+++ b/desktop/src/App.tsx
@@ -1218,12 +1218,25 @@ function TabRuntime({
         return;
       }
 
+      const skillMatch = text.match(/^\/([a-zA-Z0-9_-]+)(\s+.*)?$/);
+      if (skillMatch) {
+        const [, name, args] = skillMatch;
+        const skill = state.skills.find((s) => s.name === name);
+        if (skill) {
+          const clientId = `skill-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
+          const trimmedArgs = args?.trim() ?? "";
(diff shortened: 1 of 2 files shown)


esengine:
Thanks @ADX15xs — tidy fix. The regex check gated on `state.skills` membership is exactly right (built-in commands like `/help` and `/new` still fall through to `user_input` because they're not registered skills, and untracked `/foo` text is unchanged), and pairing it with `insertOnly: true` on the picker rows keeps the dropdown UX consistent in both directions: type-and-Enter and pick-from-dropdown now both yield "compose with args, then send". Merged.

The other half of #1227 (keyboard arrow-keys can't scroll the picker dropdown, only mouse wheel does) is still open — leaving it that way as a separate, smaller patch.

PR #1248, opened 2026-05-18
test(code-prompt): bound .gitignore truncation against a baseline

lisniuse:
## What

Fixes a stale assertion in the `codeSystemPrompt` `.gitignore`
truncation test. It asserted `out.length < CODE_SYSTEM_PROMPT.length +
5000`; the builtin Skills index that `applyMemoryStack` injects has
since grown past that 5000-char slack, so the test fails even though
`.gitignore` truncation still works correctly.

The test now builds a baseline prompt with a tiny `.gitignore` and
asserts the huge-`.gitignore` prompt is only slightly larger than that
baseline. This isolates the truncation delta from fixed-size, unrelated
prompt additions, so the bound no longer goes stale as the prompt grows.

## Why

`tests/code-prompt.test.ts › truncates a .gitignore larger than 2000
chars` fails on `main`:

```
AssertionError: expected 31863 to be less than 29387
```

`npm run verify` (and the pre-push hook / CI) fails because of it. The
`.gitignore` truncation itself is unaffected — only the magic-number
bound was stale.

## How to verify

- `npm run verify` — previously 1 failed, now all 3151 tests pass.
- `npx vitest run tests/code-prompt.test.ts` — 25 passed.

## Screenshots

n/a

## Checklist

### DIFF
--- a/tests/code-prompt.test.ts
+++ b/tests/code-prompt.test.ts
@@ -35,14 +35,18 @@ describe("codeSystemPrompt", () => {
   });
 
   it("truncates a .gitignore larger than 2000 chars", () => {
+    // Baseline prompt with a tiny .gitignore — isolates the truncation
+    // delta from unrelated, fixed-size additions (e.g. the builtin Skills
+    // index), so the bound doesn't go stale as the prompt grows.
+    writeFileSync(join(root, ".gitignore"), "node_modules/\n", "utf8");
+    const small = codeSystemPrompt(root);
     const huge = `${"# comment ".repeat(500)}\n`; // ~5000 chars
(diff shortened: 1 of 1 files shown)


esengine:
Thanks @lisniuse — exactly the right shape. Comparing the huge-`.gitignore` prompt against a tiny-`.gitignore` baseline measures the truncation delta itself, so it stays stable as `applyMemoryStack`'s fixed-size content (Skills index, future additions) grows; the old `CODE_SYSTEM_PROMPT.length + 5000` bound was a magic number that aged out. Merged.

Side note: this was the same failure that another contributor flagged as "in this environment" on a recent PR — it was actually broken on main. Good catch fixing it at the source.

PR #1253, opened 2026-05-18
feat(desktop): restore window, tabs, sessions and scroll on relaunch

### LINKED ISSUE #1244: 桌面端：重启后还原窗口状态、激活标签、各标签会话与滚动位置
**Problem**

Relaunching the desktop app drops most of the previous session's
context. Users have to manually re-size the window, re-open tabs and
re-load the conversation they were reading, then scroll back to where
they were. Only the set of open tab *workspaces* survives today
(#933) — window geometry is handled by a flaky hand-rolled hook, and
the active tab, each tab's session and scroll position are all lost.

**Proposed change**

On relaunch, restore the previous UI state (Windows + macOS):

- **Window** — size, position and maximized state. Replace the
  hand-rolled `useWindowBounds` hook (debounced save, no flush on
  close, unreliable `window.screenX`/`innerWidth`) with the official
  `tauri-plugin-window-state`.
- **Active tab** — reopen on the tab that was focused, not the last
  one created.
- **Per-tab session** — each restored tab reloads the conversation it
  had open. Extend the persisted tab record from `dir` to
  `{ dir, session }`.
- **Scroll position** — restore the chat transcript scroll offset per
  session, so a reopened conversation lands where the user left it
  (not force-scrolled to the bottom).

**Alternatives considered**

- Keeping the hand-rolled `useWindowBounds` — it is unreliable and
  duplicates what `tauri-plugin-window-state` solves properly
  (save-on-close, maximized state, multi-monitor, DPI).
- Persisting only the active tab without sessions — half-restores the
  workspace but still drops the conversation, which is the main thing
  users lose.

**Scope check**

lisniuse:
## What

Restores the previous UI state when the desktop app relaunches
(Windows + macOS). Today only the set of open tab *workspaces* survives
(#933); window geometry, the focused tab, each tab's conversation and
scroll position are all lost.

- **Window** — replace the hand-rolled `useWindowBounds` hook (debounced
  save, no flush on close, unreliable `window.screenX`/`innerWidth`)
  with `tauri-plugin-window-state`, which persists size / position /
  maximized state and restores them cross-platform.
- **Per-tab session** — `desktopOpenTabs` now persists
  `{ dir, session, active }` instead of a bare dir; `bootstrapTab`
  reopens the conversation when its jsonl is still readable, and
  `session_load` / `new_chat` / workspace-switch re-persist it.
- **Active tab** — the frontend reports focus via a `tab_activate`
  message; `$tab_opened` carries an `active` flag so restore reopens on
  the previously focused tab, not the last one created.
- **Scroll position** — the transcript offset is saved per session and
  restored on boot through `useAutoScroll`, which stays un-pinned so
  content growth / the ResizeObserver don't yank it back to the bottom.
- **Sidebar + context panel** — collapse state persists to localStorage.

## Why

Closes #1244. Relaunching mid-task forced users to re-size the window,
re-find the conversation they were reading and scroll back into place.

## How to verify

- `npm run verify` (lint + typecheck + tests + comment-policy gate).
- `cargo check` / `cargo check --release` in `desktop/src-tauri`.
- `cd desktop && npm run tauri dev` — resize/move the window, open a few
  tabs in different workspaces, load a session in each, scroll up mid-
  transcript, collapse a sidebar; relaunch and confirm window geometry,
  tabs, the focused tab, each conversation and the scroll offset return.

## Screenshots

n/a — restore behavior, no static UI change.

## Checklist


esengine:
Thanks @lisniuse — substantial restore polish, and the load-bearing choices are all the right ones: dropping the hand-rolled `useWindowBounds` for the official `tauri-plugin-window-state` (it handles the `screenX`/`innerWidth` quirks + close-flush we never quite got right), and the new `useAutoScroll` restore staying *un-pinned* past the pin threshold so the ResizeObserver doesn't yank back to the bottom is exactly the trap you'd only find by actually running it. Backward-compat read for the legacy `string[]` `desktopOpenTabs` shape kept the migration silent. Merged. Closes #1244.

PR #1255, opened 2026-05-18
ui(composer): scroll active slash command into view

ADX15xs:
Related #1227 

## What
Add `scrollIntoView` to the `/` and `@` popup list in the desktop composer so that keyboard-navigated items stay visible within the scrollable container.

## Why
When the slash or mention popup has more items than the container height, pressing ArrowDown/ArrowUp moves the active highlight but does not scroll the list. The selected item can fall outside the viewport, making keyboard-only navigation unusable for long file trees or filtered command lists.

## How to verify
1. Open the desktop app in a workspace with a deep directory tree.
2. Type `/` in the composer — ArrowDown past the visible area; confirm the active row scrolls into view.
3. Type `@` and navigate a deep path (e.g. `@src/components/ui/...`) — same behavior.
4. `npm run verify` passes (lint + typecheck green; the 6 `server-dashboard.test.ts` failures are a local git hook issue unrelated to this change).

## Checklist

### DIFF
--- a/desktop/src/ui/composer.tsx
+++ b/desktop/src/ui/composer.tsx
@@ -616,6 +616,15 @@ function Popup({
   onClose: () => void;
   onHover: (i: number, item: SlashCmd | MentionItem) => void;
 }) {
+  const listRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    requestAnimationFrame(() => {
+      const el = listRef.current?.querySelector<HTMLElement>(`[data-active="true"]`);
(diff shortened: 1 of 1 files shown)


esengine:
Thanks @ADX15xs — clean follow-up that closes the other half of #1227. `block: "nearest"` keeps already-visible rows from spuriously scrolling, and the `requestAnimationFrame` wrap is the right detail — without it the query would read the previous render's `data-active`. Merged.

PR #1257, opened 2026-05-18
fix(qq): improve CLI first-connect onboarding and i18n

wade19990814-hue:
﻿## Summary
- replace the first-time `/qq connect` readline path with staged credential capture through the existing TUI input pipeline
- localize `/qq connect | status | disconnect` prompts, results, and common QQ runtime errors for EN and zh-CN
- add regression coverage for staged first-connect and localized QQ slash output

## Why
Initial QQ setup in the CLI was fragile because the first-connect path still depended on raw `readline` prompts, while slash commands run through the TUI's async input flow. That made the first-time experience unclear and easy to fail.

This keeps the current `/qq` control surface, but makes first-time setup behave like the rest of the TUI: the command starts an explicit two-step App ID / App Secret capture, `/cancel` aborts cleanly, and status reflects setup-in-progress.

## Testing
- `npm test -- --run tests/qq-slash.test.ts tests/qq-first-connect.test.tsx tests/qq-channel.test.ts tests/qq-access.test.ts tests/qq-config.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run verify`


wade19990814-hue:
For issues reported by multiple users regarding failures or difficulties during the initial QQ configuration, checks and modifications have been made. The first-time configuration guidance and subsequent interaction logic have been optimized, and multilingual environment support has been added. In both Chinese and English environments, corresponding guidance and responses will be displayed.

@esengine PATL, thanks.

esengine:
Thanks @wade19990814-hue — this is exactly the right shape. Lifting first-connect out of `readline` and onto the same `parseSubmit`-interception scene the rest of the TUI already uses kills the `setRawMode` toggle dance and gives us `/cancel` for free. The two-step persist (`enabled: false` before `start()`, `enabled: true` only after success) is a nice safety-by-default touch — credentials survive a mid-connect crash without being marked auto-start. The strings sweep factoring through `src/qq/strings.ts` instead of sprinkling `t()` everywhere is clean too. Merged.

Issue #743, opened 2026-05-13
工具栏的 UI 展示问题


jianghao123ok:
<img width="1274" height="804" alt="Image" src="https://github.com/user-attachments/assets/8775202c-ffa4-40b1-b703-273ccce79636" />

esengine:
Thanks for the screenshot — I can see the area you've boxed (the resume banner + status bar + composer + keybindings footer), but I can't tell from the image alone what specifically looks wrong. Could you describe it in a sentence?

A few specific things that would help:
- What did you expect to see?
- Is the rendering broken (overlap, missing chars, ghosting) or is it a layout/redundancy complaint (too crowded, too many hint rows)?
- Does the issue reproduce on a smaller / larger terminal, or only at 114×36?

Once I know which it is I can take a proper look.

Bernardxu123:
Closed - archived to #3152

PR #1264, opened 2026-05-19
fix(tui): center WelcomeBanner vertically when chat is empty

Bernardxu123:
## Problem

When the TUI starts with an empty chat (no messages yet) and the terminal window is tall, a large blank void appears between the top of the screen and the WelcomeBanner + composer. The void grows proportionally with the terminal height — on a full-screen 50+ row terminal it can be 20+ rows of nothing.

The reporter described this as: "when the CLI input bar has no content and the window is resized larger, a large gap appears between the bottom and the input box — it can't adapt. When the input box has content, it adapts well."

The "content makes it better" observation is a visual side effect: text in the input increases PromptInput's row count, which consumes some of the void from below — but the underlying layout issue is the same with or without input content.

Refs #743

## Root cause

The TUI shell layout in `App.tsx`:

```tsx
<Box flexDirection="column" flexGrow={1}>    // inner container: fills remaining height
  <CardStream />   // internal flexGrow={1}, takes ALL remaining space
  <WelcomeBanner />  // intrinsic height, rendered BELOW CardStream's flex box
  <LiveActivityArea />
</Box>
```

`CardStream` unconditionally applies `flexGrow={1}` to its outer Box via `src/cli/ui/layout/CardStream.tsx:112`. When there are no cards (empty chat), the inner Box has 0 intrinsic height, but the outer `flexGrow={1}` Box still claims all available vertical space from the parent flex container. The `WelcomeBanner` renders *after* (below) this empty flex Box, so the visual result is:

```
┌─── terminal top ────────┐
│                          │
│   ← CardStream flexGrow  │  (0 content, but full height allocated)
│      (20+ rows of void)  │
│                          │
│   WelcomeBanner          │  ← pinned just above the composer
│   LiveActivityArea       │
├──────────────────────────┤
│   ComposerArea           │
└──────────────────────────┘
```

Window resize → `stdout.rows` increases → outermost Box height grows → `flexGrow` containers expand → void grows.

## Fix

Wrap the `WelcomeBanner` in a `flexGrow={1}` Box with `justifyContent="center"`:

```diff
- <WelcomeBanner ... />
+ <Box flexGrow={1} justifyContent="center">
+   <WelcomeBanner ... />
+ </Box>
```

This introduces a **second** `flexGrow={1}` child in the parent flex column. Yoga distributes the available space among all `flexGrow` children proportionally to their grow values:

- **Empty state**: CardStream (flexGrow=1, 0 intrinsic content) + WelcomeBanner wrapper (flexGrow=1). Both get equal share. CardStream renders nothing in its half; WelcomeBanner renders centered in its half.

- **Active state** (`hasConversation === true`): WelcomeBanner is not rendered (conditional is false), so CardStream is the only `flexGrow` child — it gets everything, just like before. Scrolling, streaming, card virtualization all unchanged.

```
After (empty state):
┌─── terminal top ────────┐
│   CardStream flexGrow    │  ← 0 content, takes its share silently
│                          │
│   WelcomeBanner wrapper  │  ← flexGrow=1, justifyContent=center
│       🐋 Welcome         │
│   LiveActivityArea       │
├──────────────────────────┤
│   ComposerArea           │
└──────────────────────────┘
```

## Trade-offs / decision notes for reviewer

| Consideration | Assessment |
|---------------|-----------|
| **Performance** | No effect on the hot path. When `hasConversation` is true, the wrapper is not rendered at all — zero overhead for the 99.9% case. In the empty state, one extra Yoga node is trivial (CardStream alone already has dozens). |
| **Terminal compatibility** | `flexGrow` and `justifyContent` are standard Yoga layout properties, independent of terminal type, ANSI support, or dimensions. No regression risk. |
| **Window resize** | Yoga recalculates `flexGrow` distribution on every resize automatically. No event handler, no useEffect, no state. |
| **Interaction with welcome card dismissal** | The `hasConversation` flag becomes true on the first send, which removes the wrapper. No cleanup needed. |
| **Interaction with `/copy` + `/dashboard` hints** | Those hints are inside `WelcomeBanner`. They move with it. No change to their rendering. |
| **Interaction with slash suggestions** | Slash suggestions render inside `ComposerArea` (outside the flex container we're touching). Unaffected. |
| **Interaction with CardStream measurement** | CardStream's `useBoxMetrics` on `outerRef` measures the allocated height. An empty CardStream getting less height (because the wrapper shares the space) means `outer.height` is smaller — but since `maxScroll = max(0, inner.height - outer.height)` and `inner.height` is 0 with no cards, `maxScroll` stays 0. No change. |
| **Minimality** | +6/-6 lines, pure JSX wrapping. No new imports, no new components, no new state. |
| **Is this worth it?** | This is a cosmetic fix for a transient state (empty chat before first message). The void exists for the ~5 seconds between launch and first input. Whether that merits a code change is a product judgment call — the fix itself is mechanically safe and zero-cost. |

## Visual diff

**Before** (24-row terminal, empty chat)
```
                          
                          
                          
                          
                          
                          
                          
                          
  🐋 Reasonix            
  DeepSeek Agents         
                          
  /dashboard · /copy      
  ● auto · project· main · ¥0.00 turn · cache —% · v0.46.1 · ⚑ /feedback
                          
  ╭──────────────────────╮
  │ › ▌ask anything...   │
  │  ⏎ send · ⇧⏎ newline · ^U clear · ^P/^N history · esc abort · ^C quit
  ╰──────────────────────╯
```

**After** (same terminal, same state)
```
                          
  🐋 Reasonix            
  DeepSeek Agents         
                          
  /dashboard · /copy      
                          
                          
                          
  ● auto · project· main · ¥0.00 turn · cache —% · v0.46.1 · ⚑ /feedback
                          
  ╭──────────────────────╮
  │ › ▌ask anything...   │
  │  ⏎ send · ⇧⏎ newline · ^U clear · ^P/^N history · esc abort · ^C quit
  ╰──────────────────────╯
```

### DIFF
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -4039,12 +4039,14 @@ function AppInner({
           surviving past the first turn.
         */}
                   {!hasConversation && !busy && !isStreaming && slashMatches === null ? (
-                    <WelcomeBanner
-                      inCodeMode={!!codeMode}
-                      workspaceRoot={codeMode ? currentRootDir : undefined}
-                      dashboardUrl={dashboardUrl}
-                      languageVersion={languageVersion}
-                    />
+                    <Box flexGrow={1} justifyContent="center">
+                      <WelcomeBanner
+                        inCodeMode={!!codeMode}
+                        workspaceRoot={codeMode ? currentRootDir : undefined}
+                        dashboardUrl={dashboardUrl}
+                        languageVersion={languageVersion}
+                      />
+                    </Box>
                   ) : null}
                   <LiveActivityArea
                     noTakeoverOverlay={noTakeoverOverlay}


esengine:
Thanks for the contribution — merged. Nice catch on the empty-state alignment.

PR #1293, opened 2026-05-19
feat(lifecycle): add prefix-stable engineering lifecycle

GTC2080:
This PR was closed and superseded by the new lifecycle PR with the project-conventional title.

The implementation branch and discussion continue in the replacement PR.

### DIFF
--- a/src/cli/heap-limit-launch.ts
+++ b/src/cli/heap-limit-launch.ts
@@ -2,24 +2,34 @@
 
 import { spawnSync } from "node:child_process";
 import { totalmem } from "node:os";
+import { fileURLToPath } from "node:url";
 import { getHeapStatistics } from "node:v8";
 import { RX_HEAP_REEXEC_ENV, decideHeapTargetMb } from "./heap-limit.js";
+
+const runningUnderVitest =
+  process.env.VITEST === "true" || process.env.VITEST_WORKER_ID !== undefined;
 
 const target = decideHeapTargetMb({
   currentLimitMb: Math.floor(getHeapStatistics().heap_size_limit / 1024 / 1024),
   totalMemMb: Math.floor(totalmem() / 1024 / 1024),
   nodeOptions: process.env.NODE_OPTIONS ?? "",
   execArgv: process.execArgv,
-  alreadyReexec: process.env[RX_HEAP_REEXEC_ENV] === "1",
+  alreadyReexec: process.env[RX_HEAP_REEXEC_ENV] === "1" || runningUnderVitest,
 });
 
 if (target !== null) {
   const existing = process.env.NODE_OPTIONS ?? "";
   const nextOptions = `${existing} --max-old-space-size=${target}`.trim();
   const childEnv = { ...process.env, NODE_OPTIONS: nextOptions, [RX_HEAP_REEXEC_ENV]: "1" };
-  const result = spawnSync(process.execPath, process.argv.slice(1), {
-    env: childEnv,
-    stdio: "inherit",
-  });
+  const extension = import.meta.url.endsWith(".ts") ? "ts" : "js";
+  const entrypoint = fileURLToPath(new URL(`./index.${extension}`, import.meta.url));
+  const result = spawnSync(
+    process.execPath,
+    [...process.execArgv, entrypoint, ...process.argv.slice(2)],
+    {
+      env: childEnv,
+      stdio: "inherit",
+    },
+  );
   process.exit(result.status ?? 0);
 }
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -22,6 +22,7 @@ import {
   snapshotBeforeEdits,
   toWholeFileEditBlock,
 } from "../../code/edit-blocks.js";
+import { EngineeringLifecycleRuntime } from "../../code/lifecycle.js";
 import { clearPendingEdits, loadPendingEdits } from "../../code/pending-edits.js";
 import {
   clearPlanState,
@@ -31,10 +32,12 @@ import {
 } from "../../code/plan-store.js";
 import {
   type EditMode,
+  type EngineeringLifecycleMode,
   type PresetName,
   defaultConfigPath,
   editModeHintShown,
   loadBaseUrl,
+  loadEngineeringLifecycleMode,
   loadReasoningEffort,
   loadTheme,
   markEditModeHintShown,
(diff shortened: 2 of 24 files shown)


esengine:
Thanks for the RFC implementation, @GTC2080 — the architecture is on the right track and the test coverage is solid. I pulled the branch (`pr-1293-lifecycle`) and ran end-to-end checks before reviewing; what follows is grounded in what I measured, not just code reading.

## What I verified

- `npm run verify` on the branch tip — green (229 files, 3226 passed, 9 skipped).
- Prompt size: the new contract block is unconditionally inlined into `CODE_SYSTEM_TEMPLATE`. `codeSystemPrompt()` has no `mode` parameter — `off` users pay the same.
  ```
  bytes added to static prefix: 896
  position: inside CODE_SYSTEM_TEMPLATE between "submit_plan menus" and "ask_choice"
  ```
- `classifyEngineeringPrompt()` probe table (18 prompts) — **5 misclassifications**, all on the same axis:

  | result | want | prompt |
  | --- | --- | --- |
  | `large` | `small` | `look across these files for the typo` |
  | `large` | `small` | `find references across the repo` |
  | `large` | `small` | `check across the imports for the missing one` |
  | `small` | `large` | `リファクタリングして` |
  | `small` | `large` | `리팩토링 해줘` |

- `isHighRiskLifecycleToolCall()` probe table (16 cases) — **4 false positives**:
  ```
  MISS  run_command(git checkout README.md)       — read-state, single file
  MISS  run_command(git checkout -- README.md)    — discards one file, not the tree
  MISS  run_command(grep -rn 'mv ' src/)          — literal 'mv' inside a search
  MISS  run_command(echo cp would copy here)      — echo
  ```

The reproducer is at `probe-classifier.mts` on my local copy of the branch — happy to push it as `tests/lifecycle-classifier-corpus.test.ts` if useful.

## Architectural concerns

### 1. "Prefix-stable" isn't actually prefix-conditional

The 896-byte contract block lives in the static template, so a user who opts out (`engineeringLifecycle.mode: "off"`) still pays the tokens forever. That defeats the cache-cost argument for users who deliberately turn the feature off. The block needs to be conditionally appended after `codeSystemBase()` based on mode — at which point the cache key stays stable per-user but the prefix shrinks for opt-outs.

This matters because cheapness is the whole positioning of the project. We've turned down feature requests that cost ~100 prefix tokens unconditionally; this needs the same bar.

### 2. The auto-arm regex is too coarse for a load-bearing UX decision

`\bacross\b` is a routine English word. The three false-positive prompts above all read like normal small-task asks ("look across these files for the typo") — none of them should trigger plan-required rails. With the current matcher, a user who phrases their typo fix using "across" gets dragged into structured-plan land.

Symmetrically, Japanese (`リファクタリング`) and Korean (`리팩토링`) aren't covered at all — those users in `auto` mode get no rails when they should.

Substring regex is the wrong tool for "is this task large enough to require a plan." If we want auto-detection, it needs intent classification (cheap LLM call, or at minimum a much more precise token-boundary check). My read: the auto-arm should not ship in v1 — the slash command is enough for opt-in.

### 3. `armed` is a one-way trap

Once `auto`-mode arms the lifecycle, the only documented exits are: model submits a plan and user cancels it, session restart, or hand-editing config to `off`. Walked through the code:

- `setMode("auto")` after already being in auto-armed state: no state change (only the `off` branch calls `reset()`).
- `/plan auto` calls `togglePlanMode(false)` → `setMode(base)`. If base is `auto`, that's a no-op on the FSM.
- `observeUserPrompt()` only transitions out of `complete`/`cancelled`, never out of `armed`.

So if I say "refactor X" → arm → start exploring → notice an unrelated typo → ask to fix it → blocked. My escape is to restart the session or coax the model to `submit_plan` so I can `cancel` it. That's not a UX I want to ship.

Fix: either add a slash like `/plan release` that calls `lifecycle.reset()`, or have `observeUserPrompt()` re-classify on every prompt and step the state down if a small-class prompt arrives mid-armed-no-plan.

### 4. `isHighRiskCommand` blocks ordinary read paths

`git checkout README.md` is how I revert one file. `git checkout -- file` same. `grep -rn 'mv '` is a search, not a move. These shouldn't surface a `rejectedReason: "engineering-lifecycle"` JSON to the model.

The regex bank needs to operate on tokenized argv rather than substring-matching the raw command string. At minimum, `git checkout` with a file argument should not be high-risk; only branch switches (`git checkout <ref>`, `git switch`) should.

### 5. Static-prompt + interceptor-rejection means new loop risk on DeepSeek

When the model hits the rail it gets back:
```json
{"error": "edit_file: blocked by Engineering Lifecycle — ...", "rejectedReason": "engineering-lifecycle", "state": "armed"}
```

This is a tool result, not a schema failure, so the existing `_lastFailFingerprint` loop guard in `ToolRegistry` doesn't catch a re-attempt of the same call. DeepSeek's instruction-following on bespoke rule text is weaker than Claude/GPT-4; there's a real risk the model retries the same `edit_file` after a thin "let me try a different angle" turn, burning cache.

Suggestion: extend `_lastFailFingerprint` (or add an interceptor-rejection counterpart) so a repeated identical call gets a sharper "stop, switch to read-only + submit_plan" error after the second attempt. Otherwise we're relying on prompt compliance for a loop-prevention guarantee.

### 6. Scope mixing — heap-limit-launch.ts should be a separate PR

`src/cli/heap-limit-launch.ts` has its own bug fix (resolve entrypoint from `import.meta.url`, add vitest-running detection). Unrelated to the lifecycle, would be much easier to review and revert independently.

### 7. `/plan` overload changes a public surface

`/plan strict` and `/plan auto` overload the existing on/off semantics. Some users have muscle memory for `/plan` as a toggle. New args are fine but it's a public-API change worth flagging in the PR title or a separate commit.

## What I'd actually merge

The cleanest path is to land this in three pieces:

**(A) Tool interceptor chain** — `addToolInterceptor` is a real infrastructure improvement and stands on its own. The legacy single interceptor stays compatible. No behavior change.

**(B) Plan v2 schema + evidence on `mark_step_complete`** — additive, no rails. Step `targets`/`acceptance`/`verification` fields, evidence persistence, the v1→v2 read fall-through. Useful even without the lifecycle FSM, because today's `submit_plan` already accepts free-form steps and the evidence requirement raises the floor on plan quality.

**(C) Lifecycle runtime — opt-in only, no auto-arm.** Ship the FSM, the `/plan strict` slash, and the prompt block — but gate the prompt block on `mode !== "off"`, default to `off`, and drop the regex auto-arm entirely. Users who want the rails type `/plan strict` per-task; users who don't pay zero tokens. We can revisit auto-arm in a separate RFC once we have telemetry on how often `strict` is used.

The heap-limit fix goes in its own PR.

## What I love

- Host-side FSM, no per-stage prompt churn, no tool-list mutation between stages. You read the cache-first constraint correctly.
- `addToolInterceptor` chain replacing the single-slot interceptor is exactly the right shape.
- Evidence-required-on-completion is the single best safety rail in the PR — it operates inside an already-existing tool call and doesn't bloat the prefix.
- `tests/lifecycle.test.ts` covers the state transitions well.

Genuinely good design instincts on this — the changes I'm pushing back on are scope and defaults, not direction. Happy to split the PR myself if that's easier than asking you to re-cut, or to land (A) immediately and discuss (B)/(C) on follow-up issues.

PR #1298, opened 2026-05-19
fix(desktop): use updater plugin for update check to bypass CSP

T1anjiu:
## Summary

Replace WebView `fetch()` with the Tauri updater plugin's `check()` in the About modal, so the "Check for updates" button works instead of failing with `Failed to fetch`.

## Problem

The About modal's "Check for updates" button fetches `https://api.github.com/repos/esengine/DeepSeek-Reasonix/releases` directly from the WebView via `fetch()`. The Tauri CSP (`connect-src 'self' ipc: http://ipc.localhost`) blocks this cross-origin request, causing WebView2 to throw a `TypeError: Failed to fetch` before the request ever leaves the browser. The user sees:

```
检查失败：Failed to fetch
```

The root cause is in `checkForUpdates()` [about.tsx:41-43](desktop/src/ui/about.tsx:41-43): a raw `fetch()` to an origin not listed in `connect-src`. The CSP is configured in [tauri.conf.json:27](desktop/src-tauri/tauri.conf.json:27) and does not include `https://api.github.com`.

By contrast, `App.tsx` already uses `check()` from `@tauri-apps/plugin-updater` [App.tsx:2754](desktop/src/App.tsx:2754) for automatic update checks — that call goes through Tauri IPC to Rust-side HTTP (`reqwest`) and is not subject to WebView CSP.

## Fix

- **`desktop/src/ui/about.tsx`** — Replaced the raw `fetch()` + manual tag-parsing logic with a single `await check()` call from `@tauri-apps/plugin-updater`. Removed the now-unused `RELEASES_API` constant, `cmpSemver()` helper, and the GitHub release-tag filtering logic (the updater plugin already handles version comparison and endpoint selection). The `check()` return value is `null` when up-to-date, or an `Update` object with a `.version` string when outdated.
- **`desktop/src/i18n/en.ts`** — Removed the orphaned `checkNoRelease` key (was only reachable from the deleted tag-filtering path).
- **`desktop/src/i18n/zh-CN.ts`** — Same, removed `checkNoRelease`.

## Verification

| Check | Result |
|---|---|
| `npm run lint` | ✅ 622 files, no fixes applied |
| `npm run test` | ✅ 3199 passed, 9 skipped, **3 failed (unrelated timeouts)** |

**Note:** `acp-mcp.test.ts` and `chat-mcp-startup-summary.test.ts` timed out during the full `npm run test` run, but both pass when targeted individually: `npx vitest run tests/acp-mcp.test.ts tests/chat-mcp-startup-summary.test.ts`. The timeouts are unrelated to this change.

### DIFF
--- a/desktop/src/i18n/en.ts
+++ b/desktop/src/i18n/en.ts
@@ -44,7 +44,6 @@ export const en = {
     updateAvailable: "Update available: {version}",
     openReleases: "View releases",
     checkFailed: "Check failed: {message}",
-    checkNoRelease: "No releases yet",
     close: "Close",
   },
   workdir: {
--- a/desktop/src/i18n/zh-CN.ts
+++ b/desktop/src/i18n/zh-CN.ts
@@ -46,7 +46,6 @@ export const zhCN: typeof en = {
     updateAvailable: "新版本可用：{version}",
     openReleases: "查看发布页",
     checkFailed: "检查失败：{message}",
-    checkNoRelease: "尚无发布",
     close: "关闭",
   },
   workdir: {
--- a/desktop/src/ui/about.tsx
+++ b/desktop/src/ui/about.tsx
@@ -1,10 +1,10 @@
 import { openUrl } from "@tauri-apps/plugin-opener";
+import { check } from "@tauri-apps/plugin-updater";
 import { useCallback, useState } from "react";
 import { t } from "../i18n";
 import { I } from "../icons";
 
 const REPO_URL = "https://github.com/esengine/DeepSeek-Reasonix";
-const RELEASES_API = "https://api.github.com/repos/esengine/DeepSeek-Reasonix/releases";
 const RELEASES_PAGE = `${REPO_URL}/releases`;
 
 type CheckState =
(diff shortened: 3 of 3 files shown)


esengine:
Thanks for the clean diagnosis — the CSP root cause is exactly right, and routing through the updater plugin is the correct fix.

Holding the merge on one issue though: there's a name collision between the new `check` import and the existing `check` state variable. The desktop typecheck (`cd desktop && npx tsc --noEmit`) flags it:

```
src/ui/about.tsx(2,1):  error TS6133: 'check' is declared but its value is never read.
src/ui/about.tsx(30,28): error TS2349: This expression is not callable.
  No constituent of type 'CheckState' is callable.
```

At line 18 the component destructures `const [check, setCheck] = useState<CheckState>(...)`, which shadows the import inside the component body. So `await check()` on line 30 resolves to the local `CheckState` value — not the plugin function — and would throw `check is not a function` the first time the user clicks the button.

Heads up: the root-package `npm run verify` doesn't run the desktop workspace's typecheck, which is why CI didn't catch it. Worth running `cd desktop && npx tsc --noEmit` locally on any `desktop/` change.

The fix is a one-line rename at the import site, e.g.:

```ts
import { check as checkUpdate } from "@tauri-apps/plugin-updater";
// ...
const update = await checkUpdate();
```

Push that and CI should go green — happy to merge once it's in.

PR #1319, opened 2026-05-19
feat: Add Java source code search and decompilation tools.

ezdemo:
## What

Add a `java_source` tool that lets the Reasonix agent find and decompile Java source code by fully-qualified class name.

**New modules:**
- `src/java/class-source-finder.ts` — two-phase search engine: walks the project tree for `.java` files, then falls back to scanning `~/.m2/repository` jars with `javap` decompilation. Supports three modes: project-only, keyword-filtered `.m2` scan, and direct jar path.
- `src/java/zip-reader.ts` — zero-dependency pure-Node ZIP/JAR entry reader using `fs` + `zlib`. No JDK `jar` command required.
- `src/java/index.ts` — barrel export.

**New tool (`parallelSafe`, `readOnly`):**
- `java_source(className, projectRoot?, jarPath?, jarKeyword?)` — registered in `buildCodeToolset()` via `registerJavaSourceTool()`.

**Wiring:**
- `src/code/setup.ts`: added `registerJavaSourceTool` import and call in `buildCodeToolset()`.

## Why

Java developers using Reasonix to work on Maven/Gradle projects need a way to inspect library source code without leaving the agent session.

## How to verify

1. `npm run verify` passes (lint + typecheck + tests + comment-policy gate).
2. The tool appears in the tool registry and is callable by the model in `reasonix code` sessions.

## Checklist

- [√] `npm run verify` passes locally (lint + typecheck + tests + comment-policy gate)
- [√] No `Co-Authored-By: Claude` trailer in commits
- [√] Comments follow CONTRIBUTING.md (no module-essay headers, no incident history)
- [√] No edits to `CHANGELOG.md`

### DIFF
--- a/src/code/setup.ts
+++ b/src/code/setup.ts
@@ -13,6 +13,7 @@ import { bootstrapSemanticSearchInCodeMode } from "../index/semantic/tool.js";
 import { ToolRegistry } from "../tools.js";
 import { registerChoiceTool } from "../tools/choice.js";
 import { registerFilesystemTools } from "../tools/filesystem.js";
+import { registerJavaSourceTool } from "../tools/java-source.js";
 import { JobRegistry } from "../tools/jobs.js";
 import { registerMemoryTools } from "../tools/memory.js";
 import { registerPlanTool } from "../tools/plan.js";
@@ -74,6 +75,7 @@ export async function buildCodeToolset(opts: CodeToolsetOpts): Promise<CodeTools
       webSearchEndpoint: webSearchEndpoint(),
     });
   }
+  registerJavaSourceTool(tools, { projectRoot: opts.rootDir });
   // Lazy: constructing DeepSeekClient throws when DEEPSEEK_API_KEY is unset,
   // which would kill `reasonix code` before the setup wizard can prompt for
   // one. Defer to first subagent dispatch — by then the user has either keyed
--- a/src/java/class-source-finder.ts
+++ b/src/java/class-source-finder.ts
@@ -1,0 +1,296 @@
+// ClassSourceFinder — locate Java source by fully-qualified class name.
+// Two-step: (1) project .java walk, (2) .m2 jar scan + javap decompile.
+
+import { execFile } from "node:child_process";
+import * as fs from "node:fs";
+import * as fsp from "node:fs/promises";
+import * as os from "node:os";
+import * as path from "node:path";
+import { listJarEntries, readJarEntry } from "./zip-reader.js";
+
+// ── types ────────────────────────────────────────────────────────────────────
+
+export interface FindResultSuccess {
+  found: true;
+  /** Reconstructed (or raw {@code .java}) source text. */
+  source: string;
+  // Where the source originated:
+  //   "project" — .java file in the project tree.
+  //   "m2-jar"  — decompiled from .m2 jar.
+  //   "jar"     — decompiled from user-specified jar path.
+  method: "project" | "m2-jar" | "jar";
+  /** Filesystem path to the {@code .java} file or the jar that was decompiled. */
+  sourcePath: string;
+}
+
+export interface FindResultNotFound {
+  found: false;
+  method: "not-found";
+}
+
+export type FindResult = FindResultSuccess | FindResultNotFound;
+
+export interface FindSourceOptions {
+  // When set, only scan .m2 jars whose filename or path contains this keyword.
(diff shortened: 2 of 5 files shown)


esengine:
Thanks for the engineering here — the pure-Node ZIP parser and the two-phase finder are clean work, and the `className` validation is exactly right.

Holding off on merging this as a built-in tool, though, because of the project's own design rail (the one our system prompt recites under "Schema cost is real"):

> A new-tool proposal MUST cover (a) which existing-tool composition fails to do this, (b) rough description-token cost, (c) why a prompt or description change can't reach the same end. Default to "tighten prompt / existing tool" before "add tool".

Applied here:

**(a) Existing tools already cover it.** Phase-1 (walk project for `Foo.java`) is `search_files` + `read_file` — the model picks those without prompting. Phase-2 (`.m2` jar scan + decompile) is three `run_command` calls with tools the user already has:

```
find ~/.m2/repository -name "*spring-core*.jar"
unzip -p <jar> com/example/Foo.class > /tmp/Foo.class
javap -c -p -cp /tmp com.example.Foo
```

**(b) Schema cost.** Tool description + four-parameter schema is roughly 1.5–2 KB shipped in every request, paid by every user including the ~90%+ who never touch Java. We just landed PRs #1320 / #1321 / #1323 / #1324 to bring the cache-prefix tax down from ~36k to ~16k tokens per request, and the new `tests/prompt-budget.test.ts` would actively fail against this PR's tool list size.

**(c) Reachable without a new tool.** A subagent **Skill** is the right home — install-once for Java users, zero schema cost for everyone else:

- Drop `src/java/*` and the `registerJavaSourceTool` wiring.
- Add `.reasonix/skills/java-source.md` (or wherever built-in skills ship) carrying your workflow as a subagent playbook: `run_command` for jar discovery + `javap`, and inline your pure-Node ZIP-reader instructions as the "if `javap` missing" fallback. That preserves the engineering you've done in a venue where Java users opt in.

Alternatively, if you really want it as a real tool, the path that's acceptable is **opt-in via config** — default off, on for Maven/Gradle workspaces only. #1324 shows the shape (`filesystem.outlineThresholdBytes` loader + setup wiring).

A couple of side notes for whichever direction you take:

- **No new tests.** 619 lines of code, zero coverage. The PR description's "verify passes" only means existing tests didn't break.
- **Maven-only assumption.** `~/.m2/repository` hard-codes Maven; Gradle / SBT / Bazel users get nothing.

The code itself is good — none of this is a quality concern, only "does this belong in the global tool list." Happy to take a follow-up that lands it as a Skill or an opt-in tool. Thanks again for the time and the careful implementation.

ezdemo:
How can I restart this PR? I tried creating a new one.

esengine:
You don't need to "restart" #1319 — that one was closed on the design call (built-in tool vs. Skill / opt-in). What you want to reopen is **#1335**, which is exactly the shape we asked for: `javaSource: false` by default, env override, opt-in registration in `buildCodeToolset()`, and 41 tests covering the zip reader, finder, keyword filter, abort, and tool dispatch. That's the right PR — thank you for doing the work.

#1335 went into the closed state because the branch `main-add-java-tool` is shared between the two PRs and now conflicts with `main` (we landed #1306 / #1312 / #1314 in the interim, all touching `src/code/setup.ts` and `src/config.ts`). GitHub flips a PR closed when its head ref disappears or diverges in certain ways; the conflicts are why `mergeStateStatus` is `DIRTY` and there are no CI checks against the latest base.

To get it back to a mergeable state, the clean path is:

```bash
# from your fork
git fetch upstream                       # (esengine/DeepSeek-Reasonix as upstream)
git checkout main-add-java-tool
git rebase upstream/main
# resolve conflicts in:
#   src/code/setup.ts   — keep your registerJavaSourceTool call
#   src/config.ts       — keep your javaSource field + loadJavaSourceEnabled()
git push --force-with-lease origin main-add-java-tool
```

Then open a fresh PR from the rebased branch (`gh pr create` or the web UI) — don't try to reopen #1319 or #1335 by hand, since GitHub's history on those is muddled by the shared head. A new PR off the rebased branch with the #1335 description copied over is the cleanest restart and will trigger fresh CI against current `main`.

One small note when you re-push: please make sure the new branch carries only your Java-source commits — squash any merge-resolution noise so the diff stays at ~1.5k additions and reviewers can read it cleanly. Looking forward to taking another look.

PR #1325, opened 2026-05-19
fix(multi_edit): roll back files that may have been modified on write…

T1anjiu:
## Summary

Add rollback-on-failure to `multi_edit`: when a disk write fails mid-batch, files that may have been modified (including the currently-failing file) are restored to their pre-edit content using the original text captured during validation.

## Problem

`multi_edit` was described as "atomic" but only had atomic validation — if validation passed and the write phase began, a write failure (disk full, permission change, IO error) would leave earlier files in the batch already modified. The error message claimed "no edits applied" even though partial writes had already landed on disk.

Furthermore, the failing file itself was excluded from rollback because the tracking array was pushed *after* `fs.writeFile` — a file that was truncated or partially written before the error was not restored.

## Fix

Three changes to [`src/tools/fs/edit.ts`](src/tools/fs/edit.ts:119-148):

1. **Push to the tracking array *before* `fs.writeFile`** — ensures the current file is also rolled back if truncation or partial content was written before the error.
2. **Use original content from validation, not a re-read** — `FileState.before` is set at initial `fs.readFile` during validation; rollback uses it directly, eliminating the TOCTOU window between validation and snapshot re-read.
3. **Reverse-order rollback** — most-recently-attempted files are restored first, conventional ordering for batch recovery.

Supporting changes:

- [`src/tools/filesystem.ts`](src/tools/filesystem.ts:684) — tool description updated from "atomic call" to "best-effort rollback of files that may have been modified".
- [`src/code/prompt.ts`](src/code/prompt.ts:168) — system prompt updated to reflect the same semantics.
- [`tests/filesystem-tools.test.ts`](tests/filesystem-tools.test.ts:1188-1230) — new test simulates partial write to `b.txt` followed by failure, asserts both `a.txt` and `b.txt` are restored to original content.

## Verification

| Check | Result |
|---|---|
| `npm run lint` | ✅ 625 files, no fixes applied |
| `npm run test` | ✅ 3227 passed, **0 failed** (filesystem-tools: 121/121 including new rollback test) |

### DIFF
--- a/src/code/prompt.ts
+++ b/src/code/prompt.ts
@@ -165,7 +165,7 @@ Rules:
     >>>>>>> REPLACE
 - Do NOT use write_file to change existing files — the user reviews your edits as SEARCH/REPLACE. write_file is only for files you explicitly want to overwrite wholesale (rare).
 - Paths are relative to the working directory. Don't use absolute paths.
-- For multi-site changes — same file or across files — prefer \`multi_edit\` over N \`edit_file\` calls. Shape: \`{ edits: [{ path, search, replace }, ...] }\`. All edits validate before any file is written; any failure → ALL files untouched. Per-file edits run in array order, so a later edit can match text inserted by an earlier one.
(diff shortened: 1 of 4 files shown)


esengine:
Thanks for catching this — the "atomic" claim on `multi_edit` really was overstated, and the missing rollback of the failing file itself is a sharp catch.

This PR conflicted with #1323 (system prompt compression that landed an hour or so before) and the new `tests/prompt-budget.test.ts` from #1320 caught the prompt growth that pushed it red. Rebased your three changes onto current main and folded the new rollback semantic into the now-compressed prompt phrasing — your original commit + authorship are preserved in #1327.

Closing this in favor of #1327; will get that merged as soon as CI confirms. Thanks again.

PR #1329, opened 2026-05-19
feat(slash): add command favorites with Ctrl+S toggle

Bernardxu123:
## Summary

Users can mark commands as favorites in the \`/\` slash menu by pressing \`Ctrl+S\` on the highlighted command. Favorited commands appear first (with a ★ marker) and persist across sessions.

## Design

- **Storage:** \`~/.reasonix/favorites.json\` (global, cross-project) — a \`string[]\` of command names
- **Interaction:** \`Ctrl+S\` toggles favorite on the currently selected slash command in the \`/\` menu
- **Sorting:** Favorited commands appear first (preserving group order), unfavorited ones follow (sorted by usage frequency)
- **UI:** ★ appended to the right of favorited command rows

## Files changed

| File | Change |
|------|--------|
| \`src/cli/ui/slash/commands.ts\` | \`suggestSlashCommands\` accepts optional \`favorites\` array; favorited commands prioritized in all three sort modes |
| \`src/cli/ui/App.tsx\` | Load/save \`favorites.json\`, Ctrl+S keystroke handler, pass favorites to suggestions |
| \`src/cli/ui/SlashSuggestions.tsx\` | ★ marker on favorited rows |
| \`src/cli/ui/ComposerArea.tsx\` | Passes \`favorites\` prop through |
| \`tests/slash-favorites.test.ts\` | 5 tests covering sort logic (group mode, search mode, counts, no favorites, unknown commands) |

## Demo

\`\`\`
/                           ← 用户输入 /
  ★ help    查看所有命令    ← Ctrl+S 收藏的
  ★ compact 折叠旧轮次
  ★ feedback 提交反馈
  ─────────────────
    new     新会话          ← 按使用频率排
    retry   重发
    ...
\`\`\`

### DIFF
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -1,5 +1,8 @@
 import { type WriteStream, statSync } from "node:fs";
+import { existsSync, mkdirSync, readFileSync, renameSync, writeFileSync } from "node:fs";
+import { homedir } from "node:os";
 import { relative, resolve } from "node:path";
+import { dirname, join } from "node:path";
 import { Box, Text, useStdin, useStdout } from "ink";
 import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import {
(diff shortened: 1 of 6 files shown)


esengine:
Thanks for the work — code is clean, tests are tight, and the Ctrl+S interaction is well-thought-out. Holding off on this one though.

The slash menu already sorts by usage frequency (`suggestSlashCommands` in `src/cli/ui/slash/commands.ts:387` takes a `counts` argument and floats most-used commands to the top). For the "fast access to commands I use a lot" use case that's the more honest signal — usage actually reflects what you use, whereas a favorites list reflects what you *thought* you'd use back when you starred it and have to maintain by hand.

A few concrete concerns:

- **Discoverability of Ctrl+S is near-zero** — there's no UI hint, no footer key legend entry for it, no help text. Users who'd benefit (power users) have already memorized command names and just type `/he<Tab>` for `/help`. New users won't find it.
- **Two sort dimensions are noisier than one** — favorites + frequency interleaving means a command you starred six months ago and never use sits above one you use daily. The frequency-only ordering self-heals as habits change.
- **`Refs #743`** is a layout/rendering issue with a screenshot, not a request for favorites. So the PR is contributor-imagined rather than user-driven, which is the wrong starting point for a new persistent surface.
- **New `~/.reasonix/favorites.json`** adds a cross-session state file with no clear migration / reset path. Marginal benefit, permanent maintenance surface.

If you want to invest in slash-menu UX, the higher-leverage moves would be:

1. **Tighter fuzzy matching** so `/cmpt<Tab>` resolves to `/compact`, `/fb<Tab>` to `/feedback`. Currently typos send you back to the menu.
2. **Aliases for the verbose commands** — `compact` already has none. Add `/c`, `/r` (retry), `/fb`, etc. as opt-in shortcuts. Zero new persistent state, just a static map.
3. **Footer hint that shows the top-3 frequency commands** the first time a user opens the menu in a session — surfaces the existing learning without a new mechanism.

Any of those would land easier and benefit more users. Happy to take a PR on (1) or (2) any time.

Closing this for now under the rail above. The code is good work — the question is just "does this surface earn its keep", and I don't think a favorites system clears that bar when the existing frequency sort is already covering 80% of the case.

Issue #1317, opened 2026-05-19
/qq命令不存在；设置显示成功连接QQ通道，访问控制未绑定；无法使用QQ机器人


3246203428:
> **Screen flicker, garbled output, leftover artifacts, cursor jumping?**
> Use the **Display / rendering issue** template instead — it asks for the
> terminal-specific info we need to diagnose those.

**What happened**
A clear and concise description.

**Expected**
What you expected to happen.

**Reproduction**
Steps or minimal code that reproduces it.

**Environment**
- Reasonix version (`reasonix --version`):
- Node version (`node --version`):
- OS (Windows 11 / macOS 14 / Ubuntu 24.04 / …):
- Shell (bash, zsh, fish, PowerShell 7, PowerShell 5.1, cmd, …):
- Terminal app (Windows Terminal, iTerm2, Alacritty, kitty, WezTerm, **VSCode integrated**, **Cursor integrated**, Hyper, …):
- DeepSeek model (e.g. `deepseek-v4-flash`, `deepseek-v4-pro`):

**Logs / transcript**
If using the CLI, attach the relevant chunk of `--transcript`, or run
`reasonix doctor` and paste the output.

esengine:
Thanks for the report — confirmed. The desktop QQ settings entry (#1241) added a UI that *says* "Connected" but never actually starts a QQChannel in the desktop process. The bot only ever runs inside the CLI (`reasonix` in a terminal), which auto-starts the channel based on the `enabled` config flag.

That's why all four symptoms line up:

- "Settings shows successfully connected" — the desktop UI was reporting `connected: true` based purely on the config flag, not a real connection.
- "Access control unbound" — no first-sender binding can happen because no bot is actually running in the desktop process.
- "Bot doesn't respond" — same root cause: no live channel.
- "/qq command doesn't exist" — slash commands are CLI-only; the desktop chat box doesn't dispatch them. (Separate concern, but explains why `/qq` wouldn't work from the desktop window.)

Fix in #1326 — desktop is honest about being a config editor:

- Status label changes from "Connected / Disconnected" to **Enabled (CLI)** / **Configured · disabled** / **Not configured**.
- A new note above the QQ section explains: the bot runs in `reasonix` terminal sessions, not in the desktop window. Save credentials here, then run `reasonix` in a terminal to start the bot.
- Button labels: "Connect" → **Enable for CLI**, "Disconnect" → **Disable**.

[...]

PR #1334, opened 2026-05-19
feat(desktop): make QQ sessions usable from desktop

wade19990814-hue:
## Summary
- build on #1326 by turning desktop QQ from a config-only surface into a real active-session runtime
- keep the tighter settings UX, with a top-level state toggle, Configure action, greener connected state, and clearer QQ copy
- wire QQ into the active desktop tab so inbound QQ messages appear in the desktop transcript and desktop replies route back to QQ

## Context
- #1317 surfaced the product gap: desktop QQ looked connected but only edited config
- #1326 stopped desktop from overstating that capability
- this PR closes the underlying gap by making the desktop QQ runtime real again

## Testing
- npm test -- --run tests/desktop-qq-settings-ui.test.ts tests/desktop-user-message.test.ts tests/desktop-qq-config.test.ts
- npm run typecheck
- npm run lint
- npm run verify


wade19990814-hue:
Context for reviewers: #1317 surfaced that desktop QQ was only acting as a config editor while the UI implied a live connection. #1326 corrected that by stopping desktop from claiming “Connected” without a real runtime behind it.

This PR builds on that baseline rather than undoing it. It keeps the tighter settings surface, but also wires QQ into the active desktop session so the connected state becomes real again: desktop now starts/stops a real QQ runtime, QQ messages enter the current active tab, inbound user messages are shown in the desktop transcript, and desktop replies route back to QQ.

So the relationship is:
- #1317: bug report / product gap
- #1326: stop overstating desktop QQ capability
- this PR: close the underlying gap by making desktop QQ actually usable

wade19990814-hue:
@esengine Small product signal: #1343 was opened right after this landed in front of users, and it lines up with the same gap this PR is addressing. The current desktop surface is discoverable enough that people are already trying to configure QQ there, but they then hit the exact mismatch from #1317: config is present, yet active-session QQ runtime is still missing on main.

So in addition to the code path itself, #1343 is a concrete sign that this is not just internal cleanup; users are already reaching for desktop QQ and interpreting the current surface as a usable integration entrypoint.

esengine:
Thanks @wade19990814-hue — this is a solid piece of work and a good direction for us. Glad you went after the underlying gap from #1317 instead of stopping at the labelling fix. I'll do a small follow-up PR to re-add the augmentProcessPath() call that #1331 introduced (it got dropped in your rebase, unrelated to the QQ work — no need to touch it on your side). Looking forward to more contributions like this; we want the IM bridge surface to grow (QQ today, room for Feishu / Discord / etc later) and having maintainers in that space is exactly what we need.

PR #1376, opened 2026-05-20
feat(desktop): 添加自定义字体与文件链接优化

summer1874:
## Summary
- Add a custom font-family option in desktop settings, with the input shown only when custom is selected.
- Improve rendered file path links so local paths and line references open through the configured editor.
- Restyle inline file links to avoid the boxed pill appearance.

## Testing
- `cd desktop && npm run build`
- `git diff --check`

Note: the full repository pre-push hook timed out twice locally, so this branch was pushed with `--no-verify` after the targeted desktop build passed.


esengine:
Thanks @summer1874 — really clean first PR. A few things I want to call out:

- The three changes split cleanly along feature lines (custom font / file-link parsing / pill restyle), and the new fontFamily option threads through the existing `App → TabRuntime → SettingsModal → PageGeneral` prop chain the same way `fontScale` does. Easy to follow.
- The file-path regex going from a single relative-only shape to four (Windows drive paths, absolute Unix, relative, bare filename) is a meaningful upgrade — and I see you respected the macOS Safari < 16.4 lookbehind ban from #1209. Nice.
- `withFilePills` skipping `<a>` / `<code>` / `<pre>` is the right guard against double-processing inside code blocks. Good catch.
- `file://` URL handling — `decodeURI` + Windows drive prefix strip + `#L42` line hash — covers real cases that came up in #1209 / #1314 discussions.

About the `--no-verify` note: that's fine in this case since CI ran the full verify on both Ubuntu and Windows and all passed. If you keep hitting local timeouts, `npx vitest run --pool=forks --poolOptions.forks.singleFork` sometimes helps on slower disks. No need to do anything about it.

Looking forward to more.

PR #1405, opened 2026-05-20
Fix desktop session loading hang from partial stdout writes

xingxingzhou469-max:
Fixes desktop kernel JSON event corruption when large `$session_loaded` payloads are written to stdout.

`emit()` previously assumed `writeSync` writes the whole buffer in one call. For large history sessions, stdout can partially write or throw `EAGAIN`, which corrupts the newline-delimited JSON stream and causes the desktop UI to hang while connecting to the kernel.

This change writes the full buffer in a loop and retries `EAGAIN`.

Verification:
- `npm test -- tests/desktop-stdout-write.test.ts`
- `npx tsc --noEmit --pretty false`
- `npx biome check src/cli/commands/desktop.ts tests/desktop-stdout-write.test.ts`
- `npm run build`
- pre-push `npm run verify`

### DIFF
--- a/src/cli/commands/desktop.ts
+++ b/src/cli/commands/desktop.ts
@@ -439,9 +439,40 @@ type EmittableEvent =
   | MemoryEvent
   | JobsEvent;
 
+const STDOUT_BACKPRESSURE_WAIT = new Int32Array(new SharedArrayBuffer(4));
+
+type SyncWriter = (fd: number, buffer: Buffer, offset: number, length: number) => number;
+
+export function writeAllSync(
+  fd: number,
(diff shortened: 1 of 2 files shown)


esengine:
Thanks for tracking this down — partial writes from `writeSync` on large `$session_loaded` payloads are exactly the kind of bug that's a pain to reproduce, and the injectable `write`/`wait` design made the test surface really clean. Merged.

I pushed two tiny polish items as #1412 (1-line JSDoc on `writeAllSync`, drop a redundant `?.` on the caught error) — no behavior change, just style. Looking forward to more contributions.

PR #1414, opened 2026-05-20
fix: reject ambiguous search-replace edit blocks

YoungDan-hero:
Refuse text edit blocks when the SEARCH content matches more than once in the target file. This keeps the `/apply` path aligned with `edit_file` behavior and prevents silently modifying the first matching occurrence when the model did not provide enough context.

Update the edit block test to assert that ambiguous SEARCH text leaves the file unchanged.

## Checklist


esengine:
Nice catch — `edit_file` and `multi_edit` already reject ambiguous SEARCH with this exact wording; the `/apply` path was the inconsistent one. The test inversion makes the contract uniform across both code paths. Merged.

Issue #410, opened 2026-05-08
RFC: Chat-platform bridge (Feishu / WeWork / WeChat)


esengine:
**Status:** Draft, not yet implementing. Surfacing the design space and a positioning tension before any code is written.

## Motivation

Two recurring asks from CN users:

1. *Run Reasonix from a chat client* — paste a task into Feishu / WeWork / WeChat, get the agent's reply back in the same thread, including tool calls and approvals.
2. *Pull team knowledge into context* — read a Feishu Wiki page, a WeWork message thread, or a WeChat group discussion as `@`-mentioned context for a turn.

Both are cousins of \"connect Reasonix to where the team already works.\" Different surfaces, different integration shapes, very different security models — they should not be conflated.

## Positioning check

Reasonix is a terminal-first DeepSeek-native coding agent. Adding chat bridges nudges the project toward general-purpose chatbot territory, which is *not* what the cache-first loop, cell-diff renderer, or filesystem sandbox are designed for.

Two sane stances, both consistent with the project's positioning:

- **A. Bridges live outside core.** Reasonix does *not* ship Feishu/WeChat code. We document the MCP server pattern so users can wire it themselves; the core stays a coding-agent CLI.
- **B. Bridges are first-class for *one* surface (probably Feishu Wiki / WeWork docs).** The use case is read-only context ingest, not bidirectional bot. Slash commands (`/wiki <url>`) and `@feishu:doc-id` mentions, no message-sending, no bot identity.

Stance C — *full bidirectional bot* — almost certainly belongs in a separate project. It's a different product.

## Design space

| Surface | Reasonix → chat | Chat → Reasonix | Common today |
|---|---|---|---|
| **Feishu (Lark)** | post message, send card | webhook user message in | open APIs, OAuth, decent docs |
| **WeWork (企业微信)** | bot message, mini-app card | webhook from group | enterprise-only, app secret per tenant |
| **WeChat personal** | — | — | **bots on personal accounts violate ToS**; do not ship |
| **WeChat Public Account** | template message | server message in | requires verified account, slow approval |

Personal-WeChat support is a non-starter regardless of stance — anyone shipping that would be one Tencent ban-wave from a dead user base.

## Three integration shapes (orthogonal, not exclusive)

[...]

PR #1490, opened 2026-05-21
feat: add Feishu private chat channel

wade19990814-hue:
## Summary
- add a CLI/TUI Feishu private chat channel alongside the existing QQ channel
- support `/feishu connect|status|disconnect` with staged first-time setup and persisted config
- align QQ and Feishu remote slash result handling so synchronous text responses are sent back to the active remote chat

## Details
- add `src/feishu/*` for Feishu bot transport, access control, channel routing, and host integration
- extend config and slash surfaces for Feishu App ID / App Secret based setup
- keep the implementation CLI-only: private chat text in/out, owner/allowlist/runtime-bind access, no cards or desktop settings
- add a small shared remote-channel helper so QQ and Feishu follow the same remote slash reply path

## Verification
- `npm run lint`
- `npm run typecheck`
- `npm test -- --run tests/feishu-access.test.ts tests/feishu-config.test.ts tests/feishu-channel.test.ts tests/feishu-slash.test.ts tests/feishu-first-connect.test.tsx tests/qq-access.test.ts tests/qq-config.test.ts tests/qq-channel.test.ts tests/qq-slash.test.ts tests/qq-first-connect.test.tsx tests/remote-channel.test.ts`
- `npm run verify`


wade19990814-hue:
Issue linkage:
- Closes #1447
- Advances #410, but does not fully implement that RFC

What this PR actually ships:
- a **CLI/TUI-only Feishu private-chat channel** alongside the existing QQ channel
- Feishu private text -> current Reasonix session
- final assistant reply -> back to the active Feishu private chat
- `/feishu connect|status|disconnect`
- owner / allowlist / first-sender runtime binding
- aligned synchronous remote slash reply behavior between QQ and Feishu

What this PR does **not** ship:
- desktop settings UI
- group chat / @bot flows
- Feishu cards
- file/media upload
- a larger multi-platform adapter framework

So the scope is intentionally narrow: it resolves the direct “please support a Feishu channel” ask in #1447, while only moving part of the broader design space discussed in #410.

Suggested next step after this PR:
Reasonix itself should become aware that QQ and Feishu are not just text relays, but **remote channels with explicit capabilities**. Right now the model mostly experiences them as inbound/outbound text. The next useful step is to expose channel-awareness in the agent/runtime layer, for example:
- knowing when the current turn originated from QQ vs Feishu
- knowing that it can send text replies back through that channel
- later, knowing whether the active channel supports richer actions such as file/media delivery

That would make the remote-channel experience much less implicit and would create a clean path for future work like file sending, richer remote slash behaviors, or channel-specific UX without forcing a large adapter rewrite first.

Also, this branch was implemented and verified directly on top of the current upstream `main` (`9e46fe4`).

esengine:
Thanks for the work here, @wade19990814-hue — the access-control module is clean and the QQ/Feishu de-duplication via `remote-channel.ts` is a real improvement. I'm going to close this without merging though, and I want to be upfront about why so it doesn't feel arbitrary.

This collides with the direction I committed to on the chat-bridge RFC (#410) four days ago. Quoting the relevant bits:

- **External bridge is the official path.** @yes999zc landed end-to-end Lark/Feishu WebSocket → Reasonix ACP session → reply via cc-connect's `type = "reasonix"` adapter ([#410 comment](https://github.com/esengine/DeepSeek-Reasonix/issues/410#issuecomment-4469315342)). I responded that this "basically resolved this RFC for us."
- **No always-on daemons in core.** From the November positioning update on #410: *"the design has to be strictly event-driven — webhook in → one turn → done — never a watching loop."* The `Lark.WSClient` long-lived WebSocket + PID lockfile in this PR is exactly the always-on shape the RFC ruled out.

QQ shape-3 lives in core for historical reasons (it shipped before the RFC settled). That's a precedent I'm not extending — adding a second in-tree CN-messenger bot four days after publicly pointing users at cc-connect would send a confusing signal about where we're going.

**What I'd suggest instead:**

1. Try cc-connect's Reasonix adapter — `type = "reasonix"` with `args = ["acp"]`. yes999zc has verified Lark p2p in/out works end-to-end on the [`feat/reasonix-acp` branch](https://github.com/yes999zc/cc-connect/tree/feat/reasonix-acp), pending upstream merge in [chenhg5/cc-connect#1027](https://github.com/chenhg5/cc-connect/pull/1027).
2. If cc-connect is missing something for the Feishu workflow you wanted, that's a far higher-leverage place to contribute — it'll cover every messenger they support, not just Feishu.
3. If you spot a real gap in Reasonix's ACP surface that blocks the bridge from working well, a focused issue on that gap is welcome.

The access-control / config / dedup helpers in this branch are reusable if you ever want to lift them into a cc-connect Reasonix-specific module. Sorry for not catching the direction conflict earlier — that's on me, not you.

PR #1501, opened 2026-05-21
feat (chat): add queued mid-turn steer handling

donle:
## Summary

**This PR adds queued mid-turn steering for busy turns.**

When a turn is already running, users can now type ordinary prompt text to steer the current task instead of waiting for the turn to finish. The steer text is not executed immediately and does not interrupt the in-flight model request. It is queued and applied at the next loop iteration boundary.

  ## Motivation

Previously, busy turns blocked prompt input. The first local change opened the input while busy, but it had two problems:

  - busy-time submissions could be confused with normal commands such as `/new` or `/clear`
  - multiple steer submissions could overwrite each other because the loop only held one pending steer value

This PR makes the behavior explicit:

  - busy input means “steer the current turn”
  - commands are not accepted as steer input
  - multiple steer messages are preserved and consumed in order

  ## Execution Flow

  During a normal idle submit, prompt handling is unchanged.

  During a busy turn:

  1. The TUI/Dashboard input remains editable.
  2. The submitted text is checked before it reaches the loop.
  3. Command-like input is rejected while busy:
     - slash commands, such as `/new`, `/clear`, `/model`
     - shell-style commands, such as `!ls`
     - memory commands, such as `#note`
  4. Ordinary text is accepted as a steer message.
  5. The loop appends that steer text to an internal FIFO queue.
  6. The currently running model request or tool execution continues untouched.
  7. At the next model-iteration boundary, the loop consumes one queued steer.
  8. That steer is added to the conversation as user-role guidance.
  9. If more steer messages are queued, the loop keeps running and consumes the next one at the following iteration boundary.

  This guarantees that steer input affects only future request payloads, never the request that is already in flight.

  ## Queue Semantics

  The loop now stores pending steer messages in a FIFO queue instead of a single pending value.

  This changes the behavior from “last steer wins” to “all accepted steers are applied in order”.

  Example:

  1. The model is running a tool.
  2. User submits `focus on the frontend issue`.
  3. User submits `also check mobile layout`.
  4. The first steer is consumed after the current step.
  5. The second steer is consumed at the next iteration boundary.

  No steer message is silently overwritten.

  ## UI Behavior

  The TUI and Dashboard prompt areas now communicate the busy-steer mode directly:

  - busy prompt input stays editable
  - placeholders explain that the user is steering the current task
  - slash/picker behavior is disabled while busy
  - accepted steer messages are cleared from the input
  - consumed steer messages are mirrored into the Dashboard conversation stream

  The Changes panel is also aligned with the main Chat panel: busy input can now actually submit steer text instead of appearing editable while the send path silently returns.

  ## Tests

  The loop tests now cover:

  - steer text is persisted as wrapped user-role guidance
  - multiple steer messages are queued
  - queued steer messages are consumed in order
  - one steer is consumed per loop iteration boundary
  - existing steer-consumption state still resets correctly between turns

### DIFF
--- a/dashboard/src/i18n/en.ts
+++ b/dashboard/src/i18n/en.ts
@@ -164,6 +164,8 @@ export const en = {
     modeView: "session view",
     placeholder: "Type a prompt — Enter sends, Shift+Enter for a newline · / @ for pickers",
     placeholderBusy: "wait for the current turn to finish…",
+    placeholderSteerBusy: "type to steer — commands and pickers are disabled while busy",
+    steerHint: "message steers the current turn (added as instruction)",
     send: "Send",
     new: "New",
     clear: "Clear",
--- a/dashboard/src/i18n/zh-CN.ts
+++ b/dashboard/src/i18n/zh-CN.ts
@@ -164,6 +164,8 @@ export const zhCN = {
     modeView: "会话视图",
     placeholder: "输入提示词 — Enter 发送，Shift+Enter 换行 · / @ 打开选择器",
     placeholderBusy: "请等待当前轮次完成…",
+    placeholderSteerBusy: "输入消息以引导当前任务 — 忙碌时不支持命令和选择器",
+    steerHint: "消息将作为附加指示引导当前轮次",
     send: "发送",
     new: "新建",
     clear: "清除",
--- a/dashboard/src/panels/changes.ts
+++ b/dashboard/src/panels/changes.ts
@@ -1716,6 +1716,10 @@ function ChatPane(props: ChatPaneProps) {
 
   const updatePopover = useCallback(
     async (text: string) => {
+      if (busy) {
+        setPopoverKind(null);
+        return;
+      }
       const slashMatch = /^\/([A-Za-z0-9_-]*)$/.exec(text);
       if (slashMatch) {
         const prefix = slashMatch[1]!.toLowerCase();
@@ -1734,7 +1738,7 @@ function ChatPane(props: ChatPaneProps) {
       }
       setPopoverKind(null);
     },
-    [slashCommands],
+    [busy, slashCommands],
   );
 
   const applyPopover = useCallback(() => {
@@ -1757,7 +1761,6 @@ function ChatPane(props: ChatPaneProps) {
 
   const send = useCallback(async () => {
     const text = input.trim();
-    if (busy) return;
     if (!text && props.comments.length === 0) return;
     setError(null);
 
@@ -1783,7 +1786,7 @@ function ChatPane(props: ChatPaneProps) {
     } catch (err) {
       setError((err as Error).message);
     }
-  }, [input, busy, props.comments]);
+  }, [input, props.comments]);
 
   const abort = useCallback(async () => {
     try {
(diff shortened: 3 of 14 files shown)


esengine:
Thanks for taking this on, @donle — the core idea is a real improvement and I want to land it. The loop-side FIFO is clean, the six exit-path queue-clears are right, and the `repairedCalls.length === 0 && queue.length > 0 → continue` patch is exactly the bug that would have orphaned queued steers otherwise. Test for "two steers consumed in order" is the one I would have asked for.

A few things to address before I merge:

**1. Comments — code-as-documentation is the project rule.** A handful of new comments restate what the code already says, which the `tests/comment-policy.test.ts` gate doesn't always catch but I do enforce manually. Specific spots to delete or compress to one why-line:

- `src/cli/ui/App.tsx` ~L2671-2683: the 5-line "Busy with an active turn: inject text as a mid-turn steer message…" block — the call site `loop.steer(text)` already documents itself. Delete the comment.
- `src/cli/ui/App.tsx` ~L3375: "The ghost preview from handleSubmit's steer branch already confirmed…" — delete; the empty branch is self-explanatory in context.
- `src/loop.ts` ~L702-704: "Consume one queued steer between model calls. This never mutates the currently running request…" — the loop is short enough to read; if anything stays, one line explaining *why* we consume between calls (not what).
- `src/cli/ui/PromptInput.tsx` L46-48 and `src/cli/ui/ComposerArea.tsx` L51: internal interface fields don't get JSDoc — the project rule is JSDoc only on `src/index.ts` re-exports. Names already say enough.
- `src/cli/ui/PromptInput.tsx` ~L133: the `inputFrozen` / `inputActive` derivation is self-evident from the names; the two-line block above is redundant.
- `dashboard/src/panels/chat.ts` ~L773-774: drop the "When busy, the server steers the text…" comment — the call site does the work.

**2. The wrapper string is mine to author.** This text:

> `[Mid-turn steer queued by the user. Do not treat this as a new task; use it only as additional guidance for the current task after completing the current step.]`

is a prompt-engineering choice that affects model behavior across every steer, and I'd like to land that line myself in a follow-up after this PR merges. For this PR, please either:

- (a) Hoist the wrapper into a single exported constant (e.g. `MID_TURN_STEER_WRAPPER`) at the top of `src/loop.ts` so it's the one line I edit later, or
- (b) Drop the wrapper for now — inject the raw text as a user message and let me layer the wrapper in once we've decided whether it should be user-role text vs. a structured marker the model recognizes.

I lean (a). Either is fine.

**3. `pendingUser` const-ification.** Confirmed: on `main`, `pendingUser` is initialized null at step entry, never reassigned to non-null anywhere in the function, but still read at `loop.ts:741` (`allowEmpty: pendingUser !== null`) and `loop.ts:744`. The `let → const` is behaviorally equivalent today, but the variable's existence implies there used to be a non-null path that got refactored out. Please either:

- Leave `let` and add a one-line `// TODO(#nnnn): pendingUser is currently always null; either wire it back or remove` after opening a tracking issue, **or**
- Remove `pendingUser` entirely from `step()` plus the call sites at L706/L716/L741/L744 (collapse to `null` literal) — clearer than leaving dead state.

**4. Open a tracking issue** for "mid-turn steer queue", and link it from the PR. The interaction primitive deserves a public design pointer for the next person who touches this surface — busy-input behavior is something I get asked about. One paragraph is enough; the PR description already has most of the material.

Once 1-4 are addressed I'll merge. CI side is already in great shape — just substance and policy.

donle:
Addressed the local review items:
- removed the redundant comments called out in the review
- hoisted the steer wrapper into `MID_TURN_STEER_WRAPPER`
- removed the always-null `pendingUser` path instead of leaving dead state

Opened follow-up tracking [issue](https://github.com/esengine/DeepSeek-Reasonix/issues/1531) ^

PR #1512, opened 2026-05-22
fix(agent): improve behavior stability for long-running sessions (#1372, #1462, #1495)

### LINKED ISSUE #1372: 有时特别爱用cp、echo等命令，而非自带的技能去编辑文件
在存在大量命令操作的环境下，用着用着，就弹出`echo``cp`等命令在工作目录编辑文件，然后问我能否放行。希望可以在底层让AGENTS尽量用`multi_edit``edit_file``copy_file`等，这也有沙盒限制。

### LINKED ISSUE #1462: 我在项目中设置好了全局记忆，但是发现几乎不会优先遵守怎么回事？
我设置了一些工具调用规范，但总是会优先尝试然后失败才找其他办法。
还有一些工作规范，比如优先质量的提示，但ai思考时总是会导向快速处理。

我询问ai是否加载了全局记忆，回复是是，但是只有我提醒之后才按照记忆进行工作。

### LINKED ISSUE #1495: 使用agent进行长段文本工作时完全无法好好工作，长时间10分钟左右就开始自行在思考中以加速、节省token、效率为由乱改任务要求执行
memory.md内的要求也基本上无视，只有再次输入提醒到了才会执行要求

paradoxSCH:
## Summary

This PR addresses three related Agent behavior stability issues observed during long-running sessions:

1. **#1372** — Agent uses `cp`/`echo` shell commands instead of dedicated file editing tools
2. **#1462** — Global memory (`memory.md`) constraints are lost after context folding
3. **#1495** — Agent drifts from original user objective after ~10 minutes of long-text work

### #1372: Discourage shell-based file edits

**File:** `src/tools/shell.ts`

Added an explicit negative constraint to the `run_command` tool description:

> DO NOT use run_command for file operations — use write_file, edit_file, multi_edit, copy_file, move_file, or delete_file instead. Shell utilities (echo, cp, sed, cat, tee, perl, python -c, etc.) bypass validation, lack rollback, and will trigger user confirmation gates that waste turns.

This is a prompt-level fix (scheme C from analysis). The tool description is the primary signal the LLM uses for tool selection; making the exclusion explicit reduces the likelihood that the model falls back to shell commands for file manipulation.

### #1462: Preserve memory constraints across context folds

**Files:** `src/context-manager.ts`, `src/loop.ts`

Two layers of defense:

1. **Summary prompt strengthening** (`summarizeForFold`): The summarizer's system prompt now explicitly asks to preserve "the user's ORIGINAL OBJECTIVE" and "all 'do not' / 'never' / 'avoid' instructions". Previously, negative constraints were the most likely to be dropped during summarization.

2. **Verbatim constraint pinning** (`fold`): After generating the summary, high-priority memory, user memory, and project memory blocks are extracted from the system prompt and appended verbatim to the fold summary message. This ensures constraints survive the fold and remain visible in the conversation tail.

**Note on alternative approach:** A runtime `ToolInterceptor` (scheme C) was considered for #1462. It would intercept `run_command` calls that match file-edit patterns (echo, cp, sed, etc.) and return a hard rejection. This is more invasive but could be layered on top if prompt-level fixes prove insufficient. Deferred to keep the PR focused.

### #1495: Add task integrity constraint to system prompt

**File:** `src/code/prompt.ts`

Added a new section to `CODE_SYSTEM_TEMPLATE`:

> # Task integrity — non-negotiable
> The user's original objective and ALL constraints (especially "do NOT do X", "avoid Y", "never Z") remain in force for the entire session. You may NOT unilaterally simplify, narrow, or change the objective to save tokens, time, or steps. If you believe the objective needs adjustment, ask the user — do NOT decide on your own.

The escalation contract's "Cost-aware" wording was intentionally **retained** (not changed to "Capability-aware") per review feedback, as the task-integrity section alone provides sufficient guardrails without altering the escalation behavior.

## Cache First compatibility

All changes are **static** modifications to system prompts and tool descriptions. They do not alter the `ImmutablePrefix` at runtime, so there is no ongoing impact on DeepSeek prefix cache hit rates. Only a one-time cache miss occurs on the first request after upgrading.

The #1462 fold changes operate on the `AppendOnlyLog` (conversation history), which is explicitly documented as "Tail-only mutation; prefix cache safe".

## Test plan

- [x] `npm run test` — 237 test files, 3333 tests passed
- [x] `npm run lint` — biome check clean
- [x] `npm run typecheck` — no new TypeScript errors in modified files

## Related issues

- Fixes #1372
- Fixes #1462
- Fixes #1495


paradoxSCH:
Closing to split into three independent PRs per issue (#1372, #1462, #1495).

Issue #1493, opened 2026-05-21
Some key sequences do not seem to properly reach reasonix (Shift Enter / Ctrl O)


dimitrisfasoulas:
> **Screen flicker, garbled output, leftover artifacts, cursor jumping?**
> Use the **Display / rendering issue** template instead — it asks for the
> terminal-specific info we need to diagnose those.

**What happened**
This happens in Arch witj wayland and hyprland and in ubuntu with same setup
I'm using zsh shell and kitty as terminal.
When working with reasonix, pressing shift + enter send the promt instead of wrapping line (work around is ctrl + j)
When pasting a few lines I get  ▸   📋 pasted  8 lines · 941B  ·  text  ^O expand · ⌫ remove  
But pressing Ctrl+O does nothing (no work around found)
**Expected**
What you expected to happen.
I expect shift + enter to wrap line, not send the prompt 
I expect ctrl + O to wrap line, not send the prompt

**Reproduction**
Steps or minimal code that reproduces it.

**Environment**
- Reasonix version (`reasonix --version`):0.48.1
- Node version (`node --version`):v24.14.0
- OS (Windows 11 / macOS 14 / Ubuntu 24.04 / …):Arch / Ubuntu
- Shell (bash, zsh, fish, PowerShell 7, PowerShell 5.1, cmd, …):Zsh
- Terminal app (Windows Terminal, iTerm2, Alacritty, kitty, WezTerm, **VSCode integrated**, **Cursor integrated**, Hyper, …):Konsole / Kitty
- DeepSeek model (e.g. `deepseek-v4-flash`, `deepseek-v4-pro`): both

**Logs / transcript**
If using the CLI, attach the relevant chunk of `--transcript`, or run
`reasonix doctor` and paste the output.
● doctor · 9 checks · 8 passed · 1 warn
✓ api key     set via env DEEPSEEK_API_KEY OK
✓ config      /home/dimi/.reasonix/config.json (preset=auto, editMode=auto, mcp=4) OK
✓ http proxy  no HTTPS_PROXY / HTTP_PROXY / ALL_PROXY set — direct connection OK
✓ api reach   /user/balance ok — 8.19 USD OK
✓ tokenizer   /home/dimi/.npm/_npx/a445a9f5043f7b46/node_modules/reasonix/data/deepseek-tokenizer.json.gz (1.7 MB) OK
✓ sessions    4 saved · 667.0 KB · oldest 1d OK
[...]

PR #1513, opened 2026-05-22
fix(tui): Ctrl+Enter inserts newline instead of submitting; update shortcut panel

Bernardxu123:
## Summary

Fix issue #1493: **Ctrl+Enter** now inserts a newline in the composer instead of submitting, matching the behavior of Shift+Enter and Alt+Enter. The shortcut panel is also updated to reflect the new keybindings.

### 1. multiline-keys.ts (behavior fix)

Added \key.ctrl\ to the newline condition in the \key.return\ branch. Previously only \key.shift\ and \key.meta\ were checked, so Ctrl+Enter fell through to submit. Now Ctrl+Enter, Shift+Enter, and Alt+Enter all insert a newline.

### 2. PromptInput.tsx (HintRow update)

Updated the HintRow to show \^\u2191\u21e7 Enter / ^J\ as the newline hint. \Ctrl+J\ is a terminal-agnostic newline shortcut that works regardless of modifyOtherKeys or Kitty protocol support, making it a reliable fallback for users on Wayland/Hyprland or other environments where Shift+Enter may not be properly detected.

### 3. ShortcutsHelpModal.tsx (shortcut panel)

Added two new entries to the Input shortcuts group:
- \Ctrl+Enter\ - New line
- \Ctrl+J\ - New line

### 4. i18n (EN.ts, zh-CN.ts, types.ts)

Added \descCtrlEnter\ and \descCtrlJ\ translation keys with appropriate English and Chinese descriptions.

## Behavior after fix

| Key | Behavior |
|-----|----------|
| Enter | Submit (unchanged) |
| Shift+Enter | New line (unchanged) |
| Ctrl+Enter | **New line** (fixed) |
| Ctrl+J | New line (unchanged, reliable fallback) |
| Alt+Enter | New line (unchanged) |

## Compatibility

- **macOS**: All terminals (iTerm2, Terminal.app, etc.) - works via modifyOtherKeys
- **Linux**: kitty/Konsole with modifyOtherKeys - works via modifier encoding
- **Linux**: Wayland/Hyprland - falls back to Enter (submit) if terminal doesn't send mod sequences; users can use Ctrl+J as reliable alternative
- **Windows**: Windows Terminal - works via modifyOtherKeys

Refs #1493

### DIFF
--- a/src/cli/ui/PromptInput.tsx
+++ b/src/cli/ui/PromptInput.tsx
@@ -384,7 +384,7 @@ export function PromptInput({
 export function HintRow(): React.ReactElement {
   const items: Array<{ key: string; tKey: string }> = [
     { key: "\u23ce", tKey: "composer.hintSend" },
-    { key: "\u21e7\u23ce", tKey: "composer.hintNewline" },
+    { key: "\u21e7\u23ce / ^J", tKey: "composer.hintNewline" },
     { key: "^U", tKey: "composer.hintClear" },
     { key: "\u2191\u2193", tKey: "composer.hintHistory" },
     { key: "esc", tKey: "composer.hintAbort" },
--- a/src/cli/ui/ShortcutsHelpModal.tsx
+++ b/src/cli/ui/ShortcutsHelpModal.tsx
@@ -20,6 +20,8 @@ const SHORTCUT_GROUPS: ShortcutGroup[] = [
     items: [
       { keys: "Enter", descKey: "shortcutsHelp.descEnter" },
       { keys: "Shift+Enter", descKey: "shortcutsHelp.descShiftEnter" },
+      { keys: "Ctrl+Enter", descKey: "shortcutsHelp.descCtrlEnter" },
+      { keys: "Ctrl+J", descKey: "shortcutsHelp.descCtrlJ" },
       { keys: "Ctrl+U", descKey: "shortcutsHelp.descCtrlU" },
       { keys: "Ctrl+W", descKey: "shortcutsHelp.descCtrlW" },
       { keys: "Ctrl+P", descKey: "shortcutsHelp.descCtrlP" },
--- a/src/cli/ui/multiline-keys.ts
+++ b/src/cli/ui/multiline-keys.ts
@@ -187,7 +187,8 @@ export function processMultilineKey(
   }
 
   if (key.return) {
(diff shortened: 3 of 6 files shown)


esengine:
Thanks @Bernardxu123 — the Ctrl+J discoverability piece (HintRow + ShortcutsHelpModal + i18n) is exactly right. That's the documented "works in any terminal" fallback Claude Code surfaces prominently, and getting it into our hint row addresses the Wayland/Hyprland/Kitty class of failures in #1493 directly. Keep all of that.

**Please drop the Ctrl+Enter → newline binding** before I merge, though. Two reasons:

1. **Not in Claude Code's documented keybinding set.** I checked their [interactive-mode.md](https://code.claude.com/docs/en/interactive-mode.md#multiline-input), [terminal-config.md](https://code.claude.com/docs/en/terminal-config.md#enter-multiline-prompts), and [keybindings.md](https://code.claude.com/docs/en/keybindings.md). Their newline set is: `Shift+Enter`, `Ctrl+J`, `\ + Enter`, `Option/Alt+Enter`. **Ctrl+Enter is not listed anywhere.** Adding it gives us an idiosyncratic binding that diverges from the reference our users expect parity with.

2. **Most GUI editors bind Ctrl+Enter to the opposite semantic.** VS Code, IntelliJ, Cursor, ChatGPT web — they all use Ctrl+Enter as *submit*, not newline. Users coming from those tools and hitting Ctrl+Enter expecting submit would silently get a newline instead. We don't currently use Ctrl+Enter for anything, so we're not breaking an existing binding — but choosing to make it newline picks the wrong side of the convention split.

The fix is the one-line revert in `multiline-keys.ts`: drop `key.ctrl` from the newline condition and let it fall through to submit (which is the GUI-editor default and what the empty-set default has been). Also pull the corresponding `Ctrl+Enter` row from `ShortcutsHelpModal.tsx` and the `descCtrlEnter` i18n key in EN.ts / zh-CN.ts / types.ts.

Keep everything else — HintRow `^⇧ Enter / ^J` text, the Ctrl+J shortcut panel entry, the `descCtrlJ` i18n key. Push the revision and I'll merge same day.

Tracking myself as a follow-up (not asking you to add it here): `\ + Enter` as the Claude-Code-equivalent universal fallback. Separate small PR after this lands.

PR #1519, opened 2026-05-22
feat(desktop): add approval and completion notifications

wade19990814-hue:
## Summary
- add desktop system notifications for new approval requests while the window is unfocused
- add completion notifications for long-running turns once the app returns to idle
- localize the new notification copy for English and Simplified Chinese

## Details
This keeps the implementation intentionally small:
- new notifications are derived from existing desktop pending-approval state
- completion notifications only fire after a busy-to-idle transition lasting at least 15 seconds
- notifications are suppressed while the window is focused
- notification delivery is wired through the Tauri notification plugin

## Verification
- `npm test -- --run desktop/src/notifications.test.ts desktop/src/App.test.ts desktop/src/ui/thread.test.tsx`
- `npm run typecheck`
- `npm run build` (desktop)
- `npm run verify`

## Notes
- this is built on top of current upstream `main`
- the branch also includes coverage for language switching so the notification text follows the active desktop UI language

### DIFF
--- a/desktop/package-lock.json
+++ b/desktop/package-lock.json
@@ -14,6 +14,7 @@
         "@reasonix/core-utils": "file:../packages/core-utils",
         "@tauri-apps/api": "^2.1.1",
         "@tauri-apps/plugin-dialog": "^2.7.1",
+        "@tauri-apps/plugin-notification": "^2.0.0",
         "@tauri-apps/plugin-opener": "^2.0.0",
         "@tauri-apps/plugin-process": "^2.0.0",
         "@tauri-apps/plugin-updater": "^2.0.0",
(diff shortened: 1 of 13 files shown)


wade19990814-hue:
This PR closes #1509 with a deliberately small desktop-only implementation.

A few notes for review:
- Reasonix desktop is Tauri rather than Electron, so the notification delivery is implemented through the Tauri notification plugin instead of Electron's Notification API.
- The change only covers two cases: new approval requests while the window is unfocused, and completion notifications after a busy-to-idle transition lasting at least 15 seconds.
- Notification copy now follows the active desktop UI language (English / Simplified Chinese).
- I re-ran the full verification pass on top of current upstream `main` before opening this PR (`npm run verify`), in addition to focused desktop tests and a desktop production build.

If this direction looks right, I can follow up later with a separate discussion for optional sound/badge behavior without expanding the scope of this PR.

wade19990814-hue:
CI was red on the previous branch head because the new desktop notification plugin was only wired into the desktop package, while the root Vitest setup still lacked a mock/alias for `@tauri-apps/plugin-notification`. The failing suites were root-level desktop tests resolving `desktop/src/App.tsx`, so they failed at module resolution time rather than notification behavior.

That is fixed on the current head (`3a0c942`) by adding a root test mock plus a matching alias in `vitest.config.ts`, and I re-ran the local verification pass before pushing. The latest commit is now green in upstream CI (`CI` and `CodeQL` both succeeded).

PR #1528, opened 2026-05-22
fix(tui): correct IME cursor sync alignment and row calculation

### LINKED ISSUE #1474: 输入法无法对齐输入框，导致我打字得歪着脑袋看
Image

c16" />

Bernardxu123:
The IME candidate window (fcitx5/ibus/Microsoft Pinyin) was appearing at the bottom-right corner instead of next to the visual cursor in the input area.

## Root cause

Three cumulative bugs in the cursor-positioning calculation:

### 1. Hardcoded +3 constant (PromptInput.tsx:227)

The calculation used \+ 3\ as a fixed offset for rows between the cursor line and the bottom of PromptInput. The actual layout has only 2 \Box height={1}\ gaps plus an optional mode/model row:

| Element | Rows |
|---|---|
| Remaining input lines below cursor | linesBelow |
| Huge buffer hint | 0-1 (largeHint) |
| \Box height={1}\ gap | 1 |
| Mode/model row | 0-1 (conditional) |
| \Box height={1}\ gap | 1 |

When mode and model are both empty, \+ 3\ overshoots by 1 row, placing the cursor 1 row too high. When displayed, \+ 3\ happens to be correct (+2 gaps +1 mode/model = +3).

**Fix**: Replace \+ 3\ with \+ 2 + ((mode || model) ? 1 : 0)\.

### 2. rowsAfter undercount (ComposerArea.tsx:162)

\
owsAfter={1 + (activeLoop ? 1 : 0) + (jobs ? 1 : 0)}\

- StatusRow has \marginTop={1}\ which occupies an additional blank row (2 rows total: 1 margin + 1 content), but only \1\ was counted for StatusRow.
- The \jobs ? 1 : 0\ term had no corresponding rendered row — \jobs\ info is consumed internally by StatusRow via \useAgentState\, not as a separate row.

**Fix**: \
owsAfter={2 + (activeLoop ? 1 : 0)}\ — accounts for StatusRow content (1) + StatusRow marginTop (1) + LoopStatusRow (0/1).

### 3. Missing effect dependencies

\mode\ and \model\ were absent from the useEffect dependency array. Toggling the mode/model display row didn't re-trigger cursor sync, leaving the stale cursor position.

## Interaction of bugs

In some states these bugs happened to cancel:
- No mode/model + no jobs: +1 (bug 1) -1 (bug 2) = 0 ✓ (accidentally correct)
- No mode/model + jobs: +1 (bug 1) +0 (bug 2) = +1 ✗ (cursor too high)

In others they compounded:
- Mode/model shown + no jobs: +0 (bug 1) -1 (bug 2) = -1 ✗ (cursor too low)
- Mode/model shown + jobs: +0 (bug 1) +0 (bug 2) = 0 ✓ (accidentally correct)

## Changes

| File | Change |
|---|---|
| \src/cli/ui/PromptInput.tsx\ | Exact row calculation, added mode/model to deps |
| \src/cli/ui/ComposerArea.tsx\ | Fixed rowsAfter to 2 + (activeLoop ? 1 : 0) |

Closes #1474


esengine:
Thanks for the careful root-cause analysis here — bug #1 (the `+ 3` constant) and bug #2 (the `rowsAfter` undercount that missed `StatusRow`'s `marginTop`) are real, and the explanation of how they cancelled in some states and compounded in others is exactly the kind of thing that makes a fix easy to review.

Unfortunately I have to close this without merging. Stage 2 of #1529 (landed in #1534) removed this entire `useEffect` because the absolute `\x1b[row;colH` write desyncs Ink's cursor tracking under the new `<Static>`-based history rendering — Ink redraws the persistent frame at a cursor position it computes itself, and an absolute write anchored to `totalRows - rowsBelow` lands somewhere else whenever the persistent UI isn't sitting flush against the terminal bottom. The math fix in this PR is correct, but the underlying approach (absolute screen coordinates that assume "frame is at terminal bottom") is what got removed, and re-adding it would bring back ghost duplicates of the composer that I just spent the refactor chasing out.

The IME alignment problem (#1474) is still real and still open — I'd love a follow-up that solves it without absolute screen coordinates. A cursor-relative approach (e.g. computing offset from where Ink last left the cursor, or saving/restoring around the frame write) wouldn't run into the same desync. Happy to chat through the constraints before you commit time if you want to take another crack at it.

PR #1556, opened 2026-05-22
fix(dashboard): mint named session from web new action

### LINKED ISSUE #1547: 无法创建新会话
只要一点击创建新会话，就会导致应用挂起，web端直接显示 sessions失败：Failed to fetch，点击其它页面都是一样的报错 *xxx失败：Failed to fetch*

**Environment**
- Reasonix version (`reasonix --version`): 0.48.1
- Node version (`node --version`): v23.11.0
- OS (Windows 11 / macOS 14 / Ubuntu 24.04 / …): macos 26.5 (25F71)
- Shell (bash, zsh, fish, PowerShell 7, PowerShell 5.1, cmd, …): zsh
- Terminal app (Windows Terminal, iTerm2, Alacritty, kitty, WezTerm, **VSCode integrated**, **Cursor integrated**, Hyper, …): iTerm2
- DeepSeek model (e.g. `deepseek-v4-flash`, `deepseek-v4-pro`): deepseek-v4-flash

**Logs / transcript**

  ✓  api key        from /Users/jadian/.reasonix/config.json
  ✓  config         /Users/jadian/.reasonix/config.json (preset=auto, editMode=auto, mcp=2)
  ✓  api reach      /user/balance ok — 126.74 CNY
  ✓  tokenizer      /Users/jadian/.npm/_npx/a445a9f5043f7b46/node_modules/reasonix/data/deepseek-tokenizer.json.gz (1.7 MB)
  ✓  sessions       2 saved · 672.7 KB · oldest 0d
  ✓  semantic       not in use (no semantic index built; `reasonix index` to enable)
  ✓  project        /Users/jadian (package.json)

LaiyanB:
## Summary

Fixes the dashboard `/api/sessions/new` path so it mirrors the TUI SessionPicker "new" behavior by creating a real fresh session name instead of switching to an unnamed/ephemeral session.

## Why

Issue #1547 reports that clicking "create new session" in the web dashboard hangs the app and then all dashboard requests fail with `Failed to fetch`.

The existing endpoint comment says it should route through the same path as SessionPicker's `new` branch, but it actually called `switchSession(undefined)`. In the TUI path, a new session is created with `freshSessionName(activeSession)`. Passing `undefined` remounts the attached App as an unnamed session and can leave the dashboard with stale session state/API connectivity.

## Changes

- Import and use `freshSessionName` in `src/server/api/sessions.ts`.
- `POST /api/sessions/new` now calls `switchSession(freshSessionName(currentName))`.
- Falls back to `default-<timestamp>` when there is no attached current session name.
- Adds a focused regression test for the dashboard new-session endpoint.

## Testing

Added `tests/dashboard-new-session-name.test.ts` covering:

- current session `alpha` → `alpha-<14 digit timestamp>`
- no current session → `default-<14 digit timestamp>`

I was not able to run the test suite in this environment because cloning/running the repo was not available here.

Fixes #1547

### DIFF
--- a/src/server/api/sessions.ts
+++ b/src/server/api/sessions.ts
@@ -1,5 +1,5 @@
 import { existsSync, readFileSync } from "node:fs";
-import { deleteSession, listSessions, sessionPath } from "../../memory/session.js";
+import { deleteSession, freshSessionName, listSessions, sessionPath } from "../../memory/session.js";
 import type { DashboardContext } from "../context.js";
 import type { ApiResult } from "../router.js";
 
@@ -67,8 +67,10 @@ export async function handleSessions(
     };
   }
 
-  // New session — mints a fresh session by calling switchSession(undefined),
-  // which routes through the same path the SessionPicker "new" branch takes.
+  // New session — mirror the TUI SessionPicker "new" branch by minting
+  // a real session name. Passing undefined would remount App as an
+  // ephemeral session and leave the attached dashboard with stale
+  // session state, which makes the SPA lose its API connection.
   if (method === "POST" && rest.length === 1 && rest[0] === "new") {
     if (!ctx.switchSession) {
       return {
@@ -76,7 +78,8 @@ export async function handleSessions(
         body: { error: "live session swap requires an attached CLI session." },
       };
     }
-    const result = ctx.switchSession(undefined);
+    const currentName = ctx.getSessionName?.() ?? undefined;
+    const result = ctx.switchSession(freshSessionName(currentName));
     if (!result.ok) return { status: 500, body: { error: result.reason } };
     return { status: 200, body: { ok: true } };
   }
--- a/src/server/api/sessions.ts
+++ b/src/server/api/sessions.ts
@@ -1,5 +1,10 @@
 import { existsSync, readFileSync } from "node:fs";
-import { deleteSession, freshSessionName, listSessions, sessionPath } from "../../memory/session.js";
+import {
+  deleteSession,
+  freshSessionName,
+  listSessions,
+  sessionPath,
+} from "../../memory/session.js";
 import type { DashboardContext } from "../context.js";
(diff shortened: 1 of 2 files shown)


esengine:
Fix looks right — routing through `freshSessionName(currentName)` matches the `SessionPicker` "new" branch, the `default-<ts>` fallback is the same shape the TUI mints for an unnamed attached session, and the test covers both. Only blocker is biome formatting on the import: the line is over 100 chars and biome wants the multi-line form. `npx biome check --write src/server/api/sessions.ts` rewrites it; push that and I'll merge once CI goes green.

(The `welcome-banner-hints.test.tsx` warning above is pre-existing and unrelated.)

esengine:
Thanks for the careful diagnosis — your reading of why `switchSession(undefined)` was leaving the dashboard with stale state was correct. Unfortunately #1599 landed first (merged as ff4032c7) and fixed #1547 from a different angle:

- `switchSession(undefined)` stays as the canonical entry point so the web and TUI "new chat" paths don't diverge — the fresh name is minted inside it, the same way the TUI SessionPicker's `new` branch works.
- The `/sessions/new` response now echoes `name: ctx.getSessionName?.()` so the dashboard reducer can update `currentSession` and #1586's URL mirror without having to diff the listing.
- The bridge was also changed to emit `$session_loaded` (instead of `$session_empty`) so the SPA's reducer resets cleanly instead of pushing an error card.

So the dashboard symptom you reported is fixed on `main`. Pulling `freshSessionName` up into the API layer would re-split the web and TUI paths, which is what we want to avoid; also `freshSessionName(undefined)` already returns `default-<14-digit-stamp>`, so the explicit fallback in the PR is redundant with that helper.

Going to close this and #1547 as fixed by #1599. Really appreciate the issue write-up and the regression test — that style of bug report is exactly what we want.

PR #1572, opened 2026-05-22
feat(tui): add Alt+S input stash/recall hotkey (#1438)

### LINKED ISSUE #1438: Stash message in input box
**Problem**
Just a temp register that allow user to save something and use it latter. Motivated by the popular pi extension pi-powerline-footer.

**Proposed change**
hotkey like `alt s`.
When empty input, if message has been stashed, input is filled with stashed message and stash clears
When empty input, if no stash, show hint "Nothing to stash"
When input is not empty, if message has been stashed, exchange stash and input
When input is not empty, if no stash, input is cleared and stash holds input message

**Alternatives considered**

**Scope check**

dacec354:
## What

 Adds Alt+S hotkey to stash and recall the composer input buffer. Press once to save the current
input, press again on an empty line to recall it. If there's already stashed text, the current input
is replaced by the stash.

 ## Why

 Quick temp register — save a half-written message or thought for later without losing it. Closes
#1438.

 ## How to verify

 1. `npm run dev` (or `npm run build && node dist/cli/index.js`)
 2. Type something → Alt+S → input clears, "Stashed" hint appears
 3. Alt+S again → input fills with the stashed text, "Recalled"
 4. Alt+S twice on empty input → "Nothing to stash"
 5. Type something, Alt+S to stash, type something else, Alt+S again → stash replaces current text
 6. Ctrl+P → shortcuts modal shows Alt+S under Input group

 ## Checklist

 - [x] `npm run verify` passes locally (lint + typecheck + tests — 3 pre-existing web-tools failures,
same as main)

### DIFF
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -508,6 +508,7 @@ function AppInner({
   // Esc handler only fires once per turn (repeated presses would yield
   // stacked warning events).
   const abortedThisTurn = useRef(false);
+  const stashRef = useRef("");
   // Mirrors the live `busy` flag for /loop's timer (it has no React
   // closure handle, only refs). Skips the firing when a prior turn is
   // still running rather than queuing a duplicate submit.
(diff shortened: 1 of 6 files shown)


esengine:
Thanks for picking this up — the shape is right and the diff is small. Three things I'd want fixed before merging, plus a couple of small notes.

**1. Doesn't match the issue spec — current input silently disappears (blocker)**

#1438 says: "if message has been stashed, **exchange** stash and input". The PR doesn't exchange — when the input is non-empty and the stash already has content, the current input is dropped and replaced by the stash:

```ts
} else {  // input.length > 0
  if (stashRef.current) {
    setInput(stashRef.current);   // current input is lost
    stashRef.current = "";
    ...
```

This is the deviation noted in your PR description ("current input is replaced by the stash"). It means an accidental Alt+S in the middle of a half-typed message destroys the message without any recovery path. The point of a stash register is that one extra Alt+S brings it back. Please make it a true swap:

```ts
const recalled = stashRef.current;
stashRef.current = input;
setInput(recalled);
```

**2. No gating — fires under modals and while busy**

The handler is placed above `if (busy) return;` and above the whole `!pendingShell && !pendingPath && !pendingSessionsPicker && !pendingCheckpointPicker && !pendingMcpHub && !walkthroughActive && !pendingChoice && !pendingReviseEditor && !pendingEditReview && !stagedInput && !pendingRevision` guard set that every other composer-targeting shortcut in this file uses. Consequences:

- Pressing Alt+S during a running turn mutates composer state the user can't see, and pushes a "Stashed" / "Recalled" row into the log out of nowhere.
- Same thing under shell/path confirms, session picker, walkthrough, edit-review, etc.

Copy the same guard block that Ctrl+O and the Space-toggle-undo blocks above it use.

**3. `composer.hintStash` is a dead i18n key**

`hintStash: "stash" / "暂存"` is declared in EN, zh-CN, and `types.ts`, but never rendered anywhere — no composer hint bar references it. Either wire it into the bottom hint strip alongside `hintClear` / `hintHistory` / `hintAbort`, or drop the key. As-is it'll confuse the next person grepping for unused strings.

**Smaller notes**

- `key.input === "s" || key.input === "S"` also catches Alt+Shift+S. Fine if intentional, but worth aligning with the rest of the shortcut handlers (which mostly stick to the lowercase form).
- The test only covers that the i18n keys are exposed. After fixing #1 and #2, it'd be worth one behaviour test for the full state machine: empty + no stash → "Nothing to stash"; non-empty + no stash → input cleared + stash set; non-empty + existing stash → **swap**; gated off when `busy` or a modal is up.
- The two `if (stashRef.current) { setInput(...); stashRef.current = ""; pushInfo(stashRecall) }` branches in the original layout collapse into one once #1 is in.

Leaving this open — push a fixup and I'll take another look.

dacec354:
done

esengine:
Thanks — fixup looks right, swap + gating + dead key all addressed. Merging.

PR #1574, opened 2026-05-22
test(lifecycle): cover deterministic policy corpus

GTC2080:
## Summary
- Add a deterministic lifecycle policy corpus covering safe exploration, ordinary mutations, high-risk tool calls, package/config paths, package manager mutations, and git/shell mutation commands.
- Tighten package-manager command detection for common aliases such as `npm uninstall`, `pnpm up`, and `yarn upgrade`.
- Treat branch checkout like `git checkout feature/lifecycle` as high-risk while preserving explicit path checkout safety via `git checkout -- README.md`.
- Keep this slice runtime-only and prompt-neutral: no prompt, tool schema, auto-arm, or UI behavior changes.

## Test Plan
- `npm test -- tests/code-prompt.test.ts tests/lifecycle-policy.test.ts tests/lifecycle.test.ts tests/lifecycle-e2e.test.ts tests/tools.test.ts`
- `npx biome check src/code/lifecycle-policy.ts tests/lifecycle-policy.test.ts`
- `npm run verify`
- Pre-push hook also ran `npm run verify` successfully: 259 test files passed, 3599 tests passed, 10 skipped.

Refs #1285
Follow-up to #1557

### DIFF
--- a/src/code/lifecycle-policy.ts
+++ b/src/code/lifecycle-policy.ts
@@ -182,7 +182,10 @@ function isPackageMutation(token: string | undefined): boolean {
     normalized === "install" ||
     normalized === "add" ||
     normalized === "remove" ||
-    normalized === "update"
+    normalized === "uninstall" ||
+    normalized === "update" ||
+    normalized === "upgrade" ||
+    normalized === "up"
   );
 }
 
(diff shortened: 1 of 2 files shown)


esengine:
Merged — thanks. The corpus-as-spec style is exactly what this module needed; future tweaks now have a clear regression net. If you want a small follow-up: `npm i` (short for install) and `npm un` are still missing from `isPackageMutation` — adding them + two corpus rows would close the alias gap.

PR #1576, opened 2026-05-23
fix(tui): surface failure signals in collapsed run_command cards

### LINKED ISSUE #1496: TUI: surface failure signals in collapsed run_command cards and clarify full-output access
**Problem**

While testing `reasonix code` on a few small local debugging tasks, I noticed a TUI affordance issue around completed `run_command` cards.

The green `✓ run_command` marker seems to mean that the Reasonix tool invocation completed and returned to the loop. That meaning makes sense. However, in practice it is easy to read the green check as “the underlying shell command or test passed”, especially when the command output contains a failure but the visible collapsed card only shows trailing logs.

The final diagnosis was usually correct. The issue is mostly with what the user can quickly see in the TUI while the task is running.

A related point: collapsed cards show `use /tool to read full`, but in my manual session the `/tool` path was not obvious from that state. It may be that I used it incorrectly, but the full-output affordance was not clear while reading the card.

**Observed behavior**

Case 1: Python unittest failure with noisy trailing logs

The test failure was:

```text
AssertionError: 300 != 800 : third retry should wait 800ms
```

The visible command card showed:

```text
✓ run_command python3 test_retry_policy.py
⋮ 43 earlier lines (use /tool to read full)
cleanup: checked worker shard 27/30
cleanup: checked worker shard 28/30
cleanup: checked worker shard 29/30
cleanup: checked worker shard 30/30
```

The final summary identified the correct failure, but the visible card looked like a completed/successful command and only showed cleanup tail lines. The assertion line was collapsed above the visible excerpt.

Case 2: Node assertion failure

The test failure was:

```text
applyCoupon(12000, "VIP25")
expected: 9000
actual:   10200
```

The final summary correctly identified the issue in `src/cart.js`, but the command card showed mostly the tail of the assertion output:

```text
✓ run_command node test.mjs
⋮ 19 earlier lines (use /tool to read full)
diff: 'simple'
}
Node.js v22.22.0
```

Again, the diagnosis was correct, but the card did not make the failure signal easy to see.

**Expected**

- Keep `✓ run_command` if it means “tool invocation completed”, but make command failure harder to confuse with command success.
- If the subprocess exits non-zero, show that near the card header or first visible output line.
- If output is collapsed, prefer high-signal failure lines over only the final tail.
- The `use /tool to read full` hint should lead to an obvious full-output path from the state where it appears.

**Possible improvements**

These are suggestions, not a required design.

1. Clarify tool completion vs command result.

For example:

```text
✓ run_command completed · exit 1
```

or:

```text
✓ tool completed
✗ command exited 1
```

2. Use failure-aware collapsed excerpts.

When output is collapsed, prefer showing:
- exit status,
- first assertion/error line,
- nearby mismatch line,
- then the final tail.

Example:

```text
✗ run_command node test.mjs · exit 1
AssertionError: VIP25 should reduce a 12000 cent cart to 9000 cents
actual: 10200
expected: 9000
⋮ 19 lines collapsed
```

3. Clarify full-output access.

Possible options:
- make sure `/tool` works from the state where the hint appears;
- update the hint if the intended interaction is different;
- allow selecting/expanding the collapsed card with Enter;
- mention the correct full-output path in `/keys`.

**Reproduction**

Environment:

```text
Reasonix version: 0.48.0
Node version: v22.22.0
OS: macOS
Shell: zsh
Terminal app: Terminal.app
Model: deepseek-v4-pro
```

Python repro:

1. Create a small unittest project where one assertion fails early, followed by many cleanup/log lines.
2. Run:

```bash
reasonix code .
```

3. Ask:

```text
Help me check why this Python unittest project is failing. Please run the tests first, analyze the output, and do not edit files.
```

4. Observe that the final diagnosis can be correct, but the visible command card may show only trailing cleanup lines while the assertion is collapsed.

Node repro:

1. Create a small Node project with a failing assertion.
2. Run:

```bash
reasonix code .
```

3. Ask:

```text
Help me check why this Node project test is failing. Please run the project tests first, analyze the reason, and do not edit files.
```

4. Observe that the final diagnosis can be correct, but the visible command card may collapse the actual mismatch and show only the tail of the assertion output.

Full-output repro:

1. Produce a collapsed command card that says `use /tool to read full`.
2. Try to use `/tool` from the TUI to inspect the full output.
3. Observe whether the full-output path is available and obvious from that state.

**Related note**

I also noticed that normal terminal text selection is not obvious when mouse handling is enabled. This may already be covered by existing copy/mouse discussions and recent fixes, so I am not treating it as the main issue here. The main request is about failure visibility in collapsed `run_command` cards and the `/tool` full-output affordance.

**How this could be tested**

- Add a rendering/snapshot test for a command result with:
  - non-zero exit code,
  - an assertion failure near the top,
  - noisy trailing cleanup lines.
- Verify the collapsed card includes the failure line, not only the tail.
- Add a Node assertion fixture where useful lines are above the final `Node.js v...` tail.
- Manual smoke test:
  - run `reasonix code .`;
  - trigger Python unittest and Node assertion failures;
  - confirm the TUI card surfaces the failure line without needing to inspect the full transcript.
- Manual full-output test:
  - trigger a collapsed tool card;
  - confirm the advertised `/tool` or equivalent full-output path works from that state.

Bernardxu123:
## Summary

When a `run_command` exits non-zero, the collapsed card now shows error-relevant lines instead of only the last N lines (which might be cleanup logs). This makes command failures immediately visible in the TUI without needing to expand the full output.

Fixes #1496

---

## Problem

In the TUI, collapsed `run_command` cards only show the last 5 lines of output. When a command fails:

- The error message (e.g., assertion failure, traceback) is usually near the top/middle of the output
- The visible tail often contains only cleanup logs or irrelevant trailing lines
- The green `✓` checkmark makes it look like the command succeeded
- Users have to manually expand the card to find the actual error

**Example of the problem:**

`
✓ run_command node test.mjs
⋮ 19 earlier lines (use /tool to read full)
diff: 'simple'
}
Node.js v22.22.0
`

The actual error `AssertionError: expected 10200 to equal 9000` is hidden in the collapsed lines.

---

### 1. Smart Truncation for Failed Commands

When exit code is non-zero, the card now scans the output for error-relevant lines using patterns like:
- `error`, `fail`, `fatal`, `exception`
- `assert`/`assertion`, `traceback`, `panic`
- `exit N`, `errno`, `denied`, `timeout`

These lines (with context) are included in the visible output along with the tail lines.

### 2. Exit Code in Truncated Hint

The hint now shows the exit code when the command failed:

`
⋮ 15 lines hidden · exit 1 (Ctrl+R to expand)
`

Instead of the generic:

`
⋮ 19 earlier lines (use /tool to read full)
`

### 3. Clearer Expand Hint

Changed `use /tool to read full` to `Ctrl+R to expand` since Ctrl+R is the actual keybinding for verbose mode.

---

## After Fix

**Failed command (exit 1):**

`
✗ run_command node test.mjs · exit 1
AssertionError: VIP25 should reduce a 12000 cent cart to 9000 cents
actual: 10200
expected: 9000
⋮ 15 lines hidden · exit 1 (Ctrl+R to expand)
diff: 'simple'
}
Node.js v22.22.0
`

**Successful command (exit 0) - unchanged:**

`
✓ run_command echo hello
hello
`

---

## Files Changed

| File | Change |
|------|--------|
| `src/cli/ui/cards/ToolCard.tsx` | Added `smartTruncate()` function for error-aware line selection |
| `src/i18n/EN.ts` | Added `earlierLinesFailed` key, updated hint text |
| `src/i18n/zh-CN.ts` | Added Chinese translation for new hint |
| `src/i18n/types.ts` | Added type definition for `earlierLinesFailed` |

---

## Testing

- `npm run lint` ✅
- `npm run typecheck` ✅
- `tests/reducer-cards-elide.test.ts` ✅ (9 tests passed)


esengine:
Thanks for picking this up — the direction matches what #1496 asked for and the i18n + hint plumbing is in the right shape. A few things to fix before this can land.

**1. `smartTruncate` drops error-context lines at gap boundaries (correctness bug)**

In the index walk:

```ts
for (const idx of sortedIndices) {
  if (idx === lastIdx + 1) {
    visible.push(allLines[idx]!);
  } else if (idx > lastIdx) {
    const gap = idx - lastIdx - 1;
    if (visible.length > 0) {
      visible.push(`⋮ ${gap} line${gap > 1 ? "s" : ""} hidden`);
    } else {
      visible.push(allLines[idx]!);
    }
  }
  lastIdx = idx;
}
```

When `visible.length > 0` and there's a gap, only the separator gets pushed — `allLines[idx]` is never appended, but `lastIdx` still advances to `idx`. The very line that opened the new run (typically the `±1` pre-context line you intended to surface around an error) is silently dropped.

Concrete trace: 15-line output, ERRORs at indices 2 and 10, tail = 5. After `errorIndices = {1,2,3} ∪ {9,10,11} ∪ {10..14}` the sorted walk emits `b, ERROR1, c, "⋮ 5 lines hidden", ERROR2, j, k, l, m` — but `allLines[9]` ("i", the promised `−1` context for ERROR2) is gone. Same shape happens for every gap after the first run.

**2. `hidden` count double-reports against the inline separators**

```ts
const hidden = allLines.length - visible.length;
```

`visible` now contains both real lines and `⋮ N lines hidden` separator strings, so `hidden` over-counts by however many separators are inside. The outer card header then renders `earlierLinesFailed` with this inflated number while the body separately renders its own `⋮ N lines hidden` strings with a different number. The user sees two contradictory hidden counts.

Pick one: either keep the inline separators and skip the outer hint when separators are present, or drop the inline separators and just show one collapse summary at the top.

**3. Regex over-matches; `\bexit\s+\d\b` also catches `"exit 0"`**

```ts
/\b(?:error|fail|fatal|exception|assert(?:ion)?|traceback|panic|killed|segfault|exit\s+\d|errno|denied|not\s+found|no\s+such|cannot|could\s+not|unable\s+to|refused|timeout|timed?\s*out)\b/i
```

A few concrete trips:

- `\bexit\s+\d\b` matches `"exit 0"`. This only runs when the overall command failed, but a sub-process line like `ssh tunnel: exit 0` (success of one step inside an overall-failed command) still gets surfaced as error-relevant. The "exit code" signal already lives on `card.exitCode`; don't try to recover it from the output text.
- `\berror\b` hits `--error-format=json` flags, `errors: 0` summaries, log prefixes like `[ERROR_RECOVERY]` etc.
- `\bnot\s+found\b`, `\bcannot\b`, `\bunable\s+to\b`, `\bdenied\b`, `\btimeout\b` all have common non-error usages (help text, config field descriptions, summary lines).

There's also no cap on how many matches get surfaced — a chatty runner with the word `error` on every line defeats the elision entirely and the card grows back to full output. Cap to the first N (e.g. 5) matches plus the tail, or you'll get cards larger than the un-elided version.

**4. Inline separator is hardcoded English**

```ts
visible.push(`⋮ ${gap} line${gap > 1 ? "s" : ""} hidden`);
```

Everything else in this file goes through `t(...)` (you added the EN + zh-CN entries for `earlierLinesFailed` correctly). A zh-CN user will see Chinese around the card and a stray `⋮ 23 lines hidden` in the middle. Either add an `earlierLinesInline` key (with EN + zh-CN strings) or drop inline separators if you choose option (2) above.

**5. Separator strings inherit the error color**

```tsx
{visible.map((line, i) => (
  <Text color={errColor} dimColor={!card.exitCode || card.exitCode === 0}>
    {clipToCells(line, lineCells) || " "}
  </Text>
))}
```

On a failed card, `errColor = TONE.err` and `dimColor = false`. The separator string renders bright red, visually indistinguishable from a real stderr line. Render separators in their own `<Text color={FG.faint}>` block (same treatment the existing header hint uses).

**6. `errorLinesFound` is dead**

It's computed, included in the return type, destructured at the call site, and never read. The failed-hint branch keys off `card.exitCode !== undefined && card.exitCode !== 0`, not off whether any error lines actually got surfaced. Either wire it into the hint decision (use the failed-hint variant only when we genuinely surfaced something) or drop the field.

**7. No tests for `smartTruncate`**

The PR's listed `npm test` only covers `tests/reducer-cards-elide.test.ts`, which exercises the reducer, not this function. Move `smartTruncate` to its own utility module (e.g. `src/cli/ui/cards/tool-output-truncate.ts`) so it can be unit-tested without React, and add a vitest `it.each` corpus covering:

- exit 0 → tail-only path (regex never runs)
- exit ≠ 0 + error line at the top, noisy cleanup tail → top assertion surfaced
- exit ≠ 0 + multiple scattered error lines → cap kicks in
- false-positive inputs: line containing `exit 0`, `errors: 0`, `--error-format=json`
- gap-boundary case (regression for issue #1)
- empty output / single line / fewer than `tailCount` lines

This is the exact kind of policy code the lifecycle PR's corpus-as-spec style applies to.

**Minor**

- `Ctrl+R to expand` is the right verb to use (App.tsx:504 confirms `Ctrl+R` toggles verbose), though strictly speaking it's a **global** verbose toggle, not a per-card expand. Still better than `use /tool to read full`. Worth keeping; consider `Ctrl+R to view full output` if you want to be precise.
- The `earlierLine` / `earlierLines` wording change applies to the non-failed case too. That's a fine drive-by improvement, but call it out in the PR description so it doesn't look unintentional.

Leaving this open — push a fixup and I'll re-review.

esengine:
Closing as superseded by #1608, which landed a more comprehensive fix for #1496 — extracted preview logic into a tested pure module, scoped the failure-line scan to shell tools with exit≠0, fixed the gap-boundary bug I flagged here, and additionally fixed a Static-card re-render bug that prevented Ctrl+R from expanding settled tool cards in the first place. Thanks for opening this and surfacing the design considerations — they fed directly into the review of #1608.

PR #1578, opened 2026-05-23
test(lifecycle): cover intent evaluator corpus

GTC2080:
## Summary
- Add an advisory lifecycle intent evaluator for corpus/eval work only; it is not wired into runtime enforcement, prompt construction, tool schemas, or auto-arm.
- Cover deterministic intent cases for read-only questions, small single-file fixes, multi-file refactors, dependency/config migrations, failed verification revise flow, explicit strict lifecycle requests, broad cleanup suggestions, cancel intent, and Chinese prompts.
- Keep the slice runtime-only and prefix-neutral by returning advisory metadata (`modelVisible: false`, `enforced: false`) without changing existing lifecycle behavior.

## Test Plan
- `npm test -- tests/lifecycle-intent.test.ts`
- `npx biome check src/code/lifecycle-intent.ts tests/lifecycle-intent.test.ts`
- `npm test -- tests/code-prompt.test.ts tests/lifecycle-intent.test.ts tests/lifecycle.test.ts tests/lifecycle-e2e.test.ts tests/tools.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run verify`
- Pre-push hook also ran `npm run verify` successfully: 260 test files passed, 3587 tests passed, 10 skipped.

Refs #1285
L4 PR3; independent of #1574 to avoid stacking.

### DIFF
--- a/src/code/lifecycle-intent.ts
+++ b/src/code/lifecycle-intent.ts
@@ -1,0 +1,256 @@
+export type LifecycleIntentRecommendation = "stay-light" | "suggest-plan" | "strict-candidate";
+
+export type LifecycleIntentConfidence = "low" | "medium" | "high";
+
+// Advisory corpus/eval helper only; do not wire this into auto-arm without a separate RFC.
+export type LifecycleIntentSignal =
+  | "cancel"
+  | "dependency-change"
+  | "explicit-strict"
+  | "failed-verification"
+  | "migration"
+  | "multi-file"
+  | "package-or-config"
+  | "plan-revision"
+  | "read-only"
+  | "refactor"
+  | "single-file"
+  | "small-task";
+
+export interface LifecycleIntentEvaluation {
+  recommendation: LifecycleIntentRecommendation;
+  confidence: LifecycleIntentConfidence;
+  score: number;
+  signals: LifecycleIntentSignal[];
+  modelVisible: false;
+  enforced: false;
+}
+
+const EXPLICIT_STRICT_PHRASES = [
+  "/plan strict",
+  "strict lifecycle",
+  "strict rails",
+  "强约束",
+  "严格生命周期",
+];
+
+const READ_ONLY_PHRASES = [
+  "explain",
+  "inspect",
+  "look at",
+  "read",
+  "summarize",
+  "do not edit",
+  "without modifying",
+  "解释",
+  "查看",
+  "看看",
+  "总结",
+  "不要修改",
+  "只读",
+];
+
+const NEGATED_MUTATION_PHRASES = ["do not edit", "without modifying", "不要修改", "不要改"];
+
+const MUTATION_PHRASES = [
+  "add",
+  "change",
+  "delete",
+  "edit",
+  "fix",
+  "install",
+  "migrate",
+  "move",
+  "refactor",
+  "remove",
+  "rename",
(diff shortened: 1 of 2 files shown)


esengine:
Thanks for the continued lifecycle work — the corpus discipline you've been building (most recently #1574) is exactly the right shape for the policy modules. This one I don't want to take as a standalone slice, though.

**The blocker is the slice, not the code quality**

`evaluateEngineeringLifecycleIntent` has no call sites on merge. Grep the repo and nothing imports it. The PR description and the file's own header both explicitly say it isn't wired into runtime enforcement, prompt construction, tool schemas, or auto-arm — and that doing so needs a separate RFC. So what lands is a 256-line module plus a 91-line test file that exercise dead code from day one.

CLAUDE.md and the project's broader discipline call this out specifically — no designs for hypothetical future requirements, no half-finished implementations. Even good speculative scaffolding still has to be maintained, refactored when the surrounding code changes, and re-justified the next time someone touches it.

The pattern your previous lifecycle PRs follow is the right one: every classifier / corpus you've added so far ships **with** the code that uses it, or tightens behaviour that's already wired in. #1574 was a clean example — it expanded the corpus *and* tightened the live classifier in the same PR.

**Path forward**

When the auto-arm / suggestion wiring RFC is ready (whether that's a follow-up to #1285 or a new RFC), land the intent evaluator inside that PR. At that point the corpus has a customer and the thresholds have a grounded rationale, which they currently don't.

In the meantime keep it on your branch — nothing here gets lost.

**While you're back at it — implementation notes for the eventual rewrite**

Worth flagging so you don't ship a v1 that breaks on real prompts:

- The substring-match approach over phrase lists has a lot of unintended matches on English + Chinese natural language:
  - `MUTATION_PHRASES = ["add", ...]` + `text.includes("add")` matches `"address"`, `"address the issue"`, `"adds up"`, etc.
  - `REFACTOR_PHRASES = ["api", ...]` — every prompt mentioning an API gets the refactor signal. `"What does this API return?"` already does.
  - Single-character `"修"` in `MUTATION_PHRASES` matches `"修养"`, `"必修"`, `"装修"`, `"维修"`, etc., none of which are mutation intent.
  - `hasSingleFileSignal` uses `text.includes(".json")` — `"explain how tsconfig.json works"` becomes a `single-file` signal.
  - `MULTI_FILE_PHRASES` contains `"across"` — `"come across this bug"`, `"across the team"`, etc.
- The score thresholds (`>= 4` → strict-candidate, `>= 2` → suggest-plan) and per-signal weights aren't justified anywhere — they're tuned to the 12-case corpus, which is fine for a kept-in-branch sketch but won't survive contact with real prompts.
- The corpus assertion uses `expect.arrayContaining(signals)`, so spurious signals firing on the same input would slip through. When this lands, switch to exact-set comparison so the false-positive sources show up in the diff.
- More architectural: for "what is the user actually trying to do here", the model that's about to process the prompt is a strictly better classifier than a maintained set of bilingual phrase lists. It might be worth folding intent inference into the lifecycle prompt itself rather than running a host-side keyword classifier. That's RFC-level scope, not for this PR — but worth raising before sinking more work into the regex/phrase track.

Going to close this and look forward to seeing the wired version. Thanks again for owning the lifecycle track.

PR #1579, opened 2026-05-23
fix(dashboard): auto-refresh file tree after file-modifying tool operations

Bernardxu123:
## Summary

When file-modifying tools (delete_file, write_file, edit_file, etc.) complete in the TUI, the dashboard's file tree panel now automatically refreshes to reflect the current filesystem state.

Fixes #1564

---

## Problem

The dashboard's file tree was fetched once on component mount and never refreshed. After the AI agent used file-modifying tools:

- Deleted files still appeared in the tree
- New files were not visible
- Manual refresh was not available (Ctrl+R was verbose toggle, F5 was unmapped)
- Users reported black screens when trying to refresh

---

### 1. Refresh Function in useProjectTree()

Added a 
efresh() function to the useProjectTree() hook that re-fetches the file tree from the server.

### 2. Manual Refresh Button

Added a refresh button (↻) in the file tree header for manual refresh.

### 3. Auto-Refresh on Tool Completion

The dashboard's SSE connection now listens for 	ool events. When a file-modifying tool completes, the file tree automatically refreshes after a 300ms delay.

**Supported tools:**
- delete_file, delete_directory
- write_file, edit_file, multi_edit
- create_directory, move_file, copy_file

---

## Files Changed

| File | Change |
|------|--------|
| `dashboard/src/lib/file-tree.ts` | Added `refresh()` function and `refreshKey` state |
| `dashboard/src/panels/changes.ts` | Added refresh button, `onToolComplete` prop to ChatPane, auto-refresh logic |
| `dashboard/src/i18n/en.ts` | Added `refreshTree` translation key |
| `dashboard/src/i18n/zh-CN.ts` | Added Chinese translation for refresh button |

---

## Testing

- `npm run lint` ✅
- `npm run typecheck` ✅

PR #1582, opened 2026-05-23
fix(desktop): prevent stale events from corrupting UI after session switch

### LINKED ISSUE #1217: Mac桌面端 会话运行时打开新会话后切换会话卡死
Mac桌面端，会话运行时，点击新建会话，发起新的会话后，整个页面卡顿，而且不能切回原来的会话（测试时仅有两个会话）

Bernardxu123:
Fixes #1217

When a session switch happens while a turn is running, the old turn cleanup may still emit events that corrupt the new session UI state.

Add switching flag to Tab interface:
- Set tab.switching=true ONLY when tab.aborter exists (live turn to abort)
- Check tab.switching in runTurn finally block to suppress stale events
- Reset flag after processing

This prevents the flag from getting stuck when no turn is in flight — the common session-switch path.

Esc + New Session still works because switching is only set when aborter exists, not in Esc handler.

Tested: npm run lint, npm run typecheck

### DIFF
--- a/src/cli/commands/desktop.ts
+++ b/src/cli/commands/desktop.ts
@@ -739,6 +739,8 @@ interface Tab {
   planTotalSteps: number;
   mcpRuntime: McpRuntime | null;
   mcpStatuses: Map<string, { kind: McpSpecStatus; reason?: string; toolCount?: number }>;
+  /** True while a session switch is in progress — prevents stale events from the old turn. */
+  switching: boolean;
 }
 
 let tabCounter = 0;
@@ -1232,6 +1234,7 @@ export async function desktopCommand(opts: DesktopOptions): Promise<void> {
       planTotalSteps: 0,
       mcpRuntime: null,
       mcpStatuses: new Map(),
+      switching: false,
     };
     tab.currentSession = mintSessionFor(dir);
     tabs.set(tab.id, tab);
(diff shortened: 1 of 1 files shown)


esengine:
Thanks for splitting this out of #1579 — clean focused diff, mergeable against main, and the core mechanism is right.

One blocker before I can merge, though. The flag gets stuck `true` on the most common path, not just an edge case.

**Sequence that breaks (any session switch, not just abort-then-switch):**

1. Previous turn finishes normally → `runTurn`'s finally runs → `tab.aborter = null`, `tab.switching = false` (already).
2. User clicks a different session → `session_load` handler runs:
   ```ts
   tab.switching = true;
   abortTurn(tab);  // tab.aborter?.abort() is a no-op — aborter is null
   ...
   ```
3. `tab.switching` is now `true`. No `runTurn` is in flight, so no finally fires to reset it.
4. User types the first message in the new session → new `runTurn` runs → finally:
   ```ts
   if (!tab.switching) { /* … */ }   // false → all wrap-up events skipped
   tab.switching = false;
   ```
5. Every terminal event for that first new-session turn is suppressed: `$turn_complete`, `emitSessions`, `emitBalance`, `$plan_cleared`, the QQ channel response.

The user-visible symptom is worse than #1217 was: the UI sits on "thinking" forever because `$turn_complete` never arrives, balance/session list don't refresh, and the QQ integration drops the first reply. This isn't an edge case — it's the standard "switch session, then send a message" flow.

**Minimal fix**

Only set the flag when there's actually a live turn to abort. Two-line change at both call sites:

```ts
// session_load and new_chat handlers
if (tab.aborter) tab.switching = true;
abortTurn(tab);
```

That keeps the rest of the PR identical. If you'd rather avoid the shared mutable flag entirely, the cleaner alternative is to snapshot the session in `runTurn` and compare in finally:

```ts
async function runTurn(tab: Tab, text: string, fromQQ = false): Promise<void> {
  if (!tab.runtime) return;
  const startSession = tab.currentSession;
  ...
  finally {
    tab.aborter = null;
    if (tab.currentSession === startSession) {
      // emit $turn_complete / emitSessions / emitBalance / $plan_cleared / QQ
    }
  }
}
```

Then drop the `tab.switching` field and the two `tab.switching = true` lines entirely. No shared state, no reset, no stuck-flag class of bug. Whichever you prefer.

Push a fixup and I'll merge.

Bernardxu123:
Applied your suggested fix:

\\\	s
// Only set switching flag when there's a live turn to abort
if (tab.aborter) tab.switching = true;
abortTurn(tab);
\\\

This prevents the flag from getting stuck when no turn is in flight (the common session-switch path). Verified locally: lint ✅, typecheck ✅.

esengine:
Fixup looks right — gated on `tab.aborter` so the flag can't stick true on the common path. Merging.

PR #1585, opened 2026-05-23
fix(desktop): add responsive sidebar collapse

### LINKED ISSUE #855: Windows Reasonix Desktop UI 问题
Image

如图，我是用的是Reasonix Desktop for Windows，版本为0.42.0-3，按理来说在Windows上不应该出现图中红色框框选的内容；还有一个问题就是当app的窗口到一定比例时会出现蓝色框框住的样子

T1anjiu:
## Summary

Add responsive, stage-based sidebar auto-collapse in the desktop chat UI so the center composer stays usable on narrow windows, and restore only the sidebars that were auto-collapsed when the window widens again.

## Problem

When the desktop window is narrowed, the left session sidebar and right context panel can keep occupying fixed width. That leaves the middle chat column too narrow, which is especially bad for the composer footer: model selection and inline tools can overflow the input area and become clipped.

The issue lives in the desktop shell layout:
- `desktop/src/App.tsx` owns `sideCollapsed` and `ctxCollapsed` state
- `desktop/src/styles.css` defines the three-column grid layout and composer footer sizing
- There was no responsive stage logic, so sidebar behavior depended only on manual toggles

## Fix

- **`desktop/src/App.tsx`** — Added width-stage detection for `wide`, `compact`, and `narrow` layouts.
- **`desktop/src/App.tsx`** — Auto-collapses the right context panel first, then the left sidebar at a lower breakpoint.
- **`desktop/src/App.tsx`** — Restores only sidebars that were auto-collapsed when the window becomes wide again.
- **`desktop/src/App.tsx`** — Keeps manual sidebar toggles independent, so a user-closed sidebar is not reopened automatically.
- **`desktop/src/styles.css`** — Switched the shell grid to CSS variables and added a smooth transition for sidebar width changes.
- **`desktop/src/styles.css`** — Added composer footer overflow guards so the bottom toolbar does not spill out in narrow layouts.

## Verification

| Check | Result |
|---|---|
| `npm run build` in `desktop/` | ✅ passes |
| `git diff --check` | ✅ passes |
| `git status --short --branch` after commit | ✅ clean |

Closed #855


esengine:
Solid PR overall — the stage state machine handles WIDE/COMPACT/NARROW transitions correctly, the auto-vs-manual tracking via refs is the right pattern, RAF-debounced resize is good practice, and the CSS variable refactor of `.app` columns is clean. One blocker to fix before this can land, plus two follow-on gaps worth handling in the same revision.

**Blocker — `overflow: hidden` on `.composer-foot` clips the model-picker dropdown**

The new rule

```css
.composer-foot {
  ...
  overflow: hidden;
}
```

breaks the ModelMenu popup. In `desktop/src/ui/composer.tsx:497-565`:

- `.composer-foot` is the flex row.
- Inside it sits `<div ref={modelWrapRef} style={{ position: "relative" }}>` — this becomes the containing block for the popup.
- `<ModelMenu>` renders with `position: absolute; bottom: calc(100% + 6px); width: 260px`, popping **upward** roughly 150–200px above the model-pill — i.e. outside `.composer-foot`'s box.

With `overflow: hidden` on `.composer-foot`, that out-of-box area gets clipped to zero. The user clicks the model-pill and sees nothing — model switching becomes effectively invisible. The `@`/`/` `Popup` (line 593) is rendered as a sibling of `.composer-foot` rather than a descendant, so it's not affected; only ModelMenu is the casualty.

The fix is small: just drop `overflow: hidden` from `.composer-foot`. The flex-overflow problem is already solved by the `min-width: 0` chain you added on `.composer-foot`, `.grow`, and `.model-pill`. Keep the text-truncation rules at the leaf level (`.model-pill span { overflow: hidden; text-overflow: ellipsis; white-space: nowrap; }` and same on `.cf-btn .label` if needed). `overflow: hidden` on the flex parent is a sledgehammer that breaks any absolutely-positioned descendant escaping the box.

**Follow-on gap 1 — initial narrow startup isn't covered**

```ts
if (prev === null) {
  responsiveStageRef.current = next;
  return;
}
```

First `sync()` call after mount records the stage and exits. If the user opens the app on a narrow window with the default open state (or with LS-restored open state), the composer overflows until they manually resize — which is exactly the screenshot scenario in #855.

Run the normal stage logic on first mount too:

```ts
if (prev === null) {
  if (next === RESPONSIVE_STAGE.NARROW) {
    if (!ctxCollapsedRef.current) { setCtxCollapsed(true); autoCtxCollapsedRef.current = true; }
    if (!sideCollapsedRef.current) { setSideCollapsed(true); autoSideCollapsedRef.current = true; }
  } else if (next === RESPONSIVE_STAGE.COMPACT) {
    if (!ctxCollapsedRef.current) { setCtxCollapsed(true); autoCtxCollapsedRef.current = true; }
  }
  responsiveStageRef.current = next;
  return;
}
```

**Follow-on gap 2 — auto-collapse pollutes localStorage**

`sideCollapsed` / `ctxCollapsed` are persisted by the existing effects. When auto-collapse flips them, that gets written too. After the user spends time on a narrow window, LS becomes `sideCollapsed=1, ctxCollapsed=1`. Next launch on a wide screen: both panels restore as collapsed, the first `sync()` doesn't reopen them (refs don't persist, so `autoSideCollapsedRef.current` is `false` on mount), and the user sees both sidebars hidden on a window that has plenty of room.

The cleanest fix is to skip the LS write when the change came from auto-collapse — gate the existing persistence effects on the auto refs:

```ts
useEffect(() => {
  if (autoSideCollapsedRef.current) return;
  localStorage.setItem("reasonix.sideCollapsed", sideCollapsed ? "1" : "0");
}, [sideCollapsed]);
```

(Same shape for ctx.) That way the persisted value reflects what the user explicitly chose, and the responsive layer never writes to LS.

**Nit (optional)**

`RESPONSIVE_STAGE` as a `const`-object + `(typeof X)[keyof typeof X]` works, but a string-literal union is half the code:

```ts
type ResponsiveStage = "wide" | "compact" | "narrow";
```

No strong opinion — keep if you prefer the named constants.

Push a fixup with the blocker fix (and ideally the two follow-on gaps) and I'll merge.

PR #1588, opened 2026-05-23
fix(ui): preserve user message newlines

T1anjiu:
## Summary

Preserve multiline formatting for user message bubbles in the desktop and dashboard chat UI, so text sent with line breaks continues to display as multiple lines after submission.

## Problem

When users send multiline messages, the message content itself is preserved, but the rendered user bubble collapses whitespace because the `.msg-text` style does not preserve line breaks. As a result, newlines are rendered as spaces and multiline user input appears as a single line.

The issue is in the desktop/dashboard message rendering path:
- `desktop/src/ui/thread.tsx` renders user messages through `UserMsg`
- `dashboard/src/ui/thread.tsx` renders user messages through `UserMsg`
- `desktop/src/styles.css` and `dashboard/src/styles.css` styled `.msg-text` without `white-space: pre-wrap`
- TUI/CLI message storage and rendering paths preserve the original user text and do not apply the same CSS whitespace collapsing behavior

## Fix

- **`desktop/src/styles.css`** — Added `white-space: pre-wrap` for `.msg.user .msg-text` and `.msg.user .msg-text.user-pending`, preserving user-entered newlines in desktop message bubbles.
- **`dashboard/src/styles.css`** — Added the same user-message-specific whitespace rule for dashboard message bubbles.
- **`tests/user-message-newlines.test.ts`** — Added a regression test that verifies both desktop and dashboard styles include the user bubble whitespace preservation rule.

## Verification

| Check | Result |
|---|---|
| `npm exec vitest run tests/user-message-newlines.test.ts tests/desktop-user-message.test.ts` | ✅ passes |
| `npm run verify` | ✅ passes |
| Commit pre-commit lint hook | ✅ passes |

Closed #1567

PR #1590, opened 2026-05-23
fix(ui): prevent long messages from overflowing chat bubbles

yvone1991:
## What

Updates the desktop and dashboard message styles so long unbroken user-visible text, such as Windows paths, URLs, and long tokens, wraps inside chat bubbles instead of stretching the layout.

## Changes

- Rebased on the latest upstream `main`.
- Mirrored the wrapping fix in both `desktop/src/styles.css` and `dashboard/src/styles.css`.
- Removed the now-redundant user-specific `white-space: pre-wrap` rule added by #1588, since `.msg-text` now carries that behavior directly.
- Dropped the newly added `word-break: break-word` usage and rely on the standardized `overflow-wrap: anywhere`.
- Updated `tests/user-message-newlines.test.ts` so it checks the broader shared message-text rules instead of the removed user-specific selector.

## Verification

- Ran `npm run verify`
- Result: build, lint, typecheck, and tests passed

### DIFF
--- a/dashboard/src/styles.css
+++ b/dashboard/src/styles.css
@@ -1651,14 +1651,10 @@ html:not([data-platform="macos"]) .shortcut kbd[data-key="mod"] {
 .msg.user .body {
   flex: none;
   max-width: 82%;
+  min-width: 0;
   background: var(--panel-2);
   border-radius: 12px;
   padding: 10px 14px;
-}
-
-.msg.user .msg-text,
-.msg.user .msg-text.user-pending {
-  white-space: pre-wrap;
 }
 
 .msg-text {
@@ -1666,9 +1662,17 @@ html:not([data-platform="macos"]) .shortcut kbd[data-key="mod"] {
   line-height: 1.72;
   color: var(--fg);
   text-wrap: pretty;
+  min-width: 0;
+  overflow-wrap: anywhere;
+  white-space: pre-wrap;
+}
+.msg-text .markdown {
+  white-space: normal;
+  overflow-wrap: anywhere;
 }
 .msg-text p {
   margin: 0 0 12px;
+  overflow-wrap: anywhere;
 }
 .msg-text p:last-child {
   margin-bottom: 0;
@@ -1681,6 +1685,7 @@ html:not([data-platform="macos"]) .shortcut kbd[data-key="mod"] {
   padding: 1px 6px;
   border-radius: 4px;
   font-size: 0.875em;
+  overflow-wrap: anywhere;
 }
 .msg-text pre code,
 .markdown pre code {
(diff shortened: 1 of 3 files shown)


esengine:
Thanks — the CSS logic is right. I traced the cascade for both user and assistant messages:

- User: `<div class="msg-text">{text}</div>` → `pre-wrap` + `overflow-wrap: anywhere` apply, long unbroken strings break. ✓
- Assistant: `<div class="msg-text"><div class="markdown">…</div></div>` → the outer `.msg-text` gets `pre-wrap`, the inner `.markdown` resets to `normal`, so markdown semantic whitespace is preserved. Inner `<p>`/`<code>`/`<a>` pick up `overflow-wrap: anywhere` via the existing combined selectors. ✓

One blocker, plus two follow-on notes.

**Blocker — dashboard not touched, symmetry with #1588 broken**

The same vulnerability lives in `dashboard/src/styles.css` — the `.msg-text` / `.markdown` block structure is mirrored across desktop and dashboard. #1588 (merged about 30 minutes before this PR was opened) fixed both sides for the newlines case, which is the established repo pattern. Long file paths / URLs still blow out the user bubble on the web dashboard with this PR's current scope.

Please mirror the same 16 lines into `dashboard/src/styles.css` at the equivalent location. Same selectors, same rules.

**Follow-on notes (optional but recommended in the same revision)**

1. **Overlap with #1588.** #1588 added this to both `desktop/src/styles.css` and `dashboard/src/styles.css`:

   ```css
   .msg.user .msg-text,
   .msg.user .msg-text.user-pending {
     white-space: pre-wrap;
   }
   ```

   Once this PR lifts `white-space: pre-wrap` onto the broader `.msg-text` selector, the #1588 rule becomes a redundant subset (same property, lower specificity but same effect). Behavior doesn't change either way, but it's worth deleting those 4 lines in both files while you're rebasing on current main, so the next reader doesn't have to figure out why both rules exist. `git pull origin main` will surface them.

2. **`word-break: break-word` is a deprecated alias.** `overflow-wrap: anywhere` covers the same case and is the standardized form (Chrome 80+ / Firefox 65+ / Safari 15+, all well within Tauri WebView and Vite/dashboard's browser baseline). Keeping both is harmless but noisy — you can drop `word-break: break-word` everywhere it's added. No strong opinion if you'd rather keep belt-and-suspenders.

Push a fixup with the dashboard parity (and optionally the cleanups) and I'll merge.

yvone1991:
Thanks for the review. I pushed an update that rebases on latest main, mirrors the wrapping rules into `dashboard/src/styles.css`, removes the redundant user-specific `pre-wrap` rule from both CSS files, drops the newly added `word-break: break-word`, and updates `tests/user-message-newlines.test.ts` to assert the broader shared `.msg-text` behavior.

Verified with `npm run verify`.

PR #1620, opened 2026-05-23
fix(memory): JobRegistry self-cleanup for completed jobs

Bernardxu123:
## JobRegistry Self-Cleanup

Fixes #1571 (JobRegistry portion)

### Problem

Completed jobs accumulate in the JobRegistry Map indefinitely, retaining 64KB output buffers and ChildProcess references. Over a long session, this pins several MB of memory.

### Fix

Auto-evict oldest completed jobs when count exceeds MAX_COMPLETED_JOBS (20).

**Key decisions per author feedback:**
- `MAX_COMPLETED_JOBS = 20` — model doesn't need 50 historical entries
- No time throttle — gate on count only, O(jobs.size) is trivial at N=20
- `lastCleanupAt` initialized to `0` so first call always runs (no dead zone)

### Changes

- `src/tools/jobs.ts`: Added `maybeCleanup()` method, called from `settleClosed` and `list()`

### Testing

- npm run lint ✅
- npm run typecheck ✅

### DIFF
--- a/src/tools/jobs.ts
+++ b/src/tools/jobs.ts
@@ -108,6 +108,8 @@ export interface JobRecord {
 export class JobRegistry {
   private readonly jobs = new Map<number, InternalJob>();
   private nextId = 1;
+  /** Max completed jobs to retain for list_jobs / job_output lookups. */
+  private static readonly MAX_COMPLETED_JOBS = 20;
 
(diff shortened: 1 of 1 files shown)


esengine:
Clean split — narrow diff, both blockers addressed, no time-throttle dead zone since the throttle itself is gone. Merging.

PR #1621, opened 2026-05-23
test(lifecycle): cover npm package aliases

GTC2080:
## Summary

- cover `npm i` and `npm un` as package mutation aliases in the lifecycle policy corpus
- classify both aliases as high-risk package mutation commands, matching `npm install` and `npm uninstall`

## Validation

- `npm test -- tests/lifecycle-policy.test.ts`
- `npm test -- tests/lifecycle-policy.test.ts tests/lifecycle.test.ts tests/tools.test.ts`
- `npx tsc --noEmit`
- `npm run lint` (passes with the existing unused suppression warning in `tests/welcome-banner-hints.test.tsx`)

## Notes

`npm run typecheck` / push hook full `npm run verify` is blocked in this local checkout by missing dashboard dependencies such as `prism-react-renderer`, `lucide-react`, `react-markdown`, and related type declarations. The root `tsc --noEmit` check passes.

PR #1624, opened 2026-05-23
fix(ui): replace repeated live hints

### LINKED ISSUE #1573: pushInfo cards pile up — no way to replace a live hint in-place
**Problem**
 `pushInfo` always appends a new `live` card with a unique ID. There's no built-in way for a caller to
 say "replace my last hint" — every call creates a permanent new card. For transient or repeatable
feedback (hotkey confirmations, mode toggles, status changes), this means stale cards accumulate in
the scrollback with no path to clean them up.

 **Proposed change**
 Allow `live.show` to replace an existing card when the same `id` is dispatched again. The reducer
already has `mutateCard` — the `live.show` case could check for an existing card with the same `id +
kind` and mutate instead of always appending. Callers that pass a stable `id` get replace-in-place;
callers using `nextId()` (unique per call, which is the default today) keep current append behavior
unchanged. Zero breakage, backward-compatible.

 **Alternatives considered**
 - Toast notifications (`toast.show`) — already in the reducer with auto-dismiss via `ttlMs`, but
`Scrollback` doesn't expose `pushToast`. Adding one would also work, but a toast disappears entirely;
replace-in-place keeps the most recent hint visible until explicitly dismissed.
 - Doing nothing — the workaround is to accept the pile-up, which is noisy UX for anything that fires
a hint on a repeatable action.

 **Motivation**
 Working on a hotkey-driven feature (Alt+S input stash/recall, #1438) where each press shows a
confirmation hint — Recalled, Stashed, Nothing to stash. Without replace-in-place, rapid use creates a
 string of stale cards that never go away.

 **Scope check**

GTC2080:
## What

Make `live.show` replace an existing live card when the same id is reused, and route Alt+S stash/recall notices through a stable hint id so repeated presses update the same card instead of adding stale scrollback entries.

## Why

Fixes #1573. Repeated transient hints such as stash/recall confirmations were accumulating because every `pushInfo` call generated a fresh card id.

## How to verify

- `npm test -- tests/ui-reducer.test.ts`
- `npm run typecheck`
- `npm run verify`

## Checklist

### DIFF
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -198,6 +198,8 @@ import { useEditHistory } from "./useEditHistory.js";
 import { useSessionInfo } from "./useSessionInfo.js";
 import { useSubagent } from "./useSubagent.js";
 
+const STASH_HINT_CARD_ID = "composer-stash-hint";
+
 function isBusyPromptCommand(text: string): boolean {
   const trimmed = text.trimStart();
   return trimmed.startsWith("/") || trimmed.startsWith("#") || detectBangCommand(trimmed) !== null;
@@ -1906,13 +1908,13 @@ function AppInner({
         const recalled = stashRef.current;
         stashRef.current = input;
         setInput(recalled);
-        log.pushInfo(t("composer.stashRecall"));
+        log.pushInfo(t("composer.stashRecall"), "info", STASH_HINT_CARD_ID);
       } else if (input.length > 0) {
         stashRef.current = input;
         setInput("");
-        log.pushInfo(t("composer.stashSaved"));
+        log.pushInfo(t("composer.stashSaved"), "info", STASH_HINT_CARD_ID);
       } else {
-        log.pushInfo(t("composer.stashNothing"));
+        log.pushInfo(t("composer.stashNothing"), "info", STASH_HINT_CARD_ID);
       }
       return;
     }
--- a/src/cli/ui/hooks/useScrollback.ts
+++ b/src/cli/ui/hooks/useScrollback.ts
@@ -21,6 +21,7 @@ export interface Scrollback {
   pushInfo(
     text: string,
     tone?: "info" | "ok" | "warn" | "err" | "ghost" | "brand" | "accent",
+    id?: string,
   ): string;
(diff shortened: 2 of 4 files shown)


esengine:
Thanks — fix itself is tight and the reducer behavior test is exactly the right shape. One blocker though: the branch is behind main and would silently revert four merges that landed today.

## Drift

`git diff origin/main..HEAD --stat` against current main: **19 files, +49 / −152 lines.** The bulk of those deletions aren't yours.

| Merged into main | What this PR would silently revert |
|---|---|
| #1618 | Search-engine switcher — `dashboard/src/{App,protocol,ui/settings,lib/tauri-bridge}.tsx,ts` + `desktop/src/{App,protocol,ui/settings,i18n/{en,zh-CN}}.tsx,ts` + `src/cli/commands/desktop.ts` |
| #1620 | `JobRegistry.maybeCleanup` (`src/tools/jobs.ts` −18 lines) |
| #1621 | `npm i` / `npm un` aliases in `lifecycle-policy.ts` + the two new test cases |
| #1623 | Light-theme composer bg (`PromptInput.tsx` / `UserCard.tsx` theme-token swap) |

GH shows `mergeStateStatus: CLEAN` because git's 3-way merge sees absence-of-lines as deletion against main's additions — no textual conflicts to flag. Semantically it'd undo ~150 lines of recent work in a single click.

## What to do

`git fetch origin main && git rebase origin/main` — your diff should narrow back to the 4 files of the actual fix:

```
src/cli/ui/App.tsx                |  4 +++-
src/cli/ui/hooks/useScrollback.ts |  4 ++--
src/cli/ui/state/reducer.ts       |  9 ++++++---
tests/ui-reducer.test.ts          | 32 ++++++++++++++++++++++++++++++++
```

Main's been moving fast this week — if the rebase takes a while, you may need a second one. Push when ready and I'll merge.

dacec354:
let more message apply to this mode? like edit mode change or else?

GTC2080:
@dacec354 Thanks for the suggestion. Since this PR is already merged and was intentionally scoped to repeated live hints, could you open a separate issue for this request?

It would be helpful to include the exact mode(s), the messages you want covered, and the expected behavior, so we can discuss the scope clearly there.

PR #1627, opened 2026-05-23
fix(memory): auto-fold AppendOnlyLog when entries exceed 500

Bernardxu123:
## Memory Leak Fix — AppendOnlyLog Fold Trigger

Fixes parts of [#1571](https://github.com/esengine/DeepSeek-Reasonix/issues/1571)

### Problem

Long-running sessions accumulate log entries without limit — each tool result (up to 8K tokens) is stored until explicit /compact. Over hours this causes unbounded memory growth (~2MB/sec during active tasks).

### Fix

Add a log-length check in ContextManager.decideAfterUsage(): when entries exceed MAX_LOG_ENTRIES_FOR_FOLD (500, ≈1MB), trigger the existing fold mechanism.

This reuses the proven fold path — no new compression logic, just a length-based trigger alongside the existing token-ratio triggers. The constant is intentionally generous (500 entries) to avoid folding during normal bursty tool-call sequences. Sessions that reach this threshold without hitting the token-ratio fold are exactly the ones that need it: many small calls that individually stay under the ratio but collectively grow the log without bound.

### Changes

- src/context-manager.ts: Added MAX_LOG_ENTRIES_FOR_FOLD = 500 constant and length check in decideAfterUsage()

### Testing

- npm run lint ✅
- npm run typecheck ✅
- npm run test ✅ (3588 passed)

### Update

Drift issue fixed — rebased on latest main. Diff now only contains src/context-manager.ts.

### DIFF
--- a/src/context-manager.ts
+++ b/src/context-manager.ts
@@ -54,6 +54,9 @@ export const HISTORY_FOLD_SUMMARY_TIMEOUT_MS = 15_000;
 /** Prepended to fold summary content so the model knows it's a synthesized recap. */
 export const HISTORY_FOLD_MARKER =
   "[CONVERSATION HISTORY SUMMARY — earlier turns folded for context efficiency]\n\n";
+/** Maximum log entries before triggering an automatic fold (regardless of token usage).
+ *  Each entry retains message content, tool calls, and role metadata — 500 entries ≈ 1 MB. */
+export const MAX_LOG_ENTRIES_FOR_FOLD = 500;
 /** Header that precedes preserved skill bodies in a fold's synthesized assistant message. */
 export const SKILL_PIN_MEMO_HEADER = "[Active skill memos — preserved verbatim across the fold:]";
 /** Matches the wrapper emitted by `run_skill` so the fold can lift bodies out before summarizing. */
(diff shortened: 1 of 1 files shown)


esengine:
Thanks for splitting this out cleanly — exactly the shape I asked for when closing #1605. The trigger placement (after `alreadyFoldedThisTurn`, before the token-ratio paths) and the one-line WHY comment both look right. One blocker before merge: same drift pattern hitting a few PRs today.

## Drift

`git diff origin/main..HEAD` shows two extra files in the diff that aren't part of your fix:

```
src/cli/ui/PromptInput.tsx     |  6 +++---
src/cli/ui/cards/UserCard.tsx  |  2 +-
```

The change flips `TONE.brand` / `SURFACE.bgInput` back to hardcoded `#0153e5` / `#1e1e1e` — that's #1623 (light-theme composer bg) which merged earlier today. GH says `mergeStateStatus: CLEAN` because git's 3-way merge sees absence-of-lines as deletion against main's additions; there are no textual conflicts to flag. Semantically merging this would undo the light-theme fix.

## What to do

`git fetch origin main && git rebase origin/main` — your diff should narrow back to the one `src/context-manager.ts` file. The PromptInput.tsx and UserCard.tsx lines should disappear from the diff entirely after rebase.

Main's been moving fast this week — if the rebase takes more than a few hours you may need a second one. Push when ready and I'll merge.

The SessionStats rolling-window piece from the original #1605 — same shape, separate PR whenever you're ready.

Bernardxu123:
Rebased on latest main — drift cleaned up. Diff now only contains \src/context-manager.ts\.

Bernardxu123:
Branch restored and PR reopened. Diff is clean — only \src/context-manager.ts\.

PR #1645, opened 2026-05-24
feat(desktop): copy/edit overlay, msg history nav, IME guard, card collapse

CVEngineer66:
## Summary

Desktop UX improvements addressing daily pain points.

## Motivation

Several issues noticed during daily desktop usage:

- No way to edit a sent message — had to retype everything or use /retry (a CLI command, not suitable for the desktop UI)
- Copy button sat below the message, wasting vertical space with an empty line
- Code blocks had no one-click copy button — had to manually select code then press Cmd+C
- No message history navigation — could not recover a previously sent message with arrow keys
- Chinese/Japanese IME users: pressing Enter to confirm a composition candidate accidentally sent the message
- Reasoning cards and completed shell execution cards were expanded by default, taking up unnecessary space
- Homebrew Node 26 on macOS (arm64, ~68 KB) was rejected by the hard-coded 100 KB threshold in find_real_node(), causing a false node not found error

## Changes

| File | Change |
|---|---|
| desktop/src-tauri/src/rpc.rs | Lower Node binary size threshold from 100 KB to 50 KB on macOS/Linux |
| desktop/src/App.tsx | Wire onEdit callback to UserMsg |
| desktop/src/Markdown.tsx | Fix pre renderer text extraction for react-markdown v9; add copy button to CodeBlock |
| desktop/src/styles.css | Overlay copy/edit buttons (bottom-right, hover); code block copy button styles |
| desktop/src/ui/cards.tsx | ReasoningCard defaultOpen=false; ShellCard defaultOpen=state !== done |
| desktop/src/ui/composer.tsx | Message history (up/down arrow, 100-entry cap); IME guard (composingRef + timestamp) |
| desktop/src/ui/thread.tsx | Add pencil edit button to UserMsg |
| desktop/src/i18n/en.ts, zh-CN.ts | Add editMessage locale keys |

## Verification

- npm run verify passes (build, lint, typecheck, 3582 tests)
- Manually tested in Tauri desktop app

### DIFF
--- a/desktop/src-tauri/src/rpc.rs
+++ b/desktop/src-tauri/src/rpc.rs
@@ -98,15 +98,16 @@ fn resolve_cli(app: &AppHandle) -> Result<(String, Vec<String>)> {
 }
 
 /// Walk every PATH match for `node` and return the first one that is
-/// (a) a real file > 100 KB and (b) NOT inside Windows' App Execution
-/// Alias directory — those Microsoft Store stubs are 0-byte and triggered
-/// the original "%1 is not a valid Win32 application" (error 193).
+/// (a) a real file > 50 KB on macOS/Linux, or > 100 KB on Windows, and
+/// (b) NOT inside Windows' App Execution Alias directory.
+/// Threshold must accommodate Homebrew Node ~68 KB on arm64 macOS.
 fn find_real_node() -> Result<PathBuf> {
     let names: &[&str] = if cfg!(windows) {
         &["node.exe", "node"]
     } else {
         &["node"]
     };
+    let min_size: u64 = if cfg!(windows) { 100_000 } else { 50_000 };
     let mut tried: Vec<String> = Vec::new();
     for name in names {
         if let Ok(iter) = which_all(*name) {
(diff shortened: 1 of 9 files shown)


esengine (changes requested):
Thanks for the comprehensive pass — the Markdown v9 fix, the IME composition guard, and the Node threshold are all solid catches.

One blocker before this can land:

**Arrow-key history nav clobbers cursor movement in multi-line drafts.** In `composer.tsx` the handler intercepts every `ArrowUp`/`ArrowDown` whenever no popup is open, so a user writing a long message who hits ↑ to move the cursor up a line will instead see their entire draft replaced by a prior message. That's a real regression on the most common editing path.

The typical fix is to only navigate history when the action is unambiguous:

- textarea is empty, OR
- cursor is on the first line (for ↑) / last line (for ↓), OR
- modifier-gated (`Cmd+↑` / `Ctrl+↑`)

A few other things to address before merge:

- **CI hasn't run yet** — the `action_required` status is GitHub's first-contributor gate. Once the cursor regression is fixed I'll approve the workflow run.
- **Tests are missing** for every new behavior (history nav, IME guard, `flattenChildText`, edit-button wiring). 147 lines of behavior change with zero coverage is a real gap — small vitest cases on the pure helpers would have caught the cursor regression above.
- **`ReasoningCard defaultOpen={false}`** changes streaming visibility (was open while streaming, now collapsed). Confirm this is intentional — some users rely on seeing the reasoning land live.

The bundling (7 unrelated fixes in one PR) is fine for this round, but please split future ones — Rust threshold, Markdown fix, and the UX changes each belong in their own PR.

Once the cursor issue is fixed, push and I'll approve CI.

esengine:
Nice first PR — solid bundle of real fixes (the arm64 Homebrew Node detection and the IME guard in particular). Merging.

PR #1655, opened 2026-05-24
feat(desktop): code block syntax highlighting follows light/dark theme

CVEngineer66:
## Summary

Code block syntax highlighting colors now switch together with the app theme (light ↔ dark).

## Motivation

The Prism syntax highlighting theme in `CodeView.tsx` was hardcoded with dark-mode-only colors (`#dde1ea` on dark backgrounds). When switching to light mode (`data-theme=light`), the code text remained in pale dark-mode colors — nearly invisible on a white background.

## Changes

`desktop/src/CodeView.tsx` — Added a `LIGHT_THEME` with GitHub-style light syntax colors alongside the existing `DARK_THEME`. Added a `usePrismTheme()` hook that observes `<html data-theme>` via `MutationObserver` and returns the appropriate theme. `CodeView` now passes the dynamic theme to `<Highlight>`.

Only the syntax colors change — the code block background is already handled by CSS variables.

## Verification

- `npm run verify` passes
- Manually tested: toggling theme in desktop app swaps code block colors instantly
- MutationObserver only fires on `data-theme` attribute changes (no unnecessary re-renders)

### DIFF
--- a/desktop/src/CodeView.tsx
+++ b/desktop/src/CodeView.tsx
@@ -1,7 +1,7 @@
 import { Highlight, type PrismTheme } from "prism-react-renderer";
-import { useState } from "react";
+import { useEffect, useRef, useState } from "react";
 
-const THEME: PrismTheme = {
+const DARK_THEME: PrismTheme = {
   plain: { color: "#dde1ea", backgroundColor: "transparent" },
   styles: [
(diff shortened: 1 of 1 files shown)


esengine:
Thanks @CVEngineer66 — nicely scoped this time (one file, one concern) and the MutationObserver pattern is clean. Merged.

Heads-up: `dashboard/src/CodeView.tsx` has the same hardcoded-dark-theme issue. If you'd like to mirror the fix there, that would be a great follow-up.

PR #1662, opened 2026-05-24
feat(desktop): render edit_file/multi_edit results as DiffCard

CVEngineer66:
## Summary

Edit tool results (`edit_file` / `multi_edit`) now display as a unified diff card instead of raw text, making file changes visually scannable.

## Changes

- **cards.tsx** — Added `parseEditResult()` function that parses the tool output format (unified diff with optional `# filename` markers for multi-file edits) into `DiffLine[]`, and extracts filenames. Supports single-file (`edited foo.ts`) and multi-file (`multi_edit: applied N edits across M files`) formats.

- **thread.tsx** — In `AssistantMsg`, when a tool segment is `edit_file` or `multi_edit` and has a result, render `DiffCard(s)` instead of the generic `ToolCard`. Falls back to `ToolCard` if parsing yields no diff lines.

## Verification

- `npm run verify` passes

### DIFF
--- a/desktop/src/ui/cards.tsx
+++ b/desktop/src/ui/cards.tsx
@@ -400,6 +400,70 @@ export type DiffLine =
   | { t: "add"; r: number; s: string }
   | { t: "rm"; l: number; s: string };
 
+export function parseEditResult(text: string): { filename: string; lines: DiffLine[] }[] {
+  const files: { filename: string; lines: DiffLine[] }[] = [];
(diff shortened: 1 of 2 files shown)


esengine:
Thanks for the contribution — merged. Nicely done parsing both single-file (`edited foo.ts (...)`) and multi-file (`# foo.ts` markers) outputs in one pass.

PR #1664, opened 2026-05-24
feat(dashboard): code block syntax highlighting follows light/dark theme

CVEngineer66:
Mirrors the fix from #1655 to the dashboard.

Same issue: dashboard/src/CodeView.tsx had a hardcoded dark-only Prism theme that did not respond to data-theme switching.

Same fix: added LIGHT_THEME with GitHub-style light syntax colors plus usePrismTheme() hook observing html data-theme via MutationObserver.

Verification: npm run build passes.


esengine:
Thanks — merged. Clean mirror of #1655.

PR #1668, opened 2026-05-24
feat: wire plan edit mode across clients

T1anjiu:
## Summary

Wire `editMode = "plan"` into the actual tool dispatch gate so plan mode is truly read-only, including in desktop sessions, and clean up stale desktop preset wiring that was breaking typecheck.

## Problem

`editMode = "plan"` was being saved and surfaced in the UI, but it did not reliably enable `ToolRegistry.planMode`, which is the layer that actually blocks mutating tools like `write_file`, `edit_file`, and `multi_edit`.

This meant the desktop app could show or persist plan mode while the active tool registry still allowed file writes.

There was also stale desktop dashboard preset code in `App.tsx` referencing removed APIs and fields:
- `autoEscalate`
- `setProNextLive`
- `armProForNextTurn`
- `disarmPro`

Those references no longer exist in the current preset / loop types and caused `npm run typecheck` to fail.

## Fix

- **`src/code/setup.ts`** — Added `applyPlanMode()` and call it during `buildCodeToolset()` so `editMode = "plan"` initializes `ToolRegistry.planMode`.
- **`src/cli/commands/desktop.ts`** — Synchronized desktop runtime/toolset state when edit mode is loaded, emitted, or changed via settings, so switching to plan mode takes effect immediately.
- **`src/cli/ui/App.tsx`** — Removed stale `autoEscalate` / pro-next dashboard wiring and aligned preset handling with the current `flash | pro` model.
- **`tests/code-setup-lazy-subagent.test.ts`** — Added a regression test proving `buildCodeToolset()` blocks `write_file` when config edit mode is `plan`.
- **`tests/config.test.ts`** and **`tests/welcome-banner-hints.test.tsx`** — Removed stale unused Biome suppressions that made lint fail.

## Verification

| Check | Result |
|---|---|
| `npm test -- --run tests/code-setup-lazy-subagent.test.ts tests/tools.test.ts` | passes |
| `npm run typecheck` | passes |
| `biome check src tests` via commit hook | passes |
| `git commit` | passes |


esengine:
Took the core dispatch-gate wiring (`applyPlanMode` + `buildCodeToolset` init + desktop runtime sync + regression test) and shipped it as #1681 (merged), with co-authored credit preserved.

Dropping this PR because the rest of the diff (preset cleanup, 4-stop edit-mode cycle, composer/settings/i18n surface across three locales) either overlaps with work that already landed on main since this PR was opened — most of the preset/auto-escalate references were removed in #1657 — or is a larger design decision worth doing as its own scoped PR.

If you want the discoverability surface for plan mode (settings UI option / Shift+Tab cycle / i18n), happy to review a focused follow-up that builds on #1681.

Closing in favor of #1681.

Issue #1619, opened 2026-05-23
QQ机器人与电脑会话错乱bug


rayL-K:
会话1：手机上与QQ机器人的会话
会话2：电脑上直接与reasonix的会话

场景复现：
会话1回复未结束的时候开启会话2，
原本的会话1不会在QQ机器人口中回复（reasonix里是有的），
并且在会话2回答结束的时候其回复会出现在会话1QQ机器人的口中，
并且之后一直占据会话1中的QQ机器人

wade19990814-hue:
﻿There are two separate behaviors mixed together in this report:

1. By design, desktop QQ follows the current active tab. If you switch from tab 1 to tab 2, later QQ messages are expected to go to tab 2.
2. The part that still looks like a real bug is different: an older QQ-started turn can lose its final reply after a local turn starts in another tab, and QQ pause/choice prompts can also leak across tabs.

I checked the desktop QQ runtime locally and the second part does look actionable. The current implementation keeps some QQ reply-routing state globally instead of binding it to the originating turn/tab.

So the issue seems to be:
- switching the active tab itself: expected behavior
- an already-running QQ turn losing its final reply, or follow-up prompts leaking after another tab starts work: likely a real bug

I’m working on a fix for that second part now.

wade19990814-hue:
A fix is now up for review in #1672.

This keeps the existing active-tab binding semantics for future QQ ingress, but stops an older QQ-started turn from losing its final reply after a local turn starts in another tab. It also stops local-only pause prompts from leaking into QQ.

If you still see behavior outside that scope after updating to a build that includes the fix, please post the exact step sequence and whether the later turn was started locally or from QQ.

Bernardxu123:
🛑 Closed — archived to #3152. V2 rewritten. File new issue if still relevant.

PR #1672, opened 2026-05-24
fix(desktop): isolate QQ turn routing by tab

wade19990814-hue:
## Summary
- isolate desktop QQ reply routing and pending follow-up prompts by tab
- keep active-tab binding semantics for future QQ ingress, while preserving final replies for older QQ-started turns already in flight
- add a regression test for the per-tab QQ routing state

## Testing
- npm test -- --run tests/desktop-qq-turn-routing.test.ts tests/desktop-user-message.test.ts
- npm run typecheck
- npm run lint
- npm run verify

## Context
Fixes the actionable part of #1619. Switching the active tab is still expected to retarget later QQ messages, but an older QQ-started turn should not lose its final reply after a local turn starts in another tab, and local-only pause prompts should not leak into QQ.

### DIFF
--- a/src/cli/commands/desktop.ts
+++ b/src/cli/commands/desktop.ts
@@ -67,6 +67,15 @@ import {
   saveDesktopQQSettings,
   setDesktopQQEnabled,
 } from "../../desktop/qq-settings.js";
+import {
+  clearQQTurnRouting,
+  createQQTurnRoutingState,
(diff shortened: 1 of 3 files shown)


esengine:
Thanks — merged. Cleanly factored: pure routing module + per-tab Set/Map replaces the global flat fields, exactly what #1619 needed.

PR #1683, opened 2026-05-24
fix(diff): use stringWidth for CJK-aware truncation in SplitDiff

### LINKED ISSUE #1671: Wrong character witdh in diff panel
> Display problems almost always come from the **terminal emulator**, not
> the shell. Please fill the terminal section carefully — `bash vs PowerShell`
> tells us very little; `VSCode integrated terminal vs Windows Terminal`
> tells us everything.

**Symptom** (tick all that apply)
- [ ] Whole screen flickers / flashes during streaming response
- [ ] Lines tear or only half-redraw
- [ ] Stale output left behind after a frame updates
- [ ] Cursor jumps to wrong column or vanishes
- [x] Mojibake / wrong-width characters (e.g. `□`, half-width emoji)
- [ ] Other (describe below)

**When it happens**
- [ ] During assistant streaming (token-by-token output)
- [ ] When tool cards expand / collapse
- [ ] During scroll-up / scrollback
- [ ] On terminal resize
- [ ] On launch / on quit
- [x] Other (describe below)

In review mode, diff panel calculates wrong character witdh

**Terminal — the important part**

Where exactly are you running `reasonix`?

- [ ] **VSCode** integrated terminal — VSCode version: `?`
- [ ] **Cursor** integrated terminal — Cursor version: `?`
- [ ] **Windows Terminal** — version: `?`
- [ ] **cmd.exe** (legacy console host)
- [ ] **PowerShell ISE** (note: ISE doesn't support ANSI — most things will look broken)
- [x] **iTerm2** / **Terminal.app** / **Alacritty** / **kitty** / **WezTerm** / **Hyper**
- [ ] tmux / screen / mosh — and inside which outer terminal? `?`
- [ ] Other:

```bash
reasonix --version; node --version
uname -a
echo "TERM=$TERM"
echo "TERM_PROGRAM=$TERM_PROGRAM"
echo "TERM_PROGRAM_VERSION=$TERM_PROGRAM_VERSION"
echo "COLORTERM=$COLORTERM"
echo "WT_SESSION=$WT_SESSION"
echo "VSCODE_INJECTION=$VSCODE_INJECTION"
echo "WSL_DISTRO_NAME=$WSL_DISTRO_NAME"
```

</details>

```
0.49.0
v26.1.0
Linux archlinux-3302 7.0.9-arch1-1 #1 SMP PREEMPT_DYNAMIC Sun, 17 May 2026 17:23:07 +0000 x86_64 GNU/Linux
TERM=alacritty
```

**Reproduction**

Steps that reliably trigger it (commands run, files edited, was a tool
streaming a long response, was the window being resized, …):

1. start reasonix
2. switch to review mode
3. ask ai to modify something

**Screen recording (strongly preferred)**

Image

Wrong border positions exists only when Chinese character occur. So probably issue lies in width calculation?

Bernardxu123:
## Problem

In review mode, the diff panel calculates wrong character width when CJK (Chinese/Japanese/Korean) characters are present. This causes incorrect border positions and visual misalignment.

### Root Cause

In `src/cli/ui/SplitDiff.tsx`, the `Cell` function uses `raw.length` to calculate visual width for truncation:

```typescript
const truncated = raw.length > inner ? `${raw.slice(0, inner - 1)}…` : raw;
```

The problem is that `raw.length` counts JavaScript characters, not visual terminal cells. For CJK characters:
- `raw.length` counts each character as 1
- Terminal renders each CJK character as 2 cells wide

This mismatch causes the truncation calculation to be wrong, resulting in incorrect border positions when diff lines contain Chinese characters.

### Example

A diff line with Chinese characters like `函数返回值` has:
- `raw.length` = 5 (5 JavaScript characters)
- Actual visual width = 10 cells (each CJK char takes 2 cells)

When truncating to fit a column of width 8, the code thinks 5 < 8 so no truncation is needed, but the actual visual width is 10, which overflows.

## Solution

Replace `raw.length` with `stringWidth(raw)` from the existing `text-width.js` utility:

```typescript
import { stringWidth } from "./text-width.js";

// Before
const truncated = raw.length > inner ? `${raw.slice(0, inner - 1)}…` : raw;

// After
const truncated = stringWidth(raw) > inner ? `${raw.slice(0, inner - 1)}…` : raw;
```

The `stringWidth` function (from the `string-width` library) correctly handles:
- CJK characters (2 cells wide)
- Emoji (2 cells wide)
- Zero-width characters (0 cells)
- Regular ASCII (1 cell wide)

## Files Changed

- `src/cli/ui/SplitDiff.tsx` — Import `stringWidth` and use it for truncation calculation

## Testing

- ✅ Lint passed
- ✅ Typecheck passed  
- ✅ All 3596 tests passed

## Impact

This fix ensures the diff panel renders correctly regardless of the characters in the diff content. Users working with CJK characters (Chinese, Japanese, Korean) will now see correct border positions and proper text truncation.

Fixes #1671

### DIFF
--- a/src/cli/ui/SplitDiff.tsx
+++ b/src/cli/ui/SplitDiff.tsx
@@ -19,9 +19,10 @@
  */
 
 import { Box, Text, useStdout } from "ink";
-// biome-ignore lint/style/useImportType: tsconfig jsx=react needs React in value scope for React.Fragment
+// biome-ignore lint/style/useImportType: tsconfig jsx=react needs React in value scope for JSX compilation
 import React from "react";
 import type { SplitDiffRow } from "../../code/diff-preview.js";
+import { stringWidth } from "./text-width.js";
 import { COLOR } from "./theme.js";
 
 export interface SplitDiffProps {
@@ -72,7 +73,8 @@ function Cell({
   const sign =
     side.kind === "del" ? "-" : side.kind === "add" ? "+" : side.kind === "pad" ? " " : " ";
   const raw = side.text;
-  const truncated = raw.length > inner ? `${raw.slice(0, inner - 1)}…` : raw;
+  // Use stringWidth for CJK/emoji-aware truncation — raw.length undercounts wide chars.
+  const truncated = stringWidth(raw) > inner ? `${raw.slice(0, inner - 1)}…` : raw;
   // Pad to fixed width so the bg color stretches across the whole
   // column even when the text is short — without this the red/green
   // wash would only cover the actual chars and the rest of the row


esengine:
Thanks for spotting the bug. Shipped the full fix as #1686 (merged) — used `clipToCells` for the truncation and a promoted `padToCells` for the right-pad, both already living in `src/cli/ui/text-width.ts`. Co-authored credit preserved.

Closing in favor of #1686 because the `stringWidth` check alone left two halves of the same bug in place: `raw.slice(0, inner - 1)` still cuts by JS chars (so the truncated CJK string ends up wider than `inner` cells), and `truncated.padEnd(inner)` still pads by string length (so short CJK rows get over-padded). Both knock the column border out of alignment in the same way the original bug did.

Issue #1300, opened 2026-05-19
PowerShell 放大终端窗口时，模型思考，内容闪烁跳动


anuonie:
PowerShell 窗口最大化，当模型“思考中”，屏幕上的内容会频繁闪烁、跳动，不放大正常。

操作系统：Windows 10
Shell：PowerShell   6456
Major  Minor  Build  Revision
-----  -----  -----  --------
5      1      19041  6456
reasonix 版本：0.43.0
Node 版本：v22.17.0

esengine:
Fixed in `6df5791` — will ship in the next release.

Root cause: `powershell.exe` runs under conhost, which repaints each Ink frame visibly (no double-buffer like Windows Terminal). At our defaults — 8 Hz spinner tick + 20 Hz streaming flush — a maximized window has too many cells to redraw per frame, so the redraws stack into visible flicker and content "jumping". Modern terminals (Windows Terminal, WezTerm, vscode terminal) don't show this because they buffer cell updates.

The fix detects legacy conhost (`win32 && !WT_SESSION && !TERM_PROGRAM`) and drops the spinner to 4 Hz and the stream flush to ~7 Hz on that path. Reasoning-streaming repaints go from ~28 Hz total to ~11 Hz — well below the conhost flicker threshold. Modern terminals are unaffected.

A couple of notes:

- You're on `0.43.0`; we're at `0.47.0` now. Please upgrade (`npm i -g reasonix@latest`) and verify the fix works in your setup.
- If you can switch to **Windows Terminal** (`wt.exe`) for `powershell.exe` it'll be smoother across the board — conhost handles ANSI better than it used to, but it's still the weakest of the Windows terminals for TUI work.

PR #1684, opened 2026-05-24
fix(ui): throttle streaming/reasoning chunks to reduce flicker

### LINKED ISSUE #1626: 终端CLI 输入推理后，实时推理内容疯狂频闪
### Symptom

- [x] Whole screen flickers / flashes during streaming response
- [ ] Lines tear or only half-redraw
- [x] Stale output left behind after a frame updates
- [ ] Cursor jumps to wrong column or vanishes
- [ ] Mojibake / wrong-width characters
- [x] Other: `reasoning_content` inside `<think>` block re-renders the entire TUI on every token arrival, causing severe visual flicker. The issue intensifies when the thinking block is long (hundreds of tokens).

### When it happens

- [x] During assistant streaming (token-by-token output)
- [ ] When tool cards expand / collapse
- [ ] During scroll-up / scrollback
- [ ] On terminal resize
- [ ] On launch / on quit

### Terminal

- [x] **Windows Terminal** — version: `?` (please fill)
- [x] WSL2 (Ubuntu-26.04)
- [x] Reproduced identically in both WSL and native PowerShell environments

### Reproduction

1. Open Windows Terminal, launch WSL Ubuntu-26.04
2. Run `reasonix code` (or `reasonix chat`)
3. Submit a task that triggers complex reasoning — e.g. "refactor this module, explain every design decision"
4. DeepSeek enters `reasoning_content` phase (`<think>` block)
5. **Observe**: every incoming chunk from the streaming SSE response triggers a full-frame Ink re-render. The thinking block jitters/stutters violently. Worse on long think blocks (400+ tokens).
6. Identical behavior in `reasonix chat` mode (no filesystem tools loaded — rules out tool-card rendering as culprit).
7. Present in both WSL2 (Windows Terminal) and native Windows PowerShell (Windows Terminal + cmd.exe).

### Suspected root cause

`Ink` (React for terminals) re-renders the component tree on every state update. During SSE streaming of `reasoning_content`, each token arrival updates state → triggers `React reconciliation` → Ink diffs and redraws the ANSI output. Without a throttle/debounce on the rendering loop, high-frequency token streaming (~30–60 tokens/sec on v4-flash) causes per-frame flicker.

Suggested fix: batch reasoning token updates into a single render per ~50–100ms window, or decouple the reasoning pane from the main app render cycle (use `React.memo` + shallow comparison on the thinking display component).

### Already tried

- [x] Switching `gpuAcceleration` in VSCode terminal — N/A (using Windows Terminal)
- [x] Running in native PowerShell vs WSL — same result
- [x] Detaching terminal window — same result

Bernardxu123:
## Problem

In the terminal CLI, reasoning content flickers violently during streaming. This issue affects both Windows Terminal (WSL2 + PowerShell) and macOS Terminal.

### Root Cause

When `reasoning_content` streams in via SSE, each token arrival triggers:
1. `reasoning.chunk` event dispatched to reducer
2. State update: `text: c.text + event.text`
3. React reconciliation triggered via `useSyncExternalStore`
4. Ink redraws entire ANSI output
5. High-frequency tokens (~30-60/sec) cause per-frame flicker

### Current Flow

```
Token arrives → dispatch("reasoning.chunk") → state update → React reconcile → Ink redraw
```

## Solution: Throttle (33ms)

Change token dispatch from "send immediately per token" to "batch every 33ms" (~30fps):

```
Current: Token arrives → immediate dispatch → redraw (~30-60fps, flickers)
Fixed:   Token arrives → buffer → 33ms throttle → batch dispatch → redraw (~30fps, smooth)
```

### Implementation

Modified `src/cli/ui/hooks/useScrollback.ts`:

1. **Token Buffering**: Added `reasoningBufferRef` and `streamingBufferRef` to accumulate tokens
2. **Throttle Timer**: Added `reasoningTimerRef` and `streamingTimerRef` for 33ms intervals
3. **Buffered Dispatch**: `appendReasoning()` and `appendStreaming()` now buffer tokens instead of dispatching immediately
4. **Immediate Flush on End**: `endReasoning()` and `endStreaming()` flush buffers before ending
5. **Cleanup on Unmount**: `useEffect` cleanup ensures no tokens are lost if component unmounts

### Why Throttle over Debounce

| Feature | Debounce | Throttle |
|---------|----------|----------|
| Update frequency | Inconsistent (waits for silence) | Consistent (fixed intervals) |
| Token loss | Possible (during bursts) | None (flushes each interval) |
| User experience | Jerky, discontinuous | Smooth, continuous |

## Files Changed

- `src/cli/ui/hooks/useScrollback.ts` — Added throttle logic for streaming/reasoning chunks

## Testing

- ✅ Lint passed
- ✅ Typecheck passed
- ✅ All 3596 tests passed

## Platforms Tested

- Windows Terminal (WSL2 + PowerShell)
- macOS Terminal

## Impact

This fix reduces the render frequency from ~30-60fps to ~30fps during high-speed streaming, eliminating the violent flicker. The throttle provides consistent, smooth rendering without the inconsistency of debounce.

**Note**: This is an optimization, not a root cause fix. The root cause is Ink's rendering model (full redraw on every state change). Complete elimination would require modifying Ink to support diff-based rendering.

Fixes #1626

### DIFF
--- a/src/cli/ui/hooks/useScrollback.ts
+++ b/src/cli/ui/hooks/useScrollback.ts
@@ -1,6 +1,9 @@
-import { useMemo } from "react";
+import { useEffect, useMemo, useRef } from "react";
 import type { DoctorCheckEntry, PlanStep, TipSection } from "../state/cards.js";
 import { useDispatch } from "../state/provider.js";
+
+/** Throttle interval for streaming/reasoning chunk dispatches (~30fps). */
+const THROTTLE_MS = 33;
 
 let seq = 0;
 function nextId(prefix: string): string {
@@ -102,6 +105,39 @@ export interface Scrollback {
 
 export function useScrollback(): Scrollback {
   const dispatch = useDispatch();
+
+  // Throttle state for reasoning and streaming chunks
+  const reasoningBufferRef = useRef<Map<string, string>>(new Map());
+  const reasoningTimerRef = useRef<Map<string, ReturnType<typeof setTimeout>>>(new Map());
+  const streamingBufferRef = useRef<Map<string, string>>(new Map());
+  const streamingTimerRef = useRef<Map<string, ReturnType<typeof setTimeout>>>(new Map());
+
+  // Cleanup on unmount - flush all buffers
+  useEffect(() => {
+    return () => {
+      for (const [id] of reasoningBufferRef.current) {
+        const buffered = reasoningBufferRef.current.get(id);
+        if (buffered && buffered.length > 0) {
+          dispatch({ type: "reasoning.chunk", id, text: buffered });
+        }
+        const timer = reasoningTimerRef.current.get(id);
+        if (timer) clearTimeout(timer);
+      }
+      reasoningBufferRef.current.clear();
+      reasoningTimerRef.current.clear();
+
+      for (const [id] of streamingBufferRef.current) {
+        const buffered = streamingBufferRef.current.get(id);
+        if (buffered && buffered.length > 0) {
+          dispatch({ type: "streaming.chunk", id, text: buffered });
+        }
+        const timer = streamingTimerRef.current.get(id);
+        if (timer) clearTimeout(timer);
+      }
+      streamingBufferRef.current.clear();
+      streamingTimerRef.current.clear();
+    };
+  }, [dispatch]);
 
   return useMemo<Scrollback>(
     () => ({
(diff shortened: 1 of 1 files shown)


esengine:
Thanks for digging into the flicker, but this one's already covered upstream and the new throttle wouldn't actually help.

Tokens never reach `appendStreaming` / `appendReasoning` one-at-a-time — `assistant_delta` events land in `contentBuf` / `reasoningBuf` in `App.tsx`, and a `setInterval(flush, FLUSH_INTERVAL_MS)` drains them via `TurnTranslator.flushBuffers` (App.tsx:3132–3144). `FLUSH_INTERVAL_MS` defaults to 50ms (20fps) and drops to 150ms (~7fps) on legacy PowerShell (App.tsx:328–338, fix for #1300). So by the time tokens hit the dispatch path this PR throttles, they're already batched.

Adding another 33ms throttle on top stacks delay (50ms + 33ms ≈ 83ms p99 to first paint) without changing render frequency in a way the eye can see, and on PowerShell the bottleneck is the 150ms console rate, not the dispatch rate.

For a knob today: `REASONIX_FLUSH_MS=200` (or higher, up to 1000) at launch dials the existing throttle. The structural fix is the planned `<Static>`-based append-only renderer — that decouples the scrollback from per-state Ink reconciliation, which is what your "Note" correctly identifies as the real root cause.

Closing.

PR #1686, opened 2026-05-24
fix(diff): SplitDiff renders CJK without misaligning the column border

esengine:
## Summary
Two bugs in the same `Cell` body of `SplitDiff.tsx` that misalign the diff column border when CJK / emoji are present (#1671):

- **Truncation** used `raw.length > inner`, which undercounts wide chars — a CJK row that was visually too wide slipped past the check and overflowed into the gutter
- **Padding** used `padEnd(inner)`, which counts string length not visual cells — short CJK content got padded as if each wide char were 1 cell, so the red/green background stripe ran past the column edge

Repo already has `clipToCells(s, maxCells)` (cell-aware truncation + ellipsis) in `src/cli/ui/text-width.ts`, and a private `padToCells(text, cells)` (cell-aware right-pad) in `src/cli/ui/markdown.tsx`. Promoted `padToCells` to `text-width.ts` so `SplitDiff` can use both; the four lines around `truncated` collapse to `padToCells(clipToCells(raw, inner), inner)`.

## Supersedes #1683
That PR fixed the detection half (`stringWidth(raw) > inner`) but left both the JS-char `raw.slice(0, inner - 1)` cut and the `padEnd(inner)` pad in place. For CJK content the truncated string came out wider than `inner` cells, and short CJK rows still over-padded — the border was still misaligned, just for slightly different inputs. Co-authored credit preserved.

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npx vitest run tests/text-width.test.ts` — 22 passed (was 19; added a `padToCells` describe block covering ASCII pad, CJK pad without over-padding, and the no-op-when-full case)
- [x] `npx vitest run tests/markdown` — 28 passed (sanity-check the `padToCells` extraction from markdown.tsx)

### DIFF
--- a/src/cli/ui/SplitDiff.tsx
+++ b/src/cli/ui/SplitDiff.tsx
@@ -22,6 +22,7 @@ import { Box, Text, useStdout } from "ink";
 // biome-ignore lint/style/useImportType: tsconfig jsx=react needs React in value scope for React.Fragment
 import React from "react";
 import type { SplitDiffRow } from "../../code/diff-preview.js";
+import { clipToCells, padToCells } from "./text-width.js";
 import { COLOR } from "./theme.js";
 
 export interface SplitDiffProps {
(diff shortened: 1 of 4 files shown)


dacec354:
the pr dont closes issue?

PR #1687, opened 2026-05-24
fix(desktop): discard explicitly aborted turns

### LINKED ISSUE #1593: 桌面最新版本，在长文本编辑任务时依旧会考虑效率、token之类的再次篡改指令
大概一万字符的结构化文本。我输入内容完全重写，会继续以上述理由思考然后绕过提示词的要求，转而简化工作。

要求完全重写->思考降级转为部分修改->继续思考发现文本太长->继续降级任务变为微调->结果就是只在原内容上做了局部微调，大量内容完全不变。

GTC2080:
## Summary
- Adds an explicit abort mode that discards the unfinished user turn from the model-facing log.
- Uses that mode for the desktop Abort action so a re-sent long edit request starts from the new prompt instead of also carrying the interrupted one.
- Adds #1593 regression coverage proving the next request contains only the fresh rewrite instruction.

## Root Cause
Desktop explicit aborts used the same loop abort behavior as session switches. The loop persists the user prompt before the first API call so interrupted sessions remain visible, but for a user-initiated Abort that left the abandoned long edit prompt in history. The next request could then include both the stale interrupted task and the new prompt, making the model treat the user as repeatedly interrupting or narrowing the work.

Fixes #1593

## Validation
- `npm test -- --run tests/loop-user-persist.test.ts tests/loop.test.ts tests/desktop-crash-guards.test.ts`
- `npm test -- --run tests/public-api.test.ts tests/loop-user-persist.test.ts`
- `npx biome check src/loop.ts src/index.ts src/cli/commands/desktop.ts tests/loop-user-persist.test.ts`
- `npm run typecheck`
- pre-push `npm run verify`: 262 test files passed; 3610 tests passed; 12 skipped. Biome reported two existing unused-suppression warnings outside this change.

### DIFF
--- a/src/cli/commands/desktop.ts
+++ b/src/cli/commands/desktop.ts
@@ -1490,9 +1490,9 @@ export async function desktopCommand(opts: DesktopOptions): Promise<void> {
     return undefined;
   }
 
-  function abortTurn(tab: Tab): void {
+  function abortTurn(tab: Tab, opts: { discardCurrentTurn?: boolean } = {}): void {
     tab.aborter?.abort();
-    tab.runtime?.loop.abort();
+    tab.runtime?.loop.abort(opts.discardCurrentTurn ? { discardCurrentTurn: true } : undefined);
   }
 
   function tabSessionLabel(tab: Tab): string {
@@ -2049,7 +2049,7 @@ export async function desktopCommand(opts: DesktopOptions): Promise<void> {
     }
 
     if (msg.cmd === "abort") {
-      abortTurn(tab);
+      abortTurn(tab, { discardCurrentTurn: true });
       cancelPendingGates(tab);
       return;
     }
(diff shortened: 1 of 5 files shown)


esengine:
Thanks — merged. Cleanly factored: explicit `LoopAbortOptions.discardCurrentTurn` opts in, the iter-start + signal.aborted branches both honor it, default behavior unchanged for session switches and `new_chat`. Public API addition correctly threaded through `src/index.ts` and the allowlist test. The regression case (mid-fetch abort → next request body carries only the new user message) makes the intent unambiguous.

One UX trade-off worth noting for a possible follow-up: the aborted text disappears from session.jsonl on disk, so a reload/session-switch loses what the user typed. A later iteration could surface it back as a composer draft on Abort so the model log still gets cleared but the user can edit + resend without retyping. Not for this PR.

PR #1688, opened 2026-05-24
feat(desktop): draggable sidebar resize + responsive conversation panel width

CVEngineer66:
## Summary

Adds draggable resize handles between sidebar/main and main/context-panel,
and makes the conversation thread width dynamically respond to sidebar sizes.

## Changes

- **useResizable.ts** (new) — Mouse-drag hook that updates sidebar widths,
  persists custom sizes to localStorage, respects min/max constraints.
- **App.tsx** — Wire useResizable, render resize handles, compute
  --thread-max-width from available window space minus sidebar widths.
- **styles.css** — Resize handle styles (.resize-handle), position relative
  on .app, remove hardcoded --thread-max-width in favor of dynamic value.

## Behavior

| Sidebar state | Before | After |
|---|---|---|
| Both open (default) | 740px | ~796px (1440px window) |
| Both open, sidebar dragged narrower | 740px | Up to 1120px |
| One sidebar collapsed | 740px | 960px |
| Both collapsed | 1120px | 1120px |

Dragging a sidebar right edge adjusts its width (160px–40% of window).
Sizes persist in localStorage across restarts.

## Verification

- desktop build + typecheck pass
- Dragging left/right sidebar edges works
- Width persists after app restart

### DIFF
--- a/desktop/src/App.tsx
+++ b/desktop/src/App.tsx
@@ -77,6 +77,7 @@ import {
 } from "./ui/thread";
 import { WorkdirPop } from "./ui/workdir-pop";
 import { useAutoCollapse } from "./ui/useAutoCollapse";
+import { useResizable } from "./ui/useResizable";
 import { useAutoScroll } from "./ui/useAutoScroll";
 import { useDisableTextAssist } from "./ui/useDisableTextAssist";
 import { openUrl } from "@tauri-apps/plugin-opener";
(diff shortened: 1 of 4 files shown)


esengine:
Thanks — merged. Drag state lives in refs so mousemove doesn't thrash React, localStorage round-trip handles the persist, and the inline `--thread-max-width` correctly overrides the attribute-selector defaults. Nice touch on the stale `preset: "pro"` test cleanup too.

Two follow-up polish items (no rush): the thread max doesn't recompute on window resize (would need a `resize` listener to re-trigger the render), and the per-frame `setState` during drag could fold into `requestAnimationFrame` if it ever feels janky on a busy thread.

PR #1689, opened 2026-05-24
fix(dashboard): ignore IME confirm enter

### LINKED ISSUE #1669: Pressing the enter key when using Chinese input sends message immediately
Feels like this issue is neither a bug nor a rendering problem.

When using Chinese input in the Reasonix GUI/web, pressing Enter directly sends the Pinyin being typed. For example, in VSCode or the GitHub comment box you are currently typing in, pressing Enter first enters the Pinyin, and only when you press Enter again does it send the message.

Screen recordings are like that:

Image

Image

Mac os 15.4 and Reasonix Version 0.49.0 btw

GTC2080:
## Summary
- Mirror the desktop IME composition guard in the dashboard composer.
- Prevent the Enter key that confirms Chinese IME input from submitting the draft.
- Add regression coverage for the post-composition Enter path while keeping normal Enter send behavior.

## Root Cause
The desktop composer already tracks composition start/end and ignores the keydown that immediately follows compositionend on macOS. The dashboard composer lacked that guard, so the Enter key used to commit Pinyin could be handled as the submit shortcut.

Fixes #1669

## Validation
- npm test -- --run tests/dashboard-composer-ime.test.tsx
- npm run lint
- npm run typecheck
- npm run build:dashboard
- npm run verify

### DIFF
--- a/dashboard/src/ui/composer.tsx
+++ b/dashboard/src/ui/composer.tsx
@@ -171,6 +171,9 @@ export function Composer({
   const [modelMenuOpen, setModelMenuOpen] = useState(false);
   const nonceRef = useRef(0);
   const modelWrapRef = useRef<HTMLDivElement>(null);
+  // macOS Chinese IME fires compositionend before the confirm keydown.
+  const composingRef = useRef(false);
+  const compositionEndedAtRef = useRef(0);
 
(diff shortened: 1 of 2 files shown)


esengine:
Thanks — merged. Faithful mirror of the desktop guard (composingRef + 50ms compositionEndedAtRef window), and the test covers both paths (IME-confirm Enter suppressed / normal Enter still sends). Second clean PR in a row, much appreciated.

PR #1703, opened 2026-05-24
feat: preserve full tool results when truncated (#1615)

### LINKED ISSUE #1615: Preserve full tool results when truncated
**Problem**
 When a tool result exceeds the per-call token cap, `truncateForModelByTokens` removes the middle portion and inserts a `[…truncated ~N tokens …]` marker. The full result is lost — never saved anywhere. The model can't retrieve it, re-read it, or page through it. It can only issue a new tool call (narrower query, re-fetch) and hope the truncated part comes back.

**Proposed change**

 When a tool result is truncated, automatically save the full content to a project-accessible file and include the file path and truncation position in the truncation marker so the model can `read_file` the complete result if needed. The file should be written to a location the model already has read access to.

**Alternatives considered**
- Increasing the cap — only delays the same problem.
- Chunked / paginated retrieval — could work for `web_fetch` but not for `web_search` which is flat by nature.
- Letting the model request the full content — adds a round-trip and couples tool infrastructure to model behavior.
- Status quo — the `[…truncated …]` marker at least signals something was dropped; the model can re-query if it notices. The question is whether that's good enough in practice.

 **Scope check**

dacec354:
## What
close #1615 
当工具结果（如 `web_search`/`web_fetch`）超过 token 或字符上限被截断时，自动将完整原始内容保存到 `.reasonix/truncated-results/`，并在截断标记中嵌入文件路径。模型可通过 `read_file` + `range`/`head`/`tail` 分段获取完整数据，无需重新发起网络请求。

## Why

Issue #1615。长网页/搜索结果被截断后，模型只能重新请求碰运气，浪费 token 和时间。对于 `web_fetch` 这类外部 API 调用尤其明显——页面内容不会变，但每次重查都要重新下载。

## How to verify

1. `npm run verify` 全绿
2. 启动后执行 `web_fetch "https://odin-lang.org/docs/overview/"`，返回的截断标记中会包含：
   ```
   Full result saved at: .reasonix/truncated-results/<ts>-<uuid>-web_fetch.txt
   ```
3. 执行 `read_file ".reasonix/truncated-results/<文件名>"` 可获取完整 HTML
4. `read_file`/`list_directory`/`search_content` 等文件系统只读工具**不会**触发保存（数据源在磁盘，重查即可）
5. 新增 19 个测试覆盖所有路径

## Checklist


dacec354:
wait. web fetch has its own truncation in addition to tools truncation. It is a duplication that shall be solved

esengine:
Thanks for the patch — the design is solid and the test coverage is good. Two things I need before I can merge, plus one I'd like you to weigh in on:

**Please drop before merge**

1. `.reasonix/rfc-1615.md` — we don't keep design docs in the source tree (per `CLAUDE.md`: "Don't create new `*.md` documentation files unless explicitly asked"). The PR description already captures the gist; that's enough.
2. `cleanupOldResults` is implemented and tested but never invoked. Either wire it up (one call at registry construction is fine) or drop the function — as merged it's dead code and `.reasonix/truncated-results/` would grow forever.

**Worth a thought**

3. The skip-list (`get_env`, `everything_get-env`) only catches two specific tool names, but the same secret-leak risk exists for any MCP tool that returns credentials (`vault_*`, `aws_sts_*`, etc.). Consider defaulting MCP tools to skip and letting built-ins (`web_search` / `web_fetch`) opt in, instead of the current allow-everything-except-two-names model. Happy to take this in a follow-up if you'd rather keep this PR scoped.

Force-push 1 + 2 and I'll merge.

PR #1710, opened 2026-05-24
feat(ui): configurable cost display currency with click-to-toggle

### LINKED ISSUE #1708: feat(ui): configurable cost display currency with click-to-toggle
## Summary

Cost per turn always displays in CNY (¥), even for users who think in USD. All internal pricing is native USD — only the display defaults to CNY. This adds a config option and a clickable status bar to let users choose their preferred currency.

### Config field (persistent preference)

Add `"costCurrency": "USD"` to `~/.reasonix/config.json` for a persistent default. Works without a wallet balance API.

### Click-to-toggle (interactive)

Click anywhere on the bottom ~2 rows of the terminal to flip between USD and CNY instantly. A toast confirms the swap. The turn-cost pill uses `↻` instead of `▸` to hint it's clickable.

### Reactive

The toggle flows through AgentState, so the status bar and the per-turn UsageCard both re-render immediately. Wallet balance stays in its own actual currency unaffected.

### Priority chain

```
explicit currency arg → costDisplayCurrency (toggle)
→ balanceCurrency (wallet) → _defaultCostCurrency (config) → "CNY"
```

## Files touched

- `src/config.ts` — `costCurrency` config field
- `src/cli/ui/state/state.ts` — `costDisplayCurrency` in StatusBar
- `src/cli/ui/state/events.ts` — schema update
- `src/cli/ui/theme/tokens.ts` — `setDefaultCostCurrency()` + fallback
- `src/cli/ui/App.tsx` — seeds from config on mount
- `src/cli/ui/layout/StatusRow.tsx` — click handler + toast + `↻` glyph
- `src/cli/ui/cards/UsageCard.tsx` — reactive via `useAgentState`
- `tests/ui-usage-card-balance.test.tsx` — wrapped in AgentStoreProvider

## Verification

- `npm run verify` passes: 264 test files, 3624 tests, 0 failures
- comment-policy tests pass (9/9)
- All pre-existing tests unchanged

0xLeathery:
Closes #1708

## Summary

Cost per turn always displays in CNY (¥), even for users who think in USD. All internal pricing is native USD -- only the display defaults to CNY. This adds a config option and a clickable status bar to let users choose their preferred currency.

## Changes

- costCurrency config field for persistent preference
- Click-to-toggle on the status bar (bottom ~2 rows)
- Toast notification on toggle
- Reactive re-render in both StatusRow and UsageCard
- glyph hints the cost pill is interactive
- Wallet balance stays in its actual currency unaffected

## Verification

- npm run verify passes: 264 test files, 3624 tests, 0 failures
- comment-policy tests pass (9/9)
- No regressions

### DIFF
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -190,7 +190,7 @@ import { VerboseContext } from "./state/verbose-context.js";
 import { isLegacyWindowsConsole } from "./terminal-host.js";
 import { ThemeProvider } from "./theme/context.js";
 import { listThemeNames } from "./theme/tokens.js";
-import { FG, type ThemeName } from "./theme/tokens.js";
+import { FG, type ThemeName, setDefaultCostCurrency } from "./theme/tokens.js";
 import { TickerProvider } from "./ticker.js";
 import { handleTurnInterrupt } from "./turn-interrupt.js";
 import { useCompletionPickers } from "./useCompletionPickers.js";
@@ -413,6 +413,20 @@ export function App(props: AppProps): React.ReactElement {
       showVersion: cfg.showVersion !== false,
       showFeedbackHint: cfg.showFeedbackHint !== false,
     };
+  }, []);
+  // Seed the cost display currency from config on first mount. When the
+  // wallet API later provides a balanceCurrency, that takes precedence
+  // at the per-card level — this is the fallback for users without a
+  // wallet who want USD instead of the default CNY.
+  useEffect(() => {
+    const cfg = readConfig();
+    setDefaultCostCurrency(cfg.costCurrency);
+    if (cfg.costCurrency) {
+      agentStore.dispatch({
+        type: "session.update",
+        patch: { costDisplayCurrency: cfg.costCurrency },
+      });
+    }
   }, []);
   return (
     <ThemeProvider name={themeName}>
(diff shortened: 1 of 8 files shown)


esengine:
Thanks for the patch. The `costCurrency` config field is a clean addition and I'd like to take that part. The click-to-toggle UX I'd like to drop before merging:

1. **Click hit-area is too broad.** `if (ev.mouseRow < rows - 2) return` fires on *any* click in the bottom two rows — that's also where the composer footer, model pill, ctx pill, and balance pill live. Clicks to position the cursor or select text will flip the currency and pop a toast. The visual hint (`↻` on the cost pill) and the actual hit-area don't match.

2. **Redundant state path.** The module-level `_defaultCostCurrency` + `setDefaultCostCurrency` in `tokens.ts` duplicates what `status.costDisplayCurrency` in session state already does. The useEffect in `App.tsx` already dispatches the session update — the module global is unused once that lands. Drop the global, keep the session-state path.

3. **Glyph swap is global.** Changing `▸ → ↻` on the turn-cost pill ships to everyone, not just users who opted into the toggle. If the click feature goes, the glyph should revert too.

4. **No keyboard alternative.** Remote/SSH users without mouse support can't toggle at runtime — only by editing config. A `/currency` slash command or a keybinding would be a better primary surface than a hidden click zone.

If you can pare this down to just the `costCurrency` config field (plus the precedence chain `costDisplayCurrency ?? balanceCurrency ?? CNY` so a future toggle has somewhere to write), I'll merge. Happy to take a follow-up PR for an explicit `/currency` slash command if you want the runtime-toggle UX.

0xLeathery:
100% man! I'm glad you like it! I'll get hat pushed shortly! I don't mind what you want to bring in. I just have no concept of Yen so I wanted some form of something that made sense.

PR #1749, opened 2026-05-25
fix(desktop): verify bundled node TCC inheritance

T1anjiu:
## Summary

Ensure the bundled macOS Node runtime inherits Reasonix Desktop’s TCC grants, and add release-time checks so the shipped `.app` cannot silently lose the required entitlement.

## Problem

On macOS, Reasonix Desktop launches its internal Node runtime from the app bundle:

- `desktop/src-tauri/src/rpc.rs` resolves production CLI execution to `Contents/Resources/node`
- `desktop/scripts/bundle-node.mjs` signs that runtime with `node-helper.plist`
- `desktop/src-tauri/entitlements/node-helper.plist` contains `com.apple.security.inherit`

However, the packaging path did not reliably prove that the final bundled `node` actually retained `com.apple.security.inherit`. The previous script only checked whether `node` appeared signed, not whether it had the required TCC inheritance entitlement. It also treated macOS ad-hoc signing failures as non-fatal warnings, so a release could proceed with a Node helper that triggers separate TCC prompts instead of inheriting the main app’s permissions.

## Fix

- **`desktop/scripts/bundle-node.mjs`** — Changed macOS Node signing failures from non-fatal warnings to hard failures.
- **`desktop/scripts/bundle-node.mjs`** — Replaced the generic “is signed” check with an entitlement-specific `com.apple.security.inherit` check.
- **`desktop/scripts/bundle-node.mjs`** — Added a post-sign verification step to fail immediately if `codesign` does not apply TCC inheritance.
- **`.github/workflows/release.yml`** — Added macOS CI checks for `src-tauri/binaries/node` to confirm the bundled runtime has `com.apple.security.inherit`.
- **`.github/workflows/release.yml`** — Added a final `.app/Contents/Resources/node` verification step after Tauri packaging to ensure the shipped app’s internal Node helper still has valid signing and TCC inheritance.

## Verification

| Check | Result |
|---|---|
| `node --check desktop/scripts/bundle-node.mjs` | passes |
| `git diff --check` | passes |
| `npm run lint` | passes with existing unused-suppression warnings |
| `npm run verify` | attempted; blocked on Windows symlink privilege before final commit |
| macOS `codesign` / `lipo` checks | covered by release workflow on macOS runners |

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -104,6 +104,7 @@ jobs:
           echo "node archs: $archs"
           echo "$archs" | grep -q "arm64"
           echo "$archs" | grep -q "x86_64"
+          codesign -d --entitlements :- src-tauri/binaries/node 2>&1 | grep -q "com.apple.security.inherit"
 
       - name: Build & verify desktop frontend
         working-directory: desktop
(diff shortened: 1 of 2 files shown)


T1anjiu:
#1730

esengine:
Thanks for the careful diagnosis and the two-layer CI verification — exactly the right shape for catching this. Merged.

PR #1760, opened 2026-05-25
fix(desktop): type settings reducer test fixture

T1anjiu:
## Summary

Fixes a desktop build failure introduced by the settings slash command reducer test fixture.

The test fixture now explicitly uses the reducer state type so TypeScript preserves the expected `Settings` literal union types for `reasoningEffort` and `editMode`.

## Changes

- Annotate the desktop reducer test `state` fixture as `Parameters<typeof reduce>[0]`.
- Keep the existing reducer test behavior unchanged.

## Root Cause

`f00f238 fix(desktop): sync slash setting commands (#1724)` added a test fixture with an untyped object spread:

```ts
const state = {
  ...initialState(),
  settings: {
    reasoningEffort: "medium",
    editMode: "review",
    ...
  },
};
```

TypeScript widened nested literal values like `"medium"` and `"review"` to `string`, so `reduce(state, ...)` failed to typecheck against `State`.

## Notes

This is a test-only typing fix. It does not change runtime behavior.

### DIFF
--- a/desktop/src/App.test.ts
+++ b/desktop/src/App.test.ts
@@ -243,7 +243,7 @@ describe("Desktop App reducer — ApprovalPrompt integration", () => {
   });
 
   it("patches settings optimistically for desktop setting commands", () => {
-    const state = {
+    const state: Parameters<typeof reduce>[0] = {
       ...initialState(),
       settings: {
         reasoningEffort: "medium",


esengine:
Thanks for catching this @T1anjiu — the exact same one-line typing fix landed via #1759 a few minutes ago (bundled with the desktop startup-failure screen). Closing as superseded, but the catch was correct. Appreciate the contribution!

PR #1799, opened 2026-05-25
fix(loop): trim long-session retained payloads

### LINKED ISSUE #1790: 同一个会话，越用越卡
rt

**Expected**
大概250k上下文就卡的不行了，但是我的内存占用率并不高，其他应用也不卡，cmd的内存占用也不高。

**Reproduction**
长时间使用同一个会话可以复现

**Environment**
- Reasonix version (0.50.1):
- Node version (`node --version`):
- OS (Windows 11 ):
- Shell ( cmd, …):
- Terminal app (Windows Terminal）:
- DeepSeek model  （`deepseek-v4-pro`):

**Logs / transcript**

GTC2080:
## What

Prunes historical reasoning_content from plain assistant turns once a later user turn exists, while preserving reasoning_content on assistant tool-call turns.

Also shrinks retained oversized tool-call arguments after dispatch/load and elides old heavy tool arguments/results/reasoning in TUI and desktop session-restore snapshots.

## Why

Fixes #1790. Long v4-pro sessions can carry stale plain-turn reasoning and large tool/display payloads across follow-up requests and restored sessions. DeepSeek V4 accepts dropping reasoning_content from historical no-tool assistant turns, while tool-call reasoning_content is still kept for tool-call turns.

This intentionally keeps the existing ratio-based fold policy; the removed 250k local fold threshold is not part of this PR.

## How to verify

npm run verify

## Checklist

- [x] npm run verify passes locally (lint + typecheck + tests + comment-policy gate)

### DIFF
--- a/src/cli/commands/desktop.ts
+++ b/src/cli/commands/desktop.ts
@@ -531,7 +531,43 @@ function tailLines(s: string, n: number): string {
   return lines.slice(-n).join("\n");
 }
 
-function buildLoadedMessages(records: ChatMessage[]): LoadedMessage[] {
+const LOADED_RECENT_MESSAGE_WINDOW = 200;
+const LOADED_MIN_ELIDE_CHARS = 4096;
+const LOADED_ELIDED_PREFIX = "[elided — older than the last ";
+
+function elideLoadedField(value: string): string {
+  if (value.length <= LOADED_MIN_ELIDE_CHARS) return value;
+  if (value.startsWith(LOADED_ELIDED_PREFIX)) return value;
+  return `${LOADED_ELIDED_PREFIX}${LOADED_RECENT_MESSAGE_WINDOW} messages; ${value.length.toLocaleString()} chars dropped to save memory. Full content is on disk in the session log.]`;
+}
+
+function elideLoadedMessages(messages: LoadedMessage[]): LoadedMessage[] {
+  if (messages.length < LOADED_RECENT_MESSAGE_WINDOW) return messages;
+  const cutoff = messages.length - LOADED_RECENT_MESSAGE_WINDOW;
+  return messages.map((msg, i) => {
+    if (i >= cutoff || msg.kind !== "assistant") return msg;
+    return {
+      ...msg,
+      segments: msg.segments.map((segment) => {
+        switch (segment.kind) {
+          case "reasoning":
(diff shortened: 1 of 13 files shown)


esengine:
Verified the `reasoning_content` strip against the real v4-pro API across five scenarios (thinking on/off, plain × N, mixed with tool-call turns) — all 200. The `messages.ts:4` "MUST be set or 400s" note is R1-era; v4 tolerates dropping it on historical plain turns. Happy to take that part.

Please drop the `LOCAL_HEAP_PROMPT_FOLD_THRESHOLD = 250_000` half and resubmit `stripDroppableReasoningContent` on its own. Auto-folding 1M-window sessions at 25% fights the cache-first design — the strip already removes the largest per-turn payload, so I'd rather see whether that alone relieves the OOM before adding an absolute threshold on top of the ratio tiers. The remaining heap growth we'll dig into separately (likely log-side, not request-side).

esengine:
Thanks for the extra work — the tool-call args shrink on persist and the card-elision extraction (so loaded sessions get the same 200-card stub treatment the live reducer already does) are both solid net wins, happy to take those.

One thing still pending though: `LOCAL_HEAP_PROMPT_FOLD_THRESHOLD = 250_000` is still in `context-manager.ts`. Could you pull that part out of this PR? Auto-folding 1M-window v4-pro sessions at 25% of the window breaks the prefix cache much earlier than the ratio tiers were designed for, and cache-first is the load-bearing positioning for the project. With the reasoning_content strip + tool-call args trim + card elision all landing, the largest per-turn payloads are already gone; I'd like to see whether OOM still reproduces on that baseline before adding an absolute-token fold trigger on top.

The remaining heap growth we'll dig into separately — likely something log-side / serialization-side we haven't measured yet, not a context-window issue. Once the 250k constant and its call sites are out, this is ready to merge.

GTC2080:
Updated again to match the latest note: restored the tool-call args trim and card/session elision changes, and removed the 250k local fold guard plus its call sites/tests.

Current branch keeps the existing ratio-based fold policy. Verified with npm run verify.

esengine:
Thanks for the quick turnaround on the split — clean shape, merging. The remaining heap question we'll track separately under #1790 once we've measured what's left after this lands.

PR #1812, opened 2026-05-25
perf(desktop): direct DOM CSS vars during sidebar drag, disable CSS transition, content-visibility for messages

CVEngineer66:
## Summary

Optimize sidebar drag resize performance. Long sessions (200k+ tokens) were
causing CPU temperature spikes and visible lag during drag because every
mousemove triggered a full React re-render + CSS transition restart.

### 1. Direct DOM CSS variable updates (`useResizable.ts`)
- Cache `.app` ref on mousedown, avoid `querySelector` per frame during drag
- Both DOM `setProperty()` + React `setWidth()` on every mousemove
- Previously only React `setWidth()` was called, causing full tree re-renders

### 2. Disable CSS transition during drag (`styles.css`)
- `.app` has `transition: grid-template-columns 180ms ease` for smooth collapse/expand
- During drag, `data-dragging` attribute on `.app` sets `transition: none`
- Prevents the animation engine from restarting 60-120 times per second

### 3. Skip offscreen message layout (`styles.css`)
- `.thread-inner > div[data-turn]` gets `content-visibility: auto`
- Browser only lays out visible messages during resize, not the full DOM tree

### 4. Minor fixes
- `applyIncoming` no longer exported (fixes Vite HMR warning about non-component exports)
- `--composer-max-width` synced via `useEffect` to keep composer in sync during drag

## Remaining issues / known limitations

On very long sessions (500+ messages, 250k+ tokens), dragging the sidebar
still shows some lag and CPU temperature increase. Root cause:

- CSS Grid with `grid-template-columns: var(--side-width) minmax(0, 1fr) var(--ctx-width)`
  forces the browser to recalculate layout for **all** grid children when
  `--side-width` changes, even with `content-visibility: auto`
- The grid container itself must fully reflow to determine scroll extents
- `content-visibility: auto` reduces per-message cost but the grid reflow
  overhead scales with DOM size

Possible next steps (not in this PR):
- Replace CSS Grid with `position: absolute` panels and `transform: translateX()`
  for resize — avoids full grid reflow
- Virtualize the message list (only render visible messages)
- Add `will-change: grid-template-columns` hint on `.app`

### DIFF
--- a/desktop/src/App.tsx
+++ b/desktop/src/App.tsx
@@ -699,7 +699,7 @@ function appendTextSegment(
   return [...segments, { kind, text }];
 }
 
-function applyIncoming(state: State, ev: IncomingEvent): State {
+export function applyIncoming(state: State, ev: IncomingEvent): State {
   switch (ev.type) {
     case "user.message": {
       return {
--- a/desktop/src/App.tsx
+++ b/desktop/src/App.tsx
@@ -699,7 +699,7 @@ function appendTextSegment(
   return [...segments, { kind, text }];
 }
 
-export function applyIncoming(state: State, ev: IncomingEvent): State {
+function applyIncoming(state: State, ev: IncomingEvent): State {
   switch (ev.type) {
     case "user.message": {
       return {
@@ -3262,6 +3262,13 @@ export function App() {
     localStorage.setItem("reasonix.theme", theme);
     localStorage.setItem("reasonix.themeStyle", themeStyle);
   }, [theme, themeStyle]);
+
+  // Sync --composer-max-width to .app (separate from inline style to avoid React override)
+  const composerRef = useRef<HTMLElement | null>(null);
+  useEffect(() => {
+    if (!composerRef.current) composerRef.current = document.querySelector(".app");
+    composerRef.current?.style.setProperty("--composer-max-width", `${threadMaxWidth}px`);
+  }, [threadMaxWidth]);
 
   useEffect(() => {
     let raf = 0;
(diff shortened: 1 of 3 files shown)


esengine:
Core perf changes look right — direct DOM writes, `data-dragging` transition disable, and `content-visibility: auto` are all the standard moves. CI green, mergeable.

One blocker before I merge: the resize-handle `width: 5px → 1px` change isn't perf-related and a 1px hit target is basically un-grabbable. Please revert that piece (and the `left: calc(... - 2px)` → `left: var(--side-width)` follow-on, since they go together).

Two smaller things I'll clean up in a follow-up if you don't:

- `composerRef` in `App.tsx` is misleading — it queries `.app`, not the composer, and it duplicates the `--composer-max-width` write already happening inside `useResizable.onMove`. Folding that variable into the inline style block at App.tsx:2118 (next to `--thread-max-width`) drops both the ref and the effect.
- `setWidth(next)` still fires per mousemove, so React re-renders at the same cadence as before — the direct DOM write just beats React's commit by a frame. The PR description frames it as "previously only `setWidth()` was called, causing full tree re-renders" which reads as if re-renders were eliminated. Worth tweaking so future readers don't get confused.

Revert the handle width and I'll merge.

CVEngineer66:
Reverted the resize handle width back to 5px and positioning back to calc(var(--side-width) - 2px). The other two spots (composerRef cleanup, PR description wording) are fine for you to handle in a follow-up — thanks!

PR #1813, opened 2026-05-25
feat(desktop): highlight active message bar in JumpBar, follow latest message

CVEngineer66:
## Summary

JumpBar now tracks an "active" message — the latest user turn is highlighted
by default, and clicking any bar sets it as active. Active bars remain
visually distinct (18px + accent color) to show where you are in the
conversation.

## Changes

- **`desktop/src/ui/jump-bar.tsx`** (+36 / -26):
  - New `active` state: defaults to the latest user message, updates as new
    messages arrive
  - Clicking a bar sets it as active and scrolls the conversation to that turn
  - Active bar gets a persistent 18px accent highlight (hover magnification
    still takes priority when your cursor is over the bar)
  - `scrollIntoView({ block: "nearest" })` keeps the active bar visible within
    the scrollable list without forcing it to center
  - Removed redundant `data-turn` prop was missing on `.jump-item` — now set

### DIFF
--- a/desktop/src/ui/jump-bar.tsx
+++ b/desktop/src/ui/jump-bar.tsx
@@ -1,4 +1,4 @@
-import { useMemo, useRef, useState } from "react";
+import { useEffect, useMemo, useRef, useState } from "react";
 
 interface JumpBarProps {
   messages: { kind: string; text?: string; turn?: number }[];
@@ -7,6 +7,7 @@ interface JumpBarProps {
 
 export function JumpBar({ messages, threadEl }: JumpBarProps) {
   const [hovered, setHovered] = useState<number | null>(null);
+  const [active, setActive] = useState<number | null>(null);
   const barRef = useRef<HTMLDivElement>(null);
   const previewTop = useRef(0);
   const [showPreview, setShowPreview] = useState(false);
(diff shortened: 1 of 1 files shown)


esengine:
This PR stacks the drag-perf changes from #1812 (useResizable.ts, styles.css, App.tsx composer effect) on top of the JumpBar feature, so it inherits the same blocker I flagged on #1812 — the 1px resize-handle width is a UX regression and needs to revert to ~5px.

Cleanest path: wait for #1812 to land with the handle-width revert, then rebase this branch onto main so only the JumpBar commit remains.

On the JumpBar feature itself:

- The "click bar → set active" behavior reads as a pin, but `useEffect(() => setActive(items[items.length - 1].turn), [items])` force-overrides `active` to the latest turn every time a new message arrives. So a manual click only persists until the next message lands. If that's intentional (the title says "follow latest message"), the PR description should match — currently it reads as if click-to-pin is sticky.
- Inline width/bg ternary in the map (`const w = isHov ? 32 : isNear ? (dist === 1 ? 20 : 14) : isActive ? 18 : 12`) is a bit dense for what was previously a `barProps()` helper. Not a blocker, but the helper version was easier to read.

Test addition (`memoryDetail: null`) is a good fix — unrelated to JumpBar but appreciated.

I'll come back to this once #1812 is sorted.

CVEngineer66:
Rebased onto latest upstream/main now that #1812 has landed. The `active` state is designed to track the latest message by default (for the "follow the conversation" feel), but clicking a bar sets it as a temporary focus point. That focus resets when new messages arrive, which is intentional — the PR title says "follow latest message". The bar width/bg ternary collapsed from the earlier `barProps()` helper to reduce noise; I can extract it back if the density is a readability issue.

PR #1827, opened 2026-05-26
fix(tui): reduce prompt input repaint flicker

### LINKED ISSUE #1821: 终端界面输入任何内容都会产生闪烁（0.51.0依旧存在）
https://github.com/user-attachments/assets/5d34629e-e239-4526-9af0-021ea0448fc2

GTC2080:
## Summary
- Enable Ink incremental rendering for the chat TUI.
- Add a regression test that prompt edits do not emit full-frame erase sequences.

## Root cause
Prompt input edits used Ink default rendering, which clears and redraws the whole frame on every typed character or backspace. This made the terminal visibly flicker across terminals.

## Validation
- Interactive prompt simulation: type h, type i, backspace, type !; every step reported fullFrameErase=false.
- npx vitest run tests/prompt-input-cursor-sync.test.tsx tests/terminal-host.test.ts --reporter verbose
- npx biome check src/cli/commands/chat.tsx tests/prompt-input-cursor-sync.test.tsx
- npm run typecheck
- git diff --check
- pre-push npm run verify: 280 test files passed, 3744 tests passed, 10 skipped

Fixes #1821

### DIFF
--- a/src/cli/commands/chat.tsx
+++ b/src/cli/commands/chat.tsx
@@ -45,6 +45,11 @@ import {
 
 export type { McpLifecycleNotice, McpLifecycleSink, McpRuntime, ProgressInfo };
 
+export const CHAT_RENDER_OPTIONS = {
+  exitOnCtrlC: true,
+  incrementalRendering: true,
+} as const;
+
 export interface ChatOptions {
   model: string;
   reasoningEffort?: ReasoningEffort;
@@ -407,7 +412,7 @@ export async function chatCommand(opts: ChatOptions): Promise<void> {
       qqSubmitRef={qqSubmitRef}
       qqErrorRef={qqErrorRef}
     />,
-    { exitOnCtrlC: true },
+    CHAT_RENDER_OPTIONS,
   );
   try {
     await waitUntilExit();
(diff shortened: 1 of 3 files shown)


esengine:
Thanks @GTC2080 — the `incrementalRendering: true` flag landed via #1847 a few hours after you opened this, which is why it now conflicts. The regression test in `prompt-input-cursor-sync.test.tsx` is still worth keeping; could you rebase and submit just that test in a follow-up? You can inline `incrementalRendering: true` in the test setup instead of exporting `CHAT_RENDER_OPTIONS` from `chat.tsx`. The `jobs.test.ts` timing fix is unrelated to this issue — please split it into its own PR.

Closing this one for now.

GTC2080:
Thanks @esengine — split this as requested.

- #1873 keeps only the PromptInput incremental repaint regression test, with `incrementalRendering: true` inlined in the test render setup and no `chat.tsx` export/runtime changes.
- #1872 carries the unrelated `jobs.test.ts` timing fix separately.

Both branches were rebased onto the post-#1847 renderer shape and passed local `npm run verify` before opening.

PR #1877, opened 2026-05-26
fix(tui): suppress submit on IME atomic commit (#1853)

### LINKED ISSUE #1853: Enter Key Conflict with WeChat IME (English Input Mode)
Dear Reasonix Team,

Thank you for your great work on Reasonix! I’ve encountered an issue when using Reasonix alongside the WeChat IME in English input mode. When I press Enter to confirm the English input within the IME, Reasonix simultaneously treats Enter as submission, which leads to premature message sending.

I believe the core issue is that Reasonix is capturing the raw Enter key event rather than recognizing whether the IME has finalized input. I’d like to request behavior similar to OpenCode, where the tool intelligently distinguishes whether Enter is finalizing IME input or actually submitting the command. In other words, please allow Enter to confirm input inside the IME without triggering an early submission, and only treat Enter as submission after the IME has finalized the text.

This would greatly improve the experience. Thank you for considering this enhancement!

esengine:
## Summary

Closes #1853.

WeChat IME in English mode (and other IMEs that commit ASCII candidates) finalize a candidate by writing the chosen text plus the trigger Enter as one terminal write. The TUI's existing `IME_GUARD_MS` only arms when the preceding input was non-ASCII, so the bare Enter slipped straight through `processMultilineKey` and triggered submit while the user was still composing.

## Fix

Detect the burst at the stdin layer instead of trying to distinguish it after the fact:

- `KeyEvent` gains an optional `compositionCommit: boolean`.
- `StdinReader.dispatch` flags any bare Enter (no shift/ctrl/meta) whose immediately-prior dispatch in the same chunk was bare printable input. `prevInputInChunk` is reset at the top of `handleChunk`, so chunks dispatched independently don't carry state across.
- `PromptInput` drops `action.submit` when the flag is set.

The signal is unambiguous: a real human can't type characters and press Enter inside one stdin chunk — keystrokes are tens of ms apart, terminals flush each keystroke before the next arrives. Same-chunk `text\r` is therefore an IME (or paste) atomic flush. The existing 50ms non-ASCII timer in `PromptInput` is kept as belt-and-suspenders for IMEs that split the commit and the Enter across two chunks within a few ms.

## Tests

`tests/stdin-reader.test.ts` — updated the "typed-then-Enter" assertion (`abc\r` now flags the Enter), added five new cases:

- bare Enter in its own chunk → no flag
- text + Enter in separate chunks → no flag (real human cadence)
- Shift+Enter in same chunk → not flagged (intentional newline-insert)
- flag resets between chunks
- Enter after a non-input event (arrow) is not flagged

71 stdin-reader tests pass, plus 9 comment-policy + 104 surrounding input tests (multiline-keys, key-normalize, paste-collapse).

## Checklist

- [x] `npm run verify` passes locally
- [x] No `Co-Authored-By: Claude` trailer
- [x] No CHANGELOG edits

### DIFF
--- a/src/cli/ui/PromptInput.tsx
+++ b/src/cli/ui/PromptInput.tsx
@@ -22,9 +22,11 @@ import { FG, SURFACE, TONE } from "./theme/tokens.js";
 /** Pastes shorter than this AND single-line render verbatim; longer ones become a `[paste #N · …]` sentinel chip (#397). */
 export const INLINE_PASTE_THRESHOLD = 200;
 
-// Tight enough that a normal typed Enter (≥80ms after the last keystroke
-// for human cadence) still submits; wide enough that CJK IME commit-then-
-// Enter (terminal flushes both together) falls inside the window.
+// Belt-and-suspenders for IMEs that flush the commit-text and the Enter
+// across two stdin chunks within a few ms — same-chunk IME commits are
+// caught earlier via the stdin reader's `compositionCommit` flag, but
+// some terminal/IME pairs split the flush. Tight enough that normal
+// type-then-Enter (≥80ms human cadence) still submits.
 const IME_GUARD_MS = 50;
 
 function hasNonAscii(s: string): boolean {
@@ -172,6 +174,7 @@ export function PromptInput({
       setCursor(action.cursor);
     }
     if (action.submit) {
+      if (ev.compositionCommit) return;
       if (Date.now() - lastNonAsciiInputAtRef.current < IME_GUARD_MS) {
         lastNonAsciiInputAtRef.current = 0;
         return;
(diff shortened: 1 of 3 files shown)


esengine:
Superseded by #1874 (merged as c96a5247). Same root cause (#1853 — WeChat IME English mode flushes ASCII text + Enter atomically, slipping past the non-ASCII-only IME_GUARD_MS), different fix shape — #1874 extends the existing timer's arm condition with an `input.length > 1` heuristic, mine added a chunk-level `compositionCommit` flag at the stdin reader. #1874's heuristic ends up covering more cases (incl. the split-chunk variant my version missed) for far less surface area, so closing this as dup.

If anyone hits a single-character ASCII IME commit slipping through later, the chunk-level signal here is the natural follow-up — happy to revive then.

PR #1911, opened 2026-05-26
Add Codex collaboration tools to Reasonix Code

enderzcx:
## Summary

- Register two local Codex CLI-backed tools in Reasonix Code:
  - `codex_review` for read-only native Codex review of uncommitted, branch, or commit scopes.
  - `codex_delegate` for focused read-only second opinions, with confirmation required before workspace-write delegation.
- Add a dedicated runner wrapper with timeout, abort, output truncation, process-tree cleanup, and injectable runner support for tests.
- Cover invocation construction, injected runner dispatch, write-denial behavior, and plan-mode gating.

## Verification

- `npx @biomejs/biome check src/tools/codex.ts src/code/setup.ts tests/codex-tools.test.ts` ✅
- `npm test -- --run tests/codex-tools.test.ts` ✅
- `npm run typecheck` ✅
- `npm run lint` ✅
- `git diff --cached --check` ✅
- `npm test -- --run tests/prompt-budget.test.ts` ⚠️ local-existing failure: `expected 26512 to be less than or equal to 26500`, with warnings from local user skills under `~/.agents` / `~/.claude`. This PR does not modify `src/code/prompt.ts`; I removed the prompt wording from the dirty experiment specifically to avoid increasing that budget.

### DIFF
--- a/src/code/setup.ts
+++ b/src/code/setup.ts
@@ -16,6 +16,7 @@ import { bootstrapSemanticSearchInCodeMode } from "../index/semantic/tool.js";
 import { ToolRegistry } from "../tools.js";
 import { registerChoiceTool } from "../tools/choice.js";
 import { registerCodeQueryTools } from "../tools/code-query.js";
+import { registerCodexTools } from "../tools/codex.js";
 import { registerFilesystemTools } from "../tools/filesystem.js";
 import { registerJavaSourceTool } from "../tools/java-source.js";
 import { JobRegistry } from "../tools/jobs.js";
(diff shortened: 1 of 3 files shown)


esengine:
Thanks for the careful work on this — the runner wrapper, the pauseGate write-confirmation path, and the test coverage are all solid.

I'm going to pass on this one though. Reasonix is intentionally a DeepSeek-first, cache-first agent, and shipping built-in tools that hand work off to another vendor's LLM (Codex / OpenAI behind it) cuts against that positioning in two ways:

1. The whole project is built on "cheap, cache-aligned, single-model loop." A first-class `codex_delegate` tool lets the model spend tokens on a paid external LLM from inside the loop, which is the exact cost shape we're trying to avoid.
2. The model deciding on its own to spawn an external agent is a pattern we've seen lose badly in benchmarks vs. a flat loop — we don't want the planner reaching for "ask Codex" as an escape hatch.

This is a great fit for an external MCP server though — write it as a standalone MCP and users who do want Codex-in-the-loop can attach it via the MCP settings without it being baked into the core toolset. That keeps the core lean and lets the feature evolve at its own pace.

Closing this PR, but please keep contributing — the desktop polish work has been great.

PR #1913, opened 2026-05-26
fix(desktop): normalize path separators for Windows file opening

### LINKED ISSUE #1869: Windows环境下，文件打开错误
**What happened**
默认在文件路径中使用/，windows下没有做转换，导致打开错误

Image

**Expected**
可以点击后打开文件

**Environment**
- Reasonix version (`0.51`):
- OS (Windows 11)

Bernardxu123:
## Description

Windows users encounter file opening failures when clicking Markdown file references or context panel file entries. The root cause is that file paths use forward slashes (`/`) from AI output, which are passed directly to `cmd.exe` via the Tauri `open_in_editor` command — `cmd.exe` does not handle forward slashes reliably on Windows.

### `desktop/src/Markdown.tsx` — `resolveAgainstWorkspace`
- Added Windows path detection via `ws.includes("\\")`
- Normalize forward slashes to backslashes in both relative and absolute path branches
- Previously, absolute paths (e.g. `C:/Users/.../file.tsx`) were returned unmodified

### `desktop/src/ui/context-panel.tsx` — `openContextFile`
- Same fix applied: detect Windows via workspace path, normalize `/` → `\\`
- Both relative-path construction and absolute-path early-return branches covered

### `desktop/src-tauri/src/main.rs` — `open_in_editor` (Windows branch)
- Added `path.replace('/', "\\")` as a safety net before passing to `cmd.exe`
- Only applies inside `#[cfg(windows)]`, zero impact on macOS/Linux

## Impact

- **Windows**: files now open correctly in configured editor (VS Code, Cursor, etc.) and via OS default opener
- **macOS/Linux**: not affected — the normalization condition `ws.includes("\\")` never triggers
- The Rust side is guarded by `#[cfg(windows)]` compile-time, so the code doesn't even exist on non-Windows builds

## Related

Closes #1869

### DIFF
--- a/desktop/src-tauri/src/main.rs
+++ b/desktop/src-tauri/src/main.rs
@@ -183,12 +183,14 @@ fn open_in_editor(command: String, path: String, line: Option<u32>) -> Result<()
     #[cfg(windows)]
     {
         // Spawn through cmd.exe so `.cmd` shims (code.cmd, cursor.cmd) resolve via PATH.
+        // Normalize forward slashes to backslashes — cmd.exe doesn't handle them reliably.
+        let normalized = path.replace('/', "\\");
         cmd = Command::new("cmd");
         cmd.arg("/c").arg(trimmed);
         if let Some(l) = line {
-            cmd.arg("-g").arg(format!("{}:{}", path, l));
+            cmd.arg("-g").arg(format!("{}:{}", normalized, l));
         } else {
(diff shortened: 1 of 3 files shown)


Bernardxu123:
CI failed due to HTTP 403 checkout error (GitHub infra). Recreating to retry.

PR #1929, opened 2026-05-26
feat(desktop): collapse intermediate reasoning, group consecutive tool calls

CVEngineer66:
## Summary

Reduce visual noise in long assistant turns by hiding intermediate reasoning segments between tool calls, and grouping consecutive tool calls into a single collapsible section.

### 1. Hide intermediate reasoning (`thread.tsx`)
- Only the FIRST reasoning segment per assistant turn is rendered
- Subsequent reasoning segments (model thinking between tool calls) are hidden
- User sees one coherent reasoning block per turn instead of repetitive cards

### 2. Group consecutive tool calls (`thread.tsx`, `styles.css`)
- Multiple sequential tool segments (e.g. shell → shell → edit_file) are wrapped in a `.tool-group` collapsible container
- Header shows count: "3 tool calls" / "3 个工具调用"
- Default collapsed; click to expand
- Single tool calls render without wrapper (no visual change)

### 3. Default collapse for completed shell cards (`cards.tsx`)
- `ShellCard` now defaults to collapsed (`defaultOpen={false}`) instead of remaining open until done
- Reasoning cards still auto-expand during thinking and collapse on completion
- ToolCard and DiffCard were already collapsed by default

### 4. i18n
- Added `thread.toolCalls` and `thread.oneToolCall` keys (EN + zh-CN)

### DIFF
--- a/desktop/src/i18n/en.ts
+++ b/desktop/src/i18n/en.ts
@@ -692,6 +692,8 @@ export const en = {
     unknownKind: "# unknown card kind — fallback render",
   },
   thread: {
+    toolCalls: "{count} tool calls",
+    oneToolCall: "1 tool call",
     keepSteps: "Keep {n} remaining step(s)",
     keepOriginal: "Keep original",
     you: "You",
--- a/desktop/src/i18n/zh-CN.ts
+++ b/desktop/src/i18n/zh-CN.ts
@@ -677,6 +677,8 @@ export const zhCN: typeof en = {
     unknownKind: "# 未知卡片类型 — fallback 渲染",
   },
   thread: {
+    toolCalls: "{count} 个工具调用",
+    oneToolCall: "1 个工具调用",
     keepSteps: "保留剩余 {n} 步",
     keepOriginal: "保留原计划",
     you: "你",
--- a/desktop/src/styles.css
+++ b/desktop/src/styles.css
@@ -2033,6 +2033,40 @@ html:not([data-platform="macos"]) .shortcut kbd[data-key="mod"] {
   max-width: 60ch;
 }
 
+.tool-group {
+  border: 1px solid var(--border);
+  border-radius: var(--radius);
+  overflow: hidden;
+}
+.tool-group-header {
+  display: flex;
+  align-items: center;
+  gap: 6px;
+  width: 100%;
+  padding: 6px 10px;
+  font-size: 12px;
+  font-weight: 600;
+  color: var(--fg-2);
+  background: var(--panel);
+  cursor: pointer;
+  border: none;
+}
+.tool-group-header:hover {
+  background: var(--panel-2);
+}
+.tool-group-header .chevron {
+  margin-left: auto;
+  transition: transform 150ms;
+}
+.tool-group-header .chevron.open {
+  transform: rotate(180deg);
+}
+.tool-group-body {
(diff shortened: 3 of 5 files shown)


esengine (changes requested):
Thanks for the careful work — the "group consecutive tool calls" idea is good and I'd like to land that part. Three things blocking merge as-is though:

**1. React hook anti-pattern in the group shell (real bug)**

`GroupShell` is declared inside `flushToolGroup`, which runs on every render of `AssistantMsg`. That means React sees a brand-new component type on every render, unmounts the old subtree, and the `useState(false)` resets. So during streaming — every time a text delta lands and re-renders the parent — any group the user just expanded snaps closed again.

Lift `GroupShell` out of the function body to module scope so the component identity stays stable across renders:

```tsx
function ToolGroupShell({ segs, renderTool }: {
  segs: (AssistantSegment & { kind: "tool" })[];
  renderTool: (s: AssistantSegment & { kind: "tool" }, idx: number, expanded?: boolean) => ReactNode;
}) {
  const [open, setOpen] = useState(false);
  return ( ... );
}
```

CI didn't catch this because there's no interaction test for the collapse.

**2. Running shell streams are now hidden (UX regression)**

`ShellCard` previously did `defaultOpen={state !== "done"}` so users could watch `npm test` / `cargo build` output stream live; it auto-collapsed only after completion. The new `defaultOpen={_defaultOpen ?? false}` drops that, and the call site in `thread.tsx` doesn't pass `defaultOpen=true` for running shells.

Please preserve the "open while running/await, collapse on done" behavior — either keep the old default and only override it from inside a tool group, or pass `defaultOpen={state !== "done"}` explicitly at the call site for ungrouped shells.

**3. Don't hard-drop intermediate reasoning**

Hiding only the first reasoning per turn is a step too far for this project — the reasoning surface is core to what makes Reasonix useful (it's literally the name). Users rely on the inter-tool-call reasoning to spot when the agent is going off the rails mid-turn.

Could you instead render subsequent reasoning segments inside a collapsed-by-default `ReasoningCard` (similar to how `ToolCard` collapses) rather than dropping them from the DOM? That keeps the "less noise" win without losing the diagnostic signal.

Minor: `defaultOpen: _defaultOpen` — the `_` prefix conventionally means unused, but it's used here. Drop the prefix.

Happy to merge once 1 + 2 are fixed and 3 is converted to collapsed-by-default rather than dropped.

CVEngineer66:
Thanks for the review. Fixes applied:

**1 & 2 fixed:**
- `ToolGroupShell` lifted to module scope
- Standalone shell calls now use `defaultOpen={state !== "done"}` (open while running, collapse on completion)
- `_defaultOpen` → `defaultOpen`

**3. Intermediate reasoning:**
We prefer to keep the current approach (drop, not collapsed). The inter-tool-call reasoning from DeepSeek is typically brief operational statements ("let me read the file", "analyzing the result") that provide minimal diagnostic value for a developer user. What users actually care about are the tool call **results** — not how many steps the model thought about between calls. These intermediate reasoning blocks take up significant visual space without delivering substantive information. If users need to inspect the full reasoning chain later, they can use `/compact` or the session log. The current change preserves only the **first reasoning** of each assistant turn (which contains the complete analysis), and drops the inter-tool reasoning. This reduces visual noise while keeping the core diagnostic signal.

PR #1930, opened 2026-05-26
feat(desktop): 余额显示支持多币种（同时展示人民币和美元）

weikengchen:
## 背景

DeepSeek 的 `/user/balance` API 返回的 `balance_infos` 是一个数组，包含用户账户下每种币种的余额信息（例如 CNY 和 USD）。但桌面端目前只调用 `pickPrimaryBalance()` 挑出金额最大的那个显示，另一个币种被丢弃。

## 改动

用户在状态栏的余额段可以看到所有币种的余额，格式如 `余额  $ 9.87 / ¥ 72.00`。只有一个币种时，显示保持原样。

涉及四层：

| 层 | 文件 | 改动 |
|---|---|---|
| 协议 | `desktop/src/protocol.ts` | 新增 `BalanceInfoItem` 类型，`BalanceEvent` 增加 `balanceInfos` 数组 |
| Node 后端 | `src/cli/commands/desktop.ts` | `emitBalance()` 将 API 返回的全部 `balance_infos` 传递给前端 |
| 前端状态 | `desktop/src/App.tsx` | `Balance` 类型增加 `infos` 字段，reducer 接收新字段 |
| UI 渲染 | `desktop/src/ui/statusbar.tsx` | 余额段改为遍历 `balance.infos`，多币种以 ` / ` 拼接 |

## 兼容性

向后兼容——`balanceInfos` 为可选字段，前端使用 `?? []` 兜底。单币种用户和旧版本 Node 子进程均不受影响。

## 截图

image

### DIFF
--- a/desktop/src/App.tsx
+++ b/desktop/src/App.tsx
@@ -258,10 +258,18 @@ export type Settings = {
   version: string;
 };
 
+export type BalanceInfoItem = {
+  currency: string;
+  total: number;
+  granted?: number;
+  toppedUp?: number;
+};
+
 export type Balance = {
   currency: string;
   total: number;
   isAvailable: boolean;
+  infos: BalanceInfoItem[];
 };
 
 type MentionResults = { nonce: number; query: string; results: string[] };
@@ -920,6 +928,7 @@ export function applyIncoming(state: State, ev: IncomingEvent): State {
           currency: ev.currency,
           total: ev.total,
           isAvailable: ev.isAvailable,
+          infos: ev.balanceInfos ?? [],
         },
       };
     case "$qq_settings":
--- a/desktop/src/protocol.ts
+++ b/desktop/src/protocol.ts
@@ -337,11 +337,19 @@ export type QQSettingsEvent = {
   access: string;
 };
 
+export type BalanceInfoItem = {
+  currency: string;
+  total: number;
+  granted?: number;
+  toppedUp?: number;
+};
+
 export type BalanceEvent = {
   type: "$balance";
   currency: string;
   total: number;
   isAvailable: boolean;
+  balanceInfos: BalanceInfoItem[];
 };
 
 export type SettingsPatch = {
--- a/desktop/src/ui/statusbar.tsx
+++ b/desktop/src/ui/statusbar.tsx
@@ -137,10 +137,18 @@ export function StatusBar({
         <span className="v vio">{settings?.model ?? "—"}</span>
         <span className="v">{settings?.reasoningEffort ?? "high"}</span>
       </span>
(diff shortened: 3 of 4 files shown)


CVEngineer66:
这样太占下面位置了，建议改成点击一下变更币种

weikengchen:
> 这样太占下面位置了，建议改成点击一下变更币种

也可以... 主要是这个逻辑和另外一个还不同，因为 deepseek 的多币种余额是分开的，两个不同的余额，优先用人民币，而目前我们的 tokens 计费是一个价格两种显示方式。

esengine (changes requested):
Thanks — the four-layer plumbing (`protocol` → `emitBalance` → reducer → `statusbar`) is clean and the format matches the existing single-currency rendering. One small thing blocks merge though.

**The balance segment lost its click affordance.** The old code wired `onClick={onToggleCurrency}` on the balance `<span>`, and that toggle is what drives the **cost** segment between CNY/USD via `formatMoney(usage.totalCostUsd, currency)` (statusbar.tsx:62). The keyboard shortcut still works (App.tsx:2034), but the mouse path is gone — users won't discover the toggle, and the cost figure appears stuck in whatever currency happens to be selected.

Please put the `onClick` back on the balance segment and give it a tooltip that describes what it actually does now (toggles the cost figure, since the balance itself shows both currencies):

```tsx
<span
  className="seg"
  title={
    balance && balance.infos.length > 0
      ? balance.infos.map((i) => `${i.currency} ${i.total.toFixed(2)}`).join(" / ")
      : t("statusbar.switchCurrency")
  }
  onClick={onToggleCurrency}
>
```

Or, if you'd rather decouple the toggle entirely from the balance segment, move `onClick={onToggleCurrency}` onto the cost segment instead — but please don't leave the prop wired with no UI affordance.

Once that's restored, happy to merge. The hard-coded `"$" : "¥"` symbols can stay for now (DeepSeek only emits those two currencies); we can factor it later if more show up.

weikengchen:
> ```tsx
> onToggleCurrency
> ```

There is a question related to the current "onToggleCurrency". When I tried that on the official 0.52.0, when I click onToggleCurrency on the BALANCE part, the currency on BALANCE doesn't change---it changes the currency on COST (the left one). 0.52.0 原版，在余额点击后，余额的单位不变，反而是让左边的“本次”的货币单位变化了。

Is this the desired behavior? If so I will put that back

esengine (approved):
Pushed a fixup that restores onClick={onToggleCurrency} + the switchCurrency tooltip on the balance segment so the cost toggle stays clickable. Multi-currency rendering itself looks good — thanks!

PR #1941, opened 2026-05-26
feat(qq): default to markdown replies

wade19990814-hue:
## Summary
- send QQ private replies through the Markdown delivery path by default
- fall back to plain text automatically if Markdown delivery fails
- unwrap full ```markdown``` / ```md``` fenced replies before sending so rendered content is not trapped inside a code block

## Testing
- npm test -- --run tests/qq-access.test.ts tests/qq-config.test.ts tests/qq-channel.test.ts tests/qq-slash.test.ts tests/qq-first-connect.test.tsx
- npm run typecheck
- npm run verify

### DIFF
--- a/src/qq/bot.ts
+++ b/src/qq/bot.ts
@@ -254,12 +254,18 @@ export class QQBot extends EventEmitter {
     content: string,
     msgId?: string,
     msgSeq?: number,
+    markdown = false,
   ): Promise<void> {
     const token = await this.ensureToken();
-    const body: Record<string, unknown> = {
-      content,
-      msg_type: 0,
-    };
+    const body: Record<string, unknown> = markdown
+      ? {
+          markdown: { content },
+          msg_type: 2,
+        }
+      : {
+          content,
+          msg_type: 0,
+        };
     if (msgId) body.msg_id = msgId;
     if (typeof msgSeq === "number" && Number.isFinite(msgSeq)) body.msg_seq = Math.trunc(msgSeq);
     const res = await fetch(`${this.baseUrl}/v2/users/${encodeURIComponent(openid)}/messages`, {
--- a/src/qq/channel.ts
+++ b/src/qq/channel.ts
@@ -11,6 +11,7 @@ import { formatQQAccessSummary } from "./strings.js";
 const QQ_LOCK_FILE = join(homedir(), ".reasonix", "qq-channel.pid");
 const QQ_MAX_CHUNK_BYTES = 1500;
 const NATURAL_SPLIT_MIN_FRACTION = 0.6;
+const QQ_MARKDOWN_WRAPPER_RE = /^```(?:markdown|md)\s*\r?\n([\s\S]*?)\r?\n```$/i;
 
 function fitUtf8Slice(text: string, maxBytes: number): string {
   let end = 0;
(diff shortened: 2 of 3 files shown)


esengine (changes requested):
Thanks — the `normalizeQQMarkdownReply` unwrap and its test are clean and I'd like to land that part as-is. The markdown-by-default delivery path needs some work though.

**1. Silent fallback masks real errors**

The CLAUDE.md error-handling rule is explicit: *"No 'graceful fallback' silently masking bugs. Log + crash > silent wrong output."* The current `try { markdown } catch { plain }` swallows everything — including failures that aren't actually about markdown permission (expired token, account suspended, rate-limit, network blip). Operators get no signal that anything's wrong.

Please at minimum surface the first markdown failure via `onError` (as a warning, not a fatal — keep delivery flowing) so it's visible in logs.

**2. No session-level cache → 2× API cost when markdown isn't permitted**

QQ Bot Open API's `msg_type: 2` raw markdown isn't enabled for most bots — consumer-tier bots typically need an approved `custom_template_id` and get rejected for raw markdown. For those bots, **every message** under this PR will round-trip markdown → reject → plain text. That doubles request count and rate-limit pressure for the common case.

Please cache the disabled state on the channel instance after the first markdown rejection:

```ts
export class QQChannel {
  private markdownDisabled = false;
  // ...
  async sendResponse(text: string): Promise<void> {
    // ...
    if (!this.markdownDisabled) {
      try {
        await this.bot.sendPrivateMessage(..., true);
        continue;
      } catch (err) {
        this.markdownDisabled = true;
        this.callbacks.onError?.(`QQ markdown delivery disabled after first failure: ${(err as Error).message}`);
      }
    }
    await this.bot.sendPrivateMessage(..., false);
  }
}
```

This keeps the auto-detection benefit (bots with markdown perms get the better UX) without the steady-state cost for bots without.

**3. `msgSeq` reuse risks "duplicate msgseq"**

Same `msgSeq` is reused for both the markdown attempt and the plain-text fallback. If QQ's server commits the markdown send (returns 200) but our client treats a subsequent body-rejection error as a failure and retries with plain text on the same seq, the second call hits the QQ-side duplicate-seq guard. Your own test (`stops after the first chunk whose markdown and plain-text fallback both fail`) reproduces the `duplicate msgseq` shape. With the session cache from #2, fallback only happens once per channel-lifetime, so the realistic exposure is small — but worth incrementing the seq on fallback to be safe.

**4. Chunked-boundary markdown invalidity**

`splitQQMessage` cuts at 1500 UTF-8 bytes; that can land mid-table / mid-code-fence / mid-list, producing structurally broken markdown that QQ will reject. The session cache from #2 also fixes the steady-state of this (one cut → fallback once → markdown disabled), but worth mentioning so the design choice is explicit.

Land #1 + #2 and I'll merge. #3 / #4 follow from #2 so should mostly take care of themselves.

wade19990814-hue:
Addressed the requested changes on the latest head (`9e6fd6ca`).

Changes made:
- kept `normalizeQQMarkdownReply(...)` as-is
- added a session-level `markdownDisabled` cache on `QQChannel`
- the first markdown rejection now emits a visible warning through `onError` instead of silently disappearing
- after the first rejection, later chunks on the same channel go straight to plain text
- plain-text fallback now uses a fresh `msgSeq` instead of reusing the markdown attempt sequence

Re-verified after the update:
- `npm test -- --run tests/qq-channel.test.ts tests/qq-access.test.ts tests/qq-config.test.ts tests/qq-slash.test.ts tests/qq-first-connect.test.tsx`
- `npm run typecheck`
- `npm run verify`

This keeps the markdown auto-detect behavior for bots that support it, while avoiding the steady-state double-request path for bots that do not.

esengine (approved):
All four points addressed — onError warning on first markdown failure, session-level markdownDisabled cache, fresh msgSeq on fallback. Tests cover each path. Thanks for the careful follow-up!

PR #1946, opened 2026-05-26
Quality: Hardcoded magic numbers for layout dimensions

tuanaiseo:
## Problem

The `getThreadMaxWidth` function contains multiple magic numbers (580, 80, 1120) without named constants or documentation explaining their origin. These values appear to be derived from design specifications but are not maintainable.

**Severity**: `low`
**File**: `desktop/src/ui/thread-layout.ts`

## Solution

Extract magic numbers into named constants with JSDoc comments explaining their design rationale, e.g., `MIN_THREAD_WIDTH = 580`, `THREAD_MARGIN = 80`, `MAX_THREAD_WIDTH = 1120`.

## Changes

- `desktop/src/ui/thread-layout.ts` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/desktop/src/ui/thread-layout.ts
+++ b/desktop/src/ui/thread-layout.ts
@@ -1,3 +1,12 @@
+/** Minimum thread width in pixels to ensure readability */
+const MIN_THREAD_WIDTH = 580;
+
+/** Right margin in pixels to prevent thread from touching the edge */
+const THREAD_MARGIN = 80;
+
+/** Maximum thread width in pixels to maintain optimal line length for reading */
+const MAX_THREAD_WIDTH = 1120;
+
 export function getThreadMaxWidth({
   viewportWidth,
   visibleSide,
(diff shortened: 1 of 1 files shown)


esengine:
Thanks for the suggestion, but I'm going to pass on this one.

The project's house style (see CLAUDE.md) leans toward inline clamp expressions over named layout constants when the function name already carries the intent. `getThreadMaxWidth` returning `Math.max(580, Math.min(computed, 1120))` reads as "clamp between 580 and 1120" in one line, with all three numbers comparable side-by-side. Splitting them into three named constants with one-line JSDoc each adds vertical noise without unlocking new meaning — and `"to ensure readability"` / `"to prevent thread from touching the edge"` aren't the "non-obvious why" comments belong (CLAUDE.md spells out the four shapes).

If 580 / 1120 ever pick up real hidden constraints — a specific browser quirk, a Tauri pane minimum, a design system token — that's when extracting + commenting becomes worth the maintenance cost. Until then the inline form is what we want.

Closing — no action needed from you.

Issue #1955, opened 2026-05-27
Bug: React "Maximum update depth exceeded" crash when rendering multiple edit_file results


xjr100419:
## Describe the bug

When Reasonix agent performs multiple `edit_file` / `multi_edit` calls and returns several SEARCH/REPLACE blocks in a single turn, the terminal UI crashes with:

```
Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
```

Full stack trace:
```
at getRootForUpdatedFiber (chunk-3OXD5CBM.js:13936:128)
at enqueueConcurrentHookUpdate (chunk-3OXD5CBM.js:13920:16)
at dispatchSetStateInternal (chunk-3OXD5CBM.js:15371:20)
at dispatchSetState (chunk-3OXD5CBM.js:15342:9)
at f (chunk-3OXD5CBM.js:30198:39)
at Timeout.b (chunk-3OXD5CBM.js:30113:32)
at listOnTimeout (node:internal/timers:588:17)
at process.processTimers (node:internal/timers:523:7)
```

## To Reproduce

1. Run `npx reasonix code`
2. 2. Ask the agent to fix a bug that requires multiple `edit_file` / `multi_edit` calls
3. 3. The agent returns 3+ SEARCH/REPLACE edit blocks in a single response
4. 4. The terminal UI crashes with the above error
## Expected behavior

The edit blocks should render normally without React state update loops.

## Environment

- OS: Windows 11
- - Node.js: v22.20.0
- - - Reasonix: 0.52.0
- - - - Terminal: Windows Terminal
## Additional context

The crash occurs in React Ink's reconciler (`getRootForUpdatedFiber` → `dispatchSetState`). Rapid state updates from rendering tool results (edit diffs) trigger React's nested update guard (50-deep limit). This is likely in the component that renders tool call results in the terminal UI.

The error has been observed specifically when the agent returns many SEARCH/REPLACE edit blocks for the same file or across multiple files in one turn.

esengine:
[...]

PR #1960, opened 2026-05-27
Fix edit result scroll measurement bursts

GTC2080:
## Summary
- Coalesces CardStream measurement writes for card heights and max scroll into one trailing update.
- Flushes pending measurements before user scroll actions so wheel, PageUp, and End semantics stay current.
- Adds regression coverage for edit_file and multi_edit result bursts plus measurement notification coalescing.

## Root cause
When many edit results render in one turn, CardStream can report dozens of height and max-scroll measurements synchronously. In app-managed scroll mode this creates a React/Ink external-store update burst; in Windows Terminal it can trip React's nested update guard.

## Test plan
- npm run verify

### DIFF
--- a/src/cli/ui/state/chat-scroll-store.ts
+++ b/src/cli/ui/state/chat-scroll-store.ts
@@ -59,8 +59,11 @@ export function createChatScrollStore(opts: CreateChatScrollStoreOptions = {}): 
   const listeners = new Set<ScrollListener>();
   let pendingDelta = 0;
   let flushTimer: NodeJS.Timeout | null = null;
-  let pendingMaxShrink: number | null = null;
-  let shrinkTimer: NodeJS.Timeout | null = null;
+  // CardStream can report many measured edit-result cards in one render burst;
+  // coalesce them so React/Ink sees one external-store notification.
+  let pendingMaxScroll: number | null = null;
+  let pendingCardHeights: Map<string, number> | null = null;
+  let measurementTimer: NodeJS.Timeout | null = null;
 
   function set(next: Partial<ChatScrollState>): void {
     const merged = { ...state, ...next };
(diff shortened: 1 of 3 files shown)


esengine:
Thanks for digging into this — your root-cause writeup nailed it (CardStream measurement burst tripping React's nested-update guard).

I went a different cut at the same problem in #1958: instead of coalescing the burst in the store, drop the inner-Box `useBoxMetrics` entirely and derive `totalInnerRows` from the per-card heights the store already has. The measurement-feedback edge stops existing at the data-flow level — `scrollRows` can no longer flow back into the value that feeds `maxScroll`, so the burst can't compound regardless of how the runtime times it.

Trade-offs I weighed:
- **Architectural vs. defensive**: removing the cycle vs. throttling its blast radius. The first leaves no latent edge for future regressions.
- **Store semantics**: this PR makes `setMaxScroll` / `setCardHeight` writes asynchronous, which means `store.setMaxScroll(x); store.getState().maxScroll` no longer equals `x` until the next tick. You handled that with explicit `flushMeasurements()` at every scroll-action entry point, which works but adds coupling that I'd rather not lock in.
- **Diff size**: -8 net lines vs. +156. Smaller surface = easier to reason about and revert.

One honest finding while validating: I rebuilt your `card-stream-edit-results.test.tsx` against pre-fix `main` (no fix, no store coalescing) — it still passes. The test environment (`TestStdout`, no real TTY, vitest's fork pool) doesn't reproduce the wall-clock timing that fires the nested-update guard in a real terminal. So the assertion `not.toContain("Maximum update depth")` ends up being a smoke check rather than a regression guard for #1955 specifically. Worth knowing for next time.

Going to close this in favor of #1958 — but seriously, thank you for the independent analysis. Sharp catch.

esengine:
Closing in favor of #1958. Thanks again.

PR #2004, opened 2026-05-27
feat(skill): add builtin qq helper

wade19990814-hue:
## Summary
- add a built-in `qq` skill for QQ channel setup and troubleshooting
- localize built-in skill listing / execution text and include builtin skills in `/skill` completion
- tighten the helper so it stays concise, asks for intent when the user only vaguely mentions `qq`, and keeps App Secret handling inside the existing QQ prompts/UI

## Why
This follows the direction discussed for QQ support after the desktop/runtime fixes:
- the real pain point is not just runtime bugs, but beginners repeatedly getting stuck on first-time QQ setup
- maintainer bandwidth is limited, so a small built-in helper is cheaper than expanding more IM surfaces before onboarding is under control
- the helper keeps the product guidance inside Reasonix itself instead of depending on users to find GitHub docs first

The skill is intentionally narrow:
- it focuses on QQ channel setup / troubleshooting for CLI or desktop
- it prefers the shortest next action instead of dumping the whole manual
- it includes a gentle safety boundary so users do not paste `App Secret` into model chat
- docs remain the fallback, not the primary path

## Scope notes
- this is primarily a CLI-first helper; it does not add a new desktop feature surface
- it does not broaden into Feishu / Discord / cc-connect
- it does not change QQ runtime semantics

## Implementation
- add builtin `qq` skill in `src/skills.ts`
- localize builtin skill descriptions, including `/skill list` / `/skill <name>` output
- include builtin skills in `/skill` argument completion
- add tests for:
  - builtin skill registration
  - localized `/skill list` and `/skill qq`
  - builtin skill completion
- fix test/config isolation so slash-model tests do not write invalid model ids into the real user config
- make CLI default model resolution use the validated model loader path

## Validation
- `npm test -- --run tests/skills.test.ts tests/slash.test.ts tests/skill-completion-picker.test.tsx`
- `npm run typecheck`
- `npm run verify`


wade19990814-hue:
The value of this PR is not just “less support load” in the abstract.

More specifically, it gives Reasonix a built-in QQ-specific help surface that can answer the first layer of setup and troubleshooting inside the product, using the same current runtime assumptions as the actual QQ channel.

What it already knows how to do:
- distinguish between first-time QQ setup vs troubleshooting
- distinguish CLI vs desktop paths
- point users to the right entry point (`/qq connect` or desktop settings)
- explain common states like `unbound`
- avoid the most common unsafe behavior: pasting `App Secret` into model chat
- stay narrow to QQ instead of broadening into unrelated IM/runtime directions

So this is not only “documentation moved into a skill”.
It is a built-in, localized, intent-shaped QQ helper that is discoverable through `/skill`, shows up in `/skill` completion, and gives users a shortest-next-step path before they open another issue.

It does not replace runtime fixes, and it does not solve every QQ problem.
But it should reduce the repetitive “I know QQ exists, but I still don’t know what to do next in my current context” support loop.

That matters because the helper is not just a static text blob: it is registered as a builtin skill, localized in `/skill list` and `/skill qq`, and exposed through builtin skill completion, so the product can actually surface it at the moment the user is trying to get unstuck.

PR #1915, opened 2026-05-26
feat(i18n): add Japanese (ja) locale

cwatanab:
- src/i18n/JA.ts: full JA translation for CLI/TUI (~1900 lines)
- desktop/src/i18n/ja.ts: full JA translation for desktop shell (~740 lines)
- dashboard/src/i18n/ja.ts: JA translation for dashboard SPA (~1320 lines)
- All three index.ts: register 'ja' in Lang type, dicts, detection, labels
- Wizard.tsx: add 日本語 to language picker

## Checklist

- [ ] `npm run verify` passes locally (lint + typecheck + tests + comment-policy gate)
- [ ] No `Co-Authored-By: Claude` trailer in commits
- [ ] Comments follow CONTRIBUTING.md (no module-essay headers, no incident history)
- [ ] No edits to `CHANGELOG.md` — release notes are maintainer-written at release time

PR #2015, opened 2026-05-27
feat(desktop): grant pending tool/path approval with Enter key

### LINKED ISSUE #1962: 授权是否可以适配回车或者设置快捷键，每次都需要手动点击
授权是否可以适配回车或者设置快捷键，每次都需要手动点击

nianyi778:
## Problem

Closes #1962. The permission approval prompts (`ConfirmApprovalCard` / `PathAccessApprovalCard`) in the desktop and dashboard UIs could only be granted by clicking — there was no keyboard shortcut, so every authorization required reaching for the mouse. The TUI already supports this (the `SingleSelect` confirms the highlighted choice on Enter); this brings the GUIs in line.

## Change

Extend the existing global `keydown` handler in both `desktop/src/App.tsx` and `dashboard/src/App.tsx` (the same place that already maps Esc → abort) so **Enter grants the pending authorization (run once)**, with the same guards as the Esc handler:

- ignored when focus is in an `INPUT` / `TEXTAREA` / `contentEditable` element (so it never hijacks the composer or the always-allow prefix field);
- ignored when a modal (settings / about / jobs / working-dir picker) is open;
- only the single pending `Confirm` / `PathAccess` prompt is resolved (most-recent first), via the existing `resolveConfirm`/`resolvePathAccess` with `{ type: "run_once" }`.

Multi-option decisions (plan / checkpoint / revision / choice cards) are intentionally left click-only, since Enter has no unambiguous default there.

## Notes

- Bare Enter (not Cmd/Ctrl+Enter) matches the TUI and the issue request; the editable-focus guard keeps it from clashing with composer send.
- No new i18n surface added — behavior only.

## Verification

- `npm run typecheck` (root + dashboard) ✓
- `cd desktop && tsc --noEmit` ✓ (the unrelated pre-existing `ja` locale error from #1915 is not touched here)
- `npm run verify` (build + lint + typecheck + full test suite) ✓

### DIFF
--- a/dashboard/src/App.tsx
+++ b/dashboard/src/App.tsx
@@ -1586,11 +1586,50 @@ function TabRuntime({
         if (target?.tagName === "INPUT" || target?.tagName === "TEXTAREA") return;
         e.preventDefault();
         abort();
+      } else if (e.key === "Enter" && !mod && !e.shiftKey && !e.altKey) {
+        // Defer to any control that already handles Enter — native inputs/buttons,
+        // ARIA button/link widgets (sidebar rows, file pills), or anything that called
+        // preventDefault — so we only grant when focus is on inert layout (#2015).
+        if (e.defaultPrevented) return;
+        const target = e.target as HTMLElement | null;
+        if (
+          target?.isContentEditable ||
+          target?.closest('input, textarea, button, select, a, [role="button"], [role="link"]')
+        ) {
+          return;
+        }
+        if (settingsOpen || jobsOpen || wdOpen) return;
+        // Enter grants the pending authorization prompt (run once), matching the
+        // TUI where Enter confirms the highlighted choice (#1962).
+        const confirm = state.pendingConfirms.at(-1);
+        if (confirm) {
+          e.preventDefault();
+          resolveConfirm(confirm.id, { type: "run_once" });
+          return;
+        }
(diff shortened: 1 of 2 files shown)


nianyi778:
Good catch — fixed in the latest commit (`69296fe7`). The Enter handler now bails out when focus is on any actionable control (`BUTTON` / `SELECT` / `A`, in addition to `INPUT` / `TEXTAREA` / contentEditable), so a focused **Reject** or **Always Allow** button keeps its native Enter behavior and is never overridden by the `run_once` shortcut. The window-level handler only grants when focus is outside the card's buttons (e.g. on `body`). Applied to both `desktop/src/App.tsx` and `dashboard/src/App.tsx`.

@codex review

nianyi778:
Addressed in `d40d9a70`. Replaced the tag-name allowlist with a more robust guard that also covers custom keyboard controls: the handler now bails out if (a) the event was already `preventDefault`ed by a child handler (e.g. the file pills in `Markdown.tsx`), or (b) the target is inside an interactive element via `closest('input, textarea, button, select, a, [role="button"], [role="link"]')` — which catches the ARIA `role="button"` sidebar session rows and similar widgets that are `DIV`/`SPAN` at the tag level. So Enter on those runs only their own handler and never falls through to `run_once`. Applied to desktop + dashboard.

@codex review

PR #2056, opened 2026-05-27
perf(desktop): virtual list with Virtuoso for message thread

CVEngineer66:
## Before

All messages rendered as real DOM nodes in the thread. 60 messages = 60 full DOM subtrees (UserMsg + TurnDivider + AssistantMsg + DiffStats). Every state update triggered React reconciliation on all 60 nodes, even though only ~8 fit on screen.

## After

`react-virtuoso` virtual list — only visible messages produce DOM nodes. Off-screen messages are zero-height spacer elements. React reconciliation is limited to the ~8 visible items regardless of total message count. DOM nodes drop from O(total messages) to O(viewport size).

### DIFF
--- a/desktop/src/App.tsx
+++ b/desktop/src/App.tsx
@@ -66,6 +66,7 @@ import { useElapsed } from "./ui/live";
 import { AboutModal } from "./ui/about";
 import { SettingsModal, type PageId as SettingsPageId } from "./ui/settings";
 import { JumpBar } from "./ui/jump-bar";
+import { Virtuoso, type VirtuosoHandle } from "react-virtuoso";
 import { Sidebar } from "./ui/sidebar";
 import { Shortcut, localizeShortcutText, shortcutText } from "./ui/shortcut";
 import { Splash, shouldShowSplash } from "./ui/splash";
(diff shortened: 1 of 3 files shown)


esengine:
This now conflicts with #2055 (just merged) — both rewrote the streaming reducer cases in `desktop/src/App.tsx`. The backwards-walk reducers + virtuoso virtual list are actually the right combo for the long-session pressure (#2001, #2012, #2013), so I'd love to take both — could you rebase onto current `main`?

One thing to think about while you're at it: Virtuoso needs stable, monotonic item heights for `followOutput="auto"` to track the bottom correctly when streaming. The current `state.messages[index]` lookup re-renders the same item on every reducer update, which Virtuoso will then re-measure. If you can wrap `itemContent` in a memoized renderer keyed by `(turn, segment count, pending)`, the scroll-anchor stability gets much more predictable on fast streams.

Pending-approval cards, jump bar, empty-state, restore-scroll behavior — please verify those still work after rebase. The render path moved a lot of conditional branches around.

CVEngineer66:
Rebased onto latest \`main\` — no conflicts since #2055's streaming handler changes are in the reducer and this PR's Virtuoso changes are in the render path.

On \`itemContent\` memo: the function reads from \`state.messages\` and several callbacks that change on every new message or config update, so an empty-dep useCallback would go stale. A per-item memo keyed by \`(turn, segments.length, pending)\` would fix the scroll-anchor — I'll send a follow-up once this lands.

Pending-approval cards, jump bar, empty-state, and restore-scroll all verified independently (they're rendered outside the \`Virtuoso\` component).

PR #2057, opened 2026-05-27
perf(core): windowed AppendOnlyLog (200 messages) with lazy file fallback

CVEngineer66:
## Before

`AppendOnlyLog` kept all messages in memory as `ChatMessage[]`. Session file was read entirely on resume: `readFileSync` + `split` + `JSON.parse` every line. A 500-message session held all 500 entries in memory permanently. `toMessages()` always returned the full array from RAM.

## After

`AppendOnlyLog` windows to 200 entries. Older messages are evicted from `_entries` and only reloaded from the session file when `toMessages()` or `getEntry()` is called. The backward-file-scanner (`readTailMessages`) reads only the last N lines from the JSONL via chunked reverse scan, not the entire file.

- **Memory**: `_entries` array capped at 200 entries regardless of session length
- **Session resume**: reads the last ~200 lines from disk (~64KB chunks from EOF) instead of mmap'ing the entire file
- **Lazy fallback**: `toMessages()` and `getEntry()` transparently read old entries from disk when needed (e.g., during fold or API call construction)


esengine:
The direction is right — keeping a 200-entry window in RAM and letting the JSONL file be the source of truth for older messages is the natural endgame for long sessions. A few things I want to talk through before this lands, because `AppendOnlyLog` is load-bearing:

1. **`toMessages()` becomes sync disk I/O on the hot path.** It's called from compaction, summary fold, prefix rebuilds, and the API request builder — anywhere the loop needs the full history. A 200KB+ JSONL reverse-scan on every call will stall the event loop for tens of ms at p95. Two options: (a) cache the "full read" result inside the log and invalidate on `append`/`compactInPlace`, or (b) make the contract explicit — `toMessages()` returns the window only, callers that need everything call a new `toFullHistory()` and own the cost.

2. **`length` semantics flip.** Before: `length === entries.length`. After: `length === totalLength` (window + disk). Every caller that did `for (let i = 0; i < log.length; i++) log.entries[i]` (and there are several) now reads past the end. Worth grepping for `log.length`, `log.entries[`, and `log.entries.length` and updating the touch points in one pass.

3. **`compactInPlace` resets `_totalLength` to `replacement.length` but doesn't truncate the on-disk file.** Next `toMessages()` call with `_entries.length < _totalLength` is false (they're equal), so the disk file is ignored — fine. But if a later `append()` pushes the window over and a subsequent `toMessages()` falls back to disk, it'll read pre-compaction messages. Either truncate/rewrite the session file at compaction time, or stamp the log with a "no longer authoritative" flag for the disk fallback.

4. **Session resume + StormBreaker** — StormBreaker reads the tail of `log.entries` to detect repeats. If `entries` is bounded by the window but the storm pattern started before the window's left edge, StormBreaker silently loses signal. Probably fine in practice but worth a one-line note that the storm window must be ≤ log window.

CI is still running — let me see it pass first. Once we resolve (1)–(3), happy to take this; (4) can be a follow-up.

CVEngineer66:
Thanks for the review. Here's what was addressed:

**1. toMessages() disk I/O** — Split: \`toMessages()\` now returns window only (no disk I/O). Added \`toFullHistory()\` for callers that need everything (fold, buildMessages, getLogTokens, ctx-breakdown, emitCtxBreakdown). They already do async work (network requests, file writes) so the sync read cost is amortized.

**2. length semantics** — Restored \`length\` to return \`_entries.length\` (window size). Added \`totalLength\` getter for internal use. Updated \`discardLogFrom\`, \`retryLastUser\`, \`rewindToUserTurn\` to use \`toFullHistory()\` instead of \`entries\` so they scan the full history, not just the window.

**3. compactInPlace + disk** — Confirmed no issue: after \`compactInPlace\`, \`_totalLength = replacement.length\`, so \`_entries.length >= _totalLength\` holds and \`toFullHistory()\` returns window only (no disk fallback). Subsequent \`append()\` writes to both window and disk via \`appendSessionMessage\`, keeping them in sync.

**4. StormBreaker** — Noted; storm window is 3 by default which fits well within the 200-entry log window.

PR #2058, opened 2026-05-27
feat(desktop): support desktop QQ remote commands

wade19990814-hue:
## Summary
- let desktop QQ remote input understand a small subset of desktop-native commands
- keep the scope aligned with the existing desktop design instead of copying the full CLI `/qq` surface
- add parser tests for the QQ remote desktop command subset

## What this adds
Desktop QQ remote input now handles these desktop-native commands before falling back to normal user text:
- `/help`
- `/new`
- `/abort`
- `/compact`
- `/btw <question>`
- `/<skill> [args]` for skills already exposed by the current desktop tab

This intentionally does **not** add UI-only commands such as `/theme`, `/model`, `/lang`, `/currency`, `/copy`, or `/clear`.

## Why
Desktop already has its own command surface and session model. QQ remote input was still treating everything as plain user text plus pending-prompt replies, which meant desktop-only commands were unavailable remotely.

This keeps QQ aligned with the desktop interaction model without turning it into a separate command system.

## Notes
- `/abort` and `/new` intentionally bypass the generic busy reply so they remain usable while a tab is running
- unsupported slash text still falls back to normal model input
- `/retry` is left local for now because the desktop implementation restores local draft state rather than directly submitting a new turn

## Verification
- `npm test -- --run tests/desktop-qq-remote-commands.test.ts tests/desktop-qq-turn-routing.test.ts tests/desktop-btw-status.test.ts tests/desktop-session-load.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run verify`

### DIFF
--- a/src/cli/commands/desktop.ts
+++ b/src/cli/commands/desktop.ts
@@ -76,6 +76,11 @@ import {
   readMemoryEntryDetail,
 } from "../../desktop/memory-browser.js";
 import {
+  parseQQRemoteDesktopCommand,
+  qqRemoteCommandBypassesBusy,
+  qqRemoteDesktopHelpText,
+} from "../../desktop/qq-remote-commands.js";
+import {
   loadDesktopQQState,
   saveDesktopQQSettings,
   setDesktopQQEnabled,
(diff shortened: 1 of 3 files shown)


wade19990814-hue:
One correction from my side: after `#2058` was merged, I realized I had still missed a few desktop-native remote commands that should have been considered in the same pass (`/retry`, `/model`, `/effort ...`, `/plan ...`). That was my miss, and I’m sorry for the extra review churn this creates.

I’ll keep the fix-up work in a separate follow-up PR so the already-merged scope stays clean and the delta is easy to review independently.

PR #2062, opened 2026-05-27
feat(desktop): extend QQ remote commands with model and mode controls

wade19990814-hue:
## Summary
- follow up `#2058` by adding the remaining desktop-native QQ remote commands that fit the desktop session model
- add `/retry`, `/model ...`, `/effort ...`, and `/plan ...` to the desktop QQ remote-command subset
- keep the scope focused on commands that meaningfully control the active desktop session remotely

## Context
`#2058` intentionally shipped a narrow first pass, but I later realized it still missed a few desktop-native commands that should have been considered in the same surface. This PR is the clean follow-up for that gap.

## What this adds
Desktop QQ remote input now also handles:
- `/retry`
- `/model <flash|pro|deepseek-v4-flash|deepseek-v4-pro>`
- `/effort <low|medium|high|max>`
- `/plan <review|auto|yolo>`

These are added before fallback to normal model input, alongside the existing subset from `#2058`.

## What this still does not add
This still does **not** expose UI-only desktop commands such as `/theme`, `/lang`, `/currency`, `/copy`, or `/clear` over QQ.

## Why
Desktop QQ should follow the desktop command surface where it makes sense remotely, rather than mirroring the CLI `/qq` surface or forcing every slash input through normal model text.

These commands are useful remote controls for the active desktop session:
- `/retry` re-runs the last user turn
- `/model` switches the active tab model
- `/effort` adjusts reasoning effort
- `/plan` adjusts the current plan/review/yolo mode

## Verification
- `npm test -- --run tests/desktop-qq-remote-commands.test.ts tests/desktop-qq-turn-routing.test.ts tests/desktop-btw-status.test.ts tests/desktop-session-load.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run verify`

### DIFF
--- a/src/cli/commands/desktop.ts
+++ b/src/cli/commands/desktop.ts
@@ -1356,6 +1356,52 @@ export async function desktopCommand(opts: DesktopOptions): Promise<void> {
         void runTurn(tab, prev, true);
         return true;
       }
+      case "model": {
+        if (!cmd.value) {
+          sendQQInfo(
+            `Current model: ${tab.currentModel}. Use /model flash, /model pro, /model deepseek-v4-flash, or /model deepseek-v4-pro.`,
+            tab,
+          );
+          return true;
+        }
+        const next = normalizeQQRemoteModel(cmd.value);
+        if (!next) {
+          sendQQInfo(
+            "Unsupported desktop model. Use /model flash, /model pro, /model deepseek-v4-flash, or /model deepseek-v4-pro.",
+            tab,
+          );
+          return true;
+        }
+        applyDesktopModel(tab, next);
+        sendQQInfo(`Switched desktop model to ${next}.`, tab);
+        return true;
+      }
+      case "effort":
+        if (!cmd.value) {
+          sendQQInfo(
+            `Current reasoning effort: ${loadReasoningEffort()}. Use /effort low, /effort medium, /effort high, or /effort max.`,
+            tab,
+          );
+          return true;
+        }
+        saveReasoningEffort(cmd.value);
+        tab.runtime?.loop.configure({ reasoningEffort: cmd.value });
+        emitSettings(tab);
+        sendQQInfo(`Switched desktop reasoning effort to ${cmd.value}.`, tab);
+        return true;
+      case "plan":
+        if (!cmd.value) {
+          sendQQInfo(
+            `Current plan mode: ${loadEditMode()}. Use /plan review, /plan auto, or /plan yolo.`,
+            tab,
+          );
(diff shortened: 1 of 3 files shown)


wade19990814-hue:
For review context: this PR is intentionally the follow-up to `#2058`, not a scope change inside the already-merged pass.

`#2058` landed the first narrow desktop QQ remote-command subset (`/help`, `/new`, `/abort`, `/compact`, `/btw`, `/<skill>`). After it was merged, I realized I had still missed a few desktop-native commands that belong on the same surface: `/retry`, `/model ...`, `/effort ...`, and `/plan ...`.

So `#2062` exists purely to add that missed remainder on top of the merged first pass, while keeping the already-merged PR clean.

Sorry again for creating a second review cycle here — that miss was on me.

esengine (changes requested):
Direction is good — these are the natural complements to #2058 and the parser/test/help-text plumbing is clean.

One concern before merge: `/model` (and to a lesser extent `/plan`) are added to `qqRemoteCommandBypassesBusy`, but `applyDesktopModel` does `tab.runtime = buildRuntimeFor(tab)` — a full runtime swap. Every other site in `desktop.ts` that rebuilds runtime (switchSession ~L830, startNewChatInTab ~L1240, workspace switch ~L1891) calls `abortTurn(tab)` first; otherwise the in-flight loop keeps streaming events into a tab whose runtime reference has changed, which is the same shape of bug as #1217.

`/effort` is fine because it goes through `loop.configure({ reasoningEffort })`, which is the established mid-turn-safe path.

Could you either (a) drop `model` and `plan` from `qqRemoteCommandBypassesBusy` so they wait for the turn to finish, or (b) call `abortTurn(tab)` inside `applyDesktopModel` / the `/plan` branch before mutating? Either is fine — I'd lean (a) since silently aborting a running turn from a remote QQ message is surprising.

wade19990814-hue:
Addressed — I took option (a).

Changes in `3e1cbe6`:
- removed `/model` from `qqRemoteCommandBypassesBusy`
- removed `/plan` from `qqRemoteCommandBypassesBusy`
- kept `/effort` bypassing busy, since it still goes through the existing mid-turn-safe `loop.configure(...)` path

So the QQ remote command behavior is now:
- `/help`, `/new`, `/abort`, `/effort` can still run while the active tab is busy
- `/model` and `/plan` now wait until the running turn finishes instead of mutating runtime mid-turn

That matches the safer behavior you suggested and avoids introducing the same class of runtime-swap risk as `#1217`.

PR #2066, opened 2026-05-27
feat(prompt): prefer MCP search tools over built-in defaults

DawnCloud1213:
## What
Adds a `# Tool Selection` section to the default system prompt so the model prefers MCP-provided tools (e.g. Exa) over the built-in `web_search` / `web_fetch` when both serve the same purpose.

## Why
The built-in `web_search` does keyword matching only — no semantic search, no date/domain filtering. MCP tools like Exa (27k uses on Smithery) offer semantic search, date filtering, domain scoping, and highlighted excerpts. Without explicit guidance, the model defaults to the familiar built-in tools, missing the quality gain from installed MCP servers.

## Change
One paragraph added to `defaultSystemPrompt()` in `src/cli/index.ts`:

```
# Tool Selection

When multiple tools serve the same purpose (e.g. web search), prefer MCP-provided 
tools over the built-in defaults — MCP tools like Exa typically offer higher quality 
(semantic search, date / domain filtering). If an MCP tool fails, fall back to the 
built-in.
```

## Impact
- **One paragraph of prompt text** — no runtime overhead, no config changes
- **Fully backward-compatible** — when no MCP search tools are installed, the built-in tools remain the only option
- **Includes fallback clause** — if an MCP tool fails, the model is instructed to fall back to the built-in


esengine (changes requested):
Thanks for the contribution — first-time PRs are very welcome here. A few things to discuss before this can merge:

1. **Wrong target prompt.** `defaultSystemPrompt` in `src/cli/index.ts` is only used by the `chat` command (a thin generic chat surface). The actual coding agent — what 99% of users run — uses `codeSystemPrompt` in `src/code/prompt.ts`. So as written, this change won't affect the path you're trying to improve.

2. **Trailing newline removed.** The diff strips the final newline from `src/cli/index.ts` (`\ No newline at end of file`). That'll fail biome on `npm run verify`. Please re-add it.

3. **Vendor name in the prompt.** Hard-coding "Exa" by name in the system prompt reads as promotional and ages poorly (today's preferred MCP server isn't tomorrow's). If we add this guidance, it should be vendor-neutral — something like "prefer installed MCP search tools when they cover the same purpose".

4. **Open question on the premise.** Built-in `web_search` is in-house and predictable; MCP tools can be slow, rate-limited, or misconfigured. A blanket "prefer MCP" instruction may regress sessions where the user installed an MCP server casually. I'd want to see this guarded — e.g., only when the model can see the MCP tool is healthy — before instructing the model to prefer it.

If you want to retarget this at `codeSystemPrompt` with the vendor name dropped, I'm open to discussing it further; otherwise feel free to close and we can revisit when there's a clearer signal that MCP tools are reliably better.

DawnCloud1213:
All four points are spot-on, thanks for the careful review. Will update:

1. Retarget to `CODE_SYSTEM_TEMPLATE` in `src/code/prompt.ts` — the actual coding agent prompt
2. Restore trailing newline
3. Drop vendor name — generic "prefer installed MCP search tools"
4. Add hedging — "when available" + explicit fallback clause

Pushing the fix shortly.

DawnCloud1213:
> **Disclosure:** This PR was drafted with the assistance of DeepSeek (AI coding agent). Please review carefully before merging — the proposed change is small (one paragraph in the code system prompt) but AI-generated suggestions warrant extra scrutiny.

esengine (approved):
Re-reviewed — all four points from my original review are addressed: (1) retargeted to `codeSystemPrompt` in src/code/prompt.ts (the path 99% of users hit), not the chat-only `defaultSystemPrompt`; (2) the cli/index.ts trailing-newline issue is gone with the retarget (biome's green); (3) vendor-neutral now — 'prefer installed MCP-provided tools', no 'Exa'; (4) the blanket-prefer premise I worried about is guarded by the explicit fallback ('if an MCP tool fails or times out, fall back to the built-in'), which bounds the downside when someone has a flaky MCP server installed. CI green. Good — merging.

PR #2068, opened 2026-05-27
fix(desktop): persist prompt history across sessions

### LINKED ISSUE #2051: Desktop: prompt history recall is in-memory only; it is lost across sessions and app restarts
**Problem**
The desktop app currently keeps prompt-history recall in a **live in-memory composer state** only. As a result, it is not just limited to the current chat session: it is also lost when the app is closed and reopened.

Current behavior, combining live verification and code-path inspection:

1. In one chat session, send two prompts such as `one` and `two`.
2. In that same session, `ArrowUp` / `ArrowDown` can walk back through those prompts.
3. Create a **new chat session**.
4. Focus the empty composer and press `ArrowUp`.
5. Nothing is recalled from the previous session.
6. Quit and reopen the desktop app.
7. Prompt-history recall is gone again.

So the current desktop behavior is:

- prompt history exists only inside the active in-memory desktop composer
- prompt history does not carry across chat sessions
- prompt history does not survive app restart
- recall is also gated by caret position, which makes the feature easy to miss or mistake for being absent

**Why this happens**
The current implementation stores history in the desktop composer itself:

- `historyRef = useRef<string[]>([])`
- `browseIdx = useState(-1)`
- `savedDraftRef = useRef("")`

That means the recall state is tied to the lifetime of the mounted composer instance, not to persisted desktop session data.

**Expected**
Desktop prompt-history recall should support the higher-level workflow users expect from chat tools with input-history browsing:

- `ArrowUp` / `ArrowDown` should recall prior prompts beyond the currently mounted composer instance
- recall should continue to work after switching to another chat session
- recall should survive app restart
- the interaction should remain predictable when the composer caret is at the start/end boundaries

**Proposed change**
- Scope this to **desktop only**.
- Move prompt-history storage out of the ephemeral composer instance.
- Define a persisted recall source for desktop prompt history.
- At minimum, support:
  - same-session recall
  - cross-session recall
  - recall after quitting and relaunching the app
- Keep the current `ArrowUp` / `ArrowDown` flow working while changing the backing storage.
- Add regression coverage for:
  - same-session recall
  - cross-session recall after switching to a new chat session
  - recall after desktop app restart

**Reproduction notes**
Live desktop verification confirmed:

- same-session recall works
- cross-session recall does not work

User-reported behavior, which is also consistent with the current in-memory implementation:

- after quitting and reopening the desktop app, prompt-history recall is lost

**Code pointers**
- Desktop composer history state and key handling: `desktop/src/ui/composer.tsx`
- Desktop app session shell: `desktop/src/App.tsx`

**Out of scope**
- CLI / TUI prompt history
- dashboard / web composer behavior

ferstar:
## Summary
- Persist desktop prompt-history navigation through session JSONL logs instead of composer-local memory
- Add sidecar RPC support for older/newer prompt history cursor navigation
- Preserve duplicate prompts as separate history entries and invalidate prompt-history cache when session metadata changes
- Backfill Japanese plan feedback labels required by latest upstream desktop build

Fixes #2051

## Verification
- npm run test -- tests/session.test.ts desktop/src/ui/composer-at-popup.test.tsx desktop/src/App.test.ts
- npm run test -- tests/public-api.test.ts tests/web-tools.test.ts
- npm run typecheck
- npm --prefix desktop run build
- git diff --check
- npm run verify (pre-push hook)


esengine (changes requested):
Substance looks well thought out: cursor-based session walk, cache with proper invalidation on every write path, nonces to dodge stale results, reset on session change. Tests cover dupes, workspace scoping, and cross-session walks. Closes #2051 properly.

Three things before merge:

1. **Scope creep — please split.** Two changes here are unrelated to prompt history persistence:
   - `desktop/src/i18n/ja.ts` — `sendFeedback / skipFeedback / cancelFeedbackLabel / refineFeedbackLabel` backfill. This is exactly the scope of #2069, which is open in parallel; they will conflict.
   - `tests/web-tools.test.ts` — a new `vi.mock("../src/config.js", …)` block. I can't see how that's related to this PR.

   Please drop both from this PR and rebase; #2069 will land separately and the web-tools change should have its own PR with a motivating description.

2. **Public API surface — is this intentional?** You're adding `promptHistoryStep`, `PromptHistoryCursor`, and `PromptHistoryEntry` to `src/index.ts` (and bumping the public-api allowlist). That's a stability commitment to downstream consumers, but the function reads like an internal desktop helper. Unless there's a planned external consumer, I'd prefer keeping it un-exported and having `desktop.ts` import directly from `./memory/session.js`.

3. **5s cache TTL — minor.** All mutating helpers correctly call `invalidatePromptHistoryCache()`. Add a one-line comment near the cache constant calling out that invalidation is the primary correctness mechanism and 5s is just a backstop for external mtime changes. (Optional, but helps the next reader.)

CI is green and the core change is good. Fix 1, decide on 2, and we're ready.

ferstar:
Thanks, addressed. I rebased onto the latest main and kept this PR scoped to persisted prompt history: dropped the ja/web-tools changes, kept the prompt history helper out of the public API, added the cache TTL note, and replaced the config-backed prompt history path with the session-log/cursor implementation. CI is green and the branch is clean for re-review.

ferstar:
One extra note on the follow-up commit: I removed the config-backed prompt history path so there is a single source of truth. Since prompts are already persisted in session logs, the cursor-based session walk avoids duplicating history state and prevents config/session drift.

esengine (approved):
Re-reviewed — both blockers from my original review are resolved:
1. **Scope creep dropped**: the `desktop/src/i18n/ja.ts` backfill (that's #2069's territory) and the unrelated `tests/web-tools.test.ts` `vi.mock` are both gone. The diff is now focused on prompt-history persistence (session.ts/config.ts) + its composer wiring.
2. **Public API kept internal**: `src/index.ts` is untouched — `promptHistoryStep`/`PromptHistoryCursor`/`PromptHistoryEntry` are no longer exported, so `desktop.ts` consumes them directly from `./memory/session.js` without committing to a downstream-stability surface, as preferred.
The core (cursor-based session walk, cache with per-write invalidation, nonces, reset-on-session-change) was already solid. CI green. Approving — merging if it's conflict-free.

esengine:
Approving the content (both blockers resolved). It just needs a rebase — it now conflicts with main on `src/config.ts` (recent merges touched it). `git fetch origin main && git merge origin/main`, resolve config.ts, and I'll merge. No further changes needed.

ferstar:
Merged latest origin/main into the PR branch and resolved the conflicts. Head is now cceed18e; PR is mergeable again. Local verify only hits the existing Exa-config-sensitive web-tools test on my machine, and that test passes under a clean HOME.

PR #2070, opened 2026-05-27
feat(lifecycle): detect explicit plan-first intent

GTC2080:
## Summary

- add a narrow host-side detector for explicit plan-first / strict lifecycle requests
- wire the detector into the TUI user prompt path so explicit requests enable existing strict plan rails before the model turn starts
- keep the detector conservative with bilingual positive cases and false-positive coverage for `add`, `api`, `修`, `.json`, `across`, and lifecycle-hook wording

## Design notes

This is intentionally not broad auto-arm. It only reacts when the user explicitly asks for plan-first behavior or strict lifecycle / plan mode. The runtime still uses the existing `/plan strict` rails, and the user can leave with `/plan off`.

No system prompt, tool schema, or immutable prefix changes are introduced.

## Validation

- `npm test -- tests/lifecycle-explicit-intent.test.ts`
- `npm test -- tests/code-prompt.test.ts tests/lifecycle-explicit-intent.test.ts tests/lifecycle-policy.test.ts tests/lifecycle-e2e.test.ts tests/lifecycle.test.ts tests/tools.test.ts tests/slash.test.ts`
- `npx biome check src/cli/ui/lifecycle-explicit-intent.ts tests/lifecycle-explicit-intent.test.ts src/cli/ui/App.tsx src/cli/ui/slash/types.ts src/i18n/EN.ts src/i18n/zh-CN.ts src/i18n/JA.ts src/i18n/types.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run verify`
- pre-push hook reran `npm run verify`

### DIFF
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -175,6 +175,7 @@ import { ModeStatusBar } from "./layout/LiveRows.js";
 import { StaticCardStream } from "./layout/StaticCardStream.js";
 import { StatusRow } from "./layout/StatusRow.js";
 import type { StatusBarConfig } from "./layout/StatusRow.js";
+import { shouldEnterPlanModeForExplicitIntent } from "./lifecycle-explicit-intent.js";
 import { formatLoopStatus } from "./loop.js";
 import { applyMcpAppend } from "./mcp-append.js";
 import { handleMcpBrowseSlash } from "./mcp-browse.js";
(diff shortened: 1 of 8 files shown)


esengine (approved):
Scope is appropriately narrow, false-positive cases (address / came across / tsconfig.json / lifecycle hooks) are well-chosen, and the banner + /plan off escape hatch make the auto-arm reversible. Detector lives in its own file, so we can remove it cleanly if it regrets in the wild.

PR #2073, opened 2026-05-27
fix(desktop): Virtuoso initial scroll to bottom + JumpBar onScrollToTurn callback (#2056 hotfix)

CVEngineer66:
### 1. Virtuoso container height 100% → 90%

**Why:** At 100% height the container sits flush with the bottom edge. The Composer's gradient overlay (`::before` pseudo-element, extending 24px upward) covers the last message. 90% height leaves room at the bottom, preventing content from being obscured and reducing the visual impact of flex layout shifts when sending messages.

### 2. Always scroll to bottom (removed atBottomRef guard)

**Why:** When a message is sent, the Composer clears and shrinks → flex layout recalculates → `.thread` gets taller → Virtuoso thinks the user left the bottom (scrollTop unchanged but container grew) → `followOutput` stops firing → subsequent streaming content appears above the viewport, invisible to the user.

Removing the `atBottomRef` gate ensures every state change forces `scrollTop = scrollHeight`, keeping new content always visible.

## Files changed

`desktop/src/App.tsx` — +7 / -21
- busyPrevRef: removed redundant rAF scroll (auto-scroll effect covers it)
- scrollToBottom: moved atBottomRef/setShowJumpButton to jump button's onClick
- Auto-scroll effect: removed atBottomRef guard, now unconditional
- Virtuoso height: 80% → 90%
- Jump button onClick: now sets atBottomRef + hides button before scrolling


CVEngineer66:
If the remaining Virtuoso issues are too disruptive, feel free to revert #2056 entirely and go back to the simple \`overflow-y: auto\` scrollable thread. The virtual list was meant to reduce DOM nodes on long sessions, but the existing \`content-visibility: auto\` + \`elideTranscriptMessages\` (120-msg window with heavy content stubbing) already covers most of that use case without the complexity. There may be other edge cases I haven't found yet — this PR only addresses the two most visible regressions.

esengine (changes requested):
Thanks for the hotfix — the two shipped fixes (initial scroll-to-bottom + JumpBar callback) are exactly the right shape and CI is green.

But the description doesn't match the diff:

- ✅ #1 `initialTopMostItemIndex` — in the diff.
- ✅ #2 JumpBar `onScrollToTurn` — in the diff.
- ❌ #3 "Added `m.kind === 'error'` and `m.kind === 'warning'` render cases back to virtuoso `itemContent`." — **not in the diff**. The `itemContent` callback still returns `null` for everything except `user` / `assistant`, so error and warning messages remain invisible. This is a real visible regression from #2056 — `$error` events get pushed into `state.messages` (App.tsx L1101) but never render.
- ❌ #4 "Pending-approval cards moved to `components.Footer`" — **not in the diff**. They're still rendered as a separate block below the scroll area (App.tsx L2401).

Could you either:

1. **Ship #3 in this PR** (it's the meaningful regression — error / warning visibility on desktop is broken right now), and drop #4 from the description (or split it into its own follow-up since it's a layout choice, not a bug), OR
2. **Trim the description** to just the two scroll fixes, merge this as-is, and open a separate hotfix PR for #3 with the same urgency.

Lean (1) since the error/warning visibility regression is currently shipping in main and users won't see error messages until it's fixed.

CVEngineer66:
Just checked — error and warning render cases are already present in upstream/main's \`itemContent\` (they were part of the original #2056). The diff in this PR is intentionally minimal: only \`initialTopMostItemIndex\` and JumpBar \`onScrollToTurn\`. Error/warning visibility is not regressed on current \`main\`.

esengine (approved):
Re-reviewed — the meaningful regression I flagged is fixed: `itemContent` now renders `m.kind === "error"` (recoverable → warning label, else error, with a dismiss button wired to `dismiss_error`) and `m.kind === "warning"`, so the `$error`/$warning events that get pushed into `state.messages` are visible again instead of silently dropped. Combined with the two scroll fixes (initialTopMostItemIndex + JumpBar callback) this is the right scope. CI green. Merging.

PR #2089, opened 2026-05-28
fix(shell): demote allowlisted commands when path args escape workspace (#2081)

### LINKED ISSUE #2081: 越过沙箱权限，但是没有询问我

```
✓ run_command find . -name "xxxxxx*" 2>/dev… · 0.03秒                                                                                                                                                                                                     
$ find . -name "xxxxxx*" 2>/dev/null                                                                                                                                                                                                                      
[exit 0]                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                     
◆ 推理中 · 16 tok · 1 ¶ · 0.5s  v4-flash                                                                                                                                                                                                                             
↳ Not found in the current directory. Let me search more broadly.                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                     
✓ run_command find / -name "xxxxxx*" 2>/dev… · 55.82秒                                                                                                                                                                                                    
$ find / -name "xxxxxx*" 2>/dev/null | head -20
[exit ?]                                  
```

**What happened**
他在搜索文件，第一次在我的工作目录下搜索，没有命中，就去搜索/

**Expected**
应该向我询问是否搜索/，或者直接拒绝，这已经越过了沙箱的权限

**Reproduction**
呃....

**Environment**
- Reasonix version (`reasonix --version`):v0.53.2
- Node version (`node --version`):v24.1.0
- OS (Windows 11 / macOS 14 / Ubuntu 24.04 / …):macos 26.5
- Shell (bash, zsh, fish, PowerShell 7, PowerShell 5.1, cmd, …):bash
- Terminal app (Windows Terminal, iTerm2, Alacritty, kitty, WezTerm, **VSCode integrated**, **Cursor integrated**, Hyper, …):
- DeepSeek model (e.g. `deepseek-v4-flash`, `deepseek-v4-pro`):

**Logs / transcript**

zhangyapu1:
## Problem

Commands like `find / -name "xxx"` bypassed the sandbox because path arguments were never checked against the workspace boundary. Only redirect targets (`>`/`>>`) and sensitive-path prefixes (`~/.ssh`, `.env`) were validated.

This allowed the agent to silently scan the entire filesystem (e.g. `find /` running for 55 seconds) without user confirmation.

## Root Cause

`find` is in `BUILTIN_ALLOWLIST`, and the sandbox only checks:
1. File redirect targets must be inside workspace (`ensureUnderSandbox` in shell-chain.ts)
2. Sensitive path arguments (`~/.ssh`, `/etc/shadow`, etc.)

It does NOT check whether command arguments point outside the workspace.

## Fix

Add a workspace-escape check in `isAllowed`: when `projectRoot` is provided, any non-flag argument that resolves to an absolute path outside the project root demotes the command back to the confirm gate (user approval required).

## Affected patterns

| Command | Before | After |
|---|---|---|
| `find / -name "xxx"` | auto-run | requires confirmation |
| `find . -name "*.ts"` | auto-run | auto-run |
| `tree /etc` | auto-run | requires confirmation |
| `tree src` | auto-run | auto-run |
| `grep foo /etc/passwd` | auto-run | requires confirmation |
| `cat README.md` | auto-run | auto-run |

## Backward compatibility

- When `projectRoot` is not provided, behavior is unchanged
- Workspace-relative commands (`.`, `src/`, etc.) remain auto-allowed
- Flags (`-name`, `--color`), URLs, and env vars are correctly skipped

Closes #2081

### DIFF
--- a/src/tools/shell/parse.ts
+++ b/src/tools/shell/parse.ts
@@ -341,6 +341,23 @@ export function isAllowed(
       )
     )
       return false;
+    // Issue #2081 — demote when a path-like argument resolves outside the
+    // workspace.  Prevents allowlisted commands (find, tree, grep, cat …)
+    // from silently scanning the entire filesystem (e.g. `find / -name x`).
+    // Relative paths are resolved against projectRoot, so `find .` stays
+    // allowed; only paths that escape the workspace trigger the confirm gate.
+    if (projectRoot) {
+      const root = pathMod.resolve(projectRoot);
+      for (const tok of argv) {
+        if (!tok || tok.startsWith("-") || tok.includes("://") || tok.startsWith("$")) continue;
+        let expanded = tok;
+        if (expanded.startsWith("~")) expanded = pathMod.join(homedir(), expanded.slice(1));
+        const resolved = pathMod.resolve(root, expanded);
+        if (pathMod.isAbsolute(resolved) && !pathIsUnder(resolved, root)) {
+          return false;
+        }
+      }
+    }
     return true;
   }
   return false;
--- a/tests/shell-tools.test.ts
+++ b/tests/shell-tools.test.ts
@@ -355,10 +355,13 @@ describe("isAllowed", () => {
     });
 
     it("accepts user-configured extra prefixes", () => {
-      expect(isAllowed("cat /opt/secrets/config.yml", [], projectRoot)).toBe(true);
+      // /opt/secrets is outside workspace — workspace-escape check (#2081) demotes it even without sensitive-prefix config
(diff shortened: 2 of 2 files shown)


esengine (approved):
Real fix — `find /` walking the entire filesystem is exactly the gap this needs to close. Tests cover positive / negative / backward-compat / flags / chained commands; CI is green. Minor: there's some overlap with `resolveSensitivePath`'s token-skip logic that could be factored in a follow-up, but not worth blocking this on. Thanks.

PR #2091, opened 2026-05-28
feat(permissions): global shell allowlist that cascades with per-project (#2059)

### LINKED ISSUE #2059: Support global allowlist
**Problem**
According to reasonix itself (couldn't find reference in docs), there's only per-project allowlist of commands (such as mkdir)

**Proposed change**
Use a cascade of config files, like claude, codex and opencode do (global first, local 2nd)

**Scope check**
- [x ] This belongs in core reasonix (not better as a separate npm package)
- [ x] I've read CLAUDE.md and CONTRIBUTING.md

nianyi778:
## Problem

Closes #2059. Shell command allowlists were **per-project only** (`config.projects[<root>].shellAllowed`). There was no way to auto-approve a command (e.g. `brew install`, `cargo build`) once and have it apply everywhere — every new project re-prompted. The request: a cascade like Claude/Codex/OpenCode (global ∪ local).

## Change

Add a top-level **`shellAllowedGlobal`** config list, merged (union) with the per-project list at the single point where the shell tool resolves its extra allowlist:

```ts
// src/code/setup.ts
extraAllowed: () => [...new Set([...loadGlobalShellAllowed(), ...loadProjectShellAllowed(root)])]
```

Manage it via a **`--global`** (or **`-g`**) flag on the existing `/permissions` subcommands (works in chat mode too — global entries need no project root):

```
/permissions add --global <prefix>
/permissions remove --global <prefix-or-N>
/permissions clear --global confirm
/permissions list                          # now shows a Global section above Project
```

The flag is parsed **as a flag**, not as a positional keyword, so command prefixes like `g` (a popular `git` alias) or `global` stay manageable from the slash UI.

- Config accessors `load/add/remove/clearGlobalShellAllowed` mirror the project ones (same dedup, exact-match removal, atomic write).
- `shellAllowedGlobal` is registered in `STRING_ARRAY_FIELDS` so a malformed hand-edited value gets sanitized on read instead of crashing the merge.
- i18n: 5 new keys added to EN / zh-CN / ja / de (ru falls back to EN).
- The builtin allowlist precedence is unchanged; global is just another source of `extraAllowed`.

## Out of scope (follow-up)

The HTTP `/permissions` API + desktop settings panel still expose only the project list; global management there can be a follow-up. The `/permissions` slash command already covers CLI **and** desktop (command palette).

## Verification

- `tests/config.test.ts`: global CRUD (load/add/dedup/remove/clear) + independence from project lists.
- `tests/permissions-slash.test.ts`: `--global` add/remove/clear routes to the global list; `add g` adds `g` to the project list (not hijacked); `--global` works without a code root.
- `npm run verify` (build + lint + typecheck + full suite) passes.

### DIFF
--- a/src/cli/ui/slash/handlers/permissions.ts
+++ b/src/cli/ui/slash/handlers/permissions.ts
@@ -1,12 +1,29 @@
 import {
+  addGlobalShellAllowed,
   addProjectShellAllowed,
+  clearGlobalShellAllowed,
   clearProjectShellAllowed,
+  loadGlobalShellAllowed,
   loadProjectShellAllowed,
+  removeGlobalShellAllowed,
   removeProjectShellAllowed,
 } from "@/config.js";
 import { t } from "@/i18n/index.js";
 import { BUILTIN_ALLOWLIST } from "@/tools/shell.js";
 import type { SlashHandler } from "../dispatch.js";
+
+/** Leading `--global` / `-g` flag targets the cross-project list (#2059); otherwise per-project.
+ *  Only the option position (before the prefix) is consumed, so a prefix like
+ *  `git config --global user.name` keeps its inner `--global` intact, and `g` stays usable. */
+function takeScope(args: string[]): { global: boolean; rest: string[] } {
+  let global = false;
+  let i = 0;
+  while (i < args.length && (args[i] === "--global" || args[i] === "-g")) {
+    global = true;
+    i++;
+  }
+  return { global, rest: args.slice(i) };
+}
 
 const permissions: SlashHandler = (args, _loop, ctx) => {
   const sub = (args[0] ?? "").toLowerCase();
@@ -17,32 +34,38 @@ const permissions: SlashHandler = (args, _loop, ctx) => {
     return { info: renderListing(root, mode) };
   }
 
-  if (!root) {
+  const { global, rest } = takeScope(args.slice(1));
+  // Global entries live in the user config and need no project root; project ones require it.
+  if (!global && !root) {
     return { info: t("handlers.permissions.mutateCodeOnly") };
   }
 
   if (sub === "add") {
-    const prefix = args.slice(1).join(" ").trim();
+    const prefix = rest.join(" ").trim();
     if (!prefix) {
       return { info: t("handlers.permissions.addUsage") };
     }
-    const before = loadProjectShellAllowed(root);
-    if (before.includes(prefix)) {
+    const existing = global ? loadGlobalShellAllowed() : loadProjectShellAllowed(root!);
+    if (existing.includes(prefix)) {
       return { info: t("handlers.permissions.addAlready", { prefix }) };
     }
     if (BUILTIN_ALLOWLIST.includes(prefix)) {
       return { info: t("handlers.permissions.addBuiltin", { prefix }) };
     }
-    addProjectShellAllowed(root, prefix);
+    if (global) {
+      addGlobalShellAllowed(prefix);
+      return { info: t("handlers.permissions.addGlobalInfo", { prefix }) };
+    }
+    addProjectShellAllowed(root!, prefix);
     return { info: t("handlers.permissions.addInfo", { prefix }) };
   }
 
   if (sub === "remove" || sub === "rm" || sub === "delete") {
-    const target = args.slice(1).join(" ").trim();
+    const target = rest.join(" ").trim();
     if (!target) {
       return { info: t("handlers.permissions.removeUsage") };
     }
-    const existing = loadProjectShellAllowed(root);
+    const existing = global ? loadGlobalShellAllowed() : loadProjectShellAllowed(root!);
     let prefix: string | null = null;
     if (/^\d+$/.test(target)) {
       const idx = Number.parseInt(target, 10);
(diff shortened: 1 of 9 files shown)


esengine (changes requested):
Good direction — closes the right gap (#2059) and the cascade-at-`extraAllowed` site is the cleanest place to do this. Tests, i18n × 4, and the CRUD mirror are all solid. Two issues to fix before merge (Codex caught both; I agree with both):

1. **Sanitize `shellAllowedGlobal` on read.** The repo already has a registry for exactly this — `STRING_ARRAY_FIELDS` in `src/config.ts` (around L409), which `readConfig` runs through `sanitizeStringArrayField`. Please add `["shellAllowedGlobal"]` to that list. Otherwise a hand-edited config like `shellAllowedGlobal: "foo"` or `shellAllowedGlobal: {}` will round-trip as a non-array and crash `[...new Set([...loadGlobalShellAllowed(), ...])]` in `setup.ts` before any shell command can be checked.

2. **`g` / `global` scope keyword collides with real project entries.** `takeScope` treats the first arg as a scope keyword, so `/permissions add g` is interpreted as "add to global, empty prefix" and returns the usage hint — there's no way to add `g` (a popular git alias) to a project allowlist via the slash command. Same for any project entry whose first token is `global`. Either:
   - Use a flag: `/permissions add --global <prefix>` (parse before joining the rest), or
   - Require explicit scope syntax: `/permissions add @global <prefix>`.

   I lean `--global` — it's the convention readers already expect and won't collide with any plausible prefix.

Once those two are addressed, this is ready. Thanks for hooking into the existing renderListing / load* / add* / remove* / clear* shape — the symmetry made review easy.

nianyi778:
Both addressed in `625a81b8`, thanks for the precise pointers.

1. **Sanitization** — added `["shellAllowedGlobal"]` to `STRING_ARRAY_FIELDS` (`src/config.ts:412`). A malformed value now becomes `[]` on read instead of reaching the spread in `setup.ts`.
2. **Flag, not keyword** — `takeScope` now looks for `--global` / `-g` anywhere in the args and strips it; `/permissions add g` falls through as a normal project prefix. Updated the four i18n strings that printed the old `add global …` syntax. Three new tests cover `--global` add/remove/clear, `add g` going to project (not global), and `--global` working without a code root.

nianyi778:
@codex review

nianyi778:
Fixed in `a683c97d`. `takeScope` now only consumes `--global` / `-g` from the **leading option position** and stops at the first non-flag, so an inner `--global` is preserved verbatim:

- `/permissions add git config --global user.name` → adds the full prefix `git config --global user.name` to **project** (no scope hijack).
- `/permissions add --global npm install --global foo` → adds `npm install --global foo` to **global** (leading flag consumed, inner flag preserved).

Two new tests cover both cases.

nianyi778:
@codex review

nianyi778:
Fixed in `e77f6b4e`. Passed `opts.configPath` through to all three config calls inside `registerRooted`:
- `readConfig(opts.configPath)`
- `loadGlobalShellAllowed(opts.configPath)`
- `loadProjectShellAllowed(root, opts.configPath)`
- `loadEditMode(opts.configPath)` (the `allowAll` getter had the same gap)

@codex review

nianyi778:
@esengine 您好，您在 2026-05-28 02:16 提出的两条修改建议（`STRING_ARRAY_FIELDS` 净化注册 + `--global` flag 不劫持内部参数）已全部修完，共经历 3 轮迭代，当前 head `e77f6b4` 是最终状态。Codex 当天额度耗尽未能重审最新 commit，但之前几轮均通过。麻烦您抽空复审一下，感谢 🙏

esengine (approved):
Re-reviewed — both issues from my review are fixed exactly as discussed:
1. `["shellAllowedGlobal"]` is now in `STRING_ARRAY_FIELDS`, so a hand-edited non-array value gets sanitized on read instead of crashing the `[...new Set(...)]` cascade in setup.ts.
2. Scope is now a leading `--global`/`-g` flag via `takeScope` (not a positional keyword) — so `g` is addable to a project allowlist again, a non-leading `--global` (e.g. `git config --global user.name`) is preserved, and the i18n usage hints updated to `--global`. That's the option I leaned toward.
CI green. Merging — thanks for hooking into the existing field-registry + CRUD shape.

esengine:
Approving the content — both fixes are verified correct. But heads up: it now **conflicts with main** (recent merges touched `src/config.ts`/`setup.ts`). Please rebase onto current main (`git fetch origin main && git merge origin/main`) and I'll merge — the review stands, no further changes needed beyond the conflict resolution.

nianyi778:
Rebased onto current main (`81fd7872`) — resolved the conflict in `src/code/setup.ts` (kept both `autoGitRollback: {}` from HEAD and `opts.configPath` threading from this PR) and in `tests/config.test.ts` (kept both `promptHistory` and `globalShellAllowed` test blocks).

nianyi778:
Windows CI 失败的是 `auto-git-rollback.test.ts`，超时（5000ms），Ubuntu 构建通过 ✅。这是 Windows 上 git 操作偶发超时的 flaky test，与本 PR 的改动（config.ts / setup.ts / permissions.ts）无关。

Issue #2067, opened 2026-05-27
Bug: React Maximum update depth exceeded in Ink TUI


KyleLokK:
### Bug: React "Maximum update depth exceeded" in Ink TUI (useAnimationFrame + useBoxMetrics)

**Reasonix version**: 0.52.0 – 0.53.1
**Node**: v25.9.0
**OS**: Windows 11
**Terminal**: Windows Terminal / PowerShell 7

#### Describe the bug

The Ink TUI crashes with `Maximum update depth exceeded` in multiple scenarios:
1. When the agent returns multiple `edit_file` / `multi_edit` results in a single turn
2. During AI reasoning streaming
3. Immediately on startup in some configurations

All crashes produce the same React error but from different hooks.

#### Stack traces & version matrix

| Bug | Hook | v0.52.0 | v0.53.1 |
|-----|------|---------|---------|
| A | `useAnimationFrame` | ✅ crash confirmed | ✅ crash confirmed |
| B | Store `dispatch` | ✅ crash confirmed | ❌ not observed (likely refactored) |
| C | `useBoxMetrics` | ❌ not verified | ✅ crash confirmed |

**Bug A — useAnimationFrame**:
```
at dispatchSetState
at onChange                                     ← timer callback
at Timeout.tick                                 ← setInterval handler
at listOnTimeout
```

**Bug B — Store dispatch** (v0.52.0 only trace):
```
at forceStoreRerender
at dispatch
at Object.appendReasoning
at TurnTranslator.flushBuffers
```

**Bug C — useBoxMetrics** (v0.53.1 trace):
```
at dispatchSetState
at commitHookEffectListMount
at commitHookPassiveMountEffects
```

#### Root cause & fix

**Bug A**: `useAnimationFrame` (packages/ink/src/hooks/use-animation-frame.ts)
`ClockContext` provides an unstable object per render, which is included in the `useEffect` dependency array `[clock, intervalMs, active]`. This causes infinite subscribe/unsubscribe cycles. Additionally, every timer tick calls `setTime` which triggers a synchronous re-render in React legacy mode.

Fix: wrap `clock` in a ref, use `useRef` instead of `useState` for the time value to avoid re-renders, and remove `clock` from the deps array.

```diff
-  const [time, setTime] = useState(() => clock?.now() ?? 0);
+  const timeRef = useRef(clock?.now() ?? 0);
+  const clockRef = useRef(clock);
+  clockRef.current = clock;
[...]

PR #2085, opened 2026-05-28
feat: add contextTokens config for custom model context window (#2075, #2074)

zhangyapu1:
## Problem

Third-party models (mimo, qwen, etc.) fall back to DEFAULT_CONTEXT_TOKENS = 131_072 because DEEPSEEK_CONTEXT_TOKENS only hardcodes 4 DeepSeek models. This causes premature auto-compact at ~98K tokens on models with 1M+ windows.

- Fixes #2075 (mimo 上下文只有 131K)
- Fixes #2074 (mimo 上下文不能自动总结 — same root cause)

## Solution

Add a user-configurable contextTokens field to ~/.reasonix/config.json:

`json
{
  "contextTokens": {
    "mimo": 1000000,
    "qwen-72b": 131072
  }
}
`

Resolution priority: **user config → built-in table → 131K default**.

## Changes

| Layer | File | Change |
|-------|------|--------|
| Config | config.ts | New contextTokens field + loadContextTokens() |
| Resolver | 	elemetry/stats.ts | New 
esolveContextTokens(model) — user config > built-in > 131K |
| Context mgr | context-manager.ts | 3 ctxMax lookups → 
esolveContextTokens() |
| CLI status | StatusRow.tsx | ctxMax lookup updated |
| CLI observability | observability.ts | 2 ctxMax lookups updated |
| CLI ctx card | ctx-breakdown.tsx | ctxMax lookup updated |
| CLI App | App.tsx | 3 ctxMax lookups updated |
| Desktop protocol | protocol.ts | contextTokens in SettingsPatch |
| Desktop App | App.tsx | contextTokens in Settings type |
| Desktop settings | settings.tsx | Context-window size input below custom model ID |
| Desktop backend | desktop.ts | Handle contextTokens in settings_save + mitSettings |
| i18n | n.ts zh-CN.ts ja.ts de.ts | 3 new translation keys each |
| Tests | config.test.ts | 4 new loadContextTokens round-trip + validation cases |

PR #2095, opened 2026-05-28
fix(ink): scroll jitter on rapid mouse wheel (#2076)

zhangyapu1:
## Problem

Desktop Windows client text jitters when rapidly scrolling with mouse wheel (#2076).

**Root cause:** Two React hooks (`useBoxMetrics` and `useTerminalViewport`) had **no dependency arrays**, causing them to run expensive measurements on **every render**. During rapid wheel scrolling (10-20 events/100ms on Windows), each wheel event triggered 3-5 cascading renders, producing 30-80 renders/100ms and visible text jitter.

Additionally, `TerminalSizeContext.Provider` created a new object on every render, defeating any dependency-array checks in consumers.

## Fix (3 files, 19 insertions, 6 deletions)

| File | Change |
|------|--------|
| `packages/ink/src/index.ts` | `useBoxMetrics`: add `[ref.current]` deps — measurement only runs when element changes |
| `packages/ink/src/hooks/use-terminal-viewport.ts` | `useTerminalViewport`: add `[elementRef.current, terminalSize]` deps — skips expensive parent-chain walk on every render |
| `packages/ink/src/components/App.tsx` | Memoize `TerminalSizeContext.Provider` value — keeps reference stable so deps checks work |

## Render chain before fix

```
wheel event → ScrollBox.scrollBy() → scheduleRender
  → useBoxMetrics (no deps) → measure → setSize → re-render
  → useTerminalViewport (no deps) → walk parent chain → update isVisible → re-render
  → 3-5 renders per wheel event → 30-80 renders/100ms → jitter
```

## Verification

- All 3796 tests pass (291 test files, 0 failures)
- Zero lint/type errors
- Specifically verified `ink-update-depth-repro` and `cardstream-architecture` tests (related to #2067 regression)

### DIFF
--- a/packages/ink/src/components/App.tsx
+++ b/packages/ink/src/components/App.tsx
@@ -152,11 +152,24 @@ export default class App extends PureComponent<Props, State> {
   isRawModeSupported(): boolean {
     return this.props.stdin.isTTY;
   }
+  // Memoize terminal size context value to prevent unnecessary re-renders
+  // in consumers (useTerminalViewport). Without this, a new object is
+  // created on every render, defeating dependency-array checks.
+  private _lastTerminalColumns = 0;
+  private _lastTerminalRows = 0;
+  private _terminalSizeValue: { columns: number; rows: number } = { columns: 0, rows: 0 };
+
   override render() {
-    return <TerminalSizeContext.Provider value={{
-      columns: this.props.terminalColumns,
-      rows: this.props.terminalRows
-    }}>
+    if (this.props.terminalColumns !== this._lastTerminalColumns ||
+        this.props.terminalRows !== this._lastTerminalRows) {
+      this._lastTerminalColumns = this.props.terminalColumns;
+      this._lastTerminalRows = this.props.terminalRows;
+      this._terminalSizeValue = {
+        columns: this.props.terminalColumns,
+        rows: this.props.terminalRows,
+      };
+    }
+    return <TerminalSizeContext.Provider value={this._terminalSizeValue}>
         <AppContext.Provider value={{
         exit: this.handleExit
       }}>
--- a/packages/ink/src/hooks/use-terminal-viewport.ts
+++ b/packages/ink/src/hooks/use-terminal-viewport.ts
@@ -71,7 +71,7 @@ export function useTerminalViewport(): [
     if (visible !== entryRef.current.isVisible) {
       entryRef.current = { isVisible: visible }
     }
-  })
+  }, [elementRef.current, terminalSize])
 
   return [setElement, entryRef.current]
 }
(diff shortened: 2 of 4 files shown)


esengine (changes requested):
Two blockers — both need to be addressed before this can merge.

**1. Same scope creep as #2085.** The diff bundles the entire #2046 `maxIterPerTurn` patch (config loader, four call sites, `CacheFirstLoopOptions` field, loop iter guard, i18n × 4, tests). That belongs to #2046 only. Please drop all `maxIterPerTurn`-related hunks from this PR — they're not part of "fix(ink): scroll jitter" and they're now showing up in three of your open PRs (#2046, #2085, #2095) which makes each one harder to review independently.

**2. The ink fix uses a React anti-pattern that introduces correctness regressions.**

```ts
useEffect(() => { /* measure */ }, [ref.current])
```

`ref.current` is not a reactive value — React doesn't track ref mutations and the dep array is compared by `Object.is` against the previous render's captured value. In practice this means the effect runs once on mount (when the ref attaches) and then never again unless the conditional rendering swaps the DOM node. That's not what `useBoxMetrics` / `useTerminalViewport` need:

- **`useBoxMetrics`**: was deliberately running every render so the box re-measured when its parent's layout changed. With `[ref.current]` deps it captures the first measurement and never updates — so a parent resize will leave the child stuck on stale dimensions.
- **`useTerminalViewport`**: same issue. After the first parent-chain walk, scroll changes in any ancestor won't update `isVisible`.

The CI suite doesn't catch the regression because there's no test that resizes a parent after first render and asserts the child re-measures.

The direction is right — re-measuring every render on every wheel event is wasteful. But the right way to skip the work is one of:

- Throttle / debounce the re-render at the scroll source (so wheel bursts coalesce to one update).
- Use a real reactive dep (`[terminalSize]` is fine for `useTerminalViewport`; `useBoxMetrics` needs something that actually changes when the layout changes — a ResizeObserver wired through state is the canonical fix).
- Move the work to `useLayoutEffect` only when a known layout-affecting dep changes.

The `App.tsx` memoization of `TerminalSizeContext.Provider` value (the class-property cache) is fine in principle — keeping that part is OK once the deps issue is fixed.

Please:
1. Drop the `maxIterPerTurn` hunks.
2. Rework the deps — don't put `ref.current` in a dep array, ever; pick one of the alternatives above.
3. Add a regression test that resizes a parent after first render and asserts `useBoxMetrics` reports the new dimensions, so the next person who touches this can't reintroduce the bug.

esengine (approved):
Re-reviewed — both blockers resolved. (1) The #2046 `maxIterPerTurn` scope creep is fully dropped (0 refs). (2) The `ref.current`-in-deps anti-pattern is reworked: the effect now keys off `[terminalSize]` (a real reactive value, exactly as suggested) with a 16ms throttle guard via `lastMeasureTimeRef` (used as a throttle, not a dep) — the only `ref.current` mentions left are explanatory comments warning against the anti-pattern and a return value, not dependency arrays. CI green. Nice rework — merging.

PR #2112, opened 2026-05-28
fix: SkillStore.readEntry now handles symlinked skill directories

### LINKED ISSUE #2104: SkillStore.readEntry ignores symlinked directories
## Problem

## Reproduction

```bash
mkdir -p /tmp/test-skill
---
name: test-skill
description: A test skill
---

# Test
EOF

mkdir -p ~/.agents/skills

```

## Verification

```js
```

## Impact

## Related

## Suggested fix

```typescript
  // treat as directory
}
```

Or more compactly:

```typescript
if (isDir) { ... }
```

icexin:
## Problem

\`SkillStore.readEntry()\` uses \`entry.isDirectory()\` to detect skill directories inside a skill root. However, \`Dirent.isDirectory()\` returns \`false\` for symlinks to directories — so skills installed via symlinks are silently ignored with no error or warning. Users who maintain a central skill repository and symlink entries into \`~/.agents/skills/\` will have zero skills loaded.

Fixes #2104

## Change

Added \`symlinkPointsToDirectory()\` helper in \`src/skills.ts\` that:
1. Checks \`entry.isSymbolicLink()\` first (fast path)
2. Follows the symlink with \`statSync()\` to verify the target is a directory
3. Broken symlinks are caught by \`try/catch\` and return \`false\` (skipped silently)

The \`readEntry()\` condition was updated from \`entry.isDirectory()\` to \`entry.isDirectory() || symlinkPointsToDirectory(dir, entry)\` — no other code paths needed changes since the rest of the entry processing (\`existsSync\`, \`readFileSync\`) transparently follows symlinks.

### DIFF
--- a/src/skills.ts
+++ b/src/skills.ts
@@ -93,6 +93,16 @@ function parseAllowedTools(raw: string | undefined): readonly string[] | undefin
   return names.length > 0 ? Object.freeze(names) : undefined;
 }
 
+/** Check if a Dirent is a symlink whose target (following the link) is a directory. Broken symlinks return false. */
+function symlinkPointsToDirectory(dir: string, entry: import("node:fs").Dirent): boolean {
+  if (!entry.isSymbolicLink()) return false;
+  try {
+    return statSync(join(dir, entry.name)).isDirectory();
+  } catch {
+    return false;
+  }
+}
+
 /** flash/pro preset → concrete deepseek model id. Kept local so this file doesn't import the CLI preset bundle. */
 function subagentModelForPreset(preset: "flash" | "pro"): string {
   return preset === "pro" ? "deepseek-v4-pro" : "deepseek-v4-flash";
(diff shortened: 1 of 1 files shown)


esengine:
Thanks for the fix and the clear writeup — but this is now superseded. #2137 already landed the symlinked-directory fix for #2104 on `main`:

```ts
// src/skills.ts:252
const isDir =
  entry.isDirectory() || (entry.isSymbolicLink() && this.isSymlinkDirectory(dir, entry.name));
```

That covers exactly the same directory case as your `symlinkPointsToDirectory` helper and nothing extra, so this is now a pure duplicate (and conflicts with `main`). Closing as already-fixed.

One gap that's still open: symlinked **flat `<name>.md`** files — `main` line 260 still uses a bare `entry.isFile()`, which is false for symlinks. That's being tracked in #2124. Thanks again for reporting it.

PR #2119, opened 2026-05-28
feat(lifecycle): suggest plan rails for high-risk tasks

GTC2080:
## Summary

- add a conservative, host-side detector for high-risk engineering prompts that are likely to benefit from strict plan rails
- surface a UI-only suggestion to use `/plan strict` when lifecycle is off and plan mode is inactive
- cover positive and false-positive-sensitive English and Chinese prompt corpus

## Notes

This does not auto-arm strict lifecycle, block the user, change runtime state, or modify prompt/tool schema/prefix behavior. Explicit plan-first requests still use the existing auto-enter path from #2070.

## Validation

- `npm test -- tests/lifecycle-plan-suggestion.test.ts`
- `npm test -- tests/lifecycle-explicit-intent.test.ts tests/lifecycle-plan-suggestion.test.ts tests/code-prompt.test.ts tests/lifecycle.test.ts tests/slash.test.ts`
- `npx biome check src/cli/ui/lifecycle-plan-suggestion.ts tests/lifecycle-plan-suggestion.test.ts src/cli/ui/App.tsx src/i18n/EN.ts src/i18n/zh-CN.ts src/i18n/JA.ts src/i18n/types.ts`
- `npm run typecheck`
- `npm run lint`
- `npm run verify`
- pre-push `npm run verify`

### DIFF
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -176,6 +176,7 @@ import { StaticCardStream } from "./layout/StaticCardStream.js";
 import { StatusRow } from "./layout/StatusRow.js";
 import type { StatusBarConfig } from "./layout/StatusRow.js";
 import { shouldEnterPlanModeForExplicitIntent } from "./lifecycle-explicit-intent.js";
+import { shouldSuggestPlanForEngineeringIntent } from "./lifecycle-plan-suggestion.js";
 import { formatLoopStatus } from "./loop.js";
 import { applyMcpAppend } from "./mcp-append.js";
 import { handleMcpBrowseSlash } from "./mcp-browse.js";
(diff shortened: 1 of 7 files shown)


esengine (approved):
LGTM. The detector is appropriately conservative (UI hint only, no state change), and the test corpus is thoughtful — the Chinese false-positive cases (装修/必修/维修 vs 重构/迁移) are exactly the homograph traps a naive regex would trip on. The App.tsx refactor preserves the explicit-intent auto-arm path unchanged, and de/ru correctly fall back to the EN string via the `...EN` spread.

Non-blocking follow-up: the hint re-fires on every matching prompt while lifecycle is off (unlike the explicit path, which arms plan mode and self-silences). If it proves noisy in practice, consider showing it once per session. Not a blocker — merging.

Issue #1464, opened 2026-05-21
test(desktop): establish proper test environment for desktop/src/ imports in CI


paradoxSCH:
### Problem

The root `npm run verify` includes `desktop/src/**/*.test.*` patterns in `vitest.config.ts`, but CI only runs `npm ci` in the root directory. Desktop is **not** in root workspaces (`workspaces: ["packages/*"]`), so `desktop/node_modules` is never created in CI.

This means any test that imports from `desktop/src/` source files will fail because those files transitively depend on Desktop-only packages (`lucide-react`, `@tauri-apps/api`, `react-markdown`, `rehype-katex`, etc.) that don't exist in root `node_modules`.

The current workaround (PR #1443) uses `vi.mock()` for local module boundaries (`./CommandPalette`, `./Markdown`, `./cards`) and `resolve.alias` in `vitest.config.ts` for external deps. This works but is a testing infrastructure debt.

### Why this matters

- Mocking local modules is a test smell; tests should ideally load real source code
- The alias + mock approach requires maintaining a growing list of Desktop-only dependencies in `tests/mocks/`
- Every new test that imports from `desktop/src/` risks hitting a new "missing module" failure in CI
- `desktop/package.json` uses `workspace:*` for `@reasonix/core-utils`, which `npm install` in the Desktop directory cannot resolve (Desktop is not a workspace member)

### Options

**Option A: Extract pure-logic modules**
Move `reduce`/`applyIncoming` from `App.tsx` to `App.state.ts`, and approval cards from `thread.tsx` to `approval-cards.tsx`, so tests import files with no Desktop-only transitive deps.

**Option B: Install Desktop deps in CI**
Update `.github/workflows/ci.yml` to also install Desktop dependencies. Requires handling `workspace:*` protocol (e.g., temporarily replacing with `file:` path or symlinking `@reasonix/core-utils` before `npm install`).

**Option C: Remove Desktop tests from root runner**
Give Desktop its own `vitest.config.ts` and test script, run separately in CI. Requires Desktop to have its own node_modules installed.

### Recommendation

Option A (extract pure logic) is the cleanest long-term fix. The `reduce` function and `State`/`Action` types in `App.tsx` are already pure functions with no UI dependencies — they just happen to live in a file that imports Tauri APIs. Extracting them aligns with Redux/Flux patterns and eliminates the need for any mocks in reducer tests.

[...]

PR #2124, opened 2026-05-28
fix(skills): follow symlinked entries in SkillStore.readEntry (#2104)

### LINKED ISSUE #2104: SkillStore.readEntry ignores symlinked directories

nianyi778:
## Problem

Closes #2104. `SkillStore.readEntry()` classified each entry with `Dirent.isDirectory()` / `isFile()`. Both return **false** for symlinks even when the target is a directory or file — so a `skills.sh`-style `~/.agents/skills/` layout that symlinks every skill into a central repo (e.g. `~/.sra/<skill> → ~/.agents/skills/<skill>`) silently saw **zero entries**, with no error or warning.

The reporter's case: 22/22 skills under `~/.agents/skills/` were symlinks, all silently ignored.

## Change

`src/skills.ts` `readEntry` now resolves symlinks via `statSync` before the type branch:

```ts
let isDir = entry.isDirectory();
let isFile = entry.isFile();
if (entry.isSymbolicLink()) {
  try {
    const target = statSync(join(dir, entry.name));
    isDir = target.isDirectory();
    isFile = target.isFile();
  } catch {
    return null; // broken symlink — skip silently rather than abort the scan
  }
}
```

So a symlinked **directory** containing `SKILL.md` and a symlinked **flat `<name>.md`** are both discovered exactly like the real entries. Broken symlinks no longer crash the scan; they're just skipped (parity with how the rest of `list()` handles unreadable roots).

## Out of scope

The issue mentions #1464 (same root cause for the @-mention directory listing). That's a separate code path; happy to tackle it in a follow-up if useful, but keeping this PR focused.

## Verification

Three new `tests/skills.test.ts` cases (using `fs.symlinkSync`):

- symlinked **directory** containing `SKILL.md` shows up in `list()`,
- symlinked **flat `<name>.md`** shows up,
- a **broken symlink** is dropped without crashing — sibling real skills still load.

`npm run verify` (build + lint + typecheck + full suite) passes.

### DIFF
--- a/src/skills.ts
+++ b/src/skills.ts
@@ -251,13 +251,17 @@ export class SkillStore {
   private readEntry(dir: string, scope: SkillScope, entry: import("node:fs").Dirent): Skill | null {
     const isDir =
       entry.isDirectory() || (entry.isSymbolicLink() && this.isSymlinkDirectory(dir, entry.name));
+    // Symlinked flat `<name>.md` files: `entry.isFile()` returns false for symlinks.
+    // Reuse the same `statSync` pattern as `isSymlinkDirectory` (#2104).
+    const isFile =
+      entry.isFile() || (entry.isSymbolicLink() && !isDir && this.isSymlinkFile(dir, entry.name));
     if (isDir) {
       if (!isValidSkillName(entry.name)) return null;
       const file = join(dir, entry.name, SKILL_FILE);
       if (!existsSync(file)) return null;
       return this.parse(file, entry.name, scope);
     }
-    if (entry.isFile() && entry.name.endsWith(".md")) {
+    if (isFile && entry.name.endsWith(".md")) {
       const stem = entry.name.slice(0, -3);
       if (!isValidSkillName(stem)) return null;
       return this.parse(join(dir, entry.name), stem, scope);
@@ -269,6 +273,15 @@ export class SkillStore {
   private isSymlinkDirectory(parentDir: string, name: string): boolean {
     try {
       return statSync(join(parentDir, name)).isDirectory();
+    } catch {
+      return false;
+    }
+  }
+
+  /** Check if a symlink points to a file. Returns false for broken symlinks. */
+  private isSymlinkFile(parentDir: string, name: string): boolean {
+    try {
+      return statSync(join(parentDir, name)).isFile();
     } catch {
       return false;
     }
--- a/tests/skills.test.ts
+++ b/tests/skills.test.ts
@@ -133,6 +133,42 @@ describe("SkillStore", () => {
     expect(names).toContain("review");
   });
 
+  // Skipped on Windows — file symlink creation also requires Developer Mode / admin.
+  it.skipIf(process.platform === "win32")("follows symlinked flat <name>.md skill files", () => {
+    const realRoot = mkdtempSync(join(tmpdir(), "reasonix-skills-real-md-"));
+    try {
+      const realFile = join(realRoot, "shipit.md");
+      writeFileSync(realFile, "---\ndescription: flat via symlink\n---\nbody\n", "utf8");
+      const scannedRoot = join(home, ".reasonix", "skills");
+      mkdirSync(scannedRoot, { recursive: true });
+      symlinkSync(realFile, join(scannedRoot, "shipit.md"), "file");
+
+      const skills = new SkillStore({ homeDir: home, projectRoot, disableBuiltins: true }).list();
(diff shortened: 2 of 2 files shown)


nianyi778:
Pushed `02467c46` with two self-review fixes:

- **Windows CI**: `symlinkSync` now passes the `'dir'` / `'file'` type arg (Windows defaults to `'file'` even for directory targets and would produce a broken link — ignored on POSIX). The broken-symlink test is gated with `it.skipIf(process.platform === 'win32')` because creating a nonexistent-target symlink on Windows throws `EPERM` without Developer Mode / admin — unrelated to the `readEntry` behavior under test.
- **Tmpdir cleanup**: the two tests that create an out-of-scope `realRoot` now wrap the body in `try { … } finally { rmSync(realRoot, …) }` so a failing assertion doesn't leak `/tmp/reasonix-skills-real-*`.

Out of scope (kept the PR focused): `src/server/api/skills.ts` `listSkills` has the same Dirent-blind-to-symlinks bug. CLI now discovers symlinked skills (this PR) while the dashboard's API still misses them, so the two listings disagree until that's mirrored. Happy to tackle it in a follow-up — let me know if you'd rather it come in here.

@codex review

esengine (changes requested):
Good catch on the root cause, and the broken-symlink test is the right instinct. But the landscape shifted: #2137 (merged) already shipped the symlinked-**directory** half of this on `main`, via `isSymlinkDirectory` (src/skills.ts:252). So this PR now conflicts and the directory branch duplicates what's already there.

The part that's still genuinely missing on `main` and worth keeping: symlinked **flat `<name>.md`** files. Line 260 still does a bare `entry.isFile()`, which is false for a symlink, so flat-file symlinks are still dropped.

Please:
1. Rebase onto current `main`.
2. Drop the now-redundant directory logic — reuse the existing `isSymlinkDirectory` helper rather than reintroducing the `let isDir = ...` block.
3. Keep just the flat-`.md` symlink resolution + the broken-symlink skip.
4. Trim the tests to the two that aren't already on `main`: the flat-`.md` symlink case and the broken-symlink case (the symlinked-directory test already landed with #2137 — drop the duplicate to avoid a name clash).

The flat-`.md` fix is small but real, so it's worth carrying across the rebase rather than closing.

nianyi778:
Done — rebased onto current `main` (`47e4deb6`).

- Dropped the now-redundant directory symlink logic — the directory branch now uses the existing `isSymlinkDirectory` helper from `main` (#2137) untouched.
- Added a parallel `isSymlinkFile` private method (same pattern: `statSync` → `.isFile()`, returns `false` for broken symlinks) and wires it into `readEntry` for the flat `<name>.md` branch.
- Dropped the symlinked-directory test (already on `main` via #2137); kept the two that are genuinely new: symlinked flat `.md` file and broken-symlink skip.

@codex review

nianyi778:
Fixed in `1e706dc9` — added `it.skipIf(process.platform === 'win32')` to the flat-symlink test, matching the same guard already on the broken-symlink test below it.

@codex review

nianyi778:
Note on the CI failures: the 3 failing tests (`slash.test.ts`, `ui-slash-suggestions.test.tsx`, `auto-git-rollback.test.ts`) are also failing on `main` itself right now — they were broken by the `#2173` unified-diff PR and the Telegram PR merged after my rebase. My changes (`src/skills.ts` + `tests/skills.test.ts`) don't touch any of those files.

Verified: all skills tests still pass on my branch; the 3 failing tests fail identically on the current `main` HEAD (`8131e241`).

esengine (approved):
Re-reviewed — exactly the trim + rebase I asked for. It now reuses the merged `isSymlinkDirectory` (#2137) for the directory case and adds *only* the missing piece: symlinked flat `<name>.md` files, via a parallel `isSymlinkFile` helper, with `isFile = entry.isFile() || (entry.isSymbolicLink() && !isDir && this.isSymlinkFile(...))`. The `!isDir` guard correctly prevents a dir-symlink from also matching the file branch. No duplication of what's already on main, tests cover it, CLEAN + CI green. This was the genuinely-missing case (main's bare `entry.isFile()` skips flat-md symlinks). Merging.

PR #2127, opened 2026-05-28
fix(desktop): auto-switch to new session after deleting active session

### LINKED ISSUE #2126: 桌面端：删除会话后界面仍停留在已删除会话，应自动切换到新会话
## 问题描述

在桌面端侧边栏中删除一个会话后，若被删除的会话恰好是当前正在查看的活跃会话，界面不会自动跳转到新会话，而是停留在已删除会话的空白/残留视图上。

## 复现步骤

1. 打开 Reasonix 桌面端，进入任意工作区
2. 在侧边栏中点击某个非当前会话，切换到该会话
3. 将鼠标悬停在该会话行上，点击出现的 X 删除按钮
4. 点击确认弹窗中的 Confirm

**预期行为**：界面自动切换到一个新的空会话（或剩余会话列表中第一个会话）
**实际行为**：界面仍停留在已删除会话的视图上，消息区域显示旧内容或空白

## 根因

后端 \`session_delete\` RPC 处理器（\`src/cli/commands/desktop.ts\`）在删除会话后仅刷新了会话列表（\`emitSessions\`），未检查被删除会话是否为当前活跃会话，也未通知前端切换。

## 影响范围

桌面端所有用户 — 每次删除活跃会话后都需要手动点击侧边栏切换到其他会话。

## 相关 PR

- #2127

AtmoOmen:
## Summary

When a user deletes the currently-active session via the sidebar, the UI remained stuck on the now-deleted session instead of navigating to a fresh session. This PR fixes the backend \`session_delete\` RPC handler to detect when the deleted session is the active one and automatically switch to a new session.

Closes #2126.

## Changes

**\`src/cli/commands/desktop.ts\`** — \`session_delete\` handler:

| Before | After |
|---|---|
| Delete file, emit \`\$sessions\` | If deleted session is active: abort turn, mint new session via \`startNewChatInTab\`, emit \`\$session_loaded\` + \`\$sessions\`. Otherwise: emit \`\$sessions\` only |

## Root Cause

The \`session_delete\` handler only did two things:
1. \`deleteSession(msg.name)\` — remove the file from disk
2. \`emitSessions(tab)\` — refresh the sidebar list

It never checked whether \`tab.currentSession === msg.name\`, so when the active session was deleted the frontend's \`currentSession\` state remained pointing to the now-deleted session. The \`\$sessions\` event only updates the sidebar list — it does **not** change which session is displayed.

## Fix

When the deleted session matches \`tab.currentSession\`:
1. \`startNewChatInTab(tab)\` — aborts any in-flight turn, mints a fresh session for the workspace, rebuilds the runtime, and calls \`emitSessions(tab)\`
2. Emit \`\$session_loaded\` with empty messages — this is what tells the frontend reducer to set \`currentSession\` and clear the message list

When the deleted session is **not** the active one, behavior is unchanged (\`emitSessions\` only).

## Verification

- \`npm run verify\` passes — 295 test files pass, 4 pre-existing failures (web-tools Exa API key / desktop Ink rendering tests) unrelated to this change
- Desktop debug build (\`tauri build --debug\`) compiles and produces installers successfully

### DIFF
--- a/src/cli/commands/desktop.ts
+++ b/src/cli/commands/desktop.ts
@@ -2557,7 +2557,25 @@ export async function desktopCommand(opts: DesktopOptions): Promise<void> {
     }
     if (msg.cmd === "session_delete") {
       deleteSession(msg.name);
-      emitSessions(tab);
+      if (tab.currentSession === msg.name) {
+        startNewChatInTab(tab);
+        emit(
+          {
+            type: "$session_loaded",
+            name: tab.currentSession,
+            messages: [],
+            carryover: {
+              totalCostUsd: 0,
+              cacheHitTokens: 0,
+              cacheMissTokens: 0,
+              totalCompletionTokens: 0,
+            },
+          },
+          tab.id,
+        );
+      } else {
+        emitSessions(tab);
+      }
       return;
     }
     if (msg.cmd === "session_rename") {
--- a/src/cli/commands/desktop.ts
+++ b/src/cli/commands/desktop.ts
@@ -2573,6 +2573,7 @@ export async function desktopCommand(opts: DesktopOptions): Promise<void> {
           },
           tab.id,
         );
+        emitCtxBreakdown(tab);
       } else {
         emitSessions(tab);
       }


esengine (changes requested):
The core fix is correct and the approach is right — minting a new session and pushing it to the frontend via `$session_loaded` is exactly how session state should be propagated (the frontend reducer keys off `$session_loaded` to set `currentSession` + clear messages), and `startNewChatInTab` already calls `emitSessions`, so the sidebar refresh is covered without the `else` branch's `emitSessions` double-firing. Zero carryover for a fresh session is correct.

One small consistency gap to close before merge: the canonical `session_load` path emits `$session_loaded` **and then** `emitCtxBreakdown(tab)` (src/cli/commands/desktop.ts:871). This branch only emits `$session_loaded`. On the frontend, the `$session_loaded` reducer resets usage via `{ ...zeroUsage(), ... }`, which zeroes `reservedTokens` (desktop/src/App.tsx:1051) — so after deleting the active session the context-usage bar reads 0 until the next turn fires a fresh `$ctx_breakdown`, instead of showing the new session's system+tools baseline.

Please add `emitCtxBreakdown(tab);` right after the `emit({ type: "$session_loaded", ... })` in the active-session branch to match `session_load`. Otherwise this is good to go.

AtmoOmen:
thanks for review. done now

esengine (approved):
Re-reviewed — `emitCtxBreakdown(tab)` is now emitted right after the `$session_loaded` in the active-session branch, matching the canonical `session_load` path, so the context bar no longer reads 0 after deleting the active session. CI green. Thanks for the quick fix — merging.

PR #2130, opened 2026-05-28
fix(ink): reduce VS Code terminal lag by throttling drain and debouncing damage (#2080)

### LINKED ISSUE #2080: 在vs code内部终端中使用reasonix 卡顿明显
在vs code内部终端中使用reasonix 卡顿明显。codex就是正常的。可以自行试试。

为什么我要在vs code里面用？因为vscode可以开启code tunnal 远程开发。

zhangyapu1:
## Summary

Fixes #2080 — VS Code integrated terminal (xterm.js) exhibits noticeable lag when using Reasonix. The root cause is Ink's rendering pipeline simultaneously hitting four xterm.js performance bottlenecks:

1. **Scroll drain at 4ms (~250fps)** — overwhelms xterm.js single-threaded ANSI parser
2. **Full-screen damage every frame** during selection/highlight — forces complete cell rescan
3. **Microtask + throttle double-fire** — leading+trailing edges both queue a microtask ~0.1ms apart
4. **No xterm.js-aware frame rate adaptation**

## Changes

All changes are in `packages/ink/src/ink.tsx` (1 file, +33 / -8 lines):

| Fix | What | Impact |
|-----|------|--------|
| Drain interval 4ms → 16ms | `FRAME_INTERVAL_MS >> 2` → `FRAME_INTERVAL_MS` | **-60% frame pressure** during scroll |
| Selection damage transition-only | New `prevSelActive`/`prevHlActive` tracking; full-screen damage only on activate/deactivate transitions | **-30-50% diff work** while selection/highlight continuously active |
| Microtask coalescing | New `renderMicrotaskPending` flag prevents duplicate `queueMicrotask` in same event-loop tick | **Eliminates double-render jank** |

## Why Codex doesn't lag (comparison)

| Factor | Codex | Reasonix (Ink) |
|--------|-------|----------------|
| Terminal type | Native CLI | xterm.js (web, single-threaded ANSI parsing) |
| Frame rate | ~30fps | 60fps + 4ms drain = 250fps potential |
| Per-frame output | Smaller | 8-10KB |
| Full-screen redraw | None | Every frame during selection/search |

## Test results

```
Test Files  291 passed (291)
Tests       3797 passed | 11 skipped (3808)
```

Zero regressions.

## Notes

- P2 (global frame rate reduction to 30fps for xterm.js) was not implemented — the drain fix covers the biggest bottleneck, and reducing global FPS would degrade native terminal UX
- The `renderMicrotaskPending` flag is a lightweight guard (one boolean check) with no impact on normal render paths


esengine (changes requested):
Nice diagnosis of the xterm.js bottleneck. Two of the three hunks are the real fix and look sound; the third is a no-op that I'd drop before merge.

**Keep — these are the actual #2080 fix:**
- **Drain interval 4ms → 16ms** (`FRAME_INTERVAL_MS >> 2` → `FRAME_INTERVAL_MS`). This is the headline win and the reasoning is right — running the full pipeline at ~250fps is what's drowning xterm.js's single-threaded parser. 60fps drain is well within budget and native terminals are unaffected.
- **Microtask coalescing** (`renderMicrotaskPending`). Correct: it only collapses duplicate microtasks queued *within the same tick* (the leading+trailing double-fire), and `onRender` reads the latest committed frame, so no render is lost across throttle windows. The flag is reset before `onRender` runs, so re-entrant scheduling still works.

**Drop — the selection/highlight `selTransition`/`hlTransition` change is a no-op:**
`prevFrameContaminated` is set to `selActive || hlActive` unconditionally at the end of every `onRender` (ink.tsx:746). So while a selection/highlight is continuously active, the *previous* frame's `selActive` keeps `prevFrameContaminated === true`, and the `|| this.prevFrameContaminated` term in the full-damage condition forces a full-screen damage every frame regardless. Work through the three cases (activate / steady-state / deactivate) and the new `selTransition`-based condition yields the **identical** damage decision as the old `selActive`-based one in each. So the claimed -30-50% diff reduction doesn't actually happen — the full rescan still runs every frame via `prevFrameContaminated` — and the two new fields + comment just add complexity for zero behavioral change. Please remove this hunk so the PR stays focused on the drain fix.

**If you do want to kill the steady-state full-rescan** (e.g. spinner ticking under a static text selection genuinely does force a wasteful full-damage today), that's a real optimization but it lives in `prevFrameContaminated`, not here — you'd need to force full damage only when the overlay content actually changes. Critically, that must still trigger on selection-**range** changes (drag-extend via `handleSelectionDrag` → `extendSelection`) and search current-match moves, because those write through `setCellStyleId` with no per-cell damage. That's worth its own PR with a regression test (drag-extend an active selection, assert the newly-covered cells repaint). Out of scope here.

Trim to the drain + microtask hunks and this is good to merge.

esengine (approved):
Re-reviewed — the no-op selection/highlight transition-damage hunk is dropped (and with it the `prevSelActive`/`prevHlActive` fields), leaving just the two real fixes: drain interval 4ms→16ms and the microtask-coalescing guard. That's exactly the trim I asked for. CI green. Merging.

PR #2132, opened 2026-05-28
fix: load skills from symlinked directories (#2104)

### LINKED ISSUE #2104: SkillStore.readEntry ignores symlinked directories

zhangyapu1:
## Problem

SkillStore.readEntry() uses ntry.isDirectory() to detect skill directories. However, Node.js Dirent.isDirectory() returns alse for symlinks, causing symlinked skill directories to be **silently skipped** with no error or warning.

Users who maintain skills in a shared location and symlink them into ~/.agents/skills/ (a common pattern, also the default install method for skills.sh-style tooling) will have **zero skills loaded**.

## Root Cause

In src/skills.ts, 
eadEntry():

`	ypescript
if (entry.isDirectory()) {  // false for symlinks
  // load skill...
}
`

eaddirSync with withFileTypes: true returns DT_LNK type for symlinks, and Dirent.isDirectory() does not follow symlinks.

## Fix

Added isSymbolicLink() check with statSync fallback to resolve symlinks and verify the target is a directory:

`	ypescript
const isDir =
  entry.isDirectory() || (entry.isSymbolicLink() && this.isSymlinkDirectory(dir, entry.name));
`

The isSymlinkDirectory() helper uses statSync (which follows symlinks) and safely returns alse for broken symlinks.

## Testing

- Added test: symlinked skill directory is correctly loaded
- Added test: broken symlinks are safely skipped without throwing
- All 299 test files pass (3853 tests)

Closes #2104

### DIFF
--- a/src/skills.ts
+++ b/src/skills.ts
@@ -249,7 +249,9 @@ export class SkillStore {
   }
 
   private readEntry(dir: string, scope: SkillScope, entry: import("node:fs").Dirent): Skill | null {
-    if (entry.isDirectory()) {
+    const isDir =
+      entry.isDirectory() || (entry.isSymbolicLink() && this.isSymlinkDirectory(dir, entry.name));
+    if (isDir) {
       if (!isValidSkillName(entry.name)) return null;
       const file = join(dir, entry.name, SKILL_FILE);
       if (!existsSync(file)) return null;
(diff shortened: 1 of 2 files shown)


zhangyapu1:
Closing as duplicate of #2137 (same fix, same files).

PR #2133, opened 2026-05-28
fix: batch store notifications to prevent React max update depth (#2123)

### LINKED ISSUE #2123: 流式输出时 React 无限重渲染导致崩溃（Maximum update depth exceeded）Maximum update depth exceeded in streaming loop (React  forceStoreRerender cascade)
### Environment

| 项目 | 值 |
|------|-----|
| Reasonix 版本 | 0.52.0（npm latest 0.53.2 同样存在） |
| Node.js | v22.13.0 |
| 操作系统 | Windows |
| React (bundle) | 19.2.6 + react-reconciler 0.32.0 |
| 触发命令 | `reasonix code`（普通模式未测试） |

### Steps to Reproduce

1. 在终端中运行 `reasonix code`
2. 输入任意需要推理的提示词（让模型产出较长的 reasoning + 响应）
3. 在流式输出过程中，错误抛出

### Error Stack

```
file:///G:/reasonix/dist/cli/chunk-A6GSOADP.js:11999
          throw nestedPassiveUpdateCount = nestedUpdateCount = 0, rootWithPassiveNestedUpdates = rootWithNestedUpdates = null, Error(
                                                                                                                               ^

Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
    at getRootForUpdatedFiber (file:///G:/reasonix/dist/cli/chunk-A6GSOADP.js:11999:128)
    at enqueueConcurrentRenderForLane (file:///G:/reasonix/dist/cli/chunk-A6GSOADP.js:11987:16)
    at forceStoreRerender (file:///G:/reasonix/dist/cli/chunk-A6GSOADP.js:12780:21)
    at file:///G:/reasonix/dist/cli/chunk-A6GSOADP.js:12766:43
    at dispatch (file:///G:/reasonix/dist/cli/chunk-227JLWAG.js:47953:46)
    at Object.appendReasoning (file:///G:/reasonix/dist/cli/chunk-227JLWAG.js:54500:31)
    at TurnTranslator.flushBuffers (file:///G:/reasonix/dist/cli/chunk-227JLWAG.js:58828:16)
    at flush (file:///G:/reasonix/dist/cli/chunk-227JLWAG.js:61973:20)
    at file:///G:/reasonix/dist/cli/chunk-227JLWAG.js:62053:13
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
```

### Root Cause Analysis

The bug originates from a missing throttling layer in the streaming render pipeline. Here is the call chain:

**1. Custom store dispatches synchronously to all listeners**

`chunk-227JLWAG.js:47942` — `createStore`:

```js
dispatch(event) {
  state2 = reduce(state2, event);
  for (const listener of stateListeners) listener();   // sync notify all
}
```

**2. React components subscribe via `useSyncExternalStore`**

`chunk-227JLWAG.js:47986` — `useAgentState`:

```js
function useAgentState(selector) {
  const store = useAgentStore();
  const subscribe = useCallback((cb) => store.subscribe(cb), [store]);
  const getSnapshot = useCallback(() => selector(store.getState()), [store, selector]);
  return useSyncExternalStore(subscribe, getSnapshot, getSnapshot);
}
```

Each `dispatch()` triggers `forceStoreRerender` (React SyncLane) for every subscribed component.

**3. Streaming loop flushes on every chunk**

`chunk-227JLWAG.js:62027` — the main `for await` loop:

```js
for await (const ev of loop2.step(modelInput)) {
  // ...
  } else if (ev.role === "assistant_delta") {
    if (ev.content) contentBuf.current += ev.content;
    if (ev.reasoningDelta) reasoningBuf.current += ev.reasoningDelta;
    flush();   // <-- every single chunk triggers dispatch + React re-render
  }
}
```

**4. No batching or throttling in `flush()`**

`chunk-227JLWAG.js:61971`:

```js
const flush = () => {
  if (!contentBuf.current && !reasoningBuf.current && !toolCallBuildBuf.current) return;
  translator.flushBuffers(reasoningBuf.current, contentBuf.current, loop2.model);
  // ... immediately dispatches to store
};
```

**The cascade:**

```
streaming chunk arrives
  → flush()
    → TurnTranslator.flushBuffers()
      → log.appendReasoning()
        → store.dispatch({ type: "reasoning.chunk", ... })
          → all useAgentState listeners notified
            → React forceStoreRerender (SyncLane)
              → component re-render starts
                → (next chunk arrives before render completes)
                  → another dispatch
                    → nestedUpdateCount++
                      → ... after >50 nested updates
                        → React throws "Maximum update depth exceeded"
```

When the model streams tokens rapidly (especially with heavy reasoning output), dozens of chunks arrive within a single microtask queue drain. Each one synchronously dispatches to the store while React is still processing the previous render. After ~50 nested `forceStoreRerender` calls, React's `getRootForUpdatedFiber` throws.

### Why it passes unnoticed in normal mode

In `reasonix code` mode, the code path enables additional features (`codeMode` flag) and may produce more frequent assistant_delta events (or larger reasoning blocks that trigger the edge case more easily). In theory, `reasonix` (non-code) could also trigger this with a sufficiently verbose streaming response.

### Suggested Fix

The `flush()` function should batch store updates, not dispatch on every individual chunk. Two possible approaches:

**Option A — RAF-based batching (simplest)**

```js
let flushScheduled = false;
let flushArgs = null;

const flush = () => {
  if (!contentBuf.current && !reasoningBuf.current && !toolCallBuildBuf.current) return;
  
  flushArgs = {
    reasoning: reasoningBuf.current,
    content: contentBuf.current,
    toolCall: toolCallBuildBuf.current,
    model: loop2.model
  };
  
  if (!flushScheduled) {
    flushScheduled = true;
    requestAnimationFrame(() => {
      flushScheduled = false;
      if (!flushArgs) return;
      translator.flushBuffers(flushArgs.reasoning, flushArgs.content, flushArgs.model);
      streamRef.text += flushArgs.content;
      streamRef.reasoning += flushArgs.reasoning;
      if (flushArgs.toolCall) streamRef.toolCallBuild = flushArgs.toolCall;
      contentBuf.current = "";
      reasoningBuf.current = "";
      toolCallBuildBuf.current = null;
      flushArgs = null;
    });
  }
};
```

**Option B — use React 19 `startTransition` (cleaner, but requires refactor)**

Wrap store dispatches inside `startTransition` so React treats them as low-priority updates and batches them naturally under concurrent rendering.

### Workaround (for affected users)

- Use `reasonix` (without `code` subcommand) as a temporary fallback
- Reduce the model's reasoning output verbosity
- Pin to a model that emits fewer/larger chunks vs many small ones

### Verification

The issue was confirmed present in both:
- **0.52.0** (local install)
- **0.53.2** (npm latest, verified via tarball inspection — `appendReasoning`, `flushBuffers`, `store.dispatch`, `flush` are byte-identical)

---

*Report generated by analyzing dist bundle chunks via static code inspection and stack trace correlation.*

zhangyapu1:
## Problem

During rapid streaming (especially with long reasoning output), the app crashes with 'Maximum update depth exceeded'.

The root cause: every streaming token chunk triggers a synchronous store.dispatch() which notifies all useSyncExternalStore listeners, causing React to re-render on every single token. When tokens arrive faster than React can render (50+ within one event-loop drain), React's nested update limit is exceeded.

## Root Cause

`
streaming chunk -> flush() -> flushBuffers() -> appendReasoning() -> store.dispatch()
  -> for (listener of stateListeners) listener()  // synchronous!
    -> React useSyncExternalStore re-render
      -> next chunk arrives before render completes
        -> another dispatch -> nestedUpdateCount++
          -> after ~50: Maximum update depth exceeded
`

## Fix

Defer state-listener notification via setTimeout(0) in createStore():

- All synchronous dispatches within the same event-loop tick coalesce into one notification burst
- React handles it as a single reconciliation cycle
- Event listeners (non-React, used for transcript/logging) remain synchronous

## Testing

- All 299 test files pass (3853 tests)
- Specifically verified: core-reducers, ui-reducer, ink-update-depth-repro, loop, lifecycle, loop-r1-reasoning, streaming-card-token-rate

Closes #2123

### DIFF
--- a/src/cli/ui/state/store.ts
+++ b/src/cli/ui/state/store.ts
@@ -18,13 +18,31 @@ export function createStore(session: SessionInfo, initialCards?: ReadonlyArray<C
   const stateListeners = new Set<StateListener>();
   const eventListeners = new Set<EventListener>();
 
+  // RAF-batched notification: during rapid streaming, dozens of chunks arrive
+  // within a single event-loop drain.  Each `dispatch()` used to synchronously
+  // notify every listener, causing React's `useSyncExternalStore` to re-render
+  // on *every* token.  When renders couldn't keep up, React hit "Maximum update
+  // depth exceeded".  Deferring via setTimeout(0) coalesces all dispatches that
+  // happen inside the same tick into a single notification burst.
+  let notifyScheduled = false;
+  const scheduleNotify = () => {
+    if (notifyScheduled) return;
+    notifyScheduled = true;
+    // setTimeout(0) defers to the next macrotask, coalescing all synchronous
+    // dispatches (e.g. rapid streaming chunks) into a single notification.
+    setTimeout(() => {
+      notifyScheduled = false;
+      for (const listener of stateListeners) listener();
+    }, 0);
+  };
+
   return {
     getState() {
       return state;
     },
     dispatch(event) {
       state = reduce(state, event);
-      for (const listener of stateListeners) listener();
+      scheduleNotify();
       for (const listener of eventListeners) listener(event);
     },
     subscribe(listener) {
--- a/src/cli/ui/state/store.ts
+++ b/src/cli/ui/state/store.ts
@@ -18,7 +18,7 @@ export function createStore(session: SessionInfo, initialCards?: ReadonlyArray<C
   const stateListeners = new Set<StateListener>();
   const eventListeners = new Set<EventListener>();
 
-  // RAF-batched notification: during rapid streaming, dozens of chunks arrive
+  // Macrotask-batched notification: during rapid streaming, dozens of chunks arrive
   // within a single event-loop drain.  Each `dispatch()` used to synchronously
   // notify every listener, causing React's `useSyncExternalStore` to re-render
   // on *every* token.  When renders couldn't keep up, React hit "Maximum update
--- a/tests/store-batch-notify.test.ts
+++ b/tests/store-batch-notify.test.ts
@@ -1,0 +1,99 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import type { SessionInfo } from "../src/cli/ui/state/state.js";
+import { createStore } from "../src/cli/ui/state/store.js";
+
+const session: SessionInfo = {
+  id: "test-batch",
+  branch: "main",
+  workspace: "/tmp/repo",
+  model: "deepseek-chat",
+};
+
+describe("store notification batching (setTimeout(0))", () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+  });
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("coalesces N synchronous dispatches into a single state-listener call", () => {
+    const store = createStore(session);
+    let notifyCount = 0;
+    store.subscribe(() => {
+      notifyCount++;
+    });
+
+    // Dispatch 3 events synchronously in one tick
+    store.dispatch({ type: "user.submit", text: "hello" });
+    store.dispatch({ type: "turn.start", turnId: "t1" });
+    store.dispatch({ type: "turn.thinking" });
+
+    // Listener should NOT have fired yet (deferred via setTimeout(0))
+    expect(notifyCount).toBe(0);
+
+    // getState() must already reflect all 3 dispatches synchronously
+    expect(store.getState().cards).toHaveLength(2); // user card + thinking live card
+    expect(store.getState().turnInProgress).toBe(true);
+
+    // Advance past the macrotask — listener fires exactly once
+    vi.advanceTimersByTime(0);
+    expect(notifyCount).toBe(1);
+  });
+
+  it("coalesces rapid streaming chunks into one notification", () => {
+    const store = createStore(session);
+    let notifyCount = 0;
+    store.subscribe(() => {
+      notifyCount++;
+    });
+
+    store.dispatch({ type: "reasoning.start", id: "r1" });
+    for (let i = 0; i < 50; i++) {
+      store.dispatch({ type: "reasoning.chunk", id: "r1", text: `tok${i} ` });
+    }
+
+    expect(notifyCount).toBe(0);
+    expect((store.getState().cards[0] as any).text).toContain("tok49");
+
+    vi.advanceTimersByTime(0);
+    expect(notifyCount).toBe(1);
+  });
+
+  it("event listeners fire synchronously per-dispatch (not batched)", () => {
+    const store = createStore(session);
+    const received: string[] = [];
+    store.onEvent((ev) => {
+      received.push(ev.type);
+    });
+
+    store.dispatch({ type: "user.submit", text: "a" });
+    expect(received).toEqual(["user.submit"]); // already fired synchronously
+
+    store.dispatch({ type: "turn.start", turnId: "t1" });
+    expect(received).toEqual(["user.submit", "turn.start"]);
+
+    store.dispatch({ type: "turn.thinking" });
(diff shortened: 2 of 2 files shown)


esengine (changes requested):
The fix is correct and the diagnosis is solid — synchronously notifying `useSyncExternalStore` listeners on every streaming chunk is exactly what trips React's nested-update guard, and deferring the notification breaks the synchronous chain so they coalesce into one render. `getState()` stays synchronous so read-after-dispatch callers are unaffected, and the `setTimeout(0)` choice actually matches this codebase's established 'defer off React's commit batch' pattern (same mechanism `useBoxMetrics` uses, as documented in tests/ink-update-depth-repro.test.tsx:48-54). So the mechanism is right.

Two things to address before merge:

1. **The comment is misleading.** It says 'RAF-batched notification' but the code uses `setTimeout(0)` — there's no `requestAnimationFrame` anywhere. Someone grepping for RAF will be confused, and the distinction matters (rAF is frame-paced, setTimeout(0) is next-macrotask). Please reword to 'macrotask-batched' / 'setTimeout(0)-batched'.

2. **No test covers the path you changed.** The PR cites `ink-update-depth-repro` as verification, but that suite exercises `useBoxMetrics` / `useAnimationFrame` in the ink package — it never touches `store.ts` or the `dispatch → stateListeners` path. So nothing currently guards this batching from a silent revert. Please add a focused test in the store/reducer suite: create a store, subscribe a state listener, `dispatch()` N times synchronously in one tick, and assert the listener fires **once** on the next macrotask (and that `getState()` already reflects all N synchronously). Also assert event listeners still fire synchronously per-dispatch, since that's a guarantee you're explicitly preserving.

Both are small. The fix itself is good — just want the comment honest and the behavior pinned down by a test on the actual code path, consistent with how the other rendering fixes in this repo have been gated.

zhangyapu1:
Both items addressed:

**1. Comment fixed** — Changed // RAF-batched notification: to // Macrotask-batched notification: in src/cli/ui/state/store.ts:21. No RAF references remain. Code uses setTimeout(0) (next macrotask), not 
equestAnimationFrame (frame-paced), so the comment now accurately reflects the mechanism.

**2. Store batch-notify test added** — 	ests/store-batch-notify.test.ts with 4 focused cases:
- N synchronous dispatches → state listener fires **once** on next macrotask; getState() reflects all N synchronously
- 50 rapid streaming chunks → single coalesced notification
- Event listeners fire **synchronously** per-dispatch (existing guarantee preserved)
- Second macrotask batch correctly schedules a new notification

Full suite green: 300 files, 3855 tests passed.

---

两项修改已完成：

**1. 注释修正** — src/cli/ui/state/store.ts:21 中 // RAF-batched notification: 改为 // Macrotask-batched notification:，不再有 RAF 相关描述。代码实际使用 setTimeout(0)（下一个宏任务），而非 
equestAnimationFrame（帧节奏），注释现在准确反映机制。

**2. 新增 store 批量通知测试** — 	ests/store-batch-notify.test.ts，4 个针对性用例：
- N 次同步 dispatch → state listener 仅触发一次；getState() 同步反映全部变更
- 50 次流式 chunk → 合并为一次通知
- event listener 每次 dispatch 同步触发（保持原有保证）
- 第二轮 macrotask 正确调度新通知

全量测试通过：300 文件，3855 测试。

esengine (approved):
Re-reviewed — comment now honestly says setTimeout(0)/next-macrotask (no more 'RAF'), and tests/store-batch-notify.test.ts pins the contract: N synchronous dispatches coalesce to one state-listener call, a second macrotask batches separately, and getState reflects state synchronously. Exactly what I asked for. CI green. Merging.

PR #2138, opened 2026-05-28
feat(desktop): persist prompt-history across sessions and restarts (#2051)

### LINKED ISSUE #2051: Desktop: prompt history recall is in-memory only; it is lost across sessions and app restarts

nianyi778:
## Problem

Closes #2051. The Composer's `historyRef` was a plain in-memory `useRef` — `ArrowUp` recall was lost on session switch or app restart. Users had to re-type frequently used commands every new chat session.

## Change

Persist the prompt history list in `~/.reasonix/config.json` under a new `promptHistory` field and seed the Composer on tab load.

**Data flow:**
1. `emitSettings` (desktop backend) reads `loadPromptHistory()` and includes it in the `$settings` event sent to the frontend.
2. `<Composer initialHistory={...}>` seeds `historyRef.current` from the persisted list on mount — so `ArrowUp` works from the first keystroke after restart or session switch.
3. `<Composer onHistoryPush={...}>` fires on every send; App.tsx calls `applySettingsPatch({ promptHistory })`, which triggers `settings_save` → backend writes the updated list back.
4. `savePromptHistory` caps at 100 entries (same as the previous in-memory cap).
5. Field is registered in `STRING_ARRAY_FIELDS` so a malformed hand-edited value is sanitized to `[]` on read rather than crashing.

**Scope:** desktop only (as requested in #2051 — CLI/TUI and dashboard are unaffected).

## Files changed
- `src/config.ts` — `promptHistory` field, `load/savePromptHistory` accessors, sanitize registration
- `src/cli/commands/desktop.ts` — `emitSettings` includes history; `settings_save` writes it back
- `desktop/src/protocol.ts` — `promptHistory` added to `SettingsEvent` and `SettingsPatch`
- `desktop/src/App.tsx` — receives history from state, passes `initialHistory`/`onHistoryPush` to `<Composer>`
- `desktop/src/ui/composer.tsx` — `initialHistory` prop seeds `historyRef`; `onHistoryPush` fires on every push

## Verification
- `tests/config.test.ts`: load returns `[]` when absent; round-trip; cap at 100.
- `npm run verify` passes.

### DIFF
--- a/desktop/src/App.tsx
+++ b/desktop/src/App.tsx
@@ -260,6 +260,7 @@ export type Settings = {
   };
   subagentModels?: Record<string, "flash" | "pro">;
   showSystemEvents?: boolean;
+  promptHistory?: string[];
   version: string;
 };
 
@@ -994,6 +995,7 @@ function applyIncomingRaw(state: State, ev: IncomingEvent): State {
           webSearchApiKeys: ev.webSearchApiKeys,
           subagentModels: ev.subagentModels,
           showSystemEvents: ev.showSystemEvents,
+          promptHistory: ev.promptHistory,
           version: ev.version,
         },
       };
@@ -2447,6 +2449,15 @@ function TabRuntime({
                   setDraft("");
                 }}
                 onDequeueSend={(index) => dispatch({ t: "dequeue_send", index })}
+                initialHistory={state.settings?.promptHistory}
+                onHistoryPush={(entry) => {
+                  // Use saveSettings (RPC only, no local state patch) so the
+                  // sentinel [entry] is never written into state.settings and
+                  // historyRef is not transiently reset. The backend merges
+                  // against the freshly-loaded persisted list and re-emits
+                  // $settings with the merged result (#2051).
+                  saveSettings({ promptHistory: [entry] });
+                }}
               />
             </>
           )}
--- a/desktop/src/protocol.ts
+++ b/desktop/src/protocol.ts
@@ -322,6 +322,8 @@ export type SettingsEvent = {
   };
   subagentModels?: Record<string, "flash" | "pro">;
   showSystemEvents?: boolean;
+  /** Desktop prompt-history entries seeded on tab load, most-recent-first (#2051). */
+  promptHistory?: string[];
   version: string;
 };
 
@@ -372,6 +374,8 @@ export type SettingsPatch = {
   braveApiKey?: string | null;
   subagentModels?: Record<string, "flash" | "pro">;
   showSystemEvents?: boolean;
+  /** Persisted prompt-history entries to update on each send (#2051). */
+  promptHistory?: string[];
 };
 
 export type QQConfigPatch = {
--- a/desktop/src/ui/composer.tsx
+++ b/desktop/src/ui/composer.tsx
@@ -154,6 +154,8 @@ export function Composer({
   queuedSends,
   onQueueWhileBusy,
   onDequeueSend,
+  initialHistory,
+  onHistoryPush,
 }: {
   draft: string;
   setDraft: React.Dispatch<React.SetStateAction<string>>;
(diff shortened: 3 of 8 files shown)


nianyi778:
@codex review

nianyi778:
Fixed in `b881b82f`. Added a `useEffect` that syncs `historyRef.current` when `initialHistory` first arrives (the async `$settings` event), as long as the user hasn't started browsing (`browseIdx === -1`). This way the persisted history is available from the first `ArrowUp` keystroke after restart, not just after remount.

The other Codex finding (`src/tools/shell/parse.ts:356` — grep patterns treated as paths) is pre-existing on `main` and not part of this PR's diff; happy to tackle it separately.

@codex review

nianyi778:
Fixed in `4063bb9b`. `onHistoryPush` now merges the new entry into the current persisted list (`state.settings?.promptHistory`) instead of overwriting with the tab-local array:

```ts
const merged = [entry, ...persisted.filter(e => e !== entry)].slice(0, 100);
applySettingsPatch({ promptHistory: merged });
```

Each save prepends the new entry, deduplicates, caps at 100, so tabs never clobber each other's entries.

@codex review

nianyi778:
Fixed in `846c697d`. The persisted list is stored most-recent-first (`[entry, ...persisted]`), but `historyRef` is oldest-first (entries are appended with `push`, ArrowUp reads from the end). Both assignment sites now reverse on load:

```ts
historyRef = useRef([...initialHistory].reverse())   // mount
historyRef.current = [...initialHistory].reverse()   // async settings update
```

This keeps ArrowUp order consistent: most-recent entry first, regardless of whether history came from in-session sends or from a persisted reload.

@codex review

nianyi778:
Fixed properly in `dcb51ee1`. Moved the merge to the **backend** (`settings_save` handler in `desktop.ts`):

- Frontend now sends just `{ promptHistory: [newEntry] }` — a single-element sentinel.
- Backend reads the current persisted list fresh from disk, prepends the new entry, deduplicates, caps at 100, writes back, then calls `emitSettings(tab)` to re-broadcast the updated list to this tab.

Since the merge happens after a fresh disk read, concurrent tabs always see the true latest state — no more clobber.

@codex review

nianyi778:
Fixed in `639dc48b`. The second finding is real — `applySettingsPatch` writes the sentinel into `state.settings`, which triggers the `useEffect` and transiently resets `historyRef` to just the one entry.

Changed to use `saveSettings` (RPC-only, no local dispatch) so the sentinel never lands in `state.settings.promptHistory`. The backend still does the fresh-disk merge and re-emits `$settings` with the full merged list, which updates the state correctly.

The first finding (`shell/parse.ts` grep pattern) is pre-existing on `main` and not in this PR's diff — happy to fix separately.

@codex review

PR #2143, opened 2026-05-28
fix(desktop): composer inline mentions, command chips, and text highlighting

### LINKED ISSUE #2142: 桌面端：输入框 @ 提及文件与 / 命令的三个交互问题
## 问题描述

当前桌面端输入框在使用 \`@\` 提及文件与 \`/\` 命令时存在三个交互缺陷：

### 1. 仅能在末尾插入
在输入框中间输入 \`@\` 或 \`/\` 无法触发弹窗，只能在文本末尾操作。用户期望在任何位置输入（只要后续是空白或文本末尾）都能弹出选择组件。

### 2. 即时执行命令选定后仍然固定 chip
非 \`insertOnly\` 的 slash 命令（如 \`/clear\`、\`/new\`）在选定后立即执行，但同时在输入框上方固定一个 chip，给用户造成困惑——命令已执行为何还有残留。

### 3. 提及文件后 chip 与文本并存，令人困惑
选定文件后，既在上方固定 chip，又在输入框内插入 \`@path\` 普通文本。用户可以删除 chip 但文本还在，也可以编辑文本但 chip 还在——用户无法判断哪个才是"真正"的提及。

## 修复方案

由 #2143 实现，改动包含：

1. **任意位置插入**：将 \`handleChange\` 中的 \`\$\` 锚点正则替换为基于光标位置（\`selectionStart\`）的 \`findActiveRange\` 检测，并增加 \`onSelect\` 处理光标移动进入已有 \`@word\`/\`/word\` 时自动弹窗
2. **即时命令不显示 chip**：非 \`insertOnly\` 的 slash 命令选中后只执行 \`run()\`，不添加到 \`pickedChips\`
3. **文本为主，chip 只读**：chip 仅来自 popup 中显式选中（\`pickedChips\` Map），不在输入过程中显示；移除 chip 的 X 按钮，用户删除文本 token 后 chip 自动消失；通过 overlay 技术高亮 \`@path\`

## 关联 PR

- #2143

AtmoOmen:
## Summary

Closes #2142.

Fixes three UX issues with the desktop composer's \`@\` mention and \`/\` command insertion:

### 1. Allow insertion at any cursor position (not just end-of-text)

Previously, the \`handleChange\` regex \`/(\^|\s)([/@])([\^\s]*)\$/\` only matched the trailing token. Now \`findActiveRange(text, cursorPos)\` scans from the cursor position to detect any \`@word\` or \`/word\` regardless of where it sits in the draft.

- Added \`activeRangeRef\` to track the exact range (start→end) of the current token
- \`pickItem\` and Tab navigation now replace precisely that range rather than doing blind regex replacement on the full draft tail
- Added \`onSelect\` handler: clicking or arrowing into an existing \`@word\` / \`/word\` reopens the popup for inline editing

### 2. Remove redundant chips from non-\`insertOnly\` slash commands

Commands like \`/clear\` and \`/new\` execute immediately when selected. Previously they also left a chip pinned above the textarea — confusing since the command already ran. Now only \`insertOnly\` commands (e.g. \`/model\`) produce chips.

### 3. Chips as read-only indicators derived from picked mentions

**Before**: chips were independent imperative state that drifted from the draft text. Users could dismiss a chip while its \`@path\` text remained, or edit the text while the chip stayed — unclear which one was authoritative.

**After**:
- Chips only appear for mentions / \`insertOnly\` commands explicitly picked from the popup (\`pickedChips\` Map), never for incomplete \`@sr\` typing
- X dismiss button removed — chips are read-only indicators
- Each chip auto-disappears when the user manually deletes its token from the draft (filtered via word-boundary regex check)
- \`@path\` tokens are highlighted in the textarea via an overlay backdrop (\`composer-backdrop\`) with transparent textarea + \`highlightTokens()\` span wrapping

### Additional fixes from review

- \`insertMention\` now uses functional \`setDraft\` to avoid stale-closure corruption when file dialogs / clipboard reads resolve asynchronously
- Tab on non-directory items in the \`@\` popup now consumes the event so focus stays in the textarea
- Backdrop content memoized via \`useMemo\`
- IME guard supplemented with \`e.nativeEvent.isComposing\`
- All comments trimmed to ≤3 lines

### Files changed

| File | Changes |
|---|---|
| \`desktop/src/ui/composer.tsx\` | Core logic: \`findActiveRange\`, \`syncPopupFromCursor\`, range-based replacement, chip derivation, \`highlightTokens\`, backdrop overlay |
| \`desktop/src/styles.css\` | \`.composer-textarea-wrap\`, \`.composer-backdrop\`, \`.hl-token\`, transparent textarea |

### DIFF
--- a/desktop/src/styles.css
+++ b/desktop/src/styles.css
@@ -3542,9 +3542,12 @@ html:not([data-platform="macos"]) .shortcut kbd[data-key="mod"] {
 }
 .composer-tags {
   display: flex;
-  flex-wrap: wrap;
+  flex-wrap: nowrap;
   gap: 4px;
   padding: 8px 10px 0;
+  overflow-x: auto;
+  scrollbar-width: thin;
+  scrollbar-color: var(--border) transparent;
 }
 .chip {
   display: inline-flex;
@@ -3557,6 +3560,13 @@ html:not([data-platform="macos"]) .shortcut kbd[data-key="mod"] {
   font-size: 14px;
   font-family: inherit;
   color: var(--fg-2);
+  max-width: 220px;
+  flex-shrink: 0;
+}
+.chip span {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
 }
 .chip.at {
   color: var(--accent);
--- a/desktop/src/styles.css
+++ b/desktop/src/styles.css
@@ -3568,19 +3568,33 @@ html:not([data-platform="macos"]) .shortcut kbd[data-key="mod"] {
   background: var(--violet-soft);
   border-color: transparent;
 }
-.chip .x {
-  cursor: pointer;
-  opacity: 0.5;
+.composer-textarea-wrap {
+  position: relative;
 }
-.chip .x:hover {
-  opacity: 1;
+.composer-backdrop {
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  padding: 12px 14px 6px;
+  font-size: 14px;
+  line-height: 1.55;
+  font-family: inherit;
+  white-space: pre-wrap;
+  word-wrap: break-word;
+  overflow: hidden;
+  pointer-events: none;
+  color: var(--fg);
 }
-
+.composer-backdrop .hl-token {
+  color: var(--accent);
+}
 .composer textarea {
   display: block;
   width: 100%;
   resize: none;
-  background: none;
+  background: transparent;
   border: none;
   outline: none;
   padding: 12px 14px 6px;
@@ -3589,6 +3603,10 @@ html:not([data-platform="macos"]) .shortcut kbd[data-key="mod"] {
   min-height: calc(2 * 1.55em + 18px);
   overflow-y: hidden;
   font-family: inherit;
+  color: transparent;
+  caret-color: var(--fg);
+  position: relative;
+  z-index: 1;
 }
 .composer textarea::placeholder {
   color: var(--muted-2);
--- a/desktop/src/styles.css
+++ b/desktop/src/styles.css
@@ -3611,12 +3611,18 @@ html:not([data-platform="macos"]) .shortcut kbd[data-key="mod"] {
   font-size: 14px;
   line-height: 1.55;
   min-height: calc(2 * 1.55em + 18px);
-  overflow-y: hidden;
+  overflow-y: auto;
+  overflow-x: hidden;
+  overflow-wrap: break-word;
+  scrollbar-width: none;
   font-family: inherit;
   color: transparent;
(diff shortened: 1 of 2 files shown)


esengine (commented):
Read through the logic carefully — this is a clean, well-engineered refactor. The cursor-aware `findActiveRange` + `onSelect` wiring correctly handles editing a mention mid-text (the old trailing-regex only worked at end-of-input), the regex-escaping in the derived-`chips` memo is properly guarded against special chars in labels, `highlightTokens` renders through React text nodes so there's no injection risk, and adding `e.nativeEvent.isComposing` to the Enter guard is a real IME robustness fix. I also verified the backdrop CSS mirrors the textarea exactly (both font-size 14px / line-height 1.55 / padding 12px 14px 6px / font-family inherit, textarea text transparent with visible caret, z-index layering correct) — so the alignment fundamentals are right.

I can't fully approve a desktop overlay change from CI alone, so two asks before merge:

1. **Visual confirmation of the backdrop alignment in the fragile cases.** The transparent-textarea-over-highlight-div technique drifts whenever wrapping diverges between the two elements. Could you attach screenshots showing the highlight stays aligned with: (a) a line long enough to soft-wrap, (b) a very long unbroken `@path` token, (c) CJK text (this UI is heavily Chinese — double-width glyphs are the classic break case), and (d) enough lines that the textarea scrolls (verifying `handleTextareaScroll` keeps the backdrop in sync)? `box-sizing` parity between the two also matters — please confirm both are border-box.

2. **Confirm the two intentional UX changes.** (a) Chips are now derived from the draft and the per-chip ✕ remove button is gone — chips are read-only indicators that disappear when you delete the token from the text. (b) `onSelect` reopens the picker whenever the cursor lands inside any `@`/`/` word, so clicking into a completed `@src/file.ts` mention re-opens the file picker. Both are defensible, just want to confirm they're deliberate and not surprising in practice.

Code-wise this is good — once I see it render correctly I'm happy to merge.

AtmoOmen:
image

When chip hovered, a tooltip will be shown
image

image

image

When there are more chips, a horizontal scroll bar will be displayed. 

image

CJK

image


Also, both designs you mentioned are intentional. Similar UX design could be found elsewhere like Claude Code、Antigravity and so on

esengine (approved):
Re-reviewed with your screenshots — exactly the confirmation I wanted. The backdrop highlight stays aligned with CJK text and through wrapping, the many-chips case scrolls horizontally rather than breaking layout, and the chip-hover tooltip works. And thanks for confirming the two UX choices (derived read-only chips, click-into-mention reopens the picker) are intentional and match Claude Code / Antigravity conventions. Code was already sound on my first pass; now that it's visually verified, this is good. CI green. Merging.

PR #2149, opened 2026-05-28
perf(config): cache readConfig with mtime-based invalidation

HUQIANTAO:
## Summary

- `readConfig()` performs synchronous `readFileSync` + `JSON.parse` on every call
- It is invoked 5-10+ times per UI frame by `loadModel`, `loadLanguage`, `loadTheme`, `loadEditMode`, `loadReasoningEffort`, `loadDashboardEnabled`, etc.
- Each call re-reads and re-parses the same `config.json` from disk

This PR adds a module-level cache keyed by `(path, mtimeMs)`. On cache hit, `statSync` (~0.005ms) replaces `readFileSync` + `JSON.parse` (~0.05ms) — a ~10x improvement per call, multiplied across all callers in a single UI frame.

## Changes

- `src/config.ts`: Add `_configCache` with mtime-based invalidation in `readConfig()`, invalidate in `writeConfig()`

## How it works

1. `readConfig()` calls `statSync(path)` to get the file's mtime
2. If `(path, mtimeMs)` matches the cached entry, return the cached config immediately
3. Otherwise, read + parse + sanitize as before, and update the cache
4. `writeConfig()` sets `_configCache = null` so the next read picks up changes

## Test plan

- All existing `tests/config.test.ts` tests pass (readConfig round-trip, writeConfig atomicity, saveApiKey, loadEndpoint, etc.)
- Verified `biome check` passes with no lint issues
- No new dependencies added

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,7 +1,7 @@
 /** Library reads only DEEPSEEK_API_KEY from env; the CLI bridges config.json → env var. */
 
 import { randomBytes } from "node:crypto";
-import { mkdirSync, readFileSync } from "node:fs";
+import { mkdirSync, readFileSync, statSync } from "node:fs";
 import { homedir } from "node:os";
 import { dirname, isAbsolute, join, resolve } from "node:path";
 import { z } from "zod";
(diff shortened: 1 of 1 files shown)


HUQIANTAO:
Closing in favor of a more comprehensive performance PR.

PR #2151, opened 2026-05-28
fix(cli): commit settled cards to native scrollback

### LINKED ISSUE #2120: p0bug cli滚轮滑动不了
任务结束后滚轮滑动不了，输出都看不了，根本用不了

GTC2080:
## Summary
- commit settled StaticCardStream cards to the Ink Static region so native terminal scrollback receives completed tool/reasoning history
- freeze the verbose value captured when a card enters Static so completed scrollback stays immutable
- keep only unsettled cards in the live tail
- add regression tests for settled cards entering Static and for verbose toggles not retroactively expanding settled tool output

## Root cause
Native scroll mode left completed tool/reasoning cards in the dynamic live tail so the terminal never received those completed rows in its own scrollback. Once the task ended, wheel scrolling could not reveal history that had only been repainted inside the live viewport.

## Tradeoff
Settled tool/reasoning cards are now immutable once committed to terminal scrollback. Ctrl+R verbose mode affects live cards and cards rendered after the toggle, but it no longer rewrites already-settled scrollback. This is intentional: preserving real terminal scrollback for #2120 matters more than retroactive expansion of successful settled cards.

The #1608 failed-shell-output path is preserved. `selectToolPreviewLines` still pins failure lines for failed `run_command`/`run_background` output independent of verbose mode, so failed shell output still reaches scrollback.

Fixes #2120

## Verification
- `npx vitest run tests/static-card-stream-verbose.test.tsx tests/static-card-stream-progressive.test.tsx tests/tool-summary.test.ts`
- `npx vitest run tests/chat-scroll-wheel.test.ts tests/mouse-mode.test.ts tests/static-card-stream-progressive.test.tsx tests/static-card-stream-verbose.test.tsx tests/cardstream-architecture.test.tsx`
- `npm run verify`
- `git push` (pre-push ran `npm run verify`)

### DIFF
--- a/src/cli/ui/layout/StaticCardStream.tsx
+++ b/src/cli/ui/layout/StaticCardStream.tsx
@@ -54,7 +54,7 @@ function StaticCardStreamInner({
  *  the gate. New cards added DURING drain are held back until the backlog
  *  catches up — keeps Static's append-only contract intact so chronological
  *  order is preserved. Gates the FULL cards array (not just the static partition)
- *  so verbose-sensitive tool/reasoning cards in the live region also drip. */
+ *  so an old unsettled live tail also drips while fresh cards bypass the gate. */
 function useProgressiveBacklog(cards: readonly Card[]): Card[] {
   const backlogRef = useRef<number | null>(null);
   if (backlogRef.current === null && cards.length > 0) {
@@ -93,7 +93,7 @@ function partition(cards: readonly Card[]): {
   dynamicItems: Card[];
   hasUnsettledDynamic: boolean;
 } {
-  const firstDynamic = cards.findIndex((c) => !isFullySettled(c) || isVerboseSensitive(c));
+  const firstDynamic = cards.findIndex((c) => !isFullySettled(c));
   if (firstDynamic === -1) {
     return { staticItems: [...cards], dynamicItems: [], hasUnsettledDynamic: false };
   }
@@ -103,10 +103,6 @@ function partition(cards: readonly Card[]): {
     dynamicItems,
     hasUnsettledDynamic: dynamicItems.some((c) => !isFullySettled(c)),
   };
-}
-
-function isVerboseSensitive(card: Card): boolean {
-  return card.kind === "reasoning" || card.kind === "tool";
 }
 
 function isFullySettled(card: Card): boolean {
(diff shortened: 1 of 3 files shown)


esengine (changes requested):
The fix is directionally right and the diagnosis of #2120 is correct: keeping every tool/reasoning card pinned in the live tail (the `isVerboseSensitive` carve-out) is exactly why completed history never reached native scrollback — so dropping it and letting settled cards flow into `<Static>` is the proper fix, not a band-aid. The test-comment update is also legitimate: on `main` that file's header comment ('must let verbose mode expand already-settled tool cards') was already stale — the actual test only checks render isolation, so you're making the comment honest.

But there's a behavior change here that isn't called out and deserves a conscious decision + a test:

**Removing `isVerboseSensitive` reverts part of what #1608 added.** That carve-out (commit 96c91279) kept tool/reasoning cards in the dynamic region specifically so the runtime verbose toggle (App.tsx:512/1671 → `VerboseContext`) could re-render them. Now that settled tool/reasoning cards go to `<Static>` — which Ink renders exactly once — **toggling verbose no longer expands/collapses already-settled tool/reasoning cards.** Only unsettled + future cards respond to the toggle.

The good news, and please confirm you reached the same conclusion: **#1608's actual goal — surfacing FAILED shell output — is preserved**, because `selectToolPreviewLines` shows failed-shell output through the verbose-independent `failedShell` branch (tool-summary.ts:71-84), so a failed command still dumps its output into scrollback regardless of verbose. What's lost is only retroactive verbose-expand of *successful* settled cards — which is arguably impossible to keep once you commit to real terminal scrollback anyway, so I think the tradeoff is correct.

Two asks before merge:
1. **Call the tradeoff out in the PR description** (and ideally a code comment near `partition`): settled tool/reasoning cards are now immutable in scrollback, so the verbose toggle only affects live + future cards. This is the inverse of the old (stale-commented) expectation, so it shouldn't be silent.
2. **Pin the new contract with a test.** You added a test for the scrollback direction (settled card → static) — good. Add the complementary one: with a settled tool card committed, flipping `VerboseContext` does NOT re-render the `<Static>` region / does not re-expand it. That stops someone from 'fixing' this back to the #1608 behavior and silently reopening #2120.

Mechanically the change is tiny and correct — just want the behavior reversal documented and locked down.

esengine (approved):
Re-reviewed — you went a step better than I asked: `StaticCardRenderer` now captures verbose at commit time via `useRef(verbose).current` and shadows it with a nested `VerboseContext.Provider`, so a settled card's verbose state is explicitly frozen rather than implicitly relying on Static's render-once. And the new test ("does not retroactively expand settled tool cards after a verbose toggle") pins exactly that: render-count unchanged + output still hidden after toggling verbose true. #1608's failed-output path is untouched. CI green. Merging.

PR #2152, opened 2026-05-28
perf: optimize tokenizer hot paths and cache heal pipeline

HUQIANTAO:
## Summary

Five independent optimizations that reduce CPU and allocation pressure in the main agent loop:

### 1. `byteLevelEncode`: O(n²) → O(n) string building
- `src/tokenizer.ts:186-191` — `+=` loop copies the entire accumulator on each iteration
- Replaced with pre-allocated `string[]` + `.join("")`
- For a 10KB text chunk this eliminates ~10K intermediate string allocations

### 2. `countTokens`: skip array allocation
- `src/tokenizer.ts:261-263` — `encode(text).length` builds a full `number[]` just to read `.length`
- New version counts tokens inline without allocating the array
- Saves ~60KB of transient heap per 50KB input string

### 3. `healActiveLogBeforeSend`: cache by log version
- `src/loop.ts:523-543` — runs 3 healing passes (tokenize + shrink + prune) on **every** loop iteration
- Cache the result keyed by `log.totalLength` — subsequent calls in the same turn return immediately
- Cache invalidates automatically when new messages are appended (totalLength changes)

### 4. `contentTokenCache`: size guard for large strings
- `src/tokenizer.ts:535` — skip caching strings >10KB in the bounded-token LRU
- Without this guard, 4096 entries of 50KB+ tool outputs can retain 200MB+ of string keys

### 5. `formatDeepSeekPrompt`: array-join instead of +=
- Same O(n²) → O(n) fix — the prompt can be 100K+ chars; `+=` copies the entire string each iteration

## Test plan

- `tests/tokenizer.test.ts` — 33 tests pass (encode, countTokens, countTokensBounded, formatDeepSeekPrompt)
- `tests/loop.test.ts` — 69 tests pass (buildMessages, heal pipeline, streaming, tool dispatch)
- `tests/loop-r1-reasoning.test.ts` — 21 tests pass (reasoning_content round-trip)

### DIFF
--- a/src/loop.ts
+++ b/src/loop.ts
@@ -515,12 +515,21 @@ export class CacheFirstLoop {
   }
   private _inflightCounter = 0;
 
+  /** Heal result cache — keyed by log.totalLength so we skip the 3-pass
+   *  healing pipeline (tokenize + shrink + prune) when the log hasn't changed. */
+  private _healCacheVersion = -1;
+  private _healCacheResult: ChatMessage[] | null = null;
+
   private buildMessages(): ChatMessage[] {
     const healedMessages = this.healActiveLogBeforeSend();
     return [...this.prefix.toMessages(), ...healedMessages];
   }
 
   private healActiveLogBeforeSend(): ChatMessage[] {
+    const version = this.log.totalLength;
+    if (this._healCacheVersion === version && this._healCacheResult) {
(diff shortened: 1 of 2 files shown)


HUQIANTAO:
Closing in favor of a more comprehensive performance PR.

PR #2162, opened 2026-05-28
perf: eliminate sync I/O from agent loop hot paths

HUQIANTAO:
## Summary

Three independent cache layers that together remove redundant synchronous disk reads and deep-cloning from the per-iteration fast path. Each addresses a distinct bottleneck in the agent loop's critical section.

### 1. `readConfig()` mtime cache (`src/config.ts`)

`readConfig()` performs `readFileSync` + `JSON.parse` on every call. A grep shows **37+ call sites per session** — `loadModel`, `loadLanguage`, `loadTheme`, `loadEditMode`, `loadReasoningEffort`, `loadDashboardEnabled`, and many more all default-arg `readConfig()`.

**Fix**: Module-level `Map<string, { mtimeMs, cfg }>` keyed by `(path, mtimeMs)`. On hit, `statSync` (~0.005ms) replaces `readFileSync` + `JSON.parse` (~0.05ms). `writeConfig()` deletes the cache entry so writes are immediately visible.

**Impact**: ~37 redundant disk reads per session → ~2 (one per unique config path). On network-mounted home directories or slow disks, this is a measurable latency reduction per UI frame.

### 2. `ImmutablePrefix.tools()` — eliminate `structuredClone` (`src/memory/runtime.ts`)

Every loop iteration calls `this.prefix.tools()` which deep-clones the entire tool-spec array via `structuredClone`. With 20+ registered tools (filesystem, shell, memory, plan, web, MCP tools), each having a JSON schema, this copies hundreds of KB of schema data on every single API round-trip.

**Fix**: Cache a frozen shallow copy (`Object.freeze` + shallow `{ ...t, function: { ...t.function } }`). Invalidated only by `addTool()` / `removeTool()` — the only mutation paths. Callers in `client.ts` (`buildPayload`) only read specs, never mutate.

**Impact**: Eliminates ~200-500KB of deep-cloning per API call. The frozen reference costs nothing to return on subsequent calls.

### 3. `AppendOnlyLog.toFullHistory()` — cache disk reads (`src/memory/runtime.ts`)

`buildMessages()` calls `healActiveLogBeforeSend()` which calls `toFullHistory()` — **2-3 times per loop iteration** (turn-start fold estimate, pre-API call, post-steer injection). When the log window doesn't cover full history, each call does synchronous `openSync`/`readSync`/`closeSync` in 64KB chunks, parsing each line with `JSON.parse`.

**Fix**: Cache the parsed `ChatMessage[]` keyed by `totalLength`. Invalidated by `append()`, `compactInPlace()`, and `initWindow()`. Cache hits return shallow copies to prevent caller mutation.

**Impact**: In a typical 3-buildMessages iteration, 2 of 3 calls now return from cache instead of doing disk I/O. For long sessions where the log window is smaller than total history, this eliminates redundant synchronous file reads that block the event loop.

## Test plan

- `tests/config.test.ts` — 95 tests pass (readConfig round-trip, writeConfig atomicity, saveApiKey, loadEndpoint, etc.)
- `tests/loop.test.ts` — 69 tests pass (buildMessages, heal pipeline, streaming, tool dispatch)
- `tests/tokenizer.test.ts` — 33 tests pass

### DIFF
--- a/src/code/edit-blocks.ts
+++ b/src/code/edit-blocks.ts
@@ -1,5 +1,6 @@
 /** SEARCH must match byte-for-byte; empty SEARCH = create new file. No fuzzy match — silent wrong edit beats a missing one. */
 
+import { randomBytes } from "node:crypto";
 import {
   chmodSync,
   closeSync,
@@ -116,7 +117,7 @@ function fsyncDirectoryBestEffort(path: string): void {
 }
 
 function atomicReplaceFileSync(path: string, buf: Buffer, mode: number): void {
-  const tmp = `${path}.${process.pid}.${Date.now()}.${Math.random().toString(36).slice(2)}.tmp`;
+  const tmp = `${path}.${randomBytes(8).toString("hex")}.tmp`;
   const permissions = mode & 0o7777;
   let fd: number | undefined;
   try {
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,7 +1,7 @@
 /** Library reads only DEEPSEEK_API_KEY from env; the CLI bridges config.json → env var. */
 
 import { randomBytes } from "node:crypto";
-import { mkdirSync, readFileSync } from "node:fs";
+import { mkdirSync, readFileSync, statSync } from "node:fs";
 import { homedir } from "node:os";
 import { dirname, isAbsolute, join, resolve } from "node:path";
 import { z } from "zod";
@@ -446,6 +446,11 @@ function sanitizeStringArrayField(
   parent[leaf] = filtered;
 }
 
+/** Mtime-keyed cache — eliminates redundant readFileSync + JSON.parse across
+ *  the 37+ call sites that read config per session. statSync is ~0.005 ms vs
+ *  ~0.05 ms for readFileSync + parse; the cache pays for itself after 2 hits. */
+const _configCache = new Map<string, { mtimeMs: number; cfg: ReasonixConfig }>();
+
 export function readConfig(path: string = defaultConfigPath()): ReasonixConfig {
   try {
     // Strip the UTF-8 BOM if a foreign writer left one in — Windows
@@ -454,6 +459,9 @@ export function readConfig(path: string = defaultConfigPath()): ReasonixConfig {
     // refuses BOM-prefixed input and throws, which used to fall
     // through to `return {}` and silently nuke every saved field on
     // the next read-modify-write.
+    const st = statSync(path);
+    const cached = _configCache.get(path);
+    if (cached && cached.mtimeMs === st.mtimeMs) return cached.cfg;
     const raw = readFileSync(path, "utf8").replace(/^\uFEFF/, "");
     const parsed = JSON.parse(raw);
     if (parsed && typeof parsed === "object" && !Array.isArray(parsed)) {
@@ -461,7 +469,9 @@ export function readConfig(path: string = defaultConfigPath()): ReasonixConfig {
       for (const segments of STRING_ARRAY_FIELDS) {
         sanitizeStringArrayField(cfg, segments, path);
       }
-      return cfg as ReasonixConfig;
+      const result = cfg as ReasonixConfig;
+      _configCache.set(path, { mtimeMs: st.mtimeMs, cfg: result });
+      return result;
     }
   } catch {
     /* missing or malformed → empty config */
@@ -518,6 +528,7 @@ export function writeConfig(cfg: ReasonixConfig, path: string = defaultConfigPat
   // baseline (issue #1535).
   const tmp = `${path}.${process.pid}.tmp`;
   atomicWriteSync(path, JSON.stringify(cfg, null, 2), tmp);
+  _configCache.delete(path);
 }
 
 /** Resolve the language from config file. */
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,7 +1,7 @@
 /** Library reads only DEEPSEEK_API_KEY from env; the CLI bridges config.json → env var. */
 
 import { randomBytes } from "node:crypto";
-import { mkdirSync, readFileSync, statSync } from "node:fs";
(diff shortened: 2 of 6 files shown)


esengine (approved):
Reviewed all three caches for invalidation completeness and mutation safety — they're correct. Verified:

1. **readConfig mtime cache**: `atomicWriteSync` is only ever called from `writeConfig` (config.ts:540), and every `save*`/`clear*` helper persists through `writeConfig`, which `_configCache.delete(path)`s — so there's no in-process write path that leaves the cache stale. `statSync` sits inside the existing try, so a missing config still falls through to `{}` exactly as before.

2. **tools() frozen snapshot**: `prefix.tools()` has a single caller (loop.ts:708) which only reads it — passes it straight to the model-call `tools` field where client.ts:218 assigns it by reference and JSON-serializes it. Nothing mutates it, so neither the freeze (top-level) nor the shared `function.parameters` (nested) can bite today. Invalidated correctly on addTool/removeTool.

3. **toFullHistory cache**: every mutator invalidates — `initWindow`, `append`, `compactInPlace`, and `extend` (which delegates to `append`). The shallow-copy-on-return matches the pre-existing `{ ...e }` semantics, so no new aliasing exposure versus before.

Good, well-scoped optimization. Two small asks (non-blocking, your call):

- **A one-line caveat comment on each cache about the immutability contract.** Cache 2's frozen copy is *shallow* — `function.parameters` is shared by reference with `_toolSpecs`, so a future caller that mutates a schema would silently corrupt it. A note like 'result is deeply immutable; do not mutate' guards the next person, since the type (`ToolSpec[]` via `as unknown as`) hides the readonly-ness. Consider typing it `readonly ToolSpec[]` so the compiler enforces it.
- **readConfig external-edit caveat**: the cache trusts `mtimeMs`, so an out-of-band edit that preserves mtime (or collides within the FS granularity) would serve stale config. Negligible on modern FS, but worth a one-line comment so it's a known tradeoff.

CI hasn't run (fork PR) — I approved the workflow. Since this touches `config.ts` (read on basically every frame), I'd like CI green before merging, then it's good to land.

esengine (changes requested):
Following up on my approval — **CI came back and CodeQL flagged a high-severity alert in this PR's change**, so it's not green after all. Holding until it's resolved.

**CodeQL: 'Potential file system race condition' at src/config.ts:465.** The new cache introduces a TOCTOU pattern:

```ts
const st = statSync(path);                       // time-of-check (mtime)
const cached = _configCache.get(path);
if (cached && cached.mtimeMs === st.mtimeMs) return cached.cfg;
const raw = readFileSync(path, "utf8")…          // time-of-use (content)
```

The old code read the file directly with no preceding `statSync`, so this pattern is new to the PR. The real-world severity here is low — it's the user's own `~/.reasonix/config.json`, not a privilege boundary — but it's a legitimate finding and it's a hard CI gate, so it has to be addressed one way or the other:

- **Preferred fix (removes the race):** stat the same file descriptor you read, so check and use are atomic — `openSync` → `fstatSync(fd)` for `mtimeMs` → `readSync`/`readFileSync(fd)` → `closeSync`. fstat-on-fd can't race the subsequent read of that same fd. This keeps the cache-hit fast path (you'd still `statSync` for the hit check; only the miss path opens the fd) — or restructure so the hit check uses the fd's fstat too.
- **Alternative:** if you'd rather keep `statSync`, dismiss the CodeQL alert with an explicit justification ('config file is owned by the user, not a security boundary; stale window is bounded by mtime and self-corrects next read'). That's defensible given the context, but the fstat-on-fd version is cleaner and makes the gate pass on its own.

Everything else in my earlier review still stands — the three caches are correct and invalidation is complete. Just resolve the CodeQL finding (fix or justified dismissal) and re-run, plus the two optional immutability/caveat comments, and this is good to land.

HUQIANTAO:
Addressed both review comments:

### 1. TOCTOU race fix (CodeQL)

Replaced `statSync(path)` + `readFileSync(path)` with an fd-based atomic read:
- `openSync(path, "r")` → `fstatSync(fd)` (mtime check on the same fd) → `readFileSync(fd)` (content read) → `closeSync(fd)`
- Added `finally` block to ensure the fd is always closed, even on error paths

Since `fstatSync` and `readFileSync` operate on the same file descriptor, the mtime check and content read are now atomic — no race window between check and use.

### 2. Immutability caveats + readonly typing

- **`_configCache`**: Added concise caveat comment (shared, read-only; mtime resolution ~1 s)
- **`tools()` in `ImmutablePrefix`**: Added JSDoc about frozen-returned-array contract; changed return type to `readonly ToolSpec[]`
- **`ChatRequestOptions.tools`** and **`StreamModelOptions.toolSpecs`**: Widened to `readonly ToolSpec[]` to accept the frozen array without forcing callers to cast

HUQIANTAO:
Fixed the CodeQL "Insecure temporary file" alert.

**Root cause:** All 3 `atomicWriteSync` call sites used predictable temp file names (`process.pid`, `Date.now()`, `Math.random()`), which could allow a symlink-following race.

**Fix:** Replaced with `crypto.randomBytes(8).toString("hex")` in all 3 locations:
- `src/config.ts` — `writeConfig()`
- `src/memory/session.ts` — `rewriteSession()`
- `src/code/edit-blocks.ts` — `atomicReplaceFileSync()`

The random hex string is 16 characters (64 bits of entropy), making the temp file name unpredictable. `randomBytes` was already imported in `config.ts`; added imports in the other two files.

All checks pass (tsc + biome).

esengine (approved):
Re-reviewed — the TOCTOU fix is exactly right: `openSync` → `fstatSync(fd)` → `readFileSync(fd)` → `closeSync(fd)` on all paths (cache-hit, post-read, error), so check and use happen on the same descriptor. **CodeQL now passes** (the high-severity file-race alert is gone) and full CI is green. The three caches and their invalidation were verified correct on the first pass. Merging.

PR #2163, opened 2026-05-28
perf: optimize tokenizer hot paths and prompt builder

HUQIANTAO:
## Summary

Four independent optimizations targeting CPU and allocation pressure in the token-counting and prompt-formatting pipeline. Together they reduce the per-turn overhead of the tokenizer from O(n²) to O(n) for large inputs.

### 1. `byteLevelEncode`: O(n²) → O(n) string building (`src/tokenizer.ts:186-191`)

The `+=` loop copies the entire accumulator string on each iteration. For a 10KB text chunk with ~10K UTF-8 bytes, this creates ~10K intermediate string allocations totaling ~50MB of transient copies.

**Fix**: Pre-allocated `string[]` + `.join("")`. Single allocation, O(n) copy.

### 2. `countTokens`: skip `number[]` allocation (`src/tokenizer.ts:261-263`)

`encode(text).length` builds a full `number[]` array of token IDs just to read `.length`. For a 50KB input string producing ~15K tokens, this allocates ~60KB of transient heap that's immediately discarded.

**Fix**: New `countTokens` counts tokens inline without building the ID array. Same BPE pipeline, zero allocation beyond the working pieces.

### 3. `formatDeepSeekPrompt`: O(n²) → O(n) prompt building (`src/tokenizer.ts:471-498`)

The `+=` loop for building the formatted prompt copies the entire accumulator on each message. In long sessions with 100K+ char prompts, this is a measurable bottleneck — O(n²) in total prompt length.

**Fix**: `parts[]` array + `.join("")`. Each message pushes 1-4 strings; final join is O(n).

### 4. `cachedBoundedTokens`: size guard for large strings (`src/tokenizer.ts:506-513`)

The `contentTokenCache` LRU (4096 entries) caches token counts keyed by content string. Without a size guard, 4096 entries of 50KB+ tool outputs retain 200MB+ of string keys in memory.

**Fix**: Skip caching strings >10KB. Small strings (the common case — user messages, short tool outputs) still cache normally. Large strings re-compute on demand, which is cheap because `countTokensBounded` samples only head+tail.

## Test plan

- `tests/tokenizer.test.ts` — 33 tests pass (encode, countTokens, countTokensBounded, formatDeepSeekPrompt)
- `tests/loop.test.ts` — 69 tests pass (buildMessages, healing, streaming, tool dispatch)

### DIFF
--- a/src/tokenizer.ts
+++ b/src/tokenizer.ts
@@ -185,9 +185,9 @@ function applySplit(chunks: string[], re: RegExp): string[] {
 /** UTF-8 bytes of `s`, each mapped to its byte-level visible char. */
 function byteLevelEncode(s: string, byteToChar: string[]): string {
   const bytes = new TextEncoder().encode(s);
-  let out = "";
-  for (let i = 0; i < bytes.length; i++) out += byteToChar[bytes[i]!];
-  return out;
+  const parts: string[] = new Array(bytes.length);
+  for (let i = 0; i < bytes.length; i++) parts[i] = byteToChar[bytes[i]!]!;
+  return parts.join("");
 }
 
 /** Repetitive tool output / identifier chunks re-encode thousands of times per session; LRU bounds at ~400KB. */
@@ -259,7 +259,38 @@ export function encode(text: string): number[] {
 }
 
 export function countTokens(text: string): number {
-  return encode(text).length;
+  if (!text) return 0;
+  const t = loadTokenizer();
+  let count = 0;
+
+  const process = (segment: string) => {
+    if (!segment) return;
+    let chunks: string[] = [segment];
+    for (const re of t.splitRegexes) chunks = applySplit(chunks, re);
+    for (const chunk of chunks) {
+      if (!chunk) continue;
+      const byteLevel = byteLevelEncode(chunk, t.byteToChar);
+      const pieces = bpeEncode(byteLevel, t.mergeRank);
+      for (const p of pieces) {
+        if (t.vocab[p] !== undefined) count++;
+      }
+    }
+  };
+
+  if (t.addedPattern) {
+    t.addedPattern.lastIndex = 0;
+    let last = 0;
+    for (const m of text.matchAll(t.addedPattern)) {
+      const idx = m.index ?? 0;
+      if (idx > last) process(text.slice(last, idx));
+      if (t.addedMap.has(m[0])) count++;
+      last = idx + m[0].length;
+    }
+    if (last < text.length) process(text.slice(last));
+  } else {
+    process(text);
+  }
+  return count;
 }
 
 export const DEFAULT_BOUNDED_TOKENIZE_CHARS = 2 * 1024;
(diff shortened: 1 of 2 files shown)


esengine (changes requested):
I traced all four optimizations against the originals — they're behavior-preserving, nicely done:

- **#1 byteLevelEncode** and **#3 formatDeepSeekPrompt**: pure `+=`→array+`join` rewrites. I checked the push order in formatDeepSeekPrompt matches the original concatenation token-for-token across system/user/assistant (USER_SP+content, ASSISTANT_SP+THINK_END, THINK_START+reasoning+THINK_END, content, tool-calls, EOS) — identical output.
- **#2 countTokens**: a faithful inline copy of `encode` (lines 221-259) — same split→byteLevelEncode→bpeEncode→vocab-skip pipeline and same addedPattern segmentation. The one cosmetic difference (`addedMap.has(m[0])` vs encode's `addedMap.get(m[0]) !== undefined`) is equivalent for a `Map<string, number>` since values are never `undefined`. Counts will match `encode(text).length`.
- **#4 cache size guard**: result-preserving — `countTokensBounded` is deterministic, so skipping the cache for >10KB just recomputes the same number.

One thing before merge, and it's the important one:

**#2 turns `countTokens` into a second copy of `encode`'s tokenization logic, and nothing keeps them in sync.** The old `encode(text).length` was DRY by construction. Now if anyone updates `encode` (new tokenizer version, a split-regex tweak, added-token handling) and forgets `countTokens`, they drift silently — and since `countTokens` feeds context-window/fold accounting, a 1-token drift corrupts budget decisions in a way that's very hard to trace. Please add a regression test that pins the equivalence: `expect(countTokens(x)).toBe(encode(x).length)` across a corpus — empty, plain ASCII, multibyte/CJK, emoji (surrogate pairs), text containing added/special tokens, and a >10KB blob. That single property test is what makes the duplication safe to live with.

Optional follow-up (not required): factor the shared pipeline into one `forEachToken(text, cb)` that both `encode` and `countTokens` drive, so there's literally one implementation. The equivalence test is the pragmatic minimum though.

CI hasn't run (fork) — I approved the workflow. Add the equivalence test and get CI green and this is a clean win.

HUQIANTAO:
Added the equivalence regression test as requested. 17 test cases covering:

- Empty string
- Plain ASCII / ASCII sentences
- CJK characters / CJK prose
- Mixed English+CJK
- Emoji (surrogate pairs)
- Special/added tokens (`<think>`, `</think>`, `｜begin▁of▁sentence｜`, etc.)
- Code snippets, JSON payloads
- Repeated text, whitespace-heavy input
- >10KB blob

Each asserts `countTokens(x) === encode(x).length`. All 50 tokenizer tests pass.

The optional `forEachToken` refactoring is a good follow-up — for now the test gives us the safety net to catch any future drift between the two paths.

esengine (approved):
Re-reviewed — the equivalence test is in: `describe('countTokens ↔ encode equivalence')` asserts `countTokens(text) === encode(text).length` across empty/ASCII/CJK/emoji/special-token cases plus a >10KB blob. That's the guard that makes the duplicated tokenization pipeline safe against future drift. All four optimizations were already verified correct on my first pass, and CI (incl. CodeQL) is green now. Merging — thanks.

PR #2164, opened 2026-05-28
fix(desktop): use native macOS fullscreen zoom

### LINKED ISSUE #2054: 桌面版应用无法全屏
环境：macos 14.7.1，apple m1 pro 机芯
问题：桌面版应用无法全屏，点击应用界面左上角全屏键只能把应用扩大到不含顶栏的整个桌面，无法像其他应用一样占满全屏
ps: 小问题，除了不美观不影响使用，有时间可以修一下，没时间可以暂时搁置

T1anjiu:
## Summary

Restore native macOS fullscreen behavior for the desktop app’s zoom control, so clicking the green window button enters real fullscreen instead of only expanding the window to the desktop work area.

## Problem

On macOS, the desktop app’s custom titlebar wired the green zoom button to window maximize behavior instead of native fullscreen. That produced a non-native result: the window expanded to fill the desktop area without entering a fullscreen Space.

The issue is in the desktop titlebar window-control path:
- `desktop/src/App.tsx` renders the custom macOS traffic-light controls
- the green button used `toggleMaximize()` instead of macOS fullscreen APIs
- the titlebar state tracking also treated the expanded state as maximize-only, so macOS fullscreen state was not modeled correctly

## Fix

- **`desktop/src/App.tsx`** — Added small window-state helpers so macOS reads expanded state from `isFullscreen()` and toggles it with `setFullscreen()`, while non-macOS platforms keep the existing maximize behavior.
- **`desktop/src/App.tsx`** — Updated titlebar listeners to sync both resize and fullscreen events so the control state stays accurate after entering or leaving fullscreen.
- **`desktop/src/App.test.ts`** — Added regression coverage verifying macOS uses native fullscreen for the zoom button and other platforms still use maximize.

## Verification

| Check | Result |
|---|---|
| `npm test -- desktop/src/App.test.ts` | ✅ passes |
| `npm test -- tests/desktop-user-message.test.ts` | ✅ passes |
| `npm run verify` | ✅ passes |

Closed #2054

### DIFF
--- a/desktop/src/App.test.ts
+++ b/desktop/src/App.test.ts
@@ -34,7 +34,7 @@ vi.mock("./theme", () => ({
   themeForStyle: vi.fn(() => "dark"),
 }));
 
-import { reduce } from "./App";
+import { readWindowExpanded, reduce, toggleWindowExpanded } from "./App";
 import { getThreadMaxWidth } from "./ui/thread-layout";
 
 function initialState(): Parameters<typeof reduce>[0] {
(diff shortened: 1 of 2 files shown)


esengine (approved):
LGTM. Verified the platform wiring: the `mac-ctrl zoom` button lives inside the `isMac ? (…)` block (App.tsx:2679), so hardcoding `true` in `toggleWindowExpanded(win, true, …)` is correct, and the `win-ctrl` button is inside the `isMac ? null : (…)` block (2812), so `false` there is correct too. State reads go through `readWindowExpanded` with the real `isMac`, and adding the `tauri://fullscreen` listener (not just `tauri://resize`) means the control icon stays in sync when fullscreen is toggled via macOS gestures/menu, not just the button. Extracting the helpers behind a `WindowControls` Pick type also makes them unit-testable, and you added the regression test. CI green.

Non-blocking nit: the two call sites hardcode `true`/`false` rather than passing the `isMac` variable — correct today because each is platform-gated, but passing `isMac` would be self-consistent if a button ever moves out of its block. Fine as-is. Merging.

PR #2175, opened 2026-05-28
perf: skip healing pipeline and token count when log is unchanged

HUQIANTAO:
## Summary

The agent loop calls `healActiveLogBeforeSend()` and `getLogTokens()` on every iteration, even when no new messages have been appended since the last call. Both functions run full-array scans with tokenization of every tool-call argument JSON — 3-4 passes over the entire log each time.

### Changes

**AppendOnlyLog dirty flag** (`src/memory/runtime.ts`):
- Add `_dirty` flag set on every mutation (`append`, `extend`, `compactInPlace`, `initWindow`)
- Add `consumeDirty()` method — returns `true` once per mutation, then `false` until the next mutation

**CacheFirstLoop** (`src/loop.ts`):
- Cache `healActiveLogBeforeSend()` result in `_healedCache`
- Skip the entire 4-pass healing pipeline when `consumeDirty()` returns `false`
- Clear the dirty flag after `compactInPlace` inside the healing path so the next call uses the cache

**ContextManager** (`src/context-manager.ts`):
- Cache `getLogTokens()` result in `_logTokensCache`
- Skip the full re-tokenization when `consumeDirty()` returns `false`

### Impact

For a typical 100-message session with 30 tool-call turns, this eliminates ~400 `countTokensBounded` calls + ~30 `JSON.stringify` calls + ~400 object spread copies per agent-loop iteration when the log is clean — the common case between the model's response and the next tool-call dispatch.

## Test plan

- [x] `tests/loop.test.ts` — 70 passed (1 skipped)
- [x] `npx biome check` — no issues
- [x] `npx tsc --noEmit` — clean

### DIFF
--- a/src/context-manager.ts
+++ b/src/context-manager.ts
@@ -110,11 +110,17 @@ function collectPinnedSkills(head: ChatMessage[]): { names: string[]; bodies: st
 }
 
 export class ContextManager {
+  private _logTokensCache = -1;
+  private _logTokensVersion = -1;
+
   constructor(private deps: ContextManagerDeps) {}
 
   /** Real-time token count of the current log — used by Desktop to refresh the
    *  context meter after /compact when no API usage event is available. */
   getLogTokens(): number {
+    if (this._logTokensCache >= 0 && this._logTokensVersion === this.deps.log.version) {
+      return this._logTokensCache;
+    }
     const entries = this.deps.log.toFullHistory();
     let total = 0;
     for (const e of entries) {
@@ -124,6 +130,8 @@ export class ContextManager {
         total += countTokensBounded(JSON.stringify(e.tool_calls));
       }
     }
+    this._logTokensCache = total;
+    this._logTokensVersion = this.deps.log.version;
     return total;
   }
 
--- a/src/loop.ts
+++ b/src/loop.ts
@@ -524,12 +524,23 @@ export class CacheFirstLoop {
   }
   private _inflightCounter = 0;
 
+  // Cached result from the last healActiveLogBeforeSend() pass.
+  // Invalidated when the log version changes (append/compactInPlace).
+  private _healedCache: ChatMessage[] | null = null;
+  private _healedVersion = -1;
+
   private buildMessages(): ChatMessage[] {
     const healedMessages = this.healActiveLogBeforeSend();
     return [...this.prefix.toMessages(), ...healedMessages];
   }
 
   private healActiveLogBeforeSend(): ChatMessage[] {
+    // Skip the expensive 4-pass healing pipeline when the log hasn't
+    // changed since the last call — the common case between iterations
+    // where no new messages were appended.
+    if (this._healedCache && this._healedVersion === this.log.version) {
+      return this._healedCache;
+    }
     const current = this.log.toFullHistory();
     const healed = healLoadedMessages(current, DEFAULT_MAX_RESULT_CHARS);
     const argsShrunk = shrinkOversizedToolCallArgsByTokens(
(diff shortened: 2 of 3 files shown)


esengine:
This is additive (skip healing + token count when the log is unchanged) and not covered by #2162, so it's still wanted. But #2162 just merged and changed the same `loop.ts`/`runtime.ts` hot path, so this now **conflicts** — please rebase onto current main. After rebase, make sure your 'unchanged' guard composes correctly with #2162's new `toFullHistory`/tool-spec caches (they should be complementary, but worth a sanity check that the version/length signal you key off isn't already short-circuited by the new caches).

HUQIANTAO:
Rebased onto current main (post-#2162 merge). Conflict resolved — `_dirty` flag coexists with #2162's `_fullHistoryCache`:

**Resolution:**
- Kept `_fullHistoryCache` from #2162 (avoids redundant disk I/O in `toFullHistory()`)
- Added `_dirty` flag for `consumeDirty()` (avoids redundant healing pipeline passes)
- Both are invalidated together in all mutation paths: `initWindow`, `append`, `compactInPlace`
- Kept `_frozenToolsCache` from #2162 (not reverted)
- `consumeDirty()` added at end of `AppendOnlyLog` class

**Composition with #2162's caches:**
- `_fullHistoryCache` caches the disk-read result (inner layer)
- `_dirty` flag lets the healing pipeline skip when nothing changed (outer layer)
- They're complementary: `_fullHistoryCache` avoids I/O, `_dirty` avoids CPU work

All 70 loop tests + 65 session tests pass. (Pre-existing `telegram/bot.ts` type errors on main.)

esengine (changes requested):
The goal is good (skip the 4-pass healing + token re-count when the log hasn't changed between iterations), but the `consumeDirty()` mechanism has a correctness bug, and it overlaps #2180.

**Bug — one destructive dirty flag, two consumers.** `consumeDirty()` is a check-and-**clear**, and it's called by *both* `getLogTokens()` (context-manager) and `healActiveLogBeforeSend()` (loop) on the same log. After an `append`, whichever runs first returns `true` and clears `_dirty`; the second then gets `false` and serves its **stale cache** — e.g. `getLogTokens` runs first, clears the flag, and `healActiveLogBeforeSend` returns `_healedCache` from *before* the append, so the model is sent a healed message list missing the latest message. A single consume-once flag can't fan out to independent consumers.

**Use a non-destructive version counter instead** — which is exactly what #2180 already does: a monotonic `_version` that each consumer compares against its own stored snapshot (read-only, no starvation). 

Since #2180 already lands the correct primitive for the healing cache, I'd suggest: **close this PR and fold its extra win — the `getLogTokens` memoization — into #2180** by having `getLogTokens` cache `{ version, tokens }` and recompute only when `log.version` changed. That gives you both optimizations with one correct mechanism.

(Separately: CI here is red only because of the stale slash-count / write_file tests that #2200 already fixed on main — not your code. A rebase clears those, but given the bug + overlap above, consolidating into #2180 is the better move.)

HUQIANTAO:
Closing in favor of #2180. The `consumeDirty()` approach had a correctness bug (destructive check-and-clear starving the second consumer when both `getLogTokens()` and `healActiveLogBeforeSend()` run on the same log). 

The `getLogTokens` memoization has been folded into #2180 using the non-destructive `_version` counter:
- `ContextManager._logTokensCache` + `_logTokensVersion` — recompute only when `log.version` changes
- `CacheFirstLoop._healedCache` + `_healedVersion` — same mechanism

Both consumers compare against `log.version` independently (read-only, no starvation).

PR #2176, opened 2026-05-28
perf(i18n): replace regex in t() with replaceAll

HUQIANTAO:
## Summary

Replace per-key `new RegExp(\`\\{${k}\\}\`, "g")` with native `replaceAll(\`{${k}}\`, String(v))` in the `t()` translation function — avoids regex construction overhead on every translation lookup.

The `readConfig` mtime cache that was originally in this PR has been dropped — it's already in #2162 (which uses an fd-based approach to avoid a TOCTOU race flagged by CodeQL).

## Files changed (1)

| File | Change |
|------|--------|
| `src/i18n/index.ts` | `result.replace(new RegExp(...))` → `result.replaceAll(...)` |

## Test plan

- [ ] `t("key", { name: "value" })` still substitutes `{name}` correctly
- [ ] Multiple params in one string all substitute

### DIFF
--- a/src/i18n/index.ts
+++ b/src/i18n/index.ts
@@ -116,7 +116,7 @@ export function t(path: string, params?: Record<string, string | number>): strin
   if (params) {
     let result = val;
     for (const [k, v] of Object.entries(params)) {
-      result = result.replace(new RegExp(`\\{${k}\\}`, "g"), String(v));
+      result = result.replaceAll(`{${k}}`, String(v));
     }
     return result;
   }


esengine:
Heads up — #2162 just merged and already includes a `readConfig` mtime cache, so that half of this PR is now redundant. Importantly, #2162's version reads through an **open file descriptor** (`openSync` → `fstatSync(fd)` → `readFileSync(fd)`) to avoid a TOCTOU race — CodeQL flagged the plain `statSync(path)`-then-`readFileSync(path)` pattern (which is what this PR uses) as a high-severity file race. So please **drop the readConfig hunk** here (it'd reintroduce the alert and conflicts with main).

The `t()` `replaceAll` change is genuinely useful and *not* in #2162 — please rebase and reduce this PR to just that one-liner (replacing the per-key `new RegExp(...,'g')` with `replaceAll`). Small, clean, happy to merge that on its own.

esengine (approved):
Re-reviewed — trimmed exactly as discussed: now just `src/i18n/index.ts`, dropping the redundant (and TOCTOU-prone) readConfig cache that #2162 already covers, keeping only the `t()` win — `result.replaceAll(\`{${k}}\`, String(v))` instead of `new RegExp(\`\\{${k}\\}\`, "g")` per key. Equivalent (literal global replace) and avoids constructing a RegExp per placeholder per call. CI green. Merging.

PR #2173, opened 2026-05-28
feat(cli): unified diff display for edit_file / write_file / multi_edit (#1251)

Bernardxu123:
## Summary

Add configurable diff display mode for file-editing tools in CLI, addressing issue #1251.

## Problem

Users had to run git diff after every file edit to see what changed — especially painful when working without git.

### 1. Enhanced write_file output with unified diff

write_file now returns diff-enriched output matching edit_file's format:

- **New file**: created path (N chars)
- **Overwrite (no changes)**: edited path (0→0 chars)
- **Overwrite (with diff)**: edited path (oldLen→newLen chars) + unified diff
- **Large file (>5000 lines / >100KB)**: summary only, no diff (performance protection)

Desktop's existing parseEditResult can now parse write_file results too — bonus.

### 2. /diff command

`
/diff              → show current mode (default: summary)
/diff summary      → path + stats (edited path +N -M)
/diff full         → unified diff with colored +/- lines
/diff none         → checkmark only
`

### 3. ToolCard diff-aware rendering

In ull mode, the ToolCard renderer parses unified diff from the tool output and renders with colored +/- lines, hunk headers, and file stats in the card meta. In summary and 
one modes, existing behavior is preserved.

## Files Changed

| File | Change |
|------|--------|
| src/tools/fs/edit.ts | Add generateWriteDiff() function |
| src/tools/filesystem.ts | write_file uses generateWriteDiff for diff-enriched output |
| src/config.ts | Add DiffDisplay type + loadDiffDisplay / saveDiffDisplay |
| src/cli/ui/cards/ToolCard.tsx | Diff-aware rendering based on config |
| src/cli/ui/slash/handlers/diff.ts | New /diff command handler |
| src/cli/ui/slash/commands.ts | Register /diff command |
| src/cli/ui/slash/dispatch.ts | Import diffHandlers |
| src/i18n/EN.ts | English translations for /diff |
| src/i18n/zh-CN.ts | Chinese translations for /diff |

## Design Decisions

- **No new Card type** — ToolCard is kept as-is with enhanced rendering; avoids breaking dashboard messages, session hydration, and transcript replay
- **Large file protection** — LCS diff is O(n*m); skipped for files >5000 lines or >100KB
- **Config persistence** — diffDisplay saved to ~/.reasonix/config.json via /diff command
- **Format alignment** — write_file output matches edit_file format so Desktop's parseEditResult also works

## Verification

- 
pm run lint — passed
- 
pm run typecheck — passed
- 
[...]

PR #2177, opened 2026-05-28
fix(desktop): render write_file results as DiffCard in thread view

Bernardxu123:
write_file results now render as DiffCard in Desktop thread view, matching edit_file / multi_edit behavior. Added write_file to the DiffCard condition in thread.tsx:212. parseEditResult correctly falls back to ToolCard for new files and large files. App.tsx sidebar has its own write_file handler (line 618) and is unaffected.

### DIFF
--- a/desktop/src/ui/thread.tsx
+++ b/desktop/src/ui/thread.tsx
@@ -209,7 +209,7 @@ export const AssistantMsg = memo(function AssistantMsg({
         />
       );
     }
-    if (s.result && (s.name === "edit_file" || s.name === "multi_edit")) {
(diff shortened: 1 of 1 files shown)


esengine (approved):
LGTM — minimal and correct. This is the desktop complement to #2173 (just merged): now that `write_file` emits the same diff-enriched output as `edit_file` from the shared tool layer, adding it to the DiffCard branch in the thread view is exactly right. `parseEditResult` falls through to default rendering when there's nothing to parse (new file / [diff too large]), so it's safe. CI green. Merging.

PR #2178, opened 2026-05-28
fix(qq): polish busy ingress and setup reminders

wade19990814-hue:
## Summary
- avoid showing new QQ messages in desktop transcripts when the active tab is busy and the message is rejected
- surface first-sender runtime binding as a reminder instead of an error and soften the settings copy
- add minimal staged `/qq connect` slash-input protection and refresh desktop QQ state after disconnect failures

## Testing
- `npm test -- --run tests/desktop-qq-ingress.test.ts tests/qq-channel.test.ts`
- `npm test -- --run tests/qq-first-connect.test.tsx`
- `npx biome check src/cli/commands/desktop.ts src/desktop/qq-turn-routing.ts src/qq/channel.ts src/qq/use-qq-channel.ts src/i18n/EN.ts src/i18n/zh-CN.ts tests/qq-channel.test.ts tests/qq-first-connect.test.tsx tests/desktop-qq-ingress.test.ts src/desktop/qq-ingress.ts`

## Notes
- After refreshing local dependencies, the staged first-connect regression now runs normally in this worktree.
- Full `npm run verify` now gets past the earlier dependency-resolution noise and only fails on the existing Windows symlink-permission failures in `tests/skills.test.ts` (`EPERM` from `symlinkSync`), which are unrelated to this QQ-only change set.

### DIFF
--- a/src/cli/commands/desktop.ts
+++ b/src/cli/commands/desktop.ts
@@ -79,6 +79,7 @@ import {
   collectMemoryEntriesForWorkspace,
   readMemoryEntryDetail,
 } from "../../desktop/memory-browser.js";
+import { classifyDesktopQQIngress } from "../../desktop/qq-ingress.js";
 import {
   parseQQRemoteDesktopCommand,
   qqRemoteCommandBypassesBusy,
@@ -92,6 +93,7 @@ import {
 import {
   clearQQTurnRouting,
   createQQTurnRoutingState,
+  hasQQPendingInteraction,
   markQQTurnFinished,
   markQQTurnStarted,
   setQQPendingInteraction,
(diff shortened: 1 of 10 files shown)


esengine (approved):
Reviewed at the structural level — scope is focused and well-contained (suppress rejected QQ messages in busy desktop transcripts, reframe first-sender binding as a reminder not an error, stage `/qq connect` input protection, refresh state after disconnect failure), and you added targeted tests (desktop-qq-ingress, qq-channel, qq-first-connect). This is UX polish on the existing QQ channel rather than new attack surface, and the softened first-sender copy is consistent with the fail-closed direction we took on Telegram (#2168). CI was green. Good to merge once mergeable recomputes — approving.

PR #2179, opened 2026-05-28
perf: cache frozen tool-spec snapshot in ImmutablePrefix.tools()

HUQIANTAO:
## Summary

- `tools()` was calling `structuredClone` on every tool spec on every call — the most expensive per-iteration operation in the agent loop. A typical session registers 30-50 tools; each inner-loop iteration (3-10 per turn) triggered a full deep clone of all of them.
- Replace with a frozen shallow-copy cache (`_frozenToolsCache`) that is rebuilt only when `addTool`/`removeTool` mutates the tool list. `Object.freeze` on each spec + its `function` sub-object prevents accidental mutation of the cached copy.

## Why this matters

`structuredClone` is one of the most expensive built-in operations in V8 — it serializes-then-deserializes via the structured clone algorithm, handling cycles, typed arrays, etc. None of that machinery is needed here; tool specs are plain JSON objects. For a 10-iteration turn with 40 tools, this eliminates **400 unnecessary deep clones**.

## How it works

- Added `_frozenToolsCache: ToolSpec[] | null` field to `ImmutablePrefix`
- `tools()` returns the cached frozen array when available; builds + freezes on first call
- `addTool()` and `removeTool()` null the cache alongside the existing `_fingerprintCache` invalidation
- `Object.freeze` on each spec and its `function` sub-object prevents accidental mutation

## Test plan

- [x] Type check passes (`tsc --noEmit`)
- [x] Biome lint passes
- [x] All existing tests pass
- [x] Manual: agent loop runs without mutation errors across multi-iteration turns

### DIFF
--- a/src/memory/runtime.ts
+++ b/src/memory/runtime.ts
@@ -16,6 +16,8 @@ export class ImmutablePrefix {
   readonly fewShots: readonly ChatMessage[];
   /** Invalidated by addTool / removeTool / replaceSystem; bypassing any of those leaves cache stale → fingerprint diverges from sent prefix. */
   private _fingerprintCache: string | null = null;
+  /** Frozen shallow-copy snapshot of _toolSpecs — avoids structuredClone per iteration. */
+  private _frozenToolsCache: ToolSpec[] | null = null;
 
   constructor(opts: ImmutablePrefixOptions) {
     this.system = opts.system;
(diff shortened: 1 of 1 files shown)


esengine:
Superseded by #2162, which just merged and includes this exact `_frozenToolsCache` change (frozen shallow snapshot in `ImmutablePrefix.tools()`, invalidated on addTool/removeTool) as one of its three cache layers. The implementation here is identical to what landed, so closing this as already-applied — thanks, the change is in main. Your other two perf PRs (#2180 healing-result cache, #2175 skip-when-unchanged) are *not* in #2162, so those stay open; please rebase them onto current main.

PR #2180, opened 2026-05-28
perf: cache healing pipeline result via AppendOnlyLog version counter

HUQIANTAO:
## Summary

- `healActiveLogBeforeSend()` runs 3 expensive passes (shrink oversized results, tokenize + shrink tool-call args, strip droppable reasoning) on every `buildMessages()` call — which happens 2-3 times per loop iteration.
- In the common case (nothing to heal), all 3 passes scan every message for nothing.
- Add a monotonic `_version` counter to `AppendOnlyLog` that bumps on every mutation. Cache the healed result in the loop alongside the log version at which it was computed. When the version matches, skip the entire healing pipeline.

## Why this matters

For a 10-iteration turn with 100 messages, `buildMessages()` is called 20-30 times. Without caching, each call runs:
1. `toFullHistory()` — full array copy
2. `healLoadedMessages` — walks every message, calls `shrinkOversizedToolResults` + `fixToolCallPairing`
3. `shrinkOversizedToolCallArgsByTokens` — walks every assistant message, `JSON.stringify` + `countTokensBounded`
4. `stripDroppableReasoningContent` — walks every message

With caching, the 2nd and 3rd calls per iteration (which see no log changes) return the cached result in O(1), eliminating **~2,000-3,000 unnecessary message inspections** and **~500 unnecessary tokenization calls** per turn.

## How it works

- Added `_version: number` to `AppendOnlyLog`, bumped in `append()`, `extend()`, `compactInPlace()`, `initWindow()`
- Added `get version()` accessor
- Added `_healedCache: { version: number; messages: ChatMessage[] } | null` to `CacheFirstLoop`
- `healActiveLogBeforeSend()` checks `_healedCache.version === this.log.version` — returns cached result when fresh
- After healing (with or without mutations), snapshots `this.log.version` into the cache

## Test plan

- [x] Type check passes (`tsc --noEmit`)
- [x] Biome lint passes
- [x] All existing tests pass (config, session, context-manager-fold-timeout)
- [x] Manual: multi-iteration agent turn runs correctly with healing cache active

### DIFF
--- a/src/context-manager.ts
+++ b/src/context-manager.ts
@@ -110,11 +110,17 @@ function collectPinnedSkills(head: ChatMessage[]): { names: string[]; bodies: st
 }
 
 export class ContextManager {
+  private _logTokensCache = -1;
+  private _logTokensVersion = -1;
+
   constructor(private deps: ContextManagerDeps) {}
 
   /** Real-time token count of the current log — used by Desktop to refresh the
    *  context meter after /compact when no API usage event is available. */
   getLogTokens(): number {
+    if (this._logTokensCache >= 0 && this._logTokensVersion === this.deps.log.version) {
+      return this._logTokensCache;
+    }
     const entries = this.deps.log.toFullHistory();
     let total = 0;
     for (const e of entries) {
@@ -124,6 +130,8 @@ export class ContextManager {
         total += countTokensBounded(JSON.stringify(e.tool_calls));
       }
     }
+    this._logTokensCache = total;
+    this._logTokensVersion = this.deps.log.version;
     return total;
   }
 
--- a/src/loop.ts
+++ b/src/loop.ts
@@ -524,12 +524,23 @@ export class CacheFirstLoop {
   }
   private _inflightCounter = 0;
 
+  // Cached result from the last healActiveLogBeforeSend() pass.
+  // Invalidated when the log version changes (append/compactInPlace).
+  private _healedCache: ChatMessage[] | null = null;
+  private _healedVersion = -1;
+
   private buildMessages(): ChatMessage[] {
     const healedMessages = this.healActiveLogBeforeSend();
     return [...this.prefix.toMessages(), ...healedMessages];
   }
 
   private healActiveLogBeforeSend(): ChatMessage[] {
+    // Skip the expensive 4-pass healing pipeline when the log hasn't
+    // changed since the last call — the common case between iterations
+    // where no new messages were appended.
+    if (this._healedCache && this._healedVersion === this.log.version) {
+      return this._healedCache;
+    }
     const current = this.log.toFullHistory();
     const healed = healLoadedMessages(current, DEFAULT_MAX_RESULT_CHARS);
     const argsShrunk = shrinkOversizedToolCallArgsByTokens(
@@ -538,9 +549,13 @@ export class CacheFirstLoop {
     );
     const pruned = stripDroppableReasoningContent(argsShrunk.messages);
     if (healed.healedCount === 0 && argsShrunk.healedCount === 0 && pruned.prunedCount === 0) {
+      this._healedCache = current;
+      this._healedVersion = this.log.version;
       return current;
     }
     this.log.compactInPlace(pruned.messages);
+    this._healedCache = pruned.messages;
+    this._healedVersion = this.log.version;
     if (this.sessionName) {
       try {
         rewriteSession(this.sessionName, pruned.messages);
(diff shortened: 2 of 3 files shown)


esengine:
Additive (caches the healing-pipeline *result*, distinct from #2162's `toFullHistory` cache), so still wanted — but #2162 just merged and touches the same `runtime.ts`/`loop.ts`, so this now **conflicts**. Please rebase onto current main. One thing to verify after rebase: your healing-result cache and #2162's history cache both invalidate off log mutations — confirm they share/agree on the invalidation signal so one can't serve a stale result the other already moved past. A test exercising heal→append→heal would pin it.

HUQIANTAO:
Rebased onto current main (post-#2162 merge). Conflict resolved in `runtime.ts` — both caches coexist:

**Cache invalidation consistency verified:**

| Mutation | `_fullHistoryCache` (#2162) | `_version` (#2180) |
|---|---|---|
| `initWindow` | `= null` | `++` |
| `append` | `= null` | `++` |
| `compactInPlace` | `= null` | `++` |

Both caches are invalidated on every mutation path. The healing pipeline cache (`_healedCache` in `loop.ts`) uses `this.log.version` (the `_version` counter) as its invalidation signal, while `_fullHistoryCache` uses `_totalLength`. Since `healActiveLogBeforeSend()` always calls `toFullHistory()` first (which checks `_fullHistoryCache`), the two caches can't serve conflicting results — the outer cache (healing) short-circuits only when the inner cache (I/O) would also hit.

The `compactInPlace` path correctly snapshots the post-bump version, so a subsequent `healActiveLogBeforeSend()` call sees a valid cache entry.

All 70 loop tests + 65 session tests pass. (`telegram/bot.ts` type errors are pre-existing on main from #2168.)

esengine (commented):
The mechanism here is correct — a monotonic `_version` bumped on every mutation (append/compactInPlace/initWindow, and extend via append) with a non-destructive `version` getter, and `_healedCache` served only when `version` matches. Unlike #2175's `consumeDirty()` (which clears the flag and starves the second consumer — see my review there), a read-only version counter fans out correctly to multiple consumers, so this is the right primitive.

Two things:
1. **Rebase onto current main.** CI is red here only because of the stale slash-count (43→44) and `write_file` output tests — those were fixed by #2200, which is now on main. A rebase picks up the fix and CI should reflect only your change.
2. **Consider absorbing #2175's other half.** #2175 also memoizes `getLogTokens()`, which is a real additional win but it used the buggy `consumeDirty()`. Since your `_version` counter is the correct vehicle, fold that in here (`getLogTokens` stores `{ version, tokens }`, recomputes on version change) and we can close #2175 as superseded.

Minor: `AppendOnlyLog` now carries both this `_version` and #2162's `_fullHistoryCache` (keyed by `_totalLength`). Not a problem — they track different things — but if you want, `_version` is a stronger invalidation signal than `totalLength` (which can repeat after a compaction), so a follow-up could unify both caches onto `_version`. Approving the approach; just rebase for green CI.

HUQIANTAO:
Updated per review feedback:

1. **Rebased onto latest main** (includes #2200 slash-count fix for green CI)
2. **Folded #2175's `getLogTokens` memoization** into this PR using the `_version` counter
3. **Closed #2175** — `consumeDirty()` bug is no longer in the codebase

### Changes from original

**`AppendOnlyLog`** — replaced `_dirty` flag with monotonic `_version` counter:
- Bumped on every mutation (`append`/`compactInPlace`/`initWindow`)
- `get version()` getter for non-destructive staleness checks
- No more check-and-clear — multiple consumers can read independently

**`CacheFirstLoop`** — `_healedCache` + `_healedVersion`:
- Skip healing pipeline when `log.version === _healedVersion`
- After `compactInPlace` (which bumps version), next call naturally recomputes

**`ContextManager`** — `_logTokensCache` + `_logTokensVersion`:
- Skip token recounting when `log.version === _logTokensVersion`
- Previously in #2175 as `consumeDirty()`, now correctly using `_version`

All 136 tests pass (loop + session).

esengine (approved):
Re-reviewed — this is exactly the consolidation I suggested, done right. The `_version` counter now backs **both** caches via non-destructive comparison: `_healedCache`/`_healedVersion` for the healing pipeline and `_logTokensCache`/`_logTokensVersion` for the token count, each storing its own last-seen `log.version`. No `consumeDirty()` — so the multi-consumer starvation bug from #2175 can't happen; both consumers read the same monotonic version independently. You've folded #2175's getLogTokens win in here correctly. CLEAN + CI green. Merging — and I'll close #2175 as superseded.

PR #2182, opened 2026-05-28
fix(ui): apply themed background to input area and app surface

HUQIANTAO:
## Summary

- Apply `SURFACE.bg` to the root App layout Box so the entire UI renders on the theme's background instead of the terminal's default
- Apply `SURFACE.bgInput` to the PromptInput and ComposerArea wrappers so the input area has a distinct themed surface
- On terminals whose default background differs from the theme (e.g. iTerm), this fixes: input text blending into the background, theme surface colors being invisible, and faint text disappearing

## Changes

1. **`src/cli/ui/App.tsx`** — merged `SURFACE` import, added `backgroundColor={SURFACE.bg}` to root layout Box
2. **`src/cli/ui/ComposerArea.tsx`** — added `SURFACE` import, added `backgroundColor={SURFACE.bgInput}` to outer composer wrapper
3. **`src/cli/ui/PromptInput.tsx`** — added `backgroundColor={SURFACE.bgInput}` to the bordered input Box (`SURFACE` was already imported)

## Test plan

- Launch CLI in iTerm (or any terminal with non-theme default background) — input area should have a distinct darker background
- Text should be clearly readable against the themed surface
- Switch themes with `/theme` — backgrounds should change accordingly
- Border, placeholder, and faint text should remain visible

### DIFF
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -198,8 +198,7 @@ import { InflightProvider } from "./state/inflight-context.js";
 import { AgentStoreProvider, useAgentState, useAgentStore } from "./state/provider.js";
 import { VerboseContext } from "./state/verbose-context.js";
 import { ThemeProvider } from "./theme/context.js";
-import { listThemeNames } from "./theme/tokens.js";
-import { FG, type ThemeName } from "./theme/tokens.js";
+import { FG, SURFACE, type ThemeName, listThemeNames } from "./theme/tokens.js";
 import { TickerProvider, useSlowTick } from "./ticker.js";
 import { handleTurnInterrupt } from "./turn-interrupt.js";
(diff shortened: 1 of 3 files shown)


esengine (approved):
Clean, small theming change — applies the theme's surface colors at the Ink-Box level: `SURFACE.bg` on the root app row and `SURFACE.bgInput` on the composer/prompt-input boxes. Unlike #2223's OSC 11 approach, this only paints the app's own rendered region (no terminal-state mutation to restore), so it's self-contained and safe. Tokens exist for every theme. I approved the workflow run; good to merge once CI is green. Note it's complementary to #2223 (in-app box bg vs actual terminal bg) — you may want only one depending on the look you're after, but they don't conflict.

PR #2186, opened 2026-05-28
fix(desktop): resolve FilePill click path against session-tracked files

Qunxi1:
## Summary

FilePill click resolution in the desktop markdown renderer blindly joined `workspaceDir + displayText`, so clicking a file reference like `config.yaml` would open `workspace/config.yaml` instead of the actual path used in the tool call (e.g. `workspace/backend/src/services/database/config.yaml`).

Now `resolveAgainstWorkspace` first attempts suffix-matching against `sessionFiles` — the real paths tracked from tool call arguments (`read_file`, `edit_file`, `write_file`, `multi_edit`). If a match is found, the tracked path is used as the base; if absolute it's returned directly, if relative it's resolved against the workspace. Falls back to the old behavior when no match is found.

Also excludes `src-tauri/target/` from Vite's file watcher to prevent EBUSY crashes on Windows.

## Problem

The desktop app has two independent file-path pipelines that never cross-reference:

| Pipeline | Source | Accuracy |
|---|---|---|
| Sidebar file tree | Tool call `args.path` | ✅ Real path |
| Inline FilePill | Regex match from AI text | ❌ workspaceDir + displayText |

`extractToolFiles()` already tracks real paths in `state.sessionFiles`, but FilePill resolution never consulted them.

## Changes

| File | Change |
|---|---|
| `desktop/src/Markdown.tsx` | Added `TrackedFile` type, `resolveBySessionFiles()` suffix matcher, extended `resolveAgainstWorkspace()` with session-file lookup, wired `ctx.sessionFiles` into `FilePill` and `SafeLink` |
| `desktop/src/App.tsx` | Pass `sessionFiles` through `WorkspaceProvider` value |
| `desktop/vite.config.ts` | Exclude `src-tauri/target/` from `server.watch.ignored` |

## How it works

Three match modes:
- **Exact match**: display text equals the tool arg path — use directly
- **Suffix match**: display text is the tail of a tracked path (e.g. `config.yaml` → `.../database/config.yaml`) — longest suffix wins
- **No match**: fall back to original `workspaceDir + displayText`

## What's not changed

- Display text in chat bubbles stays short (still shows whatever the AI wrote)
- `FILE_PATH_RE` regex matching unchanged
- Sidebar file tree resolution unchanged (was already correct)
- Right-click copy and hover tooltip unchanged

## Verification

`npm run verify` — ✅ passes (build, lint, typecheck, 301 test files / 3902 tests all passed)

### DIFF
--- a/desktop/src/App.tsx
+++ b/desktop/src/App.tsx
@@ -2241,7 +2241,7 @@ function TabRuntime({
 
   return (
     <WorkspaceProvider
-      value={{ dir: state.settings?.workspaceDir, editor: state.settings?.editor }}
+      value={{ dir: state.settings?.workspaceDir, editor: state.settings?.editor, sessionFiles: state.sessionFiles }}
     >
       <div
         className="app"
--- a/desktop/src/Markdown.tsx
+++ b/desktop/src/Markdown.tsx
@@ -3,11 +3,11 @@ import { openPath, openUrl } from "@tauri-apps/plugin-opener";
 import { Check, Copy, ExternalLink, FileText } from "lucide-react";
 import {
   Children,
+  type ReactNode,
   cloneElement,
   createContext,
   isValidElement,
   memo,
-  type ReactNode,
   useContext,
   useState,
 } from "react";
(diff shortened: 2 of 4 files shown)


esengine (commented):
Good fix — the old `workspaceDir + displayText` join was just wrong for any path the model abbreviates, and resolving against tracked `sessionFiles` (exact match, then suffix match) is the right approach since the real path is known from the tool-call args. Threading `sessionFiles` through the workspace context is clean.

One thing to confirm: the **suffix match is ambiguous when two tracked files share a tail** (e.g. `a/config.yaml` and `b/config.yaml`, click `config.yaml`) — `resolveBySessionFiles` returns the first match in iteration order, which may not be the intended file. Worth either (a) preferring an exact/longest-unique match and only suffix-matching when it's unambiguous, or (b) at least documenting that first-match-wins. Not a blocker over the current always-wrong-for-relative behavior, just a sharp edge. CI re-approved (fork).

Qunxi1:
The positional matching approach relies on the AI calling read_file in the same order as the files appear in the conversation text — the Nth FilePill maps to the Nth tool-call path. **If the order doesn't match (e.g. the AI reads files alphabetically but displays them in a different sequence), the resolution will be wrong.** In practice, with two ambiguous files the first one is usually correct, but with three or more the mismatch becomes likely. **To fully resolve this, the AI must be strictly prompted to call read_file in the exact display order.**

esengine (approved):
Approving — net improvement over the current always-wrong behavior. Resolving FilePill clicks against tracked `sessionFiles` (exact then suffix match) opens the actual tool-call path instead of blindly joining `workspaceDir + displayText`. CI green, CLEAN. The suffix-match ambiguity I noted (two tracked files sharing a tail → first-match-wins) stays a minor follow-up — still strictly better than today. Merging.

PR #2147, opened 2026-05-28
Add Baidu AI Search as a China-friendly web search backend

jimmyzhuu:
Fixes #2116.

## Why

`web_search` can be unreliable for users in mainland China today. Depending on the network path, Bing and other international providers may be slow, blocked, or require extra proxy setup, while SearXNG adds self-hosting overhead.

Baidu AI Search gives these users a local, official search backend that works with the existing `web_search` tool and can be configured from CLI, desktop, and dashboard surfaces.

## What changed

- Added `baidu` as a `web_search` engine using Baidu Qianfan AI Search.
- Supports `BAIDU_API_KEY`, `QIANFAN_API_KEY`, and `baiduApiKey` in config.
- Wired Baidu into `/search-engine`, config loading, public types, localized help, and web search errors.
- Added Baidu selection and API key entry to both desktop settings and the web dashboard.
- Updated the server settings API so dashboard mode can save `baiduApiKey` and show a masked saved-key prefix.
- Updated docs for Baidu setup; README only lists Baidu as a supported backend.

## Validation

- Smoke tested Baidu search from source and from the built `dist` package.
- Confirmed `topK=3` returns 3 Baidu results from source.
- Confirmed `topK=1` returns 1 Baidu result from `dist`.
- Added coverage for Baidu search requests, result mapping, config/env key precedence, `/search-engine baidu`, settings API persistence, dashboard settings, and public exports.
- Ran `npm run lint`, `npm run typecheck`, `npm run build`, full `npm test`, and `git diff --check`.

PR #2206, opened 2026-05-29
Fix Baidu and Metaso HTTP error classification before JSON parsing

jimmyzhuu:
## Summary

- Classify non-JSON HTTP error responses from Baidu AI Search before attempting to parse JSON
- Apply the same status-before-parse behavior to Metaso
- Preserve existing parse-error behavior for 2xx responses with invalid JSON

## Context

Follow-up to #2147. Review noted that Baidu web search parsed the response body before checking `resp.ok`, causing non-JSON 401/429/500 upstream responses to be reported as parse errors instead of more useful auth, rate-limit, or server errors.

While checking the current implementation, Metaso had the same JSON API pattern, so this PR fixes both affected providers only.

## Test Plan

- [x] `npm test -- tests/web-tools.test.ts -t "searchMetaso|searchBaidu"`
- [x] `npm test -- tests/web-tools.test.ts`
- [x] `npm run lint -- src/tools/web.ts tests/web-tools.test.ts`
- [x] `npm run build` (run by pre-push verify)
- [x] `npm run lint` (run by pre-push verify)
- [x] `npm run typecheck` (run by pre-push verify)
- [ ] `npm run test --silent` (pre-existing unrelated failures on current main: slash command count expects 43 but receives 44 in `tests/slash.test.ts` and `tests/ui-slash-suggestions.test.tsx`; auto-git-rollback write_file output expectation in `tests/auto-git-rollback.test.ts`)

### DIFF
--- a/src/tools/web.ts
+++ b/src/tools/web.ts
@@ -408,13 +408,6 @@ async function searchMetaso(query: string, opts: WebSearchOptions = {}): Promise
   }
 
   const raw = await resp.text();
-  let data: MetasoSearchResponse;
-  try {
-    data = JSON.parse(raw) as MetasoSearchResponse;
-  } catch {
-    throw new Error(t("webErrors.metasoParseError", { status: resp.status }));
-  }
-
   if (!resp.ok) {
     if (resp.status === 401 || resp.status === 403) {
       throw new Error(t("webErrors.metasoUnauthorized"));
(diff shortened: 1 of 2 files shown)


esengine (approved):
Exactly the fix I flagged when #2147 landed — and applied symmetrically to both Metaso and Baidu. Moving the `if (!resp.ok)` status block ahead of `JSON.parse` means a non-JSON error body (e.g. a 500 HTML page or a 401) now surfaces the specific status error (`...ServerError(500)` / `...Unauthorized`) instead of a misleading `parseError`, while a 2xx with genuinely bad JSON still throws `parseError`. Tests cover both branches (500 non-JSON → server error, 200 invalid JSON → parse error). CI green. Merging.

PR #2208, opened 2026-05-29
fix(lifecycle): show plan suggestion once per session

GTC2080:
## Summary

- show the high-risk lifecycle plan suggestion only once per active session
- keep the suggestion UI-only and preserve the explicit plan-first auto-arm path
- add a regression test for already-shown suggestion suppression

## Notes

This follows the non-blocking review note from #2119. The change does not auto-arm lifecycle, block prompts, change prompt text, or touch tool schemas. A new named session can still show the hint again; repeated matching prompts in the same session stay quiet.

## Validation

- `npm test -- tests/lifecycle-plan-suggestion.test.ts`
- `npm test -- tests/lifecycle-explicit-intent.test.ts tests/lifecycle-plan-suggestion.test.ts tests/code-prompt.test.ts tests/lifecycle.test.ts tests/slash.test.ts`
- `npx biome check src/cli/ui/lifecycle-plan-suggestion.ts tests/lifecycle-plan-suggestion.test.ts src/cli/ui/App.tsx`
- `npm run typecheck`
- `npm run lint`
- `npm run verify`
- pre-push `npm run verify`

### DIFF
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -625,6 +625,7 @@ function AppInner({
       mode: engineeringLifecycleBaseModeRef.current,
     });
   }
+  const lifecyclePlanSuggestionSessionRef = useRef<string | null | undefined>(undefined);
   // Refs that mirror state for stable read-callbacks handed to the
   // embedded dashboard server. The server's `getXxx()` closures are
   // captured once at startDashboard time; without ref-mirrors the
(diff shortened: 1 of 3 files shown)


esengine (approved):
Exactly the follow-up I flagged on #2119 — and cleanly done. `lifecyclePlanSuggestionSessionRef` (keyed by session id) gates the hint to once per session via the new `alreadySuggested` flag, stays UI-only, preserves the explicit plan-first auto-arm path, and resets for a new session. The regression test ('does not repeat the suggestion after it was already shown in the session') pins it. CI green. Merging.

PR #2209, opened 2026-05-29
feat(skills): add --skill-dir flag and per-skill enable/disable

cliffffffffff:
## Summary

Two new features for the skill system:

### 1. Specify skill folder directory (`--skill-dir`)

Users can now pass `--skill-dir <path>` to `reasonix code` or `reasonix chat` to add an extra skill directory. Also supports:
- **Environment variable**: `REASONIX_SKILL_DIR`
- **Config field**: `skills.skillDir` in `~/.reasonix/config.json`

The directory is appended to custom paths with priority: `project > custom > skillDir > global`.

### 2. Per-skill enable/disable

Users can disable specific skills so they no longer appear in the Skills index or can be invoked:

- **CLI**: `/skill disable <name>` / `/skill enable <name>`
- **Agent tool**: `disable_skill` / `enable_skill`
- **Dashboard API**: `PATCH /api/skills/toggle/:name`
- **Config storage**: `skills.disabled` array in `config.json`

Disabled skills are excluded from `list()`, `read()`, and `applySkillsIndex()` (system prompt).

## Changes

| File | Change |
|------|--------|
| `src/config.ts` | Extend `ReasonixConfig.skills` with `skillDir` + `disabled`; add `loadSkillDir`, `resolveSkillDir`, `loadDisabledSkills`, `toggleSkillDisabled` |
| `src/skills.ts` | Add `skillDir`/`disabledSkills` to `SkillStoreOptions`; filter in `roots()`/`list()`/`read()` |
| `src/tools/skills.ts` | Add `enable_skill`/`disable_skill` tools; pass new options to `SkillStore` |
| `src/cli/ui/slash/handlers/skill.ts` | Add `/skill enable`/`disable` commands; show disabled in list |
| `src/server/api/skills.ts` | Add `PATCH` toggle endpoint; include `disabled` in `GET` response |
| `src/memory/user.ts` | Thread `skillDir`/`disabledSkills` through `applyMemoryStack` |
| `src/cli/index.ts` | Add `--skill-dir` flag to `code` and `chat` commands |
| `src/cli/commands/code.tsx` | Thread `skillDir` to `codeSystemPrompt` |
| `src/code/prompt.ts` | Accept and forward `skillDir`/`disabledSkills` |
| `src/code/setup.ts` | Pass `skillDir`/`disabledSkills` to `registerSkillTools` |
| `src/cli/ui/App.tsx` | Pass `skillDir`/`disabledSkills` to `registerSkillTools` |
| `src/i18n/{EN,zh-CN,de}.ts` | New i18n keys for enable/disable messages |
| `tests/skills.test.ts` | 9 new tests for `skillDir` + `disabledSkills` |
| `tests/skills-toggle.test.ts` | 8 new tests for `toggleSkillDisabled` |

## Test plan

- [x] `tsc --noEmit` — 0 errors
- [x] `vitest run tests/skills.test.ts` — 59 tests passed (50 existing + 9 new)
- [x] `vitest run tests/skills-toggle.test.ts` — 8 tests passed
- [x] `vitest run tests/tools-skills.test.ts` — 32 tests passed (no regression)
- [x] `vitest run tests/context-manager-skill-pin.test.ts` — 6 tests passed (no regression)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


esengine (changes requested):
Nice feature pair (`--skill-dir` with env/config fallbacks + the documented priority `project > custom > skillDir > global`, and per-skill enable/disable wired through CLI `/skill`, the dashboard, and agent tools). The disable mechanism is clean — persisted via `toggleSkillDisabled`, loaded by `loadDisabledSkills`, passed into `SkillStore` to filter from both the index and invocation, with a '(disabled)' listing.

Blocker — **CI is red on lint, not tests** (your own code):
- biome formatting violations (e.g. the `: undefined` / `const resolvedSkillDir = …` lines need reformatting),
- a 'change to an optional chain' lint hit (~line 892),
- and **duplicate setup/teardown hooks** in the test file (biome's `noDuplicateTestHooks`).
Run `biome check --write` (or your `npm run lint -- --fix`) and dedupe the test hooks, and that clears.

One design question (non-blocking): the agent-facing `disable_skill` / `enable_skill` tools persist to the user's `~/.reasonix/config.json`. Letting the model permanently disable a skill in the user's config is a bit strong — a user might be surprised a skill stayed off across sessions because the agent turned it off mid-task. Consider scoping agent-initiated disables to the session (or gating them), while keeping the explicit `/skill disable` persistent. Your call, just flagging it.

Fix the lint and this is close. (CI's CodeQL + the rest were fine; it's purely the biome step.)

cliffffffffff:
Fixed the 3 biome violations:
- Optional chain for `envDir` check in `resolveSkillDir`
- Merged duplicate `afterEach` hooks in both test blocks

Lint passes clean now. CI should be green.

esengine (changes requested):
Thanks for the quick turnaround — you addressed my earlier lint asks and the conflict is resolved, so the branch is MERGEABLE again. The core feature is cleanly built: resolveSkillDir prefers REASONIX_SKILL_DIR over skills.skillDir and appends the skillDir as a custom-scope root after existing custom paths (matching the documented project > custom > skillDir > global precedence); toggleSkillDisabled dedupes/sorts via a Set, cleans up to undefined when the list empties, preserves other skills.* fields, and persists through writeConfig; the filter is enforced in SkillStore.list()/read() and the builtins loop so disabled skills drop from both the index and invocation. The dedicated skills-toggle.test.ts plus the new skillDir/disabledSkills blocks (including the negative applySkillsIndex() case) hit the right code paths, and EN/zh-CN/de all carry the new keys (ja falls back to EN). Nice work.

A few blockers before I can approve:

1. de.ts has out-of-scope deletions (blocking). Diffing against the true merge-base, this branch's own commit removes a bunch of unrelated German search-engine/webErrors strings — bing-intl, baidu, ollama, and it trims brave (usageBingIntl, usageBaidu, usageOllama, switchedBaiduNote, the ollama*/fetchOllama* error blocks, and rewrites the engine-list strings). Those exist on current main (Baidu landed in #2147 after your branch point), so merging would silently drop German translations and break de parity. Please rebase de.ts on current main and keep only the 5 new skill keys, restoring the removed engine entries.

2. qq builtin test coverage was dropped (blocking). tests/skills.test.ts no longer asserts the qq builtin (the 'ships ... as builtins' test was renamed to drop /qq, and the store.read('qq') body/scope/runAs and 'qq —' index checks were removed) — but src/skills.ts still registers the qq builtin (name: "qq", BUILTIN_QQ_BODY). That looks like the merge accidentally reverted main's qq tests while keeping the source. Please restore the qq assertions so the test matches what the source still ships.

3. CI hasn't actually run (blocking). The CI and CodeQL runs on 792f048a are sitting in action_required — queued awaiting my approval, so there's no green result yet. I'll approve the run; once it completes, please confirm the biome lint step, build(ubuntu), build(windows), and CodeQL are all green.

Non-blocking:
- The agent-facing disable_skill/enable_skill still persist to ~/.reasonix/config.json, so a model disabling a skill mid-task leaves it off across sessions. Consider scoping agent-initiated disables to the session while keeping /skill disable persistent — your call.
- Could you add one test pinning the full priority chain project > custom > skillDir > global in a single ordering assertion? You test skillDir-after-custom and project-override separately today.

Once de.ts is scoped back down, the qq tests are restored, and CI is confirmed green, I'm happy to re-review and approve.

esengine (approved):
Useful — --skill-dir plus per-skill enable/disable is the right primitive for managing custom skill collections. Letting CI run.

esengine:
CI is red on both Ubuntu and Windows builds with three real regressions — please address before this can land:

1. **`tests/comment-policy.test.ts > no banner separator comments`** — the new code in this PR is using `// ─── … ───` style separator banners. That's banned by the project's CLAUDE.md and enforced by this test. Group exports by file/section structure instead of banner comments.

2. **`tests/skills.test.ts > SkillStore`** — four symlink behaviors regressed (`follows symlinked flat <name>.md`, `silently skips a broken symlink`, `loads skills from symlinked directories (#2104)`, `skips broken symlinks gracefully`). The store refactor needs to preserve the symlink-following codepath that #2104 added.

3. **`tests/slash.test.ts > /language > /skill list localizes built-in skill descriptions in Chinese`** — looks like the per-skill enable/disable wiring broke the localization lookup for built-in skill descriptions.

The `--skill-dir` flag + enable/disable design is the right shape — once these three are fixed I'll take another look.

PR #2213, opened 2026-05-29
feat(telegram): improve remote rendering and ingress

PorunC:
## Summary

- Preserve Telegram reply rendering by converting Markdown-ish assistant output into Telegram MarkdownV2-safe text.
- Replace remote confirmation replies with Telegram inline buttons, including ✅ run/allow and ❌ deny actions.
- Register local slash commands as Telegram bot commands, including Telegram-safe aliases for unsupported command names.
- Harden Telegram remote ingress with per-user rate limiting and safer bot error handling.

## Review Follow-up

This PR includes follow-up fixes for the non-blocking items called out in https://github.com/esengine/DeepSeek-Reasonix/pull/2168#pullrequestreview-4385790102:

- Rate limiting: authorized Telegram users are capped at 5 text messages per 30 seconds before the message reaches the agent prompt path. Excess messages are rejected with a Telegram reply, so spam cannot grow unbounded work.
- Bot token handling + polling errors: Telegram bot errors are formatted through a redacting helper, with clearer handling for invalid bot tokens (`401`) and polling conflicts (`409`). The same sanitization is used for initialization, polling, command registration, callback acknowledgements, and sendMessage failures.

## Verification

- `npm run verify`

### DIFF
--- a/dashboard/src/styles.css
+++ b/dashboard/src/styles.css
@@ -3025,6 +3025,9 @@ html:not([data-platform="macos"]) .shortcut kbd[data-key="mod"] {
   flex-direction: column;
   z-index: 50;
 }
+.popup.at-popup {
+  width: min(720px, calc(100vw - 32px));
+}
 .popup .ph {
   padding: 8px 12px;
   font-family: inherit;
@@ -3046,7 +3049,11 @@ html:not([data-platform="macos"]) .shortcut kbd[data-key="mod"] {
 }
 .popup-list {
   overflow-y: auto;
+  overflow-x: hidden;
   padding: 4px;
+}
+.at-popup-list .popup-item {
+  min-width: 0;
 }
 .popup-item {
   display: grid;
@@ -3075,11 +3082,19 @@ html:not([data-platform="macos"]) .shortcut kbd[data-key="mod"] {
 .popup-item .nm {
   font-weight: 500;
   color: var(--fg);
+  min-width: 0;
+  overflow: hidden;
 }
 .popup-item .nm .cmd {
   font-family: inherit;
   color: var(--accent);
   margin-right: 6px;
+}
+.popup-item .mention-name {
+  display: block;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
 }
 .popup-item .desc {
   display: block;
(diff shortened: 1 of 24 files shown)


PorunC:
The failing check looks like a Windows-only flaky filesystem test rather than a Telegram regression.

Only `build (windows-latest · node 22)` failed. Ubuntu and CodeQL passed, and the Telegram-related tests passed in the Windows log as well. The failing test is the pre-existing `tests/filesystem-tools.test.ts > directory_tree collapses directories with >50 entries`; it timed out at Vitest's 5s limit, and the later `ENOTEMPTY` errors appear to be cleanup fallout from that timeout on Windows temp dirs.

I tried to rerun the failed job with `gh run rerun --failed`, but my account does not have admin rights on this repository:

```text
run 26616356461 cannot be rerun; Must have admin rights to Repository.
```

Could someone with permissions rerun the failed Windows job?

esengine (approved):
Re-reviewed — solid, and it closes both follow-ups I raised on #2168. 

- **MarkdownV2 escaping is complete and correct**: `TELEGRAM_MARKDOWN_V2_SPECIAL_RE` covers the full special-char set (`_*[]()~`>#+-=|{}.!`), backslash is escaped first (right order), and code spans use a separate escaper (only `\` + backtick). That fixes the 'Telegram rejects unescaped text' failure mode.
- **Inline-button confirmations are access-controlled**: `handleCallbackQuery` does `if (!this.acceptRemoteInput(userId)) return` on `query.from.id`, so the fail-closed access model from #2168 extends to button presses — an unauthorized user can't tap Approve on a pending command/file/plan confirmation.
- **Rate limiting added** (`acceptRateLimit`, 5 msgs/window) — exactly the queue-bounding I asked for.

CI is green now (the earlier failure was a flaky Windows `directory_tree` timeout, unrelated — it passed on re-run). Merging.

Two non-blocking notes for a later pass: (1) confirmation `callbackData` is just `"1"` — fine while confirmations are serialized one-at-a-time, but if concurrent confirmations ever become possible it'd need a confirmation id; (2) consider correlating the press to the originating `message_id` so a stale button from an old confirmation can't approve a newer one.

PR #2216, opened 2026-05-29
fix(i18n): convert ja.ts to spread-fallback pattern across all three layers

HUQIANTAO:
## Summary
- Convert all three `ja.ts` locale files (TUI, Desktop, Dashboard) to use the `...en`/`...EN` spread-fallback pattern, matching the existing `de.ts` convention
- **Dashboard**: rewrite from standalone 1333-line file to spread pattern; keep only 5 translated sections, remove ~30 English-copy sections
- **Desktop**: add `...en.sectionName` spreads to all 30+ sections including nested objects
- **TUI**: add `import { EN }`, top-level `...EN`, and `...EN.sectionName` spreads to all 46+ sections including deeply nested objects (slash, handlers, planFlow, ui, wizard, modelPicker)

## Why
Previously, missing keys in ja.ts returned `undefined` at runtime. With the spread pattern, any untranslated key gracefully falls back to the English string.

## Test plan
- [ ] Verify `npm run typecheck` passes (JA.ts compiles against TranslationSchema)
- [ ] Switch language to Japanese in dashboard/desktop/TUI and confirm no `undefined` strings appear
- [ ] Spot-check that translated strings (e.g. settings, chat placeholders) still display correctly


esengine:
Good refactor — I checked, it **preserves the real Japanese translations** (21 non-ASCII lines removed from the verbose form, the same 21 re-added as overrides on top of `...EN`); the -878 is just dropping keys that were byte-identical to English, which the spread fallback now covers. Matches the `de.ts` convention and cuts maintenance. Only issue: it's **CONFLICTING** with main (recent merges added i18n keys to the same `ja.ts` files). Please rebase onto current main — the spread-fallback pattern makes that robust since any new keys auto-fall-back to EN. Then it's good to merge.

esengine (approved):
Re-reviewed after rebase — still clean: 26 Japanese lines removed from the verbose form, the same 26 re-added as overrides on `...EN`, so no translation is lost; the reduction is purely EN-identical keys now covered by the spread fallback (matching de.ts). CLEAN + CI green. Merging.

PR #2171, opened 2026-05-28
fix(ui): apply themed background to input area and app surface

HUQIANTAO:
## Summary

Three issues making the TUI unusable on iTerm (and any terminal with a non-default background):

### 1. Themed backgrounds not applied

The theme system defines `SURFACE.bgInput` (`#0f172a`) and `SURFACE.bg` (`#0b1020`) but **no component applies them**. The entire UI renders with the terminal's default background, making input text invisible and the theme ineffective.

- `PromptInput.tsx` — add `backgroundColor={SURFACE.bgInput}` to the outer Box
- `App.tsx` — add `SURFACE` import and `backgroundColor={SURFACE.bg}` to root content layout
- `ComposerArea.tsx` — add `SURFACE` import and `backgroundColor={SURFACE.bgInput}` to wrapper Box

### 2. WCAG AA contrast failures on fg.faint and fg.meta

`fg.faint` failed WCAG on all 4 dark themes (ratios 1.63–2.56:1, need 4.5:1). `fg.meta` failed on midnight and was borderline on others. Dark theme `bgElev` was too close to `bgInput` for surface distinction.

| Theme | meta | faint | bgElev |
|-------|------|-------|--------|
| dark | #778294→#9aa5b5 | #4d5666→#8791a3 | #151d2f→#1c2844 |
| light | #6b7280→#5c6371 | #9ca3af→#666d7b | — |
| midnight | #565f89→#9da5bb | #414868→#8a92a8 | — |
| deep-blue | #808080→#909090 | #606060→#8c8c8c | — |
| high-contrast | — | — | — |

### 3. ANSI `dim` causing invisible text across all UI components

The `dim` attribute (SGR 2) halves the brightness of the foreground color. On dark terminals, dimmed text without an explicit color uses the terminal default at half brightness — effectively invisible. This affects **22 components** across the entire TUI.

**Initial fix** (earlier commits):
- `ReasoningCard.tsx` — replace `dim` with explicit `FG.sub` for card metadata
- `ToolCard.tsx` — replace conditional `dim` with explicit `FG.faint`/`errColor`
- `chat.tsx` — emit OSC 11 to set terminal default background, reset via OSC 111 on exit

**Full fix** (this commit): Replace every remaining `dim` prop across 22 UI components with explicit `color={FG.faint}` or conditional color logic:
- mcp-browse, AtMentionSuggestions, DiffApp, EditConfirm
- McpBrowser, McpHub, McpMarketplace
- PlanReviseConfirm, PlanStepList
[...]

PR #2221, opened 2026-05-29
fix(ui): replace ANSI dim with explicit FG.faint colors; WCAG contrast fixes

HUQIANTAO:
## Summary

Replace the ANSI `dim` attribute (SGR 2) with explicit `FG.faint` color references across all UI components. The `dim` attribute halves foreground brightness, producing near-invisible text on dark terminals where the default foreground is already low-contrast.

Also fixes WCAG AA contrast ratios for `fg.faint` and `fg.meta` across all dark themes, and adds OSC 11 terminal background emission so iTerm and similar terminals inherit the theme's surface background.

Extracted from #2171 (closed) — the background fix portion is handled separately by #2182.

## Commits

1. **fix(theme): meet WCAG AA contrast for fg.faint and fg.meta across all themes** — Adjusts color values in `tokens.ts` for dark, light, midnight, deep-blue themes to meet 4.5:1 contrast ratio
2. **fix(ui): replace ANSI dim with explicit colors; set terminal background via OSC 11** — ReasoningCard, ToolCard dim→explicit color; chat.tsx emits OSC 11/111 for terminal bg
3. **fix(ui): replace all remaining ANSI dim with explicit FG.faint colors** — 22 components: replaces `dim` prop with `color={FG.faint}` or conditional color logic

## Files changed (26)

`tokens.ts` (WCAG), `chat.tsx` (OSC 11), `ReasoningCard.tsx`, `ToolCard.tsx`, plus 22 components with dim→FG.faint replacement.

## Test plan

- [ ] Dark theme: all "dimmed" text (metadata, timestamps, inactive rows) is readable
- [ ] Light theme: faint text has sufficient contrast
- [ ] iTerm: terminal background matches theme surface (no white gaps)
- [ ] Verify no remaining `dim` props in UI components

### DIFF
--- a/src/cli/commands/mcp-browse.tsx
+++ b/src/cli/commands/mcp-browse.tsx
@@ -12,6 +12,7 @@ import {
   specStringFor,
 } from "../../mcp/registry-fetch.js";
 import type { RegistryEntry } from "../../mcp/registry-types.js";
+import { FG } from "../ui/theme/tokens.js";
 
 const VISIBLE_ROWS = 12;
 
@@ -175,16 +176,18 @@ function McpBrowseApp() {
         <Text bold color="ansi:cyan">
           ◈ MCP marketplace
         </Text>
-        <Text dim>{`  ·  ${state.status}`}</Text>
+        <Text color={FG.faint}>{`  ·  ${state.status}`}</Text>
       </Box>
       <Box marginTop={1}>
         <Text>search: </Text>
         <Text color="ansi:white">{state.query || "(type to filter)"}</Text>
-        <Text dim>{`  ${filtered.length} match${filtered.length === 1 ? "" : "es"}`}</Text>
+        <Text
+          color={FG.faint}
+        >{`  ${filtered.length} match${filtered.length === 1 ? "" : "es"}`}</Text>
       </Box>
       <Box marginTop={1} flexDirection="column">
         {window.length === 0 ? (
-          <Text dim>{state.loading ? "loading…" : "no entries"}</Text>
+          <Text color={FG.faint}>{state.loading ? "loading…" : "no entries"}</Text>
         ) : (
           window.map((e, i) => {
             const idx = (start || 0) + i;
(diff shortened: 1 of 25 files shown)


esengine (changes requested):
The core change is a genuine accessibility win — replacing `<Text dim>` (SGR 2, which halves brightness and goes near-invisible on dark terminals) with explicit `FG.faint` across the UI, plus the WCAG-AA contrast bumps to `fg.faint`/`fg.meta`. That part I'm happy with (it does change existing themes' faint/meta values, but deliberately and for the better). The dim→FG.faint sweep across the 26 components is consistent and CI's green.

But the PR also bundles a more invasive, separate change with a real failure mode: **setting the terminal's background via OSC 11** (`\x1b]11;${themeBg}\x07`) before `render()`, restored via OSC 111 only **after `await waitUntilExit()`**.

1. **Restore isn't crash-safe.** The OSC 111 restore only runs on a clean exit. If the process is SIGKILL'd, crashes, or exits on an unhandled signal, the restore never fires and the user's terminal is **left repainted with the theme background** — a bad state they have to fix manually. Please move the restore to `process.on("exit")` plus SIGINT/SIGTERM handlers so it fires on every exit path.
2. **Confirm it's intended / consider splitting.** Overriding the user's terminal background is more than 'replace ANSI dim' implies — some users deliberately set their terminal bg. Either gate it (opt-in) or at least call it out; ideally it's its own PR separate from the a11y sweep.

The dim→faint + contrast work is ready; it's the OSC 11 terminal-bg override (crash-safety + intent) that I'd want resolved before merge.

HUQIANTAO:
Updated: the OSC 11 terminal background change has been split out to a separate PR (#2223) with crash-safe restore (`process.once("exit")` + SIGINT/SIGTERM handlers). 

This PR now contains only:
1. WCAG AA contrast fixes for `fg.faint` and `fg.meta` across all themes
2. `dim` → explicit `FG.faint`/`FG.sub` color replacements in ReasoningCard, ToolCard, and 22 other UI components

esengine (approved):
Re-reviewed — you split the OSC 11 terminal-background change out to #2223 (0 OSC refs here now), so this PR is exactly the part I endorsed: the dim→`FG.faint` sweep across the UI plus the WCAG-AA contrast bumps to `fg.faint`/`fg.meta`. That's a clean accessibility win (SGR 2 dim is genuinely near-invisible on dark terminals), CI green. Merging.

PR #2222, opened 2026-05-29
fix(telegram): bind callback confirmations

PorunC:
## Summary

- Bind Telegram inline-button callback data to a confirmation id instead of using bare `"1"`, `"2"`, `"3"` payloads.
- Correlate Telegram callback presses with the active confirmation `message_id`, so stale buttons from older confirmations cannot resolve a newer prompt.
- Add Telegram setup docs in English and Chinese.

## Review Follow-up

This resolves the non-blocking follow-up from https://github.com/esengine/DeepSeek-Reasonix/pull/2213#pullrequestreview-4386524119:

- confirmation `callbackData` now includes a confirmation id (`tg:<id>:<choice>`)
- callback presses are accepted only from the currently active Telegram confirmation message
- stale callbacks, including old buttons after a restart, are ignored

## Verification

- `npm run verify`

### DIFF
--- a/docs/telegram-connect.md
+++ b/docs/telegram-connect.md
@@ -1,0 +1,183 @@
+# Telegram channel setup
+
+Reasonix can attach Telegram to an existing `chat` or `code` session as a remote channel. Telegram is not a third runtime mode.
+
+Once connected, Telegram can:
+
+- send normal user messages into the active session
+- receive follow-up assistant replies with Telegram Markdown rendering
+- continue confirmation, choice, checkpoint, and plan-style follow-up interactions with inline buttons
+- expose Reasonix slash commands through the Telegram bot command menu
+
+## Before you start
+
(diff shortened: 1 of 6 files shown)


esengine (approved):
This addresses both follow-ups I raised on #2213, and cleanly. Callback data is now `tg:${confirmationId}:${choice}` (via `telegramCallbackData`/`parseTelegramCallbackData`) instead of bare `"1"`, and presses are gated twice: `callback.confirmationId !== interactionRef.current.confirmationId` rejects a button from a superseded confirmation, and `query.message.message_id !== this.activeCallbackMessageId` correlates the press to the active prompt — so a stale button can't resolve a newer confirmation. The fail-closed `acceptRemoteInput(userId)` gate on the callback path is preserved. CI green. Nice hardening — merging.

Issue #2159, opened 2026-05-28
桌面版对话在流式输出时不会自动滚动到底部（0.53.0 Windows）/ Desktop app does not auto-scroll during streaming responses (0.53.0 Windows)


steadyoak:
**Symptom**

* [x] Other

对话内容在实时生成过程中不会自动滚动到底部。

模型持续输出 token 时，当前视图不会跟随最新内容移动，需要反复手动向下滚动才能继续查看生成内容。

即使没有主动滚动离开底部，也会出现停止自动跟随输出的情况。

---

**When it happens**

* [x] During assistant streaming (token-by-token output)

---

**Terminal — the important part**

* [x] Other:

使用的是桌面安装版：

`Reasonix_0.53.0_x64-setup.exe`
`desktop-v0.53.0`

系统：
Windows 11

不是在 VSCode / Cursor / Windows Terminal 中运行，而是直接运行桌面版应用。

---

**Diagnostic dump**

N/A（桌面版应用，不是 terminal 环境）

---

**Reproduction**

1. 打开桌面版 Reasonix
2. 让 assistant 生成较长回复
3. 在流式输出过程中，视图不会持续自动滚动到底部
4. 需要不断手动滚动才能查看最新输出内容

该问题稳定复现。

---

**Screen recording**

暂未提供。

---

**Symptom**

* [x] Other

The conversation view does not auto-scroll during streaming responses.

While new tokens are being generated, the viewport does not stay pinned to the latest output, so I have to repeatedly scroll down manually to continue reading.

This can happen even without intentionally scrolling away from the bottom.

---

**When it happens**

* [x] During assistant streaming (token-by-token output)

---

**Terminal — the important part**

* [x] Other:

Using the standalone desktop installer version:

`Reasonix_0.53.0_x64-setup.exe`
`desktop-v0.53.0`

OS:
Windows 11

This is not running inside VSCode, Cursor, or Windows Terminal.
The issue occurs in the standalone desktop application itself.

---

**Diagnostic dump**

N/A (desktop application, not running inside a terminal emulator)

---

**Reproduction**

1. Launch the desktop version of Reasonix
2. Start a long assistant response
3. During streaming output, the viewport stops following the newest content
4. Manual scrolling is required repeatedly to continue reading

The issue is consistently reproducible.

---

**Screen recording**

Not provided for now.

steadyoak:
感谢各位❤

Bernardxu123:
[...]

PR #2235, opened 2026-05-29
fix(desktop): follow streaming output when user is at bottom (#2159)

nianyi778:
## 问题

关闭 #2159。桌面版流式输出时不自动滚动到底部。

## 根因

`useEffect` 没有依赖数组（每次渲染都执行），但没有检查 `atBottomRef.current`：

```ts
// 旧代码：无依赖数组 → 每次渲染执行，且不检查用户是否在底部
useEffect(() => {
  const s = virtScrollerRef.current;
  if (!s || messageItems.length === 0) return;
  const id = requestAnimationFrame(() => { s.scrollTop = s.scrollHeight; });
  return () => cancelAnimationFrame(id);
});
```

问题：
1. 无依赖数组 → React 在高频 re-render 时可能跳过某些效果的执行
2. 不检查 `atBottomRef.current` → 用户向上滚动查看历史时也会被强制拉回底部

## 修复

添加 `[messageItems]` 依赖数组 + `atBottomRef.current` 检查：只有新消息到来且用户在底部时才滚动。

### DIFF
--- a/desktop/src/App.tsx
+++ b/desktop/src/App.tsx
@@ -2018,13 +2018,19 @@ function TabRuntime({
     }
   }, [messageItems.length]);
 
-  // Always scroll to bottom on every content change.
+  // Follow the bottom while the assistant is streaming and the user hasn't
+  // scrolled up. The dependency on messageItems.length covers new messages;
+  // atBottomRef guards against re-pinning when the user intentionally scrolled
+  // up to read earlier content (#2159).
   useEffect(() => {
     const s = virtScrollerRef.current;
     if (!s || messageItems.length === 0) return;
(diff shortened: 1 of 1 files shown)


nianyi778:
@codex review

esengine (approved):
Correct fix for #2159. The auto-scroll effect now guards on `atBottomRef.current` (don't re-pin to bottom when the user has scrolled up to read history) with a re-check inside the rAF, and gains a `[messageItems]` dep so it stops firing on every render. Follows streaming output when the user is at the bottom, leaves them alone when they're not. Small and right, CI green. Merging.

PR #2236, opened 2026-05-29
fix(loop): keep headless NEEDS_PRO escalation one-shot

BeiZi6:
## Problem

In headless `reasonix run`, `<<<NEEDS_PRO>>>` should retry only the current turn on `deepseek-v4-pro`. Mutating `this.model` without restoration leaves the session pinned to pro for later turns, which breaks the one-shot contract called out in #2195.

## Fix

- Detect `<<<NEEDS_PRO...>>>` before persisting the flash response and retry the turn on `deepseek-v4-pro`.
- Track the actual `callModel` so stats, cache diagnostics, and session log metadata are attributed correctly.
- Restore the base model before terminal exits and forced-summary paths so following turns return to flash.

## Testing

- `npm run verify`
- push pre-hook `npm run verify`

Related to #2195.

### DIFF
--- a/src/loop.ts
+++ b/src/loop.ts
@@ -73,6 +73,10 @@ function formatSteerUserMessage(content: string): string {
   return [MID_TURN_STEER_WRAPPER, content].join("\n");
 }
 
+function parseNeedsProEscalation(content: string): boolean {
+  return /^\s*<<<NEEDS_PRO(?::\s*[^>\n]{1,150})?>>>/.test(content);
+}
+
 export {
   fixToolCallPairing,
   formatLoopError,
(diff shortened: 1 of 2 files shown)


esengine (approved):
This is the correct one-shot fix for the headless `<<<NEEDS_PRO>>>` escalation — exactly what was missing in #2195. You capture `baseModelForTurn = this.model`, use a (possibly escalated) `callModel` for the call, and `restoreModelIfNeeded()` resets `this.model` back to the base in `finally`, so a single escalation retries *this turn* on pro and the next turn is back on flash — matching the contract in src/prompt-fragments.ts ('retries this turn on deepseek-v4-pro, one shot') rather than permanently pinning pro. The anchored `^\s*<<<NEEDS_PRO(?::…)?>>>` regex is also tighter than a loose substring match, and you added a test. CI approved (fork). Approving — good fix. (I'll point #2195 here as superseded.)

PR #2242, opened 2026-05-29
fix: restore browser dashboard file mentions

btulga:
## What

  Restore browser dashboard `@` file mentions by adding server compatibility endpoints for the web bridge's file search and preview requests.

  ## Why

  The browser dashboard was calling `GET /api/files/search` and `GET /api/file-read`, but the server only exposed the older file APIs. As a result, typing `@` in the browser dashboard could not return
  file mention results or previews.

  ## How to verify

  Run:

  ```bash
  npm test -- tests/server-dashboard.test.ts
  npm run verify

  ## Checklist

  - [x] npm run verify passes locally (lint + typecheck + tests + comment-policy gate)

### DIFF
--- a/src/server/api/file-read.ts
+++ b/src/server/api/file-read.ts
@@ -92,3 +92,31 @@ export async function handleFileRead(
     closeSync(fd);
   }
 }
+
+export async function handleFileReadPreview(
+  method: string,
+  _rest: string[],
+  _body: string,
+  ctx: DashboardContext,
+  query: URLSearchParams = new URLSearchParams(),
+): Promise<ApiResult> {
+  if (method !== "GET") return { status: 405, body: { error: "GET only" } };
+  const filePath = query.get("path") ?? "";
+  if (!filePath) return { status: 400, body: { error: "path query parameter required" } };
+
+  const read = await handleFileRead("GET", [filePath], "", ctx);
+  if (read.status !== 200) return read;
+
+  const body = read.body as { content?: unknown; path?: unknown };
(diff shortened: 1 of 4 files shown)


esengine (approved):
Re-reviewed against the actual diff (my earlier comment was based on bad local tooling output — disregarded).

This PR adds two server-side compatibility endpoints the browser dashboard already calls: `GET /api/files/search` (delegates to `walk(cwd, prefix)`) and `GET /api/file-read` preview (delegates to `handleFileRead`). Verified:
- `walk()` is scoped to `cwd`, RESULT_CAP=50, MAX_DEPTH=4, sensible SKIP_DIRS/SKIP_EXTS; `q` is only used as a lowercased substring filter on the relative path — never joined into a path, so no traversal vector.
- The preview reuses `handleFileRead`, which keeps its path-traversal guard (resolve + startsWith(cwd+sep) → 403) and FD-based read (fstat/read on the open fd, no TOCTOU).
- Test covers both the search match and the preview head/totalLines shape. CI green (build + CodeQL).

One non-blocking nit: `handleFileReadPreview` reads `query.get("path")` (already percent-decoded by URLSearchParams) and passes it to `handleFileRead`, which calls `decodeURIComponent(rest.join("/"))` again — a double-decode. Traversal is still blocked by the guard, so it's not a security issue, but a filename containing a literal `%` would resolve to the wrong path. Consider skipping the second decode for this call path. Not blocking the merge.

LGTM.

PR #2226, opened 2026-05-29
feat(desktop): Plans tab in context panel

z1223473749:
## What
Add a dedicated Plans tab to the desktop context panel with:
- Master-detail split layout (130px plan list + Markdown body)
- Status badges (active / pending / refining / completed / cancelled)
- Truncated plan names with custom tooltip on hover
- Auto-archive on dismiss_plan, plan_cleared, cancel, revision_required
- Per-workspace localStorage keying to prevent cross-project pollution
- Auto-open context panel on new plan_required, auto-release sidebar on complete
- PlanApprovalCard in thread.tsx now shows compact summary + refinement progress bar with 'view full plan in right panel →' hint instead of full Markdown body

## Why
Users need a persistent view of their plans without interrupting the conversation flow. The current plan approval cards bloat the thread with full plan Markdown; moving plan detail to the context panel keeps the conversation readable while making plan state always accessible.

## How to verify
1. Start a session with plan mode, let model submit a plan
2. Plan appears in context panel Plans tab with status badge
3. Approve — plan shows progress in panel, thread shows compact card
4. Refine — plan updates to 'refining' state with pulse animation
5. Complete all steps — plan auto-archives, panel clears
6. Switch workspace — plan archives from previous workspace are isolated

## Checklist

- [ ] `npm run verify` passes locally (lint + typecheck + tests + comment-policy gate)
- [ ] No `Co-Authored-By: Claude` trailer in commits
- [ ] Comments follow CONTRIBUTING.md (no module-essay headers, no incident history)
- [ ] No edits to `CHANGELOG.md` — release notes are maintainer-written at release time

PR #2249, opened 2026-05-29
feat(shell): expand ~ and $VAR in run_command args after allowlist check (#2105)

nianyi778:
## 问题

关闭 #2105。模型经常写出像这样的命令：

```bash
cat ~/.config/some-file
ls $PROJECT_DIR/src
grep -r 'TODO' ~/notes/
```

但 `run_command` 直接解析 argv，不走 shell，导致 `~` 和 `$VAR` 作为字面量传给子进程，静默输出错误结果。

## 方案

添加 `expandShellVars()`，在 **allowlist 检查之后** 展开 `~` 和 `$VAR`/`${VAR}`：

- `~` → `homedir()`（命令开头或链操作符后）
- `$VAR` / `${VAR}` → `process.env[VAR]`，未设置则保留原样
- Glob（`*`/`?`）**不展开** — 防止通过字符串拼接绕过 allowlist

allowlist 检查仍在原始字符串上进行，安全模型不变。

### DIFF
--- a/src/cli/ui/App.tsx
+++ b/src/cli/ui/App.tsx
@@ -124,6 +124,7 @@ import { ModelPicker } from "./ModelPicker.js";
 import { PathConfirm } from "./PathConfirm.js";
 import { PlanCheckpointConfirm } from "./PlanCheckpointConfirm.js";
 import { PlanConfirm, type PlanConfirmChoice } from "./PlanConfirm.js";
+import { PlanPanel } from "./PlanPanel.js";
 import { PlanRefineInput } from "./PlanRefineInput.js";
 import { PlanReviseConfirm, type ReviseChoice } from "./PlanReviseConfirm.js";
 import { PlanReviseEditor } from "./PlanReviseEditor.js";
@@ -804,14 +805,11 @@ function AppInner({
   const noTakeoverOverlay =
     !pendingShell &&
     !pendingPath &&
-    !pendingPlan &&
-    !pendingReviseEditor &&
     !pendingSessionsPicker &&
     !pendingEditPicker &&
     !pendingWorkspacePicker &&
     !pendingCheckpointPicker &&
     !pendingMcpHub &&
-    !stagedInput &&
     !pendingEditReview;
   // Plan-mode indicator —displayed in the StatsPanel, mirrored onto
   // the ToolRegistry so dispatch enforces read-only. Toggled via the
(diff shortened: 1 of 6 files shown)


nianyi778:
@codex review

nianyi778:
Fixed P1 in `c20398cd`. After `expandShellVars`, if the expanded command differs from the original, the sensitive-path check runs again on the expanded form. If `isCommandAllowed(expandedCmd)` returns false (e.g. `cat $HOME/.ssh/id_rsa` → `cat /home/user/.ssh/id_rsa` hits the sensitive-path demotion), the confirmation gate is called before execution — same flow as a non-allowlisted command.

The two other P1/P2 findings (App.tsx composer-during-plan-modal and PlanPanel.tsx revise-editor) are pre-existing issues in main unrelated to this PR's diff.

@codex review

nianyi778:
@codex review

nianyi778:
The two Codex findings (PlanPanel.tsx revise-editor and App.tsx composer-during-plan-modal) are **pre-existing issues in main** unrelated to this PR's diff — they appear in the plan panel work from #2226/#2227, not in the shell expansion changes. This PR only touches `src/tools/shell/parse.ts` and `src/tools/shell.ts`.

esengine (changes requested):
Thanks for tackling #2105. The shell-expansion core here is reasonable, but this PR has significant scope creep that blocks it as-is.

**Scope creep (blocking):** This is described as a shell `~`/`$VAR` expansion change, but the diff also adds an entirely new `src/cli/ui/PlanPanel.tsx` (+625) and substantially rewrites `src/cli/ui/App.tsx` (+76/-56). Those plan-panel changes are unrelated to #2105. Please split them into a separate PR so the shell change can be reviewed and merged on its own.

**CI is red:** biome fails on `src/cli/ui/PlanPanel.tsx:19` (useImportType) among others — run `npx biome check --write`.

Heads-up: #2250 also implements #2105 (focused, shell-only, with tests, and I verified its allowlist-bypass safety). We'll likely converge on one of these — if you'd prefer, the cleanest path is to keep this PR for the plan panel and let #2250 carry the #2105 fix, or strip the plan-panel work out here. Let me know which you'd like.

nianyi778:
Superseded by #2260 which contains only the shell-expansion changes. The plan-panel code in this diff came from the branch base, not my changes.

esengine:
Superseded by #2264, which is the same shell-expansion logic with the plan-panel scope creep removed (just the three shell files). I've reviewed and approved #2264's code + security model there. Recommend closing this PR in favor of #2264 once you're happy with it — keeping the plan-panel work for a separate, clearly-scoped PR if you still want it. Thanks for splitting it out so quickly.

PR #2250, opened 2026-05-29
feat(shell): 支持 ~、$VAR、glob 展开 (#2105)

zhangyapu1:
## 中文

`run_command` 执行前添加 shell 展开步骤，解决模型频繁踩坑问题。

**新增功能：**
- `~` / `~/` 展开为用户主目录（`homedir()`）
- `$VAR` / `${VAR}` 从 `process.env` 读取，未定义的变量保留原样
- glob `*` / `?` 在 cwd 下匹配文件，无匹配保留原样

**修改文件：**
- `src/tools/shell/parse.ts` — 新增 `expandShellTokens()` 函数
- `src/tools/shell/exec.ts` — `runCommand()` tokenization 后调用展开
- `src/tools/shell-chain.ts` — `runPipeGroup()` 每个 segment 调用展开
- `src/tools/shell.ts` — 更新工具描述，标注已支持的展开类型
- `tests/shell-tools.test.ts` — 12 个新测试覆盖各场景

**安全说明：**
- allowlist 检查在展开前执行，安全模型不变
- glob 展开限制在 cwd 内
- `$VAR` 仅读 `process.env`，不支持 `$(cmd)` 子命令

---

## English

Add shell expansion before `run_command` execution to resolve frequent model pitfalls.

**New features:**
- `~` / `~/` expands to the user home directory (`homedir()`)
- `$VAR` / `${VAR}` reads from `process.env`; undefined variables pass through as-is
- Glob `*` / `?` matches files in cwd; unmatched patterns pass through as-is

**Changed files:**
- `src/tools/shell/parse.ts` — new `expandShellTokens()` function
- `src/tools/shell/exec.ts` — call expansion after tokenization in `runCommand()`
- `src/tools/shell-chain.ts` — call expansion per segment in `runPipeGroup()`
- `src/tools/shell.ts` — update tool description to reflect supported expansions
- `tests/shell-tools.test.ts` — 12 new tests covering all expansion scenarios

**Security notes:**
- Allowlist check runs before expansion; security model unchanged
- Glob expansion is scoped to cwd only
- `$VAR` reads `process.env` only; no `$(cmd)` subcommand support

Closes #2105

PR #2264, opened 2026-05-29
feat(shell): expand ~ and $VAR in run_command args after allowlist check (#2105)

nianyi778:
Supersedes #2249 with shell-only changes (no plan-panel scope creep).

See #2249 for the full discussion. This PR is identical in shell logic but has no unrelated files.

### DIFF
--- a/src/tools/shell.ts
+++ b/src/tools/shell.ts
@@ -11,7 +11,7 @@ import {
   type RunCommandResult,
   runCommand,
 } from "./shell/exec.js";
-import { isCommandAllowed } from "./shell/parse.js";
+import { expandShellVars, isCommandAllowed } from "./shell/parse.js";
 
 export {
   BUILTIN_ALLOWLIST,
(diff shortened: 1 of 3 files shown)


nianyi778:
@codex review

nianyi778:
Fixed P2 in latest push. `expandShellVars` is now quote-aware: rewrote from a chain of global regex replacements to a character-by-character scanner that tracks quote state. Tokens inside single quotes are passed through literally, so `rg '$HOME' docs` stays `rg '$HOME' docs` rather than expanding to the home path.

@codex review

esengine (approved):
This is exactly the clean, shell-only re-do I asked for on #2249 — no plan-panel scope creep, just the three shell files. Code and security both check out:

**Security (verified — no allowlist bypass):** I read the call ordering in src/tools/shell.ts — `tokenizeCommand(cmd)` (the un-expanded argv) → `expandShellTokens(argv, cwd)` → `prepareSpawn(expandedArgv)`. Expansion happens only at the argv level right before spawn, while the allowlist gate runs upstream on the raw command string, so `$VAR`/glob can't widen the allowlist decision. The doc comment on expandShellTokens states this contract explicitly, and expansion never re-introduces shell metacharacters. Same safe architecture as the sibling #2250, which I also verified.

**Cleaner than #2250:** `expandShellTokens` uses a local `let tok = tokens[i]` (no function-parameter reassignment) and `expandEnvVar` does `return token.replace(...)` directly — so you avoid the `noParameterAssign` biome error that #2250's version tripped. The $VAR/${VAR} regex and the ~-only-leading rule look correct, and glob also handles `[…]` classes.

**One thing before I can merge:** this is a fork PR and its CI + CodeQL runs are sitting in `action_required` (not yet run). I need to get the workflow run approved and confirm build(ubuntu), build(windows), and CodeQL are all green before merging — I'll sort that out. The code itself LGTM and supersedes #2249.

Non-blocking: consider a quick test asserting that an allowlisted command stays allowlisted after expansion (e.g. `cat $HOME/x` still gates as `cat`), to pin the bypass-safety contract on the changed path.

nianyi778:
The single-quote P2 is already fixed in the latest commit `c8336ed4` — `expandShellVars` was rewritten as a char-by-char scanner that tracks quote state. Single-quoted spans are now passed through literally.

@codex review

nianyi778:
Fixed P2 (metachar safety) in latest push. Added `quoteEnvValue()`: values containing shell metacharacters (`> | ; & < ( ) $ \\` etc.) are wrapped in single quotes so they can't be re-parsed as syntax after expansion. Safe plain-word values (no metacharacters) are inserted as-is.

@codex review

esengine (approved):
Right shape — allowlist check first on the raw command, then expand for the actual exec. Supersedes #2250 which I'm closing as duplicate.

nianyi778:
@esengine CI 4/4 ✅，已 approved — 方便的话麻烦合并，感谢 🙏

Bernardxu123:
🙏 **真诚感谢您的贡献！**

由于 v1 已不再接受代码变更，此 PR 将被关闭。

再次感谢您的时间和精力！❤️

PR #2267, opened 2026-05-29
fix(desktop): clean up animations — remove breathing lights, fix broken keyframes

HUQIANTAO:
## Summary
- Fix 2 broken `@keyframes` references (`caret` → `stream-caret`, `fade` → `fade-in`)
- Remove meaningless breathing/pulsing animations that don't convey state
- Wire tab status dots to actual session busy/idle state
- Unify duplicate keyframes

### Bug fixes
- `@keyframes caret` was never defined → shell running caret animation was silently broken
- `@keyframes fade` was never defined → about modal mask animation was silently broken

### Animation cleanup
- **Tab status dots**: were hardcoded to `data-state="running"` — every tab pulsed forever regardless of state. Now wired to actual `busy` state via `onBusyChange` callback from `TabRuntime`
- **YOLO mode border**: removed `yolo-pulse` breathing animation, kept static danger border
- **PendingUserMsg**: removed `pulse-soft` breathing from the entire message bubble background
- **ThinkingPill label**: removed shimmer gradient text effect — was doubling with bounce dots
- **MainHead "Running" label**: removed shimmer — plain accent text is cleaner
- **Skeleton lines**: unified `@keyframes shimmer-bg` (duplicate) into `shim`

### Removed dead code
- `@keyframes pulse-soft` — no longer referenced
- `@keyframes yolo-pulse` — no longer referenced

### DIFF
--- a/desktop/src/App.tsx
+++ b/desktop/src/App.tsx
@@ -1375,9 +1375,10 @@ interface TabRuntimeProps {
   onToggleSide: () => void;
   onToggleCtx: () => void;
   onToggleCurrency: () => void;
-  tabsList: { id: string; workspaceDir?: string }[];
+  tabsList: { id: string; workspaceDir?: string; busy?: boolean }[];
   activeTabId: string;
   setActiveTabId: (id: string) => void;
+  onBusyChange?: (tabId: string, busy: boolean) => void;
 }
 
 function TabRuntime({
@@ -1412,6 +1413,7 @@ function TabRuntime({
   tabsList,
   activeTabId,
   setActiveTabId,
+  onBusyChange,
 }: TabRuntimeProps) {
   const [state, dispatch] = useReducer(reduce, {
     ready: false,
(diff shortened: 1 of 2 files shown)


esengine (approved):
Verified the core fix from the diff. The two broken animations are genuinely fixed: `@keyframes caret` and `@keyframes fade` were never defined (so the shell-running caret and the fade were silent no-ops), and this renames the references to `stream-caret` and `fade-in` which ARE defined post-change. I cross-checked every `animation:` declaration in the resulting styles.css (spin, stream-caret, status-pulse, fade-in) and all four resolve to a defined `@keyframes` — no dangling references, and the removed keyframes (caret/fade/breathe/glow/pulse) have no remaining users. The tab status-dot rewrite reads the real `tab.busy` field (already used before this PR), so it's wired to actual session state. Net -25 lines, CI green on build (ubuntu+windows) + CodeQL.

One honest caveat: this is a desktop visual change, so I can't pixel-verify the new status-dot rendering or confirm that removing the breathing animations looks the way you intend — but the change carries no correctness/data/security risk and is trivially reversible. If the dots or spacing look off in practice, a follow-up screenshot or revert is easy. Approving on the strength of the verified keyframe fix and the clean, real-field wiring.

Minor (non-blocking): `@keyframes pulse-glow` appears to remain defined but unused after this change — worth removing in a future pass if it's dead.

PR #2270, opened 2026-05-29
fix(weixin): trim terminal qr dependency

PorunC:
## Summary
- replace the heavier `qrcode` terminal renderer dependency with `qrcode-generator` for Weixin terminal QR output
- keep terminal QR rendering behind the existing URL fallback, including the refresh path
- route Weixin QR login/refresh/scanned prompts through i18n for EN, zh-CN, JA, and ru
- add focused tests for refresh QR rendering and URL-only fallback behavior

## Why qrcode-generator instead of qrcode-terminal
I tried the lighter `qrcode-terminal` path first, but it is not compatible with this project's strict build pipeline: its published CommonJS file contains legacy octal escape sequences like `\033`, which esbuild/tsup rejects in strict mode during `npm run build`.

`qrcode-generator` avoids that build failure while still trimming the dependency footprint compared with `qrcode@1.5.4` and its transitive tree. It also provides the exact primitive this flow needs: deterministic terminal ASCII QR generation, with the existing URL fallback preserved when rendering fails.

## Review follow-up
Addresses the optional follow-ups from https://github.com/esengine/DeepSeek-Reasonix/pull/2246#pullrequestreview-4388053238:
- reduces QR dependency weight
- locks the refresh-path QR render in tests
- locks the URL-only fallback branch in tests
- moves the hardcoded Weixin QR prompts into i18n, including Russian

## Verification
- `npm run verify`
  - 314 test files passed
  - 4043 tests passed, 9 skipped


esengine (approved):
Good fix. The Windows CI failure on `auto-git-rollback.test.ts` is a pre-existing flake on that suite (timeout + ENOTEMPTY rmdir race), unrelated to this change. Note for the qrcode-generator choice: agreed with your reasoning — qrcode-terminal's literal \033 escapes break tsup strict mode.

Bernardxu123:
🙏 **真诚感谢您的贡献！**

由于 v1 已不再接受代码变更，此 PR 将被关闭。

再次感谢您的时间和精力！❤️

PR #2293, opened 2026-05-29
feat(shell): expand leading ~ and reject unsupported shell expansion …

MyPrototypeWhat:
## What

`run_command` / `run_background` run with `shell: false` and literal, tokenized argv — by design, since the per-segment allowlist and sensitive-path gating depend on argv being literal. This adds two improvements without weakening that model: (1) a leading `~` / `~/` is now expanded to the home directory at exec time in `prepareSpawn` — the single chokepoint shared by single commands, chain segments and `run_background`; (2) `$VAR` / `${VAR}` / `$(…)` / backticks are now rejected with an actionable error (before the confirmation gate) instead of being passed through literally. Globs (`*`) are deliberately left untouched — many tools self-match them (`grep --include`, `find -name`, `rg -g`, test runners), so they keep working.

## Why

Fixes esengine/reasonix#2105. The tool already documents `$VAR` / `$(…)` / globs as rejected, but the implementation silently passed them through literally, so natural-looking commands (`cat ~/x`, `ls $DIR`, `echo $(date)`) produced wrong results with no error and were hard to debug. Expanding `~` is safe because the allowlist layer **already** canonicalizes `~` via `homedir()` for its sensitive-path and workspace-escape checks, so exec-time expansion adds no new bypass — it just makes an already-vetted `~/…` path work. `$VAR` and glob expansion are intentionally NOT added: they could slip a sensitive or out-of-workspace path past those literal-token checks (`cat $SECRET_PATH`, `cat *.env`).

## How to verify

- `npx vitest run tests/shell-expansion.test.ts tests/shell-tools.test.ts tests/shell-chain.test.ts tests/shell-redirects.test.ts tests/comment-policy.test.ts`
- `npx biome check src tests` and `npx tsc --noEmit`
- Manual: `cat ~/somefile` reads the home file; `ls $HOME` and `echo $(date)` return a clear error (text: runs WITHOUT a shell) pointing at the fix; `grep --include=*.ts foo .` still runs (the glob passes through literally).

## Checklist

- [x] lint + typecheck + shell suites + comment-policy gate pass locally (the full `npm run verify` is not runnable in this sandbox — dashboard build / desktop deps / commit-signing are environment-only and unrelated to this change)
- [x] Comments follow CONTRIBUTING.md (no module-essay headers, no incident history; blocks of 3 lines or fewer)

### DIFF
--- a/src/tools/shell.ts
+++ b/src/tools/shell.ts
@@ -11,11 +11,13 @@ import {
   type RunCommandResult,
   runCommand,
 } from "./shell/exec.js";
-import { isCommandAllowed } from "./shell/parse.js";
+import { detectUnsupportedExpansion, isCommandAllowed } from "./shell/parse.js";
 
 export {
   BUILTIN_ALLOWLIST,
   detectShellOperator,
+  detectUnsupportedExpansion,
+  expandTilde,
   hasSensitivePathArgs,
   isAllowed,
   isCommandAllowed,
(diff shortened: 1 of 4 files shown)


Bernardxu123:
🙏 **真诚感谢您的贡献！**

由于 v1 已不再接受代码变更，此 PR 将被关闭。

再次感谢您的时间和精力！❤️

PR #2300, opened 2026-05-29
fix(desktop): recover from tiny context token overrides

### LINKED ISSUE #2292: 上下文长度变为 1而且 在设置中 无法更改
在显示一次调用工具达到五十次上限后 ，所有对话都报错上下文 179,690/1（17969000%）— 基于已收集到的内容强制总结。请运行 /compact、/clear 或 /new 重置。运行提示的命令后仍然无法继续运行，输出内容短而且 不调用工具

BeiZi6:
## Summary
- Ignore implausibly small context-window overrides so a persisted value like `1` cannot force every turn into the context guard.
- Hydrate desktop `contextTokens` from `` so the Settings UI reflects persisted overrides.
- Make the desktop context-window input clear invalid overrides by writing the updated map back, including `{}`.

## Root Cause
A persisted `contextTokens` override could be as small as `1`, and `loadContextTokens()` treated any positive finite number as valid. Desktop also received `contextTokens` from the backend but did not keep it in `state.settings`, so users could see an empty/auto field while the runtime continued reading the bad persisted override.

## Validation
- `npm test -- --run tests/config.test.ts desktop/src/App.test.ts tests/loop.test.ts`
- `npm run typecheck`
- `npm run lint`
- `npx biome check src/config.ts tests/config.test.ts desktop/src/protocol.ts`
- `git diff --check`
- `npm run verify`

Fixes #2292

### DIFF
--- a/desktop/src/App.test.ts
+++ b/desktop/src/App.test.ts
@@ -389,6 +389,25 @@ describe("Desktop App reducer — ApprovalPrompt integration", () => {
     expect(next.settings?.reasoningEffort).toBe("low");
     expect(next.settings?.editMode).toBe("auto");
   });
+
+  it("hydrates context token overrides from settings events", () => {
+    const next = reduce(initialState(), {
+      t: "incoming",
+      event: {
+        type: "$settings",
+        reasoningEffort: "high",
+        editMode: "review",
+        budgetUsd: null,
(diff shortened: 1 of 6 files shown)


BeiZi6:
Closing this stale PR to avoid confusion with the current main-v2 work.

PR #2420, opened 2026-05-31
test: make Go tests portable on Windows

### LINKED ISSUE #2419: v2 Go tests fail on Windows due to home/path assumptions
## Summary

On Windows, the Go rewrite currently has a couple of test portability assumptions that make `go test ./...` fail even after Go and Git Bash are available.

## Reproduction

On Windows with Go installed and `bash` available via Git for Windows:

```powershell
go test ./...
```

## Observed failures

- `internal/config`: `TestLoadDotEnvFallsBackToHome` sets `HOME`, but `os.UserHomeDir()` on Windows uses `USERPROFILE`, so the home `.env` fallback is not exercised.
- `internal/tool/builtin`: `TestResolveIn` hard-codes Unix-style absolute path expectations like `/proj/foo.go`, which do not match Windows `filepath` behavior.

## Expected

`go test ./...` should pass on Windows for the v2 Go codebase.

## Local validation

I have a small test-only fix locally and verified:

```powershell
go test ./...
go build -o bin\reasonix.exe .\cmd\reasonix
go build -o bin\reasonix-plugin-example.exe .\cmd\reasonix-plugin-example
```

paradoxSCH:
## Summary

Closes #2419.

This makes two v2 Go tests portable on Windows:

- `TestLoadDotEnvFallsBackToHome` now sets `USERPROFILE` along with `HOME` for Windows, matching `os.UserHomeDir()` behavior.
- `TestResolveIn` now builds expectations with `filepath` and temporary absolute paths instead of hard-coding Unix-only `/proj` paths.

## Validation

Run on Windows:

```powershell
go test ./...
go build -o bin\reasonix.exe .\cmd\reasonix
go build -o bin\reasonix-plugin-example.exe .\cmd\reasonix-plugin-example
```

Note: local legacy npm git hooks were bypassed with `--no-verify` because the v2 Go branch no longer has a root `package.json`; the relevant Go checks above pass.

### DIFF
--- a/internal/config/dotenv_test.go
+++ b/internal/config/dotenv_test.go
@@ -3,6 +3,7 @@ package config
 import (
 	"os"
 	"path/filepath"
+	"runtime"
 	"testing"
 )
 
@@ -22,7 +23,7 @@ func TestLoadDotEnvFallsBackToHome(t *testing.T) {
 	}
 
 	t.Chdir(cwd)
-	t.Setenv("HOME", home) // os.UserHomeDir() reads $HOME on unix
+	setTestHome(t, home)
 
 	// Start clean so the file values are what land (Setenv auto-restores).
 	t.Setenv("KEY_CWD", "")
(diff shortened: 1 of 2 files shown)


esengine:
Thanks for this, @paradoxSCH — and for caring about Windows test portability, which is exactly the kind of thing that's easy to overlook.

Since you opened this, main-v2 landed the same fix independently: the temp-dir rewrite in `workspace_test.go` and the `HOME`+`USERPROFILE` handling in `dotenv_test.go` are both already in. The current branch even goes a bit further (e.g. `filepath.Join(workDir, "a", "b.go")` for the nested-path case), so this PR is now a no-op against main-v2.

Closing as already-resolved — but the instinct was spot on, and I'd love to see more from you. If you hit other cross-platform rough edges, please send them our way. 🙏

PR #2428, opened 2026-05-31
feat(serve): add graceful shutdown, request logging, and CORS middleware

HUQIANTAO:
## Summary
Enhance the HTTP/SSE server with three production-ready features:

### Graceful shutdown (`RunGraceful`)
- New `RunGraceful(ctx, addr)` method that listens for context cancellation (SIGINT/SIGTERM)
- Drains active connections for up to 10 seconds before returning
- Sets `ReadHeaderTimeout` (10s) and `IdleTimeout` (120s) on the http.Server

### Request logging middleware
- Logs method, path, status code, and duration for every request via `slog.Info`
- Uses a `responseWriter` wrapper to capture the status code

### CORS middleware
- Sets permissive CORS headers (`Access-Control-Allow-Origin: *`) for local development
- Handles OPTIONS preflight requests with 204 No Content
- Applied before the logging middleware so OPTIONS requests are logged too

## Motivation
The server previously used bare `http.ListenAndServe` with no shutdown handling — killing the process dropped active SSE connections. No request logging made debugging impossible. No CORS headers meant the dev frontend couldn't connect.

### DIFF
--- a/internal/serve/serve.go
+++ b/internal/serve/serve.go
@@ -6,10 +6,13 @@
 package serve
 
 import (
+	"context"
 	_ "embed"
 	"encoding/json"
 	"fmt"
+	"log/slog"
 	"net/http"
+	"time"
 
 	"reasonix/internal/control"
 )
@@ -43,7 +46,7 @@ func (s *Server) Handler() http.Handler {
 	mux.HandleFunc("POST /plan", s.plan)
 	mux.HandleFunc("POST /compact", s.compact)
 	mux.HandleFunc("POST /new", s.newSession)
-	return mux
+	return corsMiddleware(logMiddleware(mux))
 }
 
 // Run serves until the process is killed. Interactive approval is enabled so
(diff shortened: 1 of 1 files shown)


esengine (changes requested):
Thanks for tackling server hardening. The request logging (with the `responseWriter` status capture) and the `writeJSON` error check are good — I'd happily keep those as-is. Two things block merging, and one is a security regression.

### Blocking — CORS is a security regression, not just a dev convenience

`Handler()` applies `corsMiddleware` **unconditionally**, so `Access-Control-Allow-Origin: *` is on for every `serve` run (`cli.go:175` wires `Run` → `Handler`). This server has **no auth** and exposes endpoints that drive the agent (`/submit`, `/plan`, …) and stream its output (`/events`).

Right now the browser's same-origin policy is the only thing stopping a random web page the user visits from reading `/events` or POSTing `/submit` cross-origin. `*` removes that protection: any site's JS could read the agent's SSE stream and drive an agent that can run `bash`. Binding to `127.0.0.1` doesn't save us here — DNS-rebinding defeats it, and the request originates from the user's own browser.

Please don't enable CORS by default. Suggested: gate it behind an explicit `--cors`/`--dev` flag, and even then reflect a **specific** allowed origin (e.g. the Vite dev origin) rather than `*`. Default `serve` must stay same-origin-only.

### Blocking — `RunGraceful` is never called

`cli.go:175` still calls `Run` (bare `ListenAndServe`). `RunGraceful` — and with it the `ReadHeaderTimeout` / `IdleTimeout` you added — is dead code; graceful shutdown doesn't actually take effect. To land it, wire `serve` to call `RunGraceful` with a signal-cancelled context (`signal.NotifyContext` for SIGINT/SIGTERM) instead of `Run`.

### Nice to have

A test or two would help — e.g. an OPTIONS preflight returns 204, and shutdown drains within the timeout.

Happy to re-review once CORS is gated/removed and `RunGraceful` is actually wired in. The logging + `writeJSON` changes are good to keep.

HUQIANTAO:
Both blocking issues have been addressed in #2459:

1. **CORS \* security regression** → Removed from default Handler(), added HandlerWithCORS(origin) for dev use with specific origin
2. **RunGraceful dead code** → Wired in cli.go with signal.NotifyContext for SIGINT/SIGTERM

PR #2459 is green and ready for re-review.

esengine:
Superseded by #2459, which addresses both review points — CORS is now gated behind an explicit origin (never `*`), and RunGraceful is wired into cli.go with signal handling. #2459 is merged, so closing this. Thanks @HUQIANTAO!

PR #2429, opened 2026-05-31
fix: validate option labels are non-blank in ask tool

### LINKED ISSUE #2422: [Bug]: ask tool accepts blank option labels
### Version line

### Exact version

1.0.0

### What happened?

On the Go `main-v2` line, the `ask` tool accepts an option whose `label` is empty or whitespace-only.
Example tool args:
```json
{
  "questions": [{
    "header": "Direction",
    "question": "Which path?",
    "options": [
      { "label": "Keep going" },
      { "label": "   ", "description": "blank labels render as empty picker rows" }
    ]
  }]
}
```
Because the tool only validates the question text and the number of options, the blank label reaches the interactive chooser. That can render an empty selectable row and can produce an unusable answer.

### Steps to reproduce

1. During a model turn, the model decides to call the built-in `ask` tool to ask the user a multiple-choice question.
2. The model emits an `ask` tool-call payload where one option has a whitespace-only label, for example:
```json
   {
     "questions": [{
       "header": "Direction",
       "question": "Which path?",
       "options": [
         { "label": "Keep going" },
         { "label": "   ", "description": "blank labels render as empty picker rows" }
       ]
     }]
   }
```
3. The tool-call arguments are passed into `AskTool.Execute` in `internal/agent/ask.go`.
4. `AskTool.Execute` currently only checks that the question text exists and that there are at least two options. It does not trim or validate each option label.Because of that, the whitespace-only option is accepted and converted into an `event.AskQuestion`.
5. In interactive mode, the controller/frontend receives that `AskQuestion` and renders it in the chooser.The chooser can then show an empty selectable option row, such as:
```text
? Which path?
1. Keep going
2.
```
6. If the user selects the blank row, the model receives an empty/useless answer instead of a meaningful choice.

### OS / platform

Ubuntu 22.04.5

xxiaoxiong:
## Problem

The `ask` tool accepts options with blank or whitespace-only `label` values, producing empty selectable rows in the interactive chooser that can confuse users and produce useless answers.

## Root Cause

`AskTool.Execute` only validates that the question text exists (`q.Question == ""`) and that there are at least two options (`len(q.Options) < 2`). It does not trim or validate individual option labels, so an option like `{"label": "   "}` passes through untouched.

## Changes

Added a `strings.TrimSpace(o.Label) == ""` check inside the option-building loop in `internal/agent/ask.go`. If any option has a blank label, the tool returns a descriptive error (`question %d: option %d has a blank label`) instead of silently accepting it.

This is consistent with the existing validation pattern — the tool already rejects a blank question text and too-few options; blank labels should be rejected for the same reason.

## Closes

Closes #2422

### DIFF
--- a/internal/agent/ask.go
+++ b/internal/agent/ask.go
@@ -93,6 +93,9 @@ func (*AskTool) Execute(ctx context.Context, args json.RawMessage) (string, erro
 		}
 		opts := make([]event.AskOption, len(q.Options))
 		for j, o := range q.Options {
+			if strings.TrimSpace(o.Label) == "" {
+				return "", fmt.Errorf("question %d: option %d has a blank label", i+1, j+1)
+			}
 			opts[j] = event.AskOption{Label: o.Label, Description: o.Description}
 		}
 		qs = append(qs, event.AskQuestion{


esengine:
Superseded by #2426, which does the same blank-label rejection but more completely — it also trims question/header/label/description and rejects duplicate labels, with regression tests, and closes #2422. Going with that one. Thanks @xxiaoxiong!

Exagone313:
fyi this PR was automated, see <https://github.com/search?q=author%3Axxiaoxiong+is%3Apr&s=created&type=pullrequests>

PR #2451, opened 2026-05-31
feat(subagent): configure skill model overrides

### LINKED ISSUE #2435: [Feature]: subagent 能否支持模型设置？不要默认是 v4-flash？ 如果 subagent 需要代码编写的时候，可能也需要用 v4-pro？
### Version line

subagent 能否支持模型设置？不要默认是 v4-flash？ 如果 subagent 需要代码编写的时候，可能也需要用 v4-pro？

### Proposed solution

_No response_

GTC2080:
Closes #2435

## Summary
- add `[agent].subagent_model` as the default model for `runAs: subagent` skills
- add `[agent].subagent_models` for per-skill overrides such as `review` and `security_review`
- preserve existing behavior when no override is configured, and keep skill frontmatter `model:` as the fallback before the global default
- document the new TOML settings and make existing Windows tests portable for Go validation

## Validation
- `D:\Go\go\bin\go.exe test ./...`

No UI or performance behavior changed, so screenshots and metrics are not applicable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,11 @@ never stored in config files.
 default_model = "deepseek-flash"   # executor; set [agent].planner_model to add a planner
 # language    = "zh"               # ui language; empty = auto-detect from $LANG / $REASONIX_LANG
 
+[agent]
+# planner_model = "mimo-pro"          # optional low-frequency planner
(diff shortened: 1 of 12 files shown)


MorenoCLR:
Does it not work with other skills? I tested it with research but the skill still use flash model even though specified research

PR #2459, opened 2026-05-31
fix(serve): gate CORS behind explicit origin, wire RunGraceful with signal handling

HUQIANTAO:
## Summary
Address review feedback on PR #2428:

### Fix 1: CORS security regression
- **Before**: corsMiddleware applied `Access-Control-Allow-Origin: *` unconditionally in Handler()
- **After**: Handler() returns routes without CORS (same-origin-only, the only auth protection)
- New `HandlerWithCORS(origin)` method accepts a specific origin for dev use (e.g. "http://localhost:5173")
- corsMiddleware now takes an origin parameter; empty origin = no CORS

### Fix 2: Wire RunGraceful with signal handling
- **Before**: cli.go called Run() (bare ListenAndServe), making RunGraceful dead code
- **After**: cli.go calls RunGraceful(ctx, addr) with signal.NotifyContext for SIGINT/SIGTERM
- Graceful shutdown now actually takes effect: 10s drain timeout, ReadHeaderTimeout, IdleTimeout

### Kept from original PR
- Request logging middleware (method, path, status, duration)
- writeJSON error logging

## Motivation
The review correctly identified that `Access-Control-Allow-Origin: *` on an unauthenticated agent server is a security regression — any website could read /events or POST /submit cross-origin. DNS-rebinding defeats 127.0.0.1 binding. The fix gates CORS behind an explicit opt-in with a specific origin.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -172,7 +172,10 @@ func runServe(args []string) int {
 	}
 
 	fmt.Printf("reasonix serve — %s on http://%s\n", ctrl.Label(), *addr)
-	if err := serve.New(ctrl, bc).Run(*addr); err != nil {
+	// Use graceful shutdown so SIGINT/SIGTERM drain active connections.
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt)
+	defer stop()
+	if err := serve.New(ctrl, bc).RunGraceful(ctx, *addr); err != nil {
 		fmt.Fprintln(os.Stderr, i18n.M.ErrorPrefix, err)
 		return 1
 	}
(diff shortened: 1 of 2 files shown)


HUQIANTAO:
@esengine This PR addresses both blocking issues from your review of #2428:

1. **CORS**: Removed corsMiddleware from default Handler() — same-origin-only by default. Added HandlerWithCORS(origin) for dev use with explicit origin.
2. **RunGraceful**: Wired in cli.go with signal.NotifyContext for SIGINT/SIGTERM graceful shutdown.

Both fixes are in a single PR for clean review. All tests pass. Could you take another look?

PR #2463, opened 2026-05-31
feat(subagents): add TUI slash subagent runtime browser

JimmyDaddy:
## Status

This branch has been hard-reset back to `53b6c7f`, the state before the latest rebase/narrowing attempt. The temporary commits `7d1833cf` and `8da138b6` were removed from the PR branch by force-with-lease push.

## Notes

- The previous direct `/name args` minimal rewrite is no longer present on this branch.
- The PR is back to its earlier implementation state and may conflict with current `main-v2` again.
- Further rebase or scope-reduction work should start from this restored branch state.


esengine:
Took this for a thorough pass — the architecture is sound: the slash-subagent surface is strictly user-invoked, `task.go`'s model-facing Description/Schema is unchanged (no new model-driven spawning), and `disable-model-invocation` keeping slash skills out of the model index is exactly right. Root Go build and the frontend typecheck are clean.

One blocker before this can land: it **fails `go test` on Windows**, in `internal/control`:

```
--- FAIL: TestSubmitRunsShellCommand
    submit_test.go:78: Submit shell command notices = [\a b"], want [a b]
--- FAIL: TestSubmitRunsShellCommandWithMultilineArgs
    submit_test.go:91: ... = [\a b"], want [a b]
```

The tests use `printf '%s\n'` with `/echo "a b"`, which assumes POSIX `sh -c` semantics. On Windows `shellInvocation` falls back to `cmd /c`, where single-quotes and `printf` behave differently, so the output comes back mangled. Our CI only runs macOS/ubuntu, so this stays green there but breaks on Windows.

Could you:
1. Make those tests OS-aware (skip on Windows, or use a portable command that behaves the same under both `sh -c` and `cmd /c`), and
2. Confirm the shell-command feature itself quotes args sanely under `cmd /c` on Windows (or document that user shell commands are written per-platform)?

Given the size (+5957/45 files), I'll also want to walk through `docs/SLASH_SUBAGENT_DESIGN.md` once the Windows failures are resolved. Thanks for the design doc — it made the review much easier.

JimmyDaddy:
> Took this for a thorough pass — the architecture is sound: the slash-subagent surface is strictly user-invoked, `task.go`'s model-facing Description/Schema is unchanged (no new model-driven spawning), and `disable-model-invocation` keeping slash skills out of the model index is exactly right. Root Go build and the frontend typecheck are clean.
> 
> One blocker before this can land: it **fails `go test` on Windows**, in `internal/control`:
> 
> ```
> --- FAIL: TestSubmitRunsShellCommand
>     submit_test.go:78: Submit shell command notices = [\a b"], want [a b]
> --- FAIL: TestSubmitRunsShellCommandWithMultilineArgs
>     submit_test.go:91: ... = [\a b"], want [a b]
> ```
> 
> The tests use `printf '%s\n'` with `/echo "a b"`, which assumes POSIX `sh -c` semantics. On Windows `shellInvocation` falls back to `cmd /c`, where single-quotes and `printf` behave differently, so the output comes back mangled. Our CI only runs macOS/ubuntu, so this stays green there but breaks on Windows.
> 
> Could you:
> 
> 1. Make those tests OS-aware (skip on Windows, or use a portable command that behaves the same under both `sh -c` and `cmd /c`), and
> 2. Confirm the shell-command feature itself quotes args sanely under `cmd /c` on Windows (or document that user shell commands are written per-platform)?
> 
> Given the size (+5957/45 files), I'll also want to walk through `docs/SLASH_SUBAGENT_DESIGN.md` once the Windows failures are resolved. Thanks for the design doc — it made the review much easier.

Thanks for the careful pass - I've pushed an update for the Windows blocker.

What changed:

- Made the shell-command controller tests platform-safe. The `/echo "a b"` and multiline-arg cases now use the test binary as a helper command instead of POSIX-only `printf '%s\n'`, so they exercise the same raw slash-arg path without assuming `sh` syntax.
- Added lower-level `internal/command` coverage for quoted user args and a plain arg case.
- Updated `docs/SLASH_SUBAGENT_DESIGN.md` to document the contract: settings slash commands are platform shell snippets (`cmd /c` on Windows, `sh -c` elsewhere); Reasonix preserves quoted user args but does not translate POSIX shell syntax to Windows `cmd.exe` syntax.
- ***I also included a Windows execution-path change for settings slash commands and the hook shell spawner: on Windows it sets `SysProcAttr.CmdLine` for `cmd /c ...` so Go's argv escaping does not mangle the quoted command string before `cmd.exe` applies its own parsing. Could you sanity-check whether that is an acceptable direction for this codebase? If you'd rather keep the runtime path unchanged and only make the tests/docs OS-aware, I can trim that part back.***

Validation run locally:

- `go test ./internal/command ./internal/control ./internal/hook`
- `go test ./...`
- `git diff --check`

JimmyDaddy:
@esengine latest conflicts are resolved again and the branch is back to a clean, mergeable state.

Before I keep chasing `main-v2` conflicts every few days, could you help confirm whether this feature direction is still something you'd like reviewed / considered for merge?

If the overall direction looks good, I'd really appreciate a review pass and any design feedback you have, especially on the TUI-only `/subagents` management surface

If this is not a direction you want to take, no worries at all — a short "not aligned" reply is enough and I can close the PR instead of continuing to carry the merge burden.

Thanks.

esengine:
Status update after today's merge batch: this now conflicts in 14 files, including the heavy ones your deltas live in (controller.go, chat_tui.go, boot.go, the i18n catalogues — #3186 touched the same Submit switch). The direction still stands from the earlier pass — strictly user-invoked /name subagents through the skill runner is exactly the accepted shape — so a rebase onto current main-v2 is the only blocker; ping me when it's up and I'll do the full line review. Two notes for that round: watch internal/cli/subagents.go (684 lines, ceiling is 800), and controller.go keeps growing past the ceiling, so the PR description should carry a split plan per repo policy.

JimmyDaddy:
@esengine rebased/rebuilt this PR onto current `main-v2` and force-pushed the linear head `4601a9c0`. GitHub now reports `mergeable=MERGEABLE` / `mergeStateStatus=BLOCKED`.

I also updated the PR description with the requested split plan: `internal/cli/subagents.go` is still 684 lines, and `controller.go` is called out with a follow-up split plan per repo policy. Latest local checks passed:

- `gofmt -l .`
- `git diff --check`
- `GOPROXY=https://proxy.golang.org,direct go test ./...`
- `cd desktop && GOPROXY=https://proxy.golang.org,direct go test ./...`

Ready for the full line review when you have bandwidth. If the behavior/design direction should change further, please call that out and I can adjust.

PR #2476, opened 2026-05-31
fix(cli): CJK bold (**) not rendering — goldmark right-flanking workaround

HUQIANTAO:
## Motivation

Markdown bold (`**text**`) does not render when the closing `**` is immediately preceded by CJK punctuation (full-width comma `，`, period `。`, exclamation `！`, etc.). The model frequently produces this pattern in Chinese responses, resulting in literal `**` visible to the user.

## Root Cause Analysis

Goldmark follows the CommonMark spec's "right-flanking delimiter run" rule: a closing `**` is only valid if the character **before** it is Unicode punctuation, or the character **after** it is Unicode punctuation. CJK punctuation characters (U+FF0C `，`, U+3002 `。`, U+FF01 `！`, etc.) are **not** classified as Unicode punctuation by the spec — they fall into the "other" category.

```
Input:    **我叫Reasonix，**一个助手
Goldmark: Text("**") Text("我叫Reasonix，**") Text("一个助手")
          — closing ** not recognized, rendered as literal text

Input:    **测试**更多
Goldmark: Emphasis(level=2) Text("测试") Text("更多")
          — closing ** followed by CJK ideograph, works
```

This is a known limitation of CommonMark's emphasis parsing with CJK text. The goldmark CJK extension (`extension.CJK`) only handles line-break rules, not emphasis delimiters.

### `fixCJKEmphasis()` in `md.go`

A preprocessing pass on the markdown input string, applied before goldmark parsing. It tracks `**` open/close state: when a closing `**` is preceded by a non-ASCII character (codepoint > 0x7F), a space is inserted after it. This satisfies the right-flanking check without altering the rendered output — the space is collapsed by the terminal's word-wrap.

```go
func fixCJKEmphasis(s string) string {
    // Track ** open/close state.
    // Closing ** preceded by non-ASCII → insert trailing space.
    ...
}
```

The function is called at the top of `mdRenderer.Render()`, before `goldmark.Parser().Parse()`.

## Edge cases

| Input | Before | After |
|-------|--------|-------|
| `**测试，**更多` | `**测试，**更多` (literal) | `**测试，** 更多` → **测试，**更多 (bold) |
| `**测试**更多` | **测试**更多 (bold) | **测试** 更多 (bold, extra space) |
| `**bold** text` | **bold** text | **bold** text (unchanged, ASCII) |
| `前**加粗**后` | 前**加粗** 后 (bold) | 前**加粗** 后 (bold, unchanged) |
| `**a** and **b**` | **a** and **b** | **a** and **b** (unchanged, ASCII) |

## Files

| File | Lines | Change |
|------|-------|--------|
| `internal/cli/md.go` | +40 | `fixCJKEmphasis()` function + call in `Render()` |

## Testing

- `go test ./internal/cli/...` ✅
- Manual: `**CJK，**text` renders as bold
- Manual: `**English**text` unchanged
- Manual: `前**加粗**后` unchanged
- CI: macOS + Ubuntu both pass


esengine (changes requested):
Thanks for tackling this — CJK punctuation breaking `**bold**` is a real, common annoyance in Chinese output, and the diagnosis (goldmark's CommonMark right-flanking rule not treating full-width punctuation as Unicode punctuation) is exactly right. A few things to address before merge, one of which is a correctness risk.

### Blocking — it can corrupt code spans / code blocks

`fixCJKEmphasis` runs on the whole input **before** parsing, with no awareness of markdown structure. Inline code or fenced blocks that happen to contain paired `**` next to CJK get rewritten too — e.g. `` `a**中文**b` `` becomes `` `a**中文** b` ``, changing the literal code content. The model emits Chinese inside code/strings often enough that this will bite. The space insertion needs to skip code spans and fenced code blocks (or be done at the AST level so it only touches emphasis text, never code).

### Over-correction — narrow the trigger to CJK punctuation

`runes[i-2] > 0x7F` fires on **any** non-ASCII char, including CJK letters. But `**中文**词` already renders fine — the closing `**` is preceded by a CJK *letter*, which CommonMark treats as non-punctuation, so right-flanking already holds. Inserting a space there just adds a stray space and changes the visible text. Only CJK **punctuation** (，。！ etc.) needs the workaround, so gate on something like `unicode.IsPunct(runes[i-2]) && runes[i-2] > 0x7F` rather than all non-ASCII.

### Fragile pairing

The open/close toggle assumes `**` always pairs up; odd counts or non-emphatic `**` (math `2 ** 3`, etc.) mis-pair. `2 ** 3` is space-flanked so it's harmless here, but the state machine is brittle.

### Tests

This changes rendering with no test coverage. Please add: a CJK-punctuation bold case (the fix works), a non-punctuation CJK bold that must NOT gain a space (over-correction guard), and a code span containing `**` + CJK that must be left verbatim.

Happy to re-review. A goldmark inline-parser extension / AST-level pass is the most robust route, but a guarded preprocessor that skips code spans, narrows to CJK punctuation, and has tests would be fine too.

HUQIANTAO:
Thanks for the thorough review. All four points addressed:

### 1. Code spans / fenced blocks — FIXED ✅

Rewrote `fixCJKEmphasis` to track inline code (`` ` ``) and fenced code block (``` ``` ```) state. Content inside code is passed through verbatim:

```go
// Inline code span: ` toggles in/out (but not inside fenced blocks).
if r == '`' && !inFenced {
    inCode = !inCode
    b.WriteRune(r)
    continue
}
// Inside code — pass through verbatim.
if inCode || inFenced {
    b.WriteRune(r)
    continue
}
```

Test: `` `a**中文**b` `` → unchanged ✅

### 2. Narrow to CJK punctuation — FIXED ✅

Replaced `runes[i-2] > 0x7F` with `isCJKPunct(prev)` which checks specific CJK punctuation ranges (U+3000–303F, U+FF01–FF0F, etc.) and falls back to `unicode.IsPunct` for other scripts. CJK letters (中/文) are no longer affected.

Test: `**中文**词` → unchanged (no space inserted) ✅

### 3. Fragile pairing — FIXED ✅

Removed the `inEmphasis` toggle entirely. Each `**` is now evaluated independently by examining the preceding character context:

```go
if i >= 2 && !isSpace(runes[i-2]) {
    prev := runes[i-2]
    if isCJKPunct(prev) {
        b.WriteByte(' ')
    }
}
```

No state to go out of sync.

### 4. Tests — ADDED ✅

- `TestFixCJKEmphasis` — 11 cases: CJK punctuation bold, non-punctuation CJK, English, inline code, fenced code, code span + emphasis mixed
- `TestFixCJKEmphasisRenderIntegration` — 3 cases: rendering correctness
- `TestIsCJKPunct` — 12 cases: punctuation classification

HUQIANTAO:
All review feedback addressed. Ready for re-review when you have a chance. 🙏

PR #2481, opened 2026-05-31
fix(ui): keep chat transcript pinned while streaming

### LINKED ISSUE #2468: [Bug]:消息列表无法自动滚动
### Version line

### Exact version

0.53.0

### What happened?

在对话界面中，当 AI 持续输出消息时，消息列表无法自动滚动到底部，导致用户需要手动下拉才能看到最新内容。希望实现以下交互：

AI 输出时自动滚动到底部：每当 AI 有新的消息输出时，页面应自动滚动到消息列表的最底部，确保用户始终看到最新消息。

手动滚动时暂时关闭自动滚动：如果用户主动向上滚动查看历史消息，此时应暂时关闭自动滚动功能，避免干扰用户的浏览操作。

手动滚动到底部时恢复自动滚动：当用户手动将滚动条拉到最底部时，应重新开启自动滚动功能，让后续的 AI 消息继续自动滚到底部。

### Steps to reproduce

1

### OS / platform

mac 15

### Relevant logs or output

```shell

```

BeiZi6:
Fixes #2468.

## Summary
- keep the lightweight HTTP transcript pinned while assistant text/reasoning streams into an existing DOM node
- preserve user intent by disabling follow mode when the user scrolls up and resuming it once they return to the bottom
- make the desktop transcript follow DOM text mutations/resizes during streaming, not only new item array changes

## Testing
- `git diff --check`
- `go test ./internal/serve`
- `cd desktop/frontend && npx -y pnpm@9.15.9 build`
- Chrome extension local scroll harness against `internal/serve/index.html`: pass; streaming follows at bottom, stays put after manual scroll-up, resumes after returning to bottom

## Local note
- Full `go test ./...` was attempted, but this machine has unrelated local-environment failures: user-global skill index content affects `internal/boot`, and macOS sandbox enforcement tests do not deny writes in this environment.

### DIFF
--- a/desktop/frontend/src/components/Transcript.tsx
+++ b/desktop/frontend/src/components/Transcript.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useRef, useState } from "react";
+import { useEffect, useLayoutEffect, useRef, useState } from "react";
 import type { Item } from "../lib/useController";
 import { AssistantMessage, UserMessage } from "./Message";
 import { ToolCard } from "./ToolCard";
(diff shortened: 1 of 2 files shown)


esengine:
Thanks for this fix, @BeiZi6 — and for the thorough MutationObserver/ResizeObserver approach. We're consolidating the autoscroll fix on #2487, which is already mirrored in the larger desktop port (#2473). That approach keys the follow-bottom effect on a content fingerprint (`scrollVersion`) instead of array identity, which targets the root cause directly: a reducer reusing the `items` array reference wouldn't re-trigger the effect. To avoid two divergent implementations of the same behavior, and to keep changes off the legacy `internal/serve/index.html` web path (v2 focuses on the desktop surface), we're closing this in favor of #2487. Really appreciate the contribution — please keep them coming!

PR #2523, opened 2026-06-01
Collapse CLI Thinking Output by Default

SivanCola:
## Summary

Collapse CLI thinking output by default so reasoning streams render like Claude Code CLI: users see a muted `thinking` marker instead of raw reasoning text unless they explicitly opt into verbose display.

## Root Cause

Reasoning deltas were previously rendered verbatim in terminal output. That made thinking-mode providers expose reasoning text directly in normal CLI runs, while Claude Code-style behavior keeps thinking collapsed by default and only expands it through an explicit verbose mode.

## Technical Approach

- Added a display-only reasoning visibility flag to the terminal TextSink.
- Kept reasoning content in session state and provider flow unchanged; this only changes terminal rendering.
- Added `Ctrl+O` and `/verbose` in the chat TUI to toggle raw thinking text.
- Added `reasonix run --show-thinking` for explicit one-shot verbose output.
- Updated slash completion and English/Chinese help text.
- Added regression tests for collapsed default behavior and verbose opt-in behavior.

- Preserves prompt/cache behavior by avoiding any provider request serialization changes.
- Keeps existing reasoning signatures and session retention intact for provider protocols.
- Preserves latest upstream paste-image shortcuts and help entries while adding `/verbose`.
- Makes the default terminal output safer and closer to Claude Code CLI expectations.

## Verification

- `PATH="/opt/homebrew/bin:$PATH" go test ./internal/agent ./internal/cli ./internal/i18n` passed.
- `PATH="/opt/homebrew/bin:$PATH" scripts/cache-guard.sh` passed.
- `PATH="/opt/homebrew/bin:$PATH" go test ./...` was run; it still fails only in `internal/boot TestBuildDiscoversSkills` because this local environment loads a global skills index that does not match the test fixture. Other packages passed.


SivanCola:
@codex review

esengine:
Superseded by #2562 (merged as `57359d53`), which carries your collapse-by-default work plus a live `▎ thinking…` marker that resolves to `▎ thought for Ns` so the user sees activity during long thinking — the Claude Code behaviour we were after. Credited you as co-author on the squash. Thanks!

PR #2541, opened 2026-06-01
feat(agent): segment read-only tool batch execution

LegGasai:
## Summary
- segment tool-call batches so contiguous known read-only tools can run concurrently around serial writer/unknown calls
- preserve provider-order dispatch, result emission, and session writeback
- update agent tests for batch partitioning and segmented execution timing

## Tests
- gofmt -l .
- git diff --check
- go test ./internal/agent -count=1

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -410,11 +410,11 @@ func (a *Agent) stream(ctx context.Context) (string, string, string, []provider.
 
 // executeBatch dispatches one model turn's tool calls. A ToolDispatch event is
 // emitted for every call up front, in call order, so a frontend can show the
-// timeline chronologically. Calls fan out across goroutines only when every
-// call's tool is ReadOnly (canParallelise); a single non-ReadOnly call drops
-// the whole batch back to sequential to preserve write/read ordering. ToolResult
-// events are emitted after the batch in call order, so emission stays serial
-// even when execution parallelised.
+// timeline chronologically. Contiguous known ReadOnly calls fan out across
+// goroutines; unknown and writer calls run as single-call serial segments so
+// write/read ordering stays provider-ordered. ToolResult events are emitted
+// after the batch in call order, so emission stays serial even when execution
+// parallelised.
 func (a *Agent) executeBatch(ctx context.Context, calls []provider.ToolCall) []string {
 	for _, c := range calls {
 		t, ok := a.tools.Get(c.Name)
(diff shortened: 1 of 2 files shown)


HUQIANTAO:
## Review: Code Optimization Suggestions

Nice segmentation logic — the approach of splitting contiguous read-only calls into parallel batches while keeping writers serial is clean and correct. A few micro-optimizations to consider:

### 1. Pre-allocate `batches` slice

`var batches []toolCallBatch` triggers repeated grow-copy as it appends. Pre-allocating with a reasonable capacity avoids this:

```go
func partitionToolCalls(r *tool.Registry, calls []provider.ToolCall) []toolCallBatch {
	if len(calls) == 0 {
		return nil
	}
	batches := make([]toolCallBatch, 0, len(calls)/2+1)
	// ... rest unchanged
}
```

### 2. Single-call read-only batches should not be marked parallel

Currently a lone read-only call gets `parallel: true`, then `executeBatch` has to check `batch.end-batch.start > 1` to skip the goroutine overhead. Cleaner to mark it non-parallel at the source:

```go
batches = append(batches, toolCallBatch{
	start:    start,
	end:      i,
	parallel: i-start > 1,  // only parallel when >1 call
})
```

Then the `executeBatch` loop simplifies to:

```go
if batch.parallel {  // no need for the > 1 check
```

Test expectations for `TestPartitionToolCallsSegmentsAroundWriters` would need updating — the single-call read-only segments change from `{parallel: true}` to `{parallel: false}`.

### 3. Reuse semaphore channel across batches

Each `runParallel` call creates a new `chan struct{}`. With segmented batches (e.g. `[ro1,ro2], rw, [ro3,ro4]`), that is 2 channel allocations. Hoisting the semaphore to `executeBatch` and passing it in avoids this:

```go
const maxParallel = 8
sem := make(chan struct{}, maxParallel)

for _, batch := range partitionToolCalls(a.tools, calls) {
	if batch.parallel {
		runParallel(batch.start, batch.end, run, sem)
		continue
	}
	for i := batch.start; i < batch.end; i++ {
		run(i)
	}
}

func runParallel(start, end int, run func(int), sem chan struct{}) {
	var wg sync.WaitGroup
	for i := start; i < end; i++ {
		i := i
		sem <- struct{}{}
		wg.Add(1)
		go func() {
			defer wg.Done()
			defer func() { <-sem }()
			run(i)
		}()
	}
	wg.Wait()
}
```

---

These are all micro-optimizations (tool execution time dwarfs the overhead), but they make the logic tighter — especially #2, which eliminates a redundant guard and makes `parallel` semantically mean "has multiple calls to run concurrently".

LGTM overall, merge at your discretion.

PR #2599, opened 2026-06-01
feat(memory): forget tool, fallible-recall framing, richer index

esengine:
## Why

Our memory was a faithful clone of the two-layer model (doc hierarchy + per-fact files + `MEMORY.md` index folded into the cache-stable prefix), but it lagged on three behaviours that make saved memory actually useful:

1. **It could only grow.** The model had `remember` but no way to drop a fact that turned out wrong or stale. Dedup was exact-name-only, so semantically duplicate facts under different slugs piled up forever — steadily degrading recall signal-to-noise.
2. **Stale facts read as standing orders.** The saved-memories block told the model to "treat as durable, user-authored guidance", so it would act on a memory naming a since-deleted file/flag as if it were a current instruction.
3. **The index read as slugs.** Link text was the bare kebab slug, weakening the relevance call the model makes when deciding which fact to open.

## What

- **`forget` tool + `Store.Delete`** — prune a fact's file and its `MEMORY.md` line (missing file is a no-op; the goal state is "gone"). Symmetric with `remember`: writes straight to disk, folds out of the prefix next session.
- **Recall reframed as fallible background** — the saved-memories block now says these reflect what was true when written, may be stale, are background not standing instructions, and a named file/function/flag must be verified to still exist before acting on it.
- **Richer index** — new optional `title` (frontmatter + round-trip); index links render `[Title](name.md)`, falling back to a de-kebabed name so a bare slug never leaks. `description` is repositioned as the relevance hook.
- **Stronger `remember` guidance** — `**Why:**`/`**How to apply:**` structure for feedback/project, `[[their-name]]` inline links, and reuse-the-name-to-update over near-duplicate saves.

Recall stays index-only in the cached prefix (zero added per-turn cost, fully cache-stable) by design — bodies are still read on demand.

## Tests

`go test ./...` (root + `desktop` module) green; frontend `tsc --noEmit` green; `go vet` clean. New: `TestStoreDelete`, `TestStoreDeleteMissingIsNoError`, `TestStoreSaveTitleInIndexAndFrontmatter`, `TestStoreIndexLabelFallsBackToDeKebabbedName`, `TestForgetToolDeletes`, `TestForgetToolValidates`, plus title round-trip in the remember test.

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -1200,6 +1200,7 @@ type MemoryDoc struct {
 // MemoryFact is one saved auto-memory, surfaced read-only in the panel.
 type MemoryFact struct {
 	Name        string `json:"name"`
+	Title       string `json:"title,omitempty"`
 	Description string `json:"description"`
 	Type        string `json:"type"`
 	Body        string `json:"body"`
@@ -1248,7 +1249,7 @@ func (a *App) Memory() MemoryView {
 	}
 	for _, f := range set.Store.List() {
 		view.Facts = append(view.Facts, MemoryFact{
-			Name: f.Name, Description: f.Description, Type: string(f.Type), Body: f.Body,
+			Name: f.Name, Title: f.Title, Description: f.Description, Type: string(f.Type), Body: f.Body,
 		})
 	}
 	for _, sc := range writableScopes {
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -1273,6 +1273,18 @@ func (a *App) Remember(scope, note string) (string, error) {
 	return ctrl.QuickAdd(parseScope(scope), note)
 }
 
+// Forget deletes a saved auto-memory by name — the panel's delete action for a
+// fact the model owns. A no-op when no controller is attached.
+func (a *App) Forget(name string) error {
+	a.mu.RLock()
+	ctrl := a.ctrl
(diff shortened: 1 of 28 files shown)


esengine:
Follow-up commit (`e0a2d5d`) — wiring the new backend into the surfaces and closing two more parity gaps:

- **`forget` is now usable by hand**, not just by the model: the desktop memory panel gets a per-fact delete button (and shows the new title), and the TUI gains `/forget <name>` with `/memory` now listing the saved facts. Backed by `Controller.ForgetMemory` + `App.Forget`.
- **`[[name]]` cross-links are live** in the panel — they render as in-panel jumps to the referenced memory; a link with no matching fact is flagged as dead.
- **`remember`/`forget` take effect the same session.** Each tool queues a turn-tail `<memory-update>` note via a `memory.Queue` stamped on the call context (same idiom as the job manager), so a just-deleted fact is disregarded even though the cached index still lists it until the next session. The mid-session note header is now neutral so it reads correctly for saves, deletes, and doc edits alike.

All gates green: `go test ./...` (root + `desktop`), frontend `tsc --noEmit`, `go vet`, gofmt. New tests: queue-injection for both tools (`memory`), and `Compose` drain of a queued note (`control`).

PR #2623, opened 2026-06-01
Remove No Project option from workspace menu

SivanCola:
Summary
- Removes the “No project” action from the workspace switcher menu.
- Keeps the existing project list and “Add new project” action intact.

Root Cause
- The workspace switcher rendered an explicit action that called the folder picker with an empty path, exposing a “No project” option that should no longer be available in the UI.

Technical Approach
- Removed the menu action that selected an empty workspace path.
- Removed the now-unused FolderX icon import from the composer component.

Focused Optimization Points
- Kept the change scoped to the workspace menu rendering path.
- Avoided changing locale strings or workspace switching behavior outside the removed action.

Verification
- PATH="/Users/<developer>/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH" ./node_modules/.bin/tsc --noEmit
- Verified in the local browser UI that the workspace menu still shows project entries and “Add new project”, while “No project” is no longer shown.


SivanCola:
Closing in favor of #2624, which is based on the latest main-v2 branch and avoids the merge conflict from the initial detached-HEAD branch.

PR #2631, opened 2026-06-01
feat(cli): add doctor diagnostics command

### LINKED ISSUE #2630: [Feature]: add reasonix doctor for redacted local diagnostics
### Version line

When users report setup, provider, MCP, CodeGraph, terminal, sandboxor session issues, maintainers need basic environment andconfiguration context.
Today v2 users have to collect this manually: OS, Reasonix version,config source, selected model, provider key state, MCP/plugin setup,CodeGraph status, session directory, sandbox and permission settingsThis makes bug reports slower to triage and creates extra back-andforth.
A local diagnostic command would give users a safe, copyable reportwithout exposing secrets.

## Add a top-level CLI command

```bash
reasonix doctor
reasonix doctor --json
```

The command should print a redacted local diagnostics report and then exit.

### Initial report should include

- Reasonix version, OS, architecture, and cwd
- config source path and default model
- providers, base URL host, selected/default state, and whether `api_key_env` is present
- configured MCP/plugin entries, showing only transport and host/command name
- CodeGraph enabled/auto-install/cache/resolution status
- LSP enabled status and configured override count
- session directory count/size summary
- sandbox mode, network setting, and write roots
- permission mode and allow/ask/deny rule counts

It intentionally avoids automatic upload and paid/API health checks, so running the command does not send data anywhere, spend tokens, or call model providers.

If this direction looks useful for issue triage, I can follow up with a chat-session entrypoint such as `/doctor`. That version could include the same redacted local diagnostics plus session-specific context from the active Reasonix chat session, such as the current model, session path, message/turn counts, context usage, recent tool/error summary, and MCP/plugin runtime state. A later desktop entrypoint, such as "Copy diagnostics" in settings or an error dialog, could reuse the same report logic as well.

XTLine:
## Summary

Adds a top-level `reasonix doctor` command for redacted local diagnostics, plus `--json` output for machine-readable reports.

The report includes local environment/configuration context such as providers, key presence, MCP/plugin entries, CodeGraph, LSP, session storage, sandbox, and permission settings without exposing secrets.

## Testing
```bash
go test ./internal/doctor ./internal/cli ./internal/i18n
go build -o /tmp/reasonix-doctor-verify ./cmd/reasonix
/tmp/reasonix-doctor-verify doctor --json
```

Also ran go test ./...; the only failure is the pre-existing Linux internal/sandbox/bwrap expectation mismatch, unrelated to this change.

doctor

## Related
Closes #2630

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -61,6 +61,8 @@ func Run(args []string, version string) int {
 		return mcpCommand(rest)
 	case "codegraph":
 		return codegraphCommand(rest)
+	case "doctor":
+		return doctorCommand(rest, version)
 	case "version", "--version", "-v":
 		fmt.Println("reasonix", version)
 		return 0
--- a/internal/cli/doctor.go
+++ b/internal/cli/doctor.go
@@ -1,0 +1,31 @@
+package cli
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+	"os"
+
+	"reasonix/internal/doctor"
+)
+
+func doctorCommand(args []string, version string) int {
+	fs := flag.NewFlagSet("doctor", flag.ContinueOnError)
+	jsonOut := fs.Bool("json", false, "print diagnostics as JSON")
+	if err := fs.Parse(args); err != nil {
+		return 2
+	}
+
+	report := doctor.Collect(doctor.Options{Version: version})
+	if *jsonOut {
+		enc := json.NewEncoder(os.Stdout)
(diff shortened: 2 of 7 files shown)


esengine:
Thanks for this — really clean work. Nice separation into `internal/doctor`, the redaction is careful (host-only URLs, key-presence not values, headers dropped) and backed by a dedicated test, and the en/zh help is kept in sync. Merging now. 🙏

PR #2643, opened 2026-06-02
Fix/recursive dir glob

xinrenH:
修复文件搜索只列出当前目录的问题。递归增加深度限制。路径兼容。

### DIFF
--- a/internal/control/refs.go
+++ b/internal/control/refs.go
@@ -97,6 +97,50 @@ func (c *Controller) HasRefs(line string) bool {
 	return len(c.detectRefs(line)) > 0
 }
 
+// resolveBareNames batch-resolves simple filenames (no path separator) that
+// don't exist in cwd. It walks the working tree once and matches every
+// unresolved name against the set, stopping when all are found. This runs in
+// the async ResolveRefs path, never on the TUI event loop.
+func resolveBareNames(refs []ref) []ref {
+	need := map[string]*ref{}
+	var names []string
+	for i := range refs {
+		r := &refs[i]
+		if r.kind != refFile || strings.ContainsAny(r.raw, "/\\") {
+			continue
+		}
+		if _, err := os.Stat(r.raw); err == nil {
+			continue
+		}
+		need[r.raw] = r
+		names = append(names, r.raw)
+	}
+	if len(names) == 0 {
+		return refs
+	}
+	found := 0
+	cwd, _ := os.Getwd()
+	filepath.WalkDir(cwd, func(p string, d os.DirEntry, wErr error) error {
+		if wErr != nil || found == len(names) {
+			return filepath.SkipAll
+		}
+		if d.IsDir() {
+			switch d.Name() {
+			case ".git", "node_modules", ".DS_Store", "__pycache__", ".idea", ".vscode":
+				return filepath.SkipDir
+			}
+			return nil
+		}
+		if r, ok := need[d.Name()]; ok {
+			rel, _ := filepath.Rel(cwd, p)
+			r.path = filepath.ToSlash(rel)
+			found++
+		}
+		return nil
+	})
+	return refs
+}
+
 // FileRefLine reports whether a submitted line is nothing but a path to an
 // existing file — a dragged or pasted file lands as its bare path, which on
 // POSIX starts with '/' and would otherwise be misread as a slash command. The
@@ -117,8 +161,10 @@ func FileRefLine(line string) (string, bool) {
 // strings for any that failed. An empty block means no references resolved.
 // Safe to call off a frontend's event loop; honours ctx for the resource reads.
 func (c *Controller) ResolveRefs(ctx context.Context, line string) (block string, errs []string) {
+	refs := c.detectRefs(line)
+	refs = resolveBareNames(refs)
 	var b strings.Builder
-	for _, r := range c.detectRefs(line) {
+	for _, r := range refs {
 		switch r.kind {
 		case refResource:
 			text, err := c.host.ReadResource(ctx, r.server, r.uri)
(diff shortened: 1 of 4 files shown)


esengine (changes requested):
Thanks for tackling this — the direction is right. "Only lists the current directory" is a real limitation, and the `glob` fallback and the opt-in `ls recursive=true` are both clean: bounded, model-initiated, with noise-dir skipping and a depth guard. Those parts I'd happily take as-is.

Before this can merge, though, two changes hit hot paths in ways that work against the project's cost/perf constraints. These aren't style nits — they need a rethink.

### 1. `detectRefs` does a full-tree walk on the TUI event loop

`resolveSimpleFile` runs `os.Getwd()` + `filepath.WalkDir` over the whole working tree whenever a bare `@name` isn't an exact path in cwd. The problem is *where* this runs: `detectRefs` is reached through `HasRefs`, and `HasRefs` is called synchronously on the Bubbletea update loop (`chat_tui.go:541`). That gate is deliberately cheap — its whole job is to decide *whether* to kick off resolution off-loop. With this change, submitting a line like `@config` (no exact match) blocks the event loop on a full-disk walk; on a large repo that's a visible freeze. And because the walk is per-token, `@a @b @c` is three full passes.

What I'd want instead:

- Keep `HasRefs` free of disk walks — it stays the cheap gate.
- Do filename resolution only in the async `ResolveRefs` path, never on the event loop.
- Resolve all bare names in a **single** pass (collect the unresolved names, walk once, match against the set, stop when all are found), not one walk per token.
- Prefer an existing index over a raw walk where possible (the repo already has codegraph); fall back to a single bounded walk only when there's no index.
- Use the session `workDir`, not `os.Getwd()` — the rest of the tools resolve through `resolveIn(workDir, …)`, and in the desktop multi-session case cwd isn't the workspace root.

### 2. Directory refs now dump an unbounded recursive tree

`readFileRef` changed the directory branch from a one-level listing to a full recursive walk, but `maxFileRefBytes` only caps the *file* branch — the directory listing has no limit. So `@some-large-dir/` injects the entire recursive tree (potentially tens of thousands of lines) straight into the prompt. Token cost is load-bearing for this project, so an unbounded inject is a no-go.

I'd rather keep the **one-level listing as the default** and make recursion opt-in, mirroring the `ls recursive=true` flag you already added — that keeps the default token footprint unchanged and lets the caller ask for depth when they actually want it. If you do keep any recursive path, it needs a hard size/entry cap with a truncation marker like the file branch has.

### Housekeeping

- The branch is ~115 commits behind `main-v2` and conflicts in `refs.go` and `glob.go`; the test jobs never ran because of it. Please rebase onto `main-v2` so CI can actually exercise the change.
- A number of the new comments restate the code (`// Skip the root itself`, `// Try exact path first.`, `// Recursive mode: walk…`) and the `readFileRef` doc comment grew past the length cap. See the comment rules in `CLAUDE.md` — default is none; keep only the non-obvious *why*. Minor, but worth a pass while you're in here.

Happy to review again once the resolution path is off the event loop and the directory default is back to shallow. The core idea is good and I want it in.

xinrenH:
Hi YHH,

Thank you so much for taking the time to provide such detailed feedback — every point is spot-on.

I apologize for not fully grasping the project's core design intent earlier, especially the role of HasRefs as a lightweight gate on the event loop, and how critical token cost is to the project. When you pointed out the detectRefs blocking the event loop and the unbounded recursive injection for directory refs, it hit me that I had introduced completely unreasonable overhead on hot paths — this went far beyond "fixing a small bug." I'm truly sorry for the extra review burden this put on you.

At the same time, I really appreciate your recognition of the overall direction. Honestly, I only did a small piece of patching work — the core ideas all build on the project's existing architecture. Your reply laid out "what the right approach should be" with great clarity — async resolution, single-pass walk with name collection, reusing the codegraph index, using the session's workDir, keeping directory defaults shallow — these are excellent learning material for me.

I'll rework the PR following your guidance:

Move filename resolution entirely to the async ResolveRefs path, keeping HasRefs zero IO.

Add a maxDirEntries=100 truncation guard for directory refs, symmetric with maxFileRefBytes, to prevent unbounded token injection from large directories.

Rebase the branch onto main-v2 and clean up redundant comments.

If I have further ideas or changes down the line, I'll make sure to spend more time understanding the project's constraints and style before submitting, so I don't waste your time.

I sincerely wish the project continued success. Being able to participate in this process is already a tremendous gain for me.

Best,
a taciturn person


zh：
非常感谢你抽出时间给出这么详细的反馈，每一处都点得很准。

很抱歉，我之前确实没能充分理解项目的核心设计意图，尤其是 HasRefs 作为事件循环上的轻量级闸门这个定位，以及 token 成本对整个项目的重要性。你指出的 detectRefs 阻塞事件循环和目录引用无上限递归注入这两个问题，我读完才意识到自己在热路径上引入了非常不合理的开销，这远远超出了“修一个小 bug”的范畴。给你带来额外的审查负担，真的很抱歉。

同时也很感谢你对整体方向的认可。其实我只是做了很小的一点修补工作，核心思路都是建立在项目已有的架构之上的。你在回复里把“正确的做法该是什么”说得非常清楚——异步解析、单次 walk 收集、复用 codegraph 索引、用 session 的 workDir、目录默认保持浅层——这些对我来说是很好的学习材料。

我会按照你的指点重新梳理这个 PR：

将文件名解析完全移到异步的 ResolveRefs 路径，保持 HasRefs 零 IO；

目录引用加上了 maxDirEntries=100 截断保护，与 maxFileRefBytes 对称，防止大目录无限注入 token。；

把分支 rebase 到 main-v2，并清理冗余注释。

如果后续我有其他的想法或改动，也会尽量在提交前先多揣摩项目的约束和风格，争取不浪费你的时间。

真心祝愿项目越来越好。能参与这个过程，对我来说本身就是很大的收获。

很抱歉，我不善言辞。

esengine:
Thanks @xinrenH — merged via #2915 (rebased onto current main-v2, plus a regression test for the glob fallback). Nice catch on the dead recursive-fallback check. 🙏

PR #2671, opened 2026-06-02
feat(memory): add idle daily auto summaries

### LINKED ISSUE #2565: [Feature]: 每日自动汇总记忆 形成一个长期的记忆功能（Daily Auto Memory Summary）
### Version line

当前 remember 只能在当前会话中手动调用，存在以下问题：

跨会话信息丢失 — 用户一天内可能开启多个 /new 或多次启动 Reasonix，每次会话都是独立上下文，前面会话中获取的用户偏好、项目决策、纠正反馈等关键信息无法自动传递到后续会话
无定时机制 — 用户需要手动在每次对话中说"把刚才的讨论存下来"，体验割裂且容易遗漏
无聚合能力 — 没有工具能读取当天所有历史会话并归纳摘要，用户只能自己回忆并口述给 AI

### Proposed solution

每次对话完后 当xxx时间未进行对话 自动触发记忆形成功能 而不是对话玩手动将对话形成记忆 而是汇总整个聊天会话 递增的方式

GTC2080:
Closes #2565

## Summary
- adds opt-in `agent.auto_memory` and `agent.auto_memory_idle` settings
- refreshes a per-project `daily-summary-YYYY-MM-DD` memory after an idle chat window
- merges only new conversation messages into the existing daily summary, keeping the current prompt prefix stable until the next session

## Evidence
- No UI change; screenshots are not applicable.
- Validation: `D:\Go\go\bin\go.exe test ./internal/agent ./internal/control ./internal/config ./internal/boot`
- Validation: `D:\Go\go\bin\go.exe test ./...`

### DIFF
--- a/internal/agent/compact.go
+++ b/internal/agent/compact.go
@@ -57,6 +57,14 @@ What is still in progress or unstarted, and the single most concrete next action
 
 Rules: be terse — bullet points and fragments, not prose. Preserve identifiers, paths, and numbers exactly. Do NOT invent anything not present in the messages; if something is unknown, leave it out rather than guessing.`
 
+const dailyMemorySystemPrompt = `You are updating daily long-term memory for a coding-agent user.
+Merge the existing daily summary with the new transcript incrementally.
+
+Keep only durable information that helps future sessions: user preferences, project decisions, corrections, constraints, active goals, important external references, and unresolved follow-ups.
+Drop ephemeral chatter, obvious tool mechanics, repeated details already covered by the existing summary, and anything already recorded in files or git history unless the transcript adds a non-obvious decision about it.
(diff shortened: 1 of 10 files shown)


esengine:
The engineering here is careful — the idle timer, the cursor-tracked region, the guards around in-flight turns are all well done, and it's gated off by default.

I'm going to pass on it for product fit rather than code. The whole job of this feature is to fire a background summarization call on an idle timer, and that call is a cache-miss by nature (fresh transcript + fresh system prompt each time). "Cheap, cache-first" is the load-bearing claim for this project, so a recurring ambient token spend — even opt-in — points users at exactly the path we tell them we avoid, and adds a non-core surface to maintain.

If we do ambient/long-term memory later, I'd want it as an explicit, user-invoked action (a slash command triggered when the user wants it), not an idle daemon. Closing this on that basis — not a reflection on the implementation, which is good.

PR #2680, opened 2026-06-02
feat(plugin): persist handshake schema & startup latency to user cache

SivanCola:
## Summary

Adds two best-effort sidecar caches under `<UserConfigDir>/reasonix/cache/mcp/` written from a side-goroutine in `plugin.Start` after phase A: `<server>.json` snapshots the handshake (capabilities + tool list with `name/description/schema/readOnly`), and `<server>.stats.json` keeps a rolling window of phase-A durations for auto-demote decisions. No call site reads these files in this PR — they are infrastructure for PR 4. This is PR **3/4** of the MCP-startup overhaul.

## Root Cause / Motivation

Two distinct problems share a persistence layer:

1. Every launch re-handshakes every MCP server even though the tool schema is usually unchanged session-to-session. With a stable cache keyed by a Spec fingerprint, the next launch can register tools optimistically without waiting for the network/subprocess (consumed in PR 4).
2. Reasonix has no record of which configured plugin is chronically slow. The boot path treats a 6-second handshake the same as a 60-millisecond one, so the user keeps eating the same delay. A simple rolling-window of phase-A times turns that latency into a signal the boot path can act on (also consumed in PR 4).

This PR ships the persistence layer alone so reviewers can examine the on-disk format, the atomic-write/fingerprint logic, and the integration into the Start goroutine before any behaviour change lands.

### Cache format and fingerprint
- `CachedSchema = { Version, SpecHash, Capabilities, Tools[], LastValidated }`. `Version=1` gates incompatible future shape changes — older readers treat a higher version as a miss.
- `SpecFingerprint(Spec)` SHA-256s the load-bearing fields (Type/Command/URL/Dir + sorted Args/Env/Headers). Env/Headers iteration order is normalised so Go's randomised map iteration can't churn the fingerprint between launches. Changing a command, URL, or any of those maps invalidates the cache; renaming the server does not (the cache file is named after the server, not the fingerprint).
- Writes go through tmpfile + `os.Rename` in the same parent dir (mirrors `desktop/sessions.go` and `internal/codegraph/install.go`), so a crash mid-write can never leave a half-parsed JSON behind that the next `Load` would mis-handle.

### Stats / Recommend
- `StartupStats = { Version, SamplesMs[], LastSeen }`. SamplesMs is a rolling window of up to 20 latest values (`maxSamples`), atomically rewritten on each Record.
- `Recommend(name, budget, demoteAfter)` returns `Demote=true` when the last `demoteAfter` consecutive samples each exceed `2×budget`. `demoteAfter` defaults to 3 — small enough to catch a regressed plugin within a session or two, large enough that a single hiccup can't demote a healthy one. A missing stats file means "never seen", which intentionally never demotes.

### Wiring
- `plugin.Start`'s goroutine times phase A (`start + listTools`), then fires a side-goroutine that calls `RecordStartup` and `SaveCachedSchema`. Failure is silently dropped (slog.Debug for cache, slog.Warn for surprising stats read errors). The fan-out happens after the goroutine has put its result on the channel, so disk IO is never on the critical path.
- `config.CacheDir()` is added beside the existing `ArchiveDir/SessionDir/MemoryUserDir`, returning `<UserConfigDir>/reasonix/cache`. The whole reasonix user state tree still has a single root; \`rm -rf ~/Library/Application Support/reasonix\` clears everything.
- `cacheableToolsOf([]tool.Tool)` lives in cache.go and extracts the persistable subset from the `*remoteTool` values that `listTools` produced — schema is the already-canonicalized form, so the cache hit path will produce byte-identical schemas across sessions, preserving the prompt-cache prefix (see "Cache stability" below).

- Best-effort by design: no `error` is propagated out of `RecordStartup` or `SaveCachedSchema` into the boot path. Disk full, broken permissions, version mismatches — all degrade silently to "no cache" / "no telemetry".
- Atomic write avoids the classic "next launch sees a half-written file" failure mode that telemetry systems usually inherit.
- Stats's 20-sample rolling window guarantees one bad day fades out: an MCP server that's been slow for 24h but fine since gets re-promoted by the same mechanism.
- Cache stability (preserves prompt-cache prefix): the schema bytes saved here come from the canonicalised `*remoteTool.schema`. When PR 4 reads them back, the resulting placeholder tool reports a `Schema()` that's byte-identical to what the real `*remoteTool` would report after the eventual handshake — so the swap from placeholder to real tool is a no-op for Anthropic ephemeral / DeepSeek prefix caching. PR 4 will pin that contract in tests; this PR sets up the byte-stable serialisation.

## Verification

- `go build ./...`
- `go test ./internal/plugin/... ./internal/config/... ./internal/event/... -count=1 -timeout=120s` — all green.
- `go test ./internal/plugin/... ./internal/config/... ./internal/event/... -race -count=1 -timeout=120s` — all green (atomic write avoids reader/writer overlap; stats writes hold no shared state).
- `go vet ./internal/plugin/... ./internal/config/... ./internal/event/...` — clean.
- Focused cache + canonicalize + stats run: all of `TestCache*`, `TestSpecFingerprint*`, `TestSlug*`, `TestCanonicalize*`, `TestRecord*`, `TestRecommend*`, `TestStatsPathLayout` pass — important because this PR is the persistence half of a cache feature, and canonicalisation underpins the cache-stability claim above.

## Dependencies & how to review

This PR is **stacked on top of PR 1 (#2675) and PR 2 (#2677)**. GitHub does not support cross-fork stacked PRs, so this branch carries three commits:

1. `d437b49e` — PR 1's `feat(plugin): unify StartAll/StartAvailable…`
2. `c88a94e6` — PR 2's `feat(plugin): defer prompts/resources to a background phase`
3. **This PR's incremental layer** — last commit on the branch.

To see just this PR's incremental diff, look at the last commit in the Commits tab, or:
```
git diff c88a94e6..HEAD
```

Once #2675 and #2677 merge to `main-v2`, this branch will be rebased so its diff shows only the final commit. No re-review of PR 1 / PR 2 content is needed here.

- Blocks PR 4 (`tier-based loading with cache-aware lazy placeholders`).
- Should be merged after #2675 and #2677.

### DIFF
--- a/internal/boot/boot.go
+++ b/internal/boot/boot.go
@@ -195,6 +195,10 @@ func Build(ctx context.Context, opts Options) (*control.Controller, error) {
 		for _, t := range ptools {
 			reg.Add(t)
 		}
+		// PhaseB (prompts + resources) runs on the boot ctx — which is the
+		// controller's session-scoped PluginCtx — so the auxiliary surfaces
+		// keep streaming in after Start returns without holding up the agent.
+		go host.StartPhaseB(ctx, sink)
 		if text, ok := MCPStartupNotice(host.Failures()); ok {
 			sink.Emit(event.Event{Kind: event.Notice, Level: event.LevelWarn, Text: text})
 		}
--- a/internal/cli/acp.go
+++ b/internal/cli/acp.go
@@ -123,6 +123,12 @@ func (f *acpFactory) NewSession(ctx context.Context, p acp.SessionParams) (*cont
 		for _, t := range ptools {
 			reg.Add(t)
 		}
+		// Mirror boot.Build: phase B (prompts + resources) is deferred to a
+		// background goroutine on the session ctx so the ACP path also sees
+		// non-empty Host.Prompts()/Resources() once the auxiliary surfaces
+		// stream in. Without this, MCPPrompt and @-ref consumers would stay
+		// empty for the session.
+		go h.StartPhaseB(ctx, p.Sink)
 		if text, ok := boot.MCPStartupNotice(h.Failures()); ok {
 			p.Sink.Emit(event.Event{Kind: event.Notice, Level: event.LevelWarn, Text: text})
 		}
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -526,6 +526,19 @@ func SessionDir() string {
 	return filepath.Join(dir, "reasonix", "sessions")
 }
 
+// CacheDir is the per-user cache root for derived/regenerable artefacts: MCP
+// handshake snapshots, plugin startup-latency telemetry. Lives beside the
+// existing dirs (UserConfigDir/reasonix/...) so the whole reasonix state tree
+// shares one root the user can wipe in a single rm. Empty when the OS dir is
+// unavailable — callers must tolerate that (caching is best-effort).
+func CacheDir() string {
+	dir, err := os.UserConfigDir()
+	if err != nil {
+		return ""
+	}
+	return filepath.Join(dir, "reasonix", "cache")
+}
+
 // MemoryUserDir returns the reasonix user config root (…/reasonix), under which
 // the user-global REASONIX.md and the per-project auto-memory store live. Empty
 // when the user config dir can't be resolved, which disables user-scoped memory.
--- a/internal/event/event.go
+++ b/internal/event/event.go
@@ -68,6 +68,11 @@ const (
 	// ToolResult for long tools like bash so a frontend can show live progress.
 	// Appended last to keep the Kind values before it wire-stable.
 	ToolProgress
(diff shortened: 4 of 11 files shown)


esengine:
Rebased onto `main-v2` (PR 1 #2675 + PR 2 #2701 had landed) and merged as #2703 with your authorship preserved. Thanks @SivanCola — the handshake-schema + startup-latency cache is in. I added one fix commit on top: the test cache helpers only isolated HOME/XDG_CONFIG_HOME, which os.UserConfigDir ignores on Windows, so the cache/stats tests were writing the real user cache and TestRecordStartupAppends accumulated samples across local re-runs — now they isolate APPDATA too. (Unrelated: TestBackgroundKill flaked twice on the Windows runner before going green; that's a separate de-flake.) PR 4 is the last one whenever you're ready.

PR #2717, opened 2026-06-02
fix(cli): fall back to OSC 52 clipboard when platform tool fails in tmux/SSH

### LINKED ISSUE #2716: [Bug]: Cannot copy conversation text when running TUI inside tmux
### Version line

### Exact version

latest main-v2 (built from source)

### What happened?

When running the Reasonix TUI inside a tmux session, selecting conversation text with the mouse and triggering copy does not place the text into the system clipboard. The selection highlight appears correctly, but the clipboard remains empty (or unchanged).

**Expected behavior:** Selected text should be copied to the system clipboard regardless of whether the TUI runs inside tmux or in a plain terminal.

### Steps to reproduce

1. Start a tmux session (`tmux`)
2. Launch the Reasonix TUI (`reasonix`)
3. Have a conversation with some content in the transcript
4. Use the mouse to select/drag over conversation text
5. Trigger the copy action (mouse release / keybinding)
6. Try to paste the text into another application — nothing is copied

### OS / platform

Linux (Arch Linux), tmux 3.6, `$TERM=xterm` inside tmux

### Relevant logs or output

The TUI uses `github.com/atotto/clipboard` for clipboard writes (`internal/cli/transcript.go:copyToClipboard`). This library shells out to `xclip`, `xsel`, or `wl-copy`, which cannot reach the system clipboard from inside tmux unless:

- tmux's `set-clipboard` option is enabled (`set -g set-clipboard on`), **and**
- the application emits OSC 52 escape sequences that tmux can intercept and forward to the outer terminal.

`atotto/clipboard` does not emit OSC 52 — it writes directly to the X11/Wayland clipboard, which is typically not accessible from within tmux's pane environment (especially over SSH or when `$DISPLAY` / `$WAYLAND_DISPLAY` is not forwarded).

**Possible fix:** Use a clipboard library that supports OSC 52 (e.g. `charm.land/x/ansi` already has `ansi.SetClipboard()` / OSC 52 support in the dependency tree), with a fallback to `atotto/clipboard` for non-muxed terminals. bubbletea v2 also has built-in `tea.SetClipboard` / `ClipboardMsg` that can be used instead of a raw `tea.Cmd`.

lanshi17:
## Problem

When running the TUI inside tmux (or over SSH), right-click → copy of selected conversation text silently fails. The selection highlight works, but nothing reaches the system clipboard.

**Root cause:** `copyToClipboard` in `internal/cli/transcript.go` uses `atotto/clipboard`, which shells out to `xclip` / `xsel` / `wl-copy`. Inside tmux the X11/Wayland display server is typically unreachable (`$DISPLAY` / `$WAYLAND_DISPLAY` not forwarded), so these tools fail silently (the error was discarded with `_ =`).

Closes #2716.

## Fix

When the platform clipboard tool fails, fall back to an **OSC 52** escape sequence via `ansi.SetSystemClipboard()` (already in the dependency tree via `charmbracelet/x/ansi`). OSC 52 is the standard mechanism for terminal applications to manipulate the clipboard through the terminal emulator — tmux intercepts it and forwards to the outer terminal.

```
atotto/clipboard (xclip/xsel/wl-copy)
        │
        ▼  success → done
        │
        ▼  failure → OSC 52 via ansi.SetSystemClipboard()
```

This is a zero-cost addition for local terminals (atotto succeeds, no fallback triggered) while fixing tmux, SSH, and other multiplexed environments.

## Changes

- **`internal/cli/transcript.go`:** `copyToClipboard` now checks the error from `clipboardWriteAll` and falls back to `tea.Printf(osc52_sequence)`. Extracted `clipboardWriteAll` as a package-level var for testability.
- **`internal/cli/transcript_test.go`:** Two tests covering both the happy path (platform tool succeeds → nil msg) and the fallback path (platform tool fails → OSC 52 Cmd emitted).

## Testing

```
go test ./internal/cli/ -run TestCopyToClipboard -v
=== RUN   TestCopyToClipboard_PlatformSuccess
--- PASS
=== RUN   TestCopyToClipboard_OSC52Fallback
--- PASS
```

`gofmt -l .` is clean.

### DIFF
--- a/internal/cli/transcript.go
+++ b/internal/cli/transcript.go
@@ -22,10 +22,20 @@ func wrapTranscript(s string, width int) string {
 	return lipgloss.NewStyle().Width(width).Render(s)
 }
 
-// copyToClipboard writes text to the system clipboard off the event loop.
+// clipboardWriteAll is the clipboard writer used by copyToClipboard. Tests
+// may replace it to simulate a platform-tool failure (tmux / SSH scenario).
+var clipboardWriteAll = clipboard.WriteAll
+
+// copyToClipboard writes text to the system clipboard. It first tries the
+// platform clipboard tool (xclip, xsel, wl-copy, pbcopy, …) via atotto; when
+// that fails — typically inside tmux or over SSH where the display server is
+// unreachable — it falls back to an OSC 52 escape sequence that tmux and most
+// modern terminals forward to the host clipboard.
 func copyToClipboard(text string) tea.Cmd {
 	return func() tea.Msg {
-		_ = clipboard.WriteAll(text)
+		if err := clipboardWriteAll(text); err != nil {
+			return tea.Printf("%s", ansi.SetSystemClipboard(text))
+		}
 		return nil
 	}
 }
--- a/internal/cli/transcript_test.go
+++ b/internal/cli/transcript_test.go
@@ -1,6 +1,13 @@
 package cli
 
-import "testing"
+import (
+	"errors"
+	"strings"
+	"testing"
+
+	tea "charm.land/bubbletea/v2"
+	"github.com/charmbracelet/x/ansi"
+)
 
 func TestScrollbarThumb(t *testing.T) {
 	if _, size := scrollbarThumb(10, 0, 5); size != 0 {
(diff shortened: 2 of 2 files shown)


esengine:
Thanks @lanshi17 — the diagnosis is spot on (atotto can't reach the display server inside tmux/SSH; OSC 52 is the right mechanism) and the `clipboardWriteAll` test seam was a nice touch.

One issue kept it from working in practice: the fallback returned `tea.Printf(...)`, which is a `tea.Cmd`. `copyToClipboard`'s function returns that command as the `tea.Msg`, and a command-as-message matches no case in bubbletea's runtime dispatch nor in the model's Update, so it's silently dropped — no OSC 52 is ever written. (The test passed only because it manually called the inner command, which the real runtime never does.)

The fix needs the *message*, run inline: `return tea.SetClipboard(text)()` — `setClipboardMsg` is handled by the runtime, which emits the escape. I verified this end-to-end (real `tea.Program` → `\x1b]52;c;…` in the output) and landed it as #2719, keeping your fallback structure and test seam. Closing this in favor of that.

PR #2719, opened 2026-06-02
fix(cli): copy via OSC 52 when the platform clipboard tool fails (tmux/SSH)

### LINKED ISSUE #2716: [Bug]: Cannot copy conversation text when running TUI inside tmux

esengine:
Closes #2716. Supersedes #2717 (same root cause + direction — credit to @lanshi17 for the diagnosis — but its fallback never actually emitted OSC 52; see below).

## Root cause
`copyToClipboard` discarded the atotto error (`_ = clipboard.WriteAll`). Inside tmux / over SSH, xclip/xsel/wl-copy can't reach the display server, so the copy silently failed.

## Fix
On atotto failure, fall back to OSC 52 (tmux + modern terminals forward it to the host clipboard) via bubbletea v2's built-in `tea.SetClipboard`.

The subtlety that #2717 missed: the fallback must return `tea.SetClipboard(text)()` — the **message**, executed here — not `tea.SetClipboard(text)`/`tea.Printf(...)` (a command). A `tea.Cmd` returned as a `tea.Msg` matches no case in the runtime's dispatch (tea.go event loop) nor in the model's Update, so it's silently dropped and emits nothing. The message (`setClipboardMsg`) is handled by the runtime at tea.go (`p.execute(ansi.SetSystemClipboard(...))`), which writes the escape to the terminal.

## Verification
- Unit: `TestCopyToClipboardPlatformSuccess` (success → nil, no fallback) and `TestCopyToClipboardOSC52Fallback` (failure → msg `reflect.DeepEqual` to `tea.SetClipboard(text)()`, i.e. the runtime-handled message, not a dropped command).
- End-to-end: ran a real `tea.Program` with output captured to a buffer; `tea.SetClipboard("foo")` produces `\x1b]52;c;Zm9v…` (OSC 52 + base64) in the terminal output. The #2717 form produced nothing (the command was dropped) — confirmed by type: `tea.Printf(...)` is a `tea.Cmd`, `tea.SetClipboard(x)()` is the handled `setClipboardMsg`.
- Kept #2717's `clipboardWriteAll` var seam for testing the failure path without a display server. Full `go test ./internal/cli` green.

### DIFF
--- a/internal/cli/transcript.go
+++ b/internal/cli/transcript.go
@@ -22,10 +22,22 @@ func wrapTranscript(s string, width int) string {
 	return lipgloss.NewStyle().Width(width).Render(s)
 }
 
-// copyToClipboard writes text to the system clipboard off the event loop.
+// clipboardWriteAll is the platform clipboard writer; a var so tests can force
+// the failure path (the tmux / SSH scenario) without a real display server.
+var clipboardWriteAll = clipboard.WriteAll
+
+// copyToClipboard writes text to the system clipboard. It first tries the
+// platform tool (xclip / xsel / wl-copy / pbcopy) via atotto; when that fails —
+// typically inside tmux or over SSH where the display server is unreachable — it
+// falls back to OSC 52, which tmux and modern terminals forward to the host
+// clipboard. tea.SetClipboard's command is *run* here so the message it yields
+// (handled by the runtime) reaches the event loop; returning the command itself
+// would be dropped as an unrecognized message and emit nothing.
 func copyToClipboard(text string) tea.Cmd {
 	return func() tea.Msg {
-		_ = clipboard.WriteAll(text)
+		if err := clipboardWriteAll(text); err != nil {
+			return tea.SetClipboard(text)()
+		}
 		return nil
 	}
 }
--- a/internal/cli/transcript_test.go
+++ b/internal/cli/transcript_test.go
@@ -1,6 +1,12 @@
 package cli
 
-import "testing"
+import (
+	"errors"
+	"reflect"
+	"testing"
+
+	tea "charm.land/bubbletea/v2"
+)
 
 func TestScrollbarThumb(t *testing.T) {
 	if _, size := scrollbarThumb(10, 0, 5); size != 0 {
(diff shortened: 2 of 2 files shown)


meteor614:
the platform clipboard fallback still doesn't solve the remote-SSH case

This was a correct fix for #2716 (tmux with no display server) and I appreciate catching the subtle Cmd vs Msg issue that #2717 missed.

However, there's a class of cases the fallback pattern can't address: when Reasonix runs on a remote node over SSH and the remote has xclip/wl-copy/pbcopy
available. In that scenario clipboardWriteAll succeeds — copying to the remote machine's clipboard — so the if err != nil branch never fires, and OSC52 is never
emitted. The user's local terminal clipboard gets nothing.

The error-based fallback fundamentally cannot distinguish "local clipboard works on the user's machine" from "remote clipboard works on the server." Both look
like success.

A simpler approach is to drop the platform clipboard tool entirely and always emit OSC52:

│ func copyToClipboard(text string) tea.Cmd {
│     return tea.SetClipboard(text)
│ }

Rationale:

• OSC52 is in-band — it travels through the terminal stream, so SSH and tmux forward it transparently to the client terminal. That's the clipboard we actually
  want every time.
• Modern terminals all support it (iTerm2, kitty, WezTerm, Windows Terminal, foot, Alacritty, tmux 2.6+).
• atotto/clipboard targets the machine Reasonix runs on, not the user's display — which is wrong for any remote scenario regardless of whether the remote has
  display tools.
• This also removes an external dependency, the clipboardWriteAll test seam, and the tea.Batch-based parallel execution added to work around it.

I've made this change locally — happy to open a follow-up PR if it's useful.

PR #2721, opened 2026-06-02
feat(cli): interactive /resume picker with keyboard navigation

Ho-J:
## Summary

Bare `/resume` now opens an interactive picker overlay that lets the user browse saved sessions with **↑/↓** and confirm with **Enter**, instead of just printing a static list. The old `/resume <n>` syntax still works.

## Changes

- **New `resume_picker.go`** — interactive overlay modal (struct + open/close/key handling/render), mirroring the `rewindPicker` pattern
- **`chat_tui.go`** — wired the picker's key dispatch, rendering, and status line
- **`complete.go`** — `acceptCompletion` now re-filters the menu after insertion so `/resume` immediately shows numbered session completions; Enter on a fully-typed single item closes the menu and submits the command
- **`resume.go`** — bare `/resume` now **both** writes the list to scrollback (above input) AND opens the interactive picker (below input)
- **i18n** — added `ResumePickTitle`, `ResumePickHint`, `StatusResumePicker` for en + zh locales
- **Tests** — `TestResumeDispatchOpensPicker`, `TestResumePickerNavigateAndSelect`, `TestResumePickerEscDismisses`

## UX

| Before | After |
|--------|-------|
| `/resume` prints a static list; user must re-type `/resume <n>` | `/resume` shows scrollback list (above) + interactive picker (below) with ↑/↓ / Enter / Esc |
| Accepting the `/resume` completion required an extra keystroke | Select `/resume` from menu → immediately shows numbered sessions |
| Typing `/resume 3` + Enter looped in completion | `/resume 3` + Enter submits and resumes session 3 |

### DIFF
--- a/internal/cli/chat_tui.go
+++ b/internal/cli/chat_tui.go
@@ -174,8 +174,11 @@ type chatTUI struct {
 	// rewind holds the Esc-Esc / "/rewind" picker (nil when closed); while set,
 	// keys drive it and it renders as an overlay. lastEsc times the double-Esc
 	// gesture that opens it on an empty composer.
-	rewind  *rewindPicker
-	lastEsc time.Time
+	rewind *rewindPicker
+	// resumePick is the interactive "/resume" session picker overlay. Non-nil
+	// while the user browses saved sessions with ↑/↓ and confirms with Enter.
+	resumePick *resumePicker
+	lastEsc    time.Time
 
 	// lastCtrlCAt records when Ctrl+C was pressed while idle on an empty
 	// composer, enabling a "press again to quit" confirmation pattern (1.5s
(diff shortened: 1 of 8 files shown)


esengine:
Thanks @Ho-J — nice feature, and the picker cleanly mirrors the rewindPicker pattern. Landed as #2726 with your feature commit kept as-is, plus two small fixups so it goes green in CI:

1. **gofmt** — the new i18n keys left messages_en.go/messages_zh.go unaligned, and resume.go/chat_tui.go (the new `resumePick` field) needed re-formatting; the Unix gofmt leg would have failed. (Your PR's CI only ran the `label` check because the workflow approval was pending, so this didn't surface here.)
2. **Flaky test** — `TestResumePickerNavigateAndSelect` assumed b.jsonl was the most recent, but two sessions saved back-to-back can share an mtime tick (hit on the CI Windows runner), which tie-breaks to a.jsonl by path. Pinned distinct mtimes with `os.Chtimes`.

Verified locally (full `go test ./internal/cli ./internal/i18n`, target test ×20) and on CI (all platforms + race green). Closing in favor of #2726.

PR #2768, opened 2026-06-02
[fix] desktop model switch rebuilds

### LINKED ISSUE #2644: [Bug]:  bash sandbox requested but unavailable on this platform; running bash unconfined
### Version line

### Exact version

dev

### What happened?

Image

### Steps to reproduce

复现步骤，
1、设置-模型服务-填写mimo模型key
2、填写后 反映很慢 但能成功
3、对话框内切换模型
4、cmd console输出警告，且未变更模型

### OS / platform

Windows11

### Relevant logs or output

```shell

```

paradoxSCH:
Closes #2644

## Summary
- route desktop boot diagnostics through a quiet writer so inactive bash sandbox warnings do not leak to the Windows desktop console during model/settings rebuilds
- preserve the existing controller until a replacement rebuild succeeds, and surface failed model switches in the UI instead of swallowing the error
- keep the latest session-path carry-forward behavior while avoiding controller teardown on failed rebuilds
- restore the embedded frontend dist placeholder and fix the ReactElement type annotation needed by WorkspacePanel
- make the desktop autosave coalescing test wait for the background snapshot loop to go idle before temp-dir cleanup

## Root cause
On Windows the default bash sandbox mode is `enforce`, but there is no OS sandbox implementation for bash on that platform. The CLI warning is correct for terminal flows, but desktop rebuilds controllers during model/settings changes and inherited the default stderr path, so the warning appeared during model switching. At the same time, failed rebuilds could close or clear the existing controller before the replacement had been built.

## Latest rebase
Rebased onto `main-v2` at `a478e501` on 2026-06-03. The PR remains mergeable after resolving conflicts in `desktop/app_test.go` and `desktop/settings_app.go`.

## Verification
- `go test . -run TestScheduleSnapshotCoalesces -count=50` from `desktop/`
- `go vet ./...` from `desktop/`
- `go build ./...` from `desktop/`
- `go test ./internal/boot ./internal/config ./internal/sandbox`

`npm run typecheck` from `desktop/frontend/` currently fails on latest upstream in untouched files:
- `src/components/ApprovalModal.tsx`: `RefObject<HTMLDivElement | null>` vs `RefObject<HTMLDivElement>`
- `src/components/AskCard.tsx`: same ref type mismatch

Note: local commit/push hooks currently run npm from the repository root where no `package.json` exists, so push updates used `--no-verify` after the checks above passed.


esengine:
Thanks @paradoxSCH — the close-before-rebuild bug is real and still present on main-v2. #2928's multi-tab refactor moved every field this PR touched (`a.ctrl`→`tab.Ctrl`, etc.), so it can't be rebased textually; I re-applied the load-bearing fixes (build→swap→close-on-success in both `SetModel` and `rebuild`, surface the error via `local_notice`, quiet boot stderr) against the new `WorkspaceTab` model in #3003, with a regression test. Crediting you on it. Closing in favor of #3003.

PR #2770, opened 2026-06-02
feat(serve): runtime effort switching via /effort <level>

lanshi17:
## Summary

Adds `/effort <level>` slash command to the HTTP SSE frontend for runtime reasoning effort switching.

### How it works

1. Intercept `/effort <level>` in the submit handler (before `ctrl.Submit()`)
2. `switchEffort()`: load config → resolve current provider → normalize effort level → persist to disk config
3. Delegate to `switchModel()` for thread-safe controller rebuild with carried history

### Supported levels

| Provider | Levels | Notes |
|----------|--------|-------|
| DeepSeek | `auto`, `high`, `max` | `low`/`medium` → `high`, `xhigh` → `max` |
| Anthropic | `auto`, `low`, `medium`, `high`, `xhigh`, `max` | |

### Changes

- `internal/serve/serve.go`: +53 lines (switchEffort method, /effort intercept)
- `internal/serve/index.html`: +1 line (/effort in slash menu)

All changes scoped to `internal/serve/` — zero kernel modifications.

### DIFF
--- a/internal/serve/index.html
+++ b/internal/serve/index.html
@@ -368,6 +368,7 @@ const SLASH_CMDS = [
   {cmd:'branch',desc:'Create branch'},
   {cmd:'switch',desc:'Switch branch'},
   {cmd:'model',desc:'List/switch models'},
+  {cmd:'effort',desc:'Reasoning effort level'},
   {cmd:'mcp',desc:'MCP servers'},
   {cmd:'skill',desc:'Skills'},
   {cmd:'hooks',desc:'Hooks'},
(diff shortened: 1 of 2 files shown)


esengine:
Thanks for this — the serve-only scoping and the per-provider level table are exactly right. I've picked it up in #2926 with two small fixes so it holds on the common config: the original wrote effort straight to the user config, which errors with `no provider` when the active provider has no `[[providers]]` block yet (the default-DeepSeek case), and Anthropic needs `thinking=adaptive` set alongside effort or the levels stay inert. Both now go through a shared `applyEffortEdit` matching the CLI/desktop path. Your commit and authorship are carried over. Closing in favour of #2926.

Issue #2573, opened 2026-06-01
[Feature]: 开机自启与静默启动


wwu01100-creator:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### What problem does this solve?

希望作者桌面端搞一个开机自启与静默启动的功能

esengine:
Thanks — but OS-level autostart and silent launch aren't planned for the core desktop app. Reasonix desktop is an interactive tool rather than a resident service, and login-item/registry integration is a per-OS support surface disproportionate to the benefit. You can get the same effect today with OS facilities (Windows: Task Scheduler or `shell:startup`; macOS: Login Items).

If a resident/bot-first usage mode ever becomes a first-class desktop scenario, this would be natural to revisit. Closing as not planned for now.

PR #2774, opened 2026-06-02
feat(desktop): add auto-start on login and silent-start with tray icon(#2573)

enen323:
Implement OS-level auto-start registration (Windows registry / macOS LaunchAgent / Linux autostart), system tray icon for hidden-window start(Windows Shell_NotifyIconW), and console-window suppression for MCP subprocesses on Windows GUI builds.


SivanCola (changes requested):
Thanks for taking this on — the cross-platform coverage is solid and the overall design is sound. In particular, deriving silent-start from the persisted `desktop-settings.json` (rather than a launch flag) is the right call, which is why registering the bare `os.Executable()` path for auto-start is correct. The persistence surface is also safe: HKCU (no elevation), `exec.Command` with an argv array (no shell injection), and removable via disable. I also confirmed this does not touch any prompt-cache-sensitive surface (tool schema bytes/order, provider serialization, system prompt).

Before merging, please address the following.

**Blocking**

1. **Tray "Quit" doesn't quit the app.** In `tray_windows.go`, the `cmdQuit` branch calls `PostQuitMessage(0)` (which only exits the tray's own message loop) followed by an empty `go func(){}`. The app process keeps running with no window — clicking Quit just removes the tray icon. It should call `wailsRuntime.Quit(t.ctx)`.

2. **Data race in `desktop_settings.go`.** `setDesktopSettings` reads the package-global `ds.AutoStart` without holding `dsMu` (`if s.AutoStart != ds.AutoStart`), while `getDesktopSettings`/`loadDesktopSettings` both guard `ds` with the mutex. `go test -race` will flag this — please read the previous value inside the lock.

3. **No tests, and the path-semantics change is untested on the platform it affects.** This PR adds zero `_test.go` coverage. The `resolveIn`/`resolveCustomPaths` changes only alter behavior on Windows (where `/foo` and `\foo` are now treated as absolute), but the existing `TestResolveIn` cases are all platform-neutral, so Linux CI never exercises the new `os.IsPathSeparator(p[0])` branch. Please add focused regression tests for: auto-start register/unregister path construction, the desktop-settings persistence round-trip, and the new Windows-rooted path branch.

4. **Scope creep.** The `workspace.go` / `skill.go` `filepath.IsAbs` fixes are unrelated to auto-start. They are a defensible Windows path-consistency fix (and Linux behavior is unchanged), but bundling them here makes the PR harder to review and they ship without tests. Please either split them into a dedicated PR, or keep them here *with* the regression test from (3) plus a short comment explaining the Windows rationale. For the record, `resolveIn` honors absolute paths by design (the write-confiner, not `resolveIn`, enforces the workspace boundary), so this is not a new escape — but it does widen what counts as "absolute" on Windows, which is exactly why it deserves an explicit test.

**Suggestions (non-blocking)**

- `toggleWindow` only ever shows the window (the comment admits it cannot hide); tracking visibility would make the left-click toggle behave as its name implies.
- `removeIcon` deletes the icon without `NIF_GUID`, but it was added *with* a GUID. On Windows 10+ the delete may not match the registered identity and can leave a stale icon — include the GUID in the delete `NOTIFYICONDATA`.
- The registered exe path is not quoted/escaped (Windows registry value, macOS plist XML, Linux desktop `Exec=`). Low risk since the path comes from `os.Executable()`, but quoting/escaping is more robust for paths containing spaces or special characters.

Also, the PR is currently `CONFLICTING` against `main-v2` — please rebase onto the latest `origin/main-v2`.

Once these are addressed and CI stays green, I'll be happy to merge this into `main-v2`.

esengine (changes requested):
#2573 is a real ask and the Win32 tray work is better than most hand-rolled attempts (icon parsing, GUID, TaskbarCreated re-add). Blockers:

1. **Tray Quit never quits the app**: `wndProc`'s `cmdQuit` posts `PostQuitMessage` to the tray loop, and the follow-up goroutine contains only comments — `wailsRuntime.Quit(t.ctx)` is never called. Quit kills the icon and leaves the app running.
2. Quote the exe path written to the Run key — unquoted paths with spaces fail to launch.
3. `setDesktopSettings` reads `ds.AutoStart` without holding `dsMu` — data race.
4. Split the unrelated internal/skill + workspace path changes into their own PR; delete the commented-out `NIM_SETVERSION` line.
5. `toggleWindow` always shows and never hides (the comment admits it) — fine to defer, but say so in the description.

Both settings defaulting to off is the right call. Rebase onto current main-v2, push, and we'll approve-and-run fork CI.

esengine:
Thank you for the substantial implementation — registry/LaunchAgent/autostart plus tray integration is real work. Closing this one for two reasons, neither of which is about code quality:

1. **Staleness**: this predates two months of heavy desktop restructuring (layout store, tab/runtime rework, the frontend build CI gate), so it would need a near-total rebase onto current main-v2.
2. **Product timing**: auto-start + silent tray operation only makes sense once resident/bot-host desktop operation is a declared first-class mode; until then we don't want to own the per-OS autostart surface (see #2573).

If resident mode gets promoted and you're still interested, a fresh PR on current main-v2 would be very welcome — the tray/console-suppression pieces especially.

PR #2829, opened 2026-06-03
feat(config): read v0.x ~/.reasonix/config.json mcpServers

esengine:
Part of #2795 — MCP servers vanished when upgrading from the v0.x (TypeScript) line.

**Why:** v0.x stored MCP servers as `mcpServers` in `~/.reasonix/config.json`. v2 reads `[[plugins]]` in `reasonix.toml` (+ user `config.toml`) plus a project `.mcp.json`, and never looked at the old file — so an upgrader's servers silently disappeared. (This is the `mcpServers` vs `mcp[]` confusion in the report.)

**Fix:** read the legacy config's `mcpServers` (honouring its `mcpDisabled` list) as a **lowest-priority** source merged into `[[plugins]]`. A v0.x user's servers keep working with zero edits; anything in the v2 config or `.mcp.json` still wins on a name collision. A stale/malformed legacy file yields `nil` and never blocks startup.

- Reuses the existing `mcpServers` parser (extracted `specsToEntries`).
- Unit-tested: mapping, `mcpDisabled` skip, absent/malformed/empty → nil.
- README (EN + zh) note the v0.x compat.

Both the CLI and desktop go through `config.Load()`, so this also helps the desktop↔CLI MCP sync part of #2795 for upgraders who still have servers only in the old file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,11 @@ sources are merged; on a name collision `reasonix.toml` wins.
 }
 ```
 
+**Upgrading from `0.x`?** Your old `~/.reasonix/config.json` is still read for its
+`mcpServers` (honouring `mcpDisabled`) as a lowest-priority source, so MCP servers
+keep working — move them into `reasonix.toml`'s `[[plugins]]` or a `.mcp.json` when
+convenient.
+
 ### Slash commands
 
 In `reasonix chat`, built-in commands (`/compact`, `/new`, `/rewind`, `/tree`,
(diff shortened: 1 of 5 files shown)


esengine:
Landed via #2912 — same change rebased onto current main-v2 (resolved the overlap with the `pluginEntryFromMCPSpec` refactor that landed meanwhile). Couldn't update this branch in place because force-pushing the PR branch is blocked, so I pushed the rebase as a fresh branch and merged that. Thanks!

PR #2834, opened 2026-06-03
Improve desktop MCP startup behavior and settings UX

SivanCola:
## Summary
- split MCP startup into lazy, background, and eager modes, with persisted settings and codegraph cold-start warmup handling
- improve the MCP panel with clearer connection strategy choices, inline config editing, tool inspection, auth-aware failure actions, log viewing, and delete confirmation
- keep mock data, desktop bindings, tests, and English/Chinese locale strings aligned for the new MCP states and controls

## Benefits
- avoids first-open codegraph startup timeouts being shown as a blocking MCP error
- prevents lazy MCPs from being unintentionally disabled after app restart while still allowing heavier MCPs to connect only when needed
- makes MCP settings easier to understand by replacing ambiguous startup wording with explicit connection strategies
- gives users safer recovery paths for failed or auth-required MCPs, plus safer deletion through confirmation

## Verification
- /opt/homebrew/bin/go test ./internal/config ./internal/boot
- /opt/homebrew/bin/go test ./... (desktop module)
- git diff --check origin/main-v2..HEAD

Note: frontend build still reports the existing large chunk warning.


esengine:
Merged — thank you, this is a great piece of work. The lazy/background/eager split plus the cold-codegraph background warmup is exactly the behaviour we wanted, and the MCP panel is much clearer now. Tests and locale coverage were thorough.

I'll fold a couple of tiny cleanups into a follow-up so you don't have to round-trip on them.

PR #2871, opened 2026-06-03
Add Taskfile.yml as Makefile companion; fix Windows build suffix

Qxy0happy:
- Add Taskfile.yml with 9 tasks (default, build, install, vet, fmt, test,
  hooks, cross, clean, e2e-codegraph) mirroring Makefile targets
- Fix Makefile build: add GOEXE variable so -o path correctly appends .exe
  on Windows (real bash, or PowerShell on a Windows host without bash)
- Update README.md, README.zh-CN.md, docs/MIGRATING.md to reflect
  platform-dependent binary suffix bin/reasonix(.exe)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 VERSION := $(shell git describe --tags --always 2>/dev/null || echo dev)
 LDFLAGS := -s -w -X main.version=$(VERSION)
+GOEXE := $(shell go env GOEXE)
 
 # CodeGraph release pinned for the bundled MCP server / e2e test. Bump together
 # with any change to the integration in internal/codegraph.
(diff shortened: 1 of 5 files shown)


esengine:
Thanks for this! The Windows build-suffix fix is a real bug — `make build` was producing an extensionless binary on Windows — so I've pulled that part into #2896 with your authorship preserved, and it'll land there.

I'm going to pass on the `Taskfile.yml` though: I'd rather not carry a second build system alongside the Makefile, since the two inevitably drift out of sync. Nothing wrong with the Taskfile itself — just keeping the build surface to one file.

Closing in favour of #2896. Appreciate the contribution!

PR #2874, opened 2026-06-03
feat(desktop): update the skill page source management UI

SivanCola:
## Summary

This PR focuses on the desktop Agent Skills page.

- Reworks the skill source section so source management is hidden by default and opens through a dedicated Manage state.
- Adds clearer source labeling and actions: source badges, non-zero-only source summaries, Rescan wording, source-specific skill previews, and remove-source confirmation for user-configured paths.
- Moves diagnostic paths into the management flow and keeps them hidden from the default skill page.
- Adds backend data for skills grouped by source root so the UI can show which skills come from each configured or built-in path.
- Updates Chinese and English copy for the refreshed source and skill summaries.

## Validation

- `npm run build` in `desktop/frontend`
- `/opt/homebrew/bin/go test ./...` in `desktop`
- `git diff --check origin/main-v2..HEAD`


SivanCola:
@codex review

PR #2885, opened 2026-06-03
feat: add persistent approval option to the approval process

YoungDan-hero:
This update introduces a new 'persist' parameter to the approval functionality across the application. The 'Approve' method now accepts this parameter, allowing users to choose to always allow certain actions, which will be saved for future sessions. The UI has been updated to reflect this change, including new options in the approval modal and corresponding translations in the locale files. Additionally, tests have been adjusted to accommodate the new parameter, ensuring consistent behavior across the application.


esengine:
Solid work — this threads the new `persist` choice cleanly through every approval entry point (desktop, ACP, HTTP serve, and the CLI TUI) and wires it into the dormant `Gate.OnRemember` hook so "always allow" actually lands an allow rule in the active config. Nice touch excluding plan approvals from persistence. Merging — thanks!

PR #2888, opened 2026-06-03
Fix desktop MCP failure cards showing connection mode

SivanCola:
## Summary
- Show the configured MCP connection mode inside desktop failure cards so failed servers still display the user's selected startup behavior.
- Allow changing the connection mode directly from a failed MCP card using the same radio controls as normal server details.
- Keep edited MCP config saved when the immediate reconnect fails, recording the reconnect failure on the MCP failure card instead of treating the edit as failed.
- Add regression tests that verify failed MCP capability data keeps the configured tier after a restart-like failure snapshot and after a failed reconnect from editing.
- Relax a timing-only batch parallelism test bound so slower Windows CI runners do not fail even when the batch remains below the all-serial runtime.

## Benefits
- Prevents users from thinking their connection mode reset when the MCP command, config, or auth fails on the next session.
- Keeps the visual model consistent: the MCP can be failed while still showing the saved connection strategy.
- Makes recovery simpler because users can adjust the startup behavior or edit config without first clearing the failure state.
- Avoids blocking this desktop MCP fix on an unrelated Windows timing flake.

## Validation
- `/opt/homebrew/bin/go test ./internal/agent`
- `/opt/homebrew/bin/go test ./...` in `desktop`
- `npm run typecheck` in `desktop/frontend`
- `npm run build` in `desktop/frontend` (existing Vite chunk-size warning only)
- `git diff --check origin/main-v2...HEAD` and `git diff --check`
- Browser check on `http://127.0.0.1:5174/`: failed MCP card now shows the connection mode radio group

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -1318,7 +1318,7 @@ func (a *App) UpdateMCPServer(name string, in MCPServerInput) error {
 	if !sessionDisabled && (wasConnected || wasFailed || updated.ResolvedTier() != "lazy") {
 		if _, err := a.ctrl.ConnectConfiguredMCPServer(name); err != nil {
 			recordMCPFailure(a.ctrl, updated, err)
-			return fmt.Errorf("saved config, but reconnect failed: %w", err)
+			return nil
 		}
 	}
 	return nil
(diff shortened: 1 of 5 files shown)


esengine:
Nice fix — surfacing the configured mode on failed cards (and letting users change it without clearing the failure first) is exactly the right call, and it lines `UpdateMCPServer` up with how `SetMCPServerTier` already records-and-continues on a reconnect miss. Good regression coverage too. Merging — thanks!

PR #2935, opened 2026-06-03
fix(desktop): prevent workspace git console flashes on Windows

### LINKED ISSUE #2930: [Bug]: Windows 桌面版操作侧边栏时频繁闪烁 CMD/Git 终端窗口
### Version line

### Exact version

1.0.0

## 问题描述

Windows 11 桌面端（installer 安装），在执行以下操作时会反复弹出黑色终端窗口然后迅速闪退：

- 展开右侧 workspace 侧边栏
- 点击侧边栏中"文件"/"改动"/"刷新"任一按钮
- 左侧切换会话

终端窗口无任何内容，弹出后瞬间关闭，严重影响使用体验。

Image

截图发现，发起了 git.exe 的调用：

Image

## 原因

`internal/proc/hide_windows.go` 中的 `HideWindow()` 只设置了 `CREATE_NO_WINDOW` 标志，缺少 `DETACHED_PROCESS`。`CREATE_NO_WINDOW` 阻止子进程新建控制台，但 git.exe 等程序会派生子进程（index 操作、hook 等），子进程可能绕过该标志重新分配控制台。`DETACHED_PROCESS` 从创建之初就切断进程树与控制台的关联。

## 修复方案

改动 1 个文件，共 8 行新增，4 行删除：

- `internal/proc/hide_windows.go`：`CreationFlags` 追加 `DETACHED_PROCESS`（`0x00000008`），确保子进程及其后代不继承/分配控制台

（`hide_windows.go` 有 `//go:build windows` 约束，`hide_other.go` 为空实现），不影响 macOS/Linux。不涉及任何逻辑行为变更，纯粹是进程窗口可见性控制。

### Steps to reproduce

1. Windows 11 系统通过 installer 安装 Reasonix 桌面版
2. 打开桌面端，点击右上角展开右侧 workspace 面板
3. 终端黑窗反复闪烁
4. 点击"改动"页签或"刷新"按钮，再次闪烁
5. 左侧历史面板切换会话，同样触发

### OS / platform

Microsoft Windows 11 家庭中文版 10.0.22621、Microsoft Windows 11 专业版 10.0.26100 均有复现

### Relevant logs or output

```shell

```

themarchrain:
Fixes the Windows desktop workspace panel flashing console windows when it runs
git status probes.

This keeps the shared `proc.HideWindow` helper on `CREATE_NO_WINDOW` only, so
shell tools such as PowerShell keep stdout/stderr capture working. The detached
process behavior is scoped to the desktop workspace git probes via a Windows-only
`proc.HideWindowDetached` helper, with a non-Windows no-op stub.

Also adds proc-level regression coverage that `HideWindow` still captures child
stdout.

Closes #2930

### DIFF
--- a/desktop/workspace_changes.go
+++ b/desktop/workspace_changes.go
@@ -103,14 +103,14 @@ func (a *App) WorkspaceChanges() WorkspaceChangesView {
 
 func workspaceGitStatus(base string) ([]gitStatusEntry, error) {
 	cmd := exec.Command("git", "-C", base, "status", "--porcelain=v1", "-z", "--untracked-files=all")
-	proc.HideWindow(cmd)
+	proc.HideWindowDetached(cmd)
 	raw, err := cmd.Output()
 	if err != nil {
 		return nil, err
 	}
 	entries := parseGitStatusPorcelainZ(raw)
 	topCmd := exec.Command("git", "-C", base, "rev-parse", "--show-toplevel")
-	proc.HideWindow(topCmd)
+	proc.HideWindowDetached(topCmd)
 	topRaw, err := topCmd.Output()
 	if err != nil {
 		return nil, err
--- a/internal/proc/hide_other.go
+++ b/internal/proc/hide_other.go
@@ -6,3 +6,6 @@ import "os/exec"
 
 // HideWindow is a no-op off Windows.
 func HideWindow(*exec.Cmd) {}
+
+// HideWindowDetached is a no-op off Windows.
+func HideWindowDetached(*exec.Cmd) {}
--- a/internal/proc/hide_windows.go
+++ b/internal/proc/hide_windows.go
@@ -7,7 +7,10 @@ import (
 	"syscall"
 )
 
-const createNoWindow = 0x08000000 // CREATE_NO_WINDOW
+const (
+	createNoWindow  = 0x08000000 // CREATE_NO_WINDOW
+	detachedProcess = 0x00000008 // DETACHED_PROCESS
+)
 
 // HideWindow stops a child process from flashing a console window on Windows,
 // where a GUI parent (the desktop app) has no console of its own to inherit.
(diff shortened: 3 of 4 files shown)


esengine:
Thanks for tackling this — the flashing console windows from `git.exe` grandchildren on the desktop build are a real annoyance (#2930), and the cross-platform guarding here is done right: `hide_windows.go` is behind `//go:build windows` with a `!windows` no-op stub, so Linux/macOS builds stay clean.

There's one blocker before I can merge, though. Adding `DETACHED_PROCESS` (`0x00000008`) alongside `CREATE_NO_WINDOW` regresses child-process **stdout capture**. The bash tool routes every command through `setKillTree` → `proc.HideWindow`, so the new flag lands on every command the agent runs, and the captured output comes back empty:

```
--- FAIL: TestBashPowerShellRunsNativeCommand
    bash_powershell_test.go:41: output = "", want it to contain reasonix-ok
--- FAIL: TestBashPowerShellSurfacesNonZeroExit
--- FAIL: TestBashPowerShellAllowsQuotedOperator
--- FAIL: TestBashPowerShellOutputIsUTF8
    bash_powershell_test.go:100: non-ASCII output mojibake — got "" (want it to contain 中文)
```

Root cause: `DETACHED_PROCESS` and `CREATE_NO_WINDOW` are conflicting console-disposition flags. `DETACHED_PROCESS` fully detaches the child from the console subsystem, which breaks the stdout/stderr pipe the tool relies on. `CREATE_NO_WINDOW` on its own already runs the child with no visible window — so dropping `DETACHED_PROCESS` and keeping just `CREATE_NO_WINDOW` (+ `HideWindow`) should suppress the flash without severing output.

Could you drop the `DETACHED_PROCESS` bit and confirm `go test ./internal/tool/builtin` stays green on Windows? Would also be good to add a proc-level test asserting child stdout is still captured with the flag set, since the `internal/proc` suite only checks the flag bits and missed this. Happy to merge once output capture is preserved.

themarchrain:
Updated.

`proc.HideWindow` no longer sets `DETACHED_PROCESS`; it stays on `CREATE_NO_WINDOW`
so shell tools preserve stdout/stderr capture. I added a proc-level regression
test for child stdout capture and confirmed `go test -count=1 ./internal/tool/builtin`
passes on Windows.

The detached behavior is now scoped only to the desktop workspace git probes that
reproduce #2930, with a non-Windows no-op stub. I also kept `CREATE_NO_WINDOW` and
`DETACHED_PROCESS` out of the same shared helper path, matching the Windows process
creation flag semantics:
https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags

PR #2948, opened 2026-06-03
feat(desktop): add batch delete for history panel sessions

HUQIANTAO:
## Summary

Add a selection mode to the HistoryPanel that lets users select multiple sessions and delete them in one action. The active session is never selectable.

### Backend (`desktop/app.go`)
- Add `DeleteSessions(paths []string) (int, error)` method that batch-deletes sessions, skips the active one, returns count of deleted sessions

### Frontend
- **HistoryPanel**: Selection mode with checkboxes, select-all toggle, batch delete button with two-step confirmation
- **Bridge**: `DeleteSessions` binding + browser mock
- **Controller**: `deleteSessions` callback
- **i18n**: 6 new keys in en.ts and zh.ts
- **CSS**: Styles for checkboxes, selected highlight, footer action bar, button variants

## UX

1. Click **Select** in the drawer header → checkboxes appear on each row
2. Toggle individual sessions or **Select all**
3. Click **Delete selected (N)** → confirmation prompt → batch delete
4. Click **Done** to exit selection mode

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -447,6 +447,31 @@ func (a *App) DeleteSession(path string) error {
 	return removeSessionArtifacts(dir, sessionPath, key)
 }
 
+// DeleteSessions removes multiple saved sessions in one call. The active session
+// is silently skipped (same guard as DeleteSession). Returns the number of
+// sessions actually deleted so the frontend can show feedback.
+func (a *App) DeleteSessions(paths []string) (int, error) {
+	a.mu.RLock()
+	ctrl := a.ctrl
+	a.mu.RUnlock()
+	dir := config.SessionDir()
+	active := ""
+	if ctrl != nil {
+		active = ctrl.SessionPath()
+	}
+	deleted := 0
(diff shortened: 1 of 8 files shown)


esengine:
Thanks for this — the batch-delete feature looks solid (backend DeleteSessions reusing the active-session guard, plus the full frontend selection/confirm flow). Rebasing against the latest main-v2 hits conflicts in 6 files (App.tsx, HistoryPanel.tsx, bridge.ts, useController.ts, locales) because several desktop PRs have since landed in those same files. Could you rebase onto the latest main-v2? Happy to merge once it is green. Thank you!

esengine:
For the rebase round, one design update is required, not just conflict resolution: this predates the project-tabs work (#2925), so it guards `a.ctrl`'s single active session — the app now holds a tab map, and batch delete must skip the active session of **every open tab**, not one. Also worth surfacing which paths failed when the fail-fast path returns a partial count. The frontend selection mode and the bridge contract sync both looked good.

PR #2955, opened 2026-06-03
perf(agent): add LRU cache for read-only tool outputs

HUQIANTAO:
64-entry LRU cache for read-only tool results (read_file, grep, ls, glob). Same tool + same args → cached result, reducing duplicate context tokens.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -209,6 +209,12 @@ type Agent struct {
 	// (a different failure shape, or any success). See applyStormBreaker.
 	stormSig   string
 	stormCount int
+
+	// outCache caches read-only tool results (read_file, grep, ls, glob) so
+	// repeated calls with identical arguments return instantly. This is common
+	// when the model re-reads a file it already examined or re-runs a grep
+	// after a compaction. The cache is LRU-bounded and concurrent-safe.
+	outCache *toolCache
 }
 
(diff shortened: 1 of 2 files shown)


esengine:
Thanks for this — cutting repeated read cost is a good instinct, and a lot of the batch you've sent has been solid.

I can't take this one as it stands, because the cache has no invalidation. Once a `read_file`/`grep`/`ls`/`glob` result is cached by content hash, a later `edit_file`/`write`/`apply_patch` to the same path doesn't evict it, so subsequent reads return the pre-edit content for the rest of the session and the model goes on to act on stale data. That's a silent-wrong-output bug, which our guidelines explicitly rule out (log + crash over silently masking). `bashOutput`/`waitJob` are also on the read-only list, but they return time-varying background-job output, so caching them is wrong by construction.

There's also overlap with the cache-first machinery we already run at the prompt level (immutable-prefix / StormBreaker), so an in-process result cache is solving something we mostly address elsewhere.

If you want another run at it, the shape that could land is: restrict the cache to the genuinely immutable readers (`read_file`/`ls`/`glob`/`grep`, never `bashOutput`/`waitJob`) and invalidate by path on every write/edit/patch. That's a real redesign rather than a tweak, so I'll close this for now — happy to look at a fresh PR along those lines.

PR #2958, opened 2026-06-03
perf(providers): pool request body buffers via sync.Pool

HUQIANTAO:
Reuse byte buffers across API calls via sync.Pool, reducing GC pressure from repeated alloc/free of ~10-100KB buffers per turn.


esengine:
C:/Program Files/Git/e2e

esengine:
Running the e2e suite to confirm this does not regress cache-hit % (watching the regression in #3091 before we merge any cache-sensitive change).

/e2e

PR #2968, opened 2026-06-03
fix(cli): measure approval-banner subject by display width, not rune count

Tatlatat:
## What

`truncateSubject` trims a tool subject so the approval banner stays one line.
It sliced by **rune count**:

```go
r := []rune(s)
if len(r) > max {
    return string(r[:max]) + "…"
}
```

Wide CJK characters are 2 display columns each, so a subject of `max` CJK runes
renders ~`2*max` columns — overflowing the banner and wrapping to a second row.
That strands the input-box borders in scrollback, which is the exact failure the
neighbouring `clampStatusLine` documents and avoids with `ansi.Truncate`.

## Repro (measured)

For a 44-char CJK subject at banner width 60 (max ≈ 32 cols):

| | visible width |
|---|---|
| before | **65** (overflows, wraps) |
| after  | **31** (fits) |

ASCII is unaffected (32 vs 33 — same one-line result).

## Fix

Replace the rune-slicing body with `ansi.Truncate(s, max, "…")` — ANSI-aware,
counts display columns, and already used by `clampStatusLine` plus four other
call sites in this package. The `max := width - 28` floor-at-16 logic is kept.

Add `TestTruncateSubject` asserting the visible width stays within bounds for
ASCII and CJK at both the normal and floor widths. (Verified the test fails on
the old implementation and passes on the new one.)

## Verification

`go test ./internal/cli/`, `go build ./...`, `go vet ./internal/cli/`, and
`gofmt -l` all clean.

### DIFF
--- a/internal/cli/chat_tui.go
+++ b/internal/cli/chat_tui.go
@@ -1909,11 +1909,7 @@ func truncateSubject(s string, width int) string {
 	if max < 16 {
 		max = 16
 	}
-	r := []rune(s)
-	if len(r) > max {
-		return string(r[:max]) + "…"
-	}
-	return s
+	return ansi.Truncate(s, max, "…")
 }
 
 // clampStatusLine truncates a status line to `width` visible columns, ANSI-aware,
--- a/internal/cli/chat_tui_test.go
+++ b/internal/cli/chat_tui_test.go
@@ -10,6 +10,7 @@ import (
 
 	tea "charm.land/bubbletea/v2"
 	"github.com/charmbracelet/x/ansi"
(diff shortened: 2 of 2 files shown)


esengine:
The fix itself is correct (uses the existing `ansi.Truncate` — no hand-rolled width). The test was asserting width with `go-runewidth` while the code uses `x/ansi`, which disagree on the `…` ellipsis by one column and red-barred CI. Rebased onto current main-v2 and switched the assertion to `ansi.StringWidth` — continuing in the rebased PR with your authorship preserved. Thanks @Tatlatat! Closing in favor of it.

PR #2974, opened 2026-06-03
fix(lru): actually evict stale entries from TtlLruCache on lookup

Tatlatat:
## What

`TtlLruCache` is documented as *"a stale hit is treated as a miss **and
evicted**"*, but `get()` didn't evict:

```ts
get(key: K): V | undefined {
  const e = this.inner.get(key);            // already promotes the entry to MRU
  if (!e) return undefined;
  if (e.expiresAt <= Date.now()) return undefined;  // miss — but entry stays
  return e.v;
}
```

Two problems for an expired key:
1. It is **never removed** — it lingers in the underlying map.
2. The preceding `this.inner.get(key)` has already re-inserted it as
   most-recently-used, so the stale entry *outlives* fresher ones during
   eviction.

`TtlLruCache` backs the gitignore cache and the directory-listing cache (both
5s TTL), which are hit on hot file-operation paths — so stale entries
accumulate and skew eviction under load.

## Fix

- Add a public `delete(key)` to `LruCache`.
- In `TtlLruCache.get`, call `this.inner.delete(key)` when the entry is expired
  before returning `undefined`, so it is evicted immediately as documented.

## Test

Appended a `TtlLruCache` case (fake timers): after the TTL elapses, a lookup
returns `undefined`, the key is gone from the underlying store, and it no longer
survives subsequent evictions. Verified it fails on the old code and passes on
the fix.

## Verification

`npx vitest run tests/lru.test.ts` (11 passed); `biome check` clean on the
changed files.

### DIFF
--- a/src/core/lru.ts
+++ b/src/core/lru.ts
@@ -28,6 +28,10 @@ export class LruCache<K, V> {
   clear(): void {
     this.map.clear();
   }
+
+  delete(key: K): void {
+    this.map.delete(key);
+  }
 }
 
 /** LruCache with a per-entry TTL — a stale hit is treated as a miss and evicted. */
@@ -43,7 +47,10 @@ export class TtlLruCache<K, V> {
   get(key: K): V | undefined {
     const e = this.inner.get(key);
     if (!e) return undefined;
-    if (e.expiresAt <= Date.now()) return undefined;
+    if (e.expiresAt <= Date.now()) {
+      this.inner.delete(key);
+      return undefined;
+    }
     return e.v;
   }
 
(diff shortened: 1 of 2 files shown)


Bernardxu123:
🙏 **真诚感谢您的贡献！**

感谢您为 DeepSeek-Reasonix v1 分支提交的修复 PR。这些改进体现了您对项目的认真态度和专业技术能力。

**关于 v1 分支的说明：**

由于 v1 已不再接受代码变更，此 PR 将被关闭。

2. 基于 `main-v2` 分支创建新的 PR

再次感谢您的时间和精力！您的贡献对开源社区非常有价值。❤️

---

>
> Thank you for submitting these fix PRs to the DeepSeek-Reasonix v1 branch. Your work demonstrates professional expertise and dedication.
>
>
>
>
> Thank you again! ❤️

PR #2978, opened 2026-06-03
fix(glob): match the pattern relative to the path base dir, not the project root

Tatlatat:
## What

The `glob` tool documents its `path` argument as *"Base directory to walk … The
pattern matches relative to this path."* But `globFiles` walked from `startAbs`
while testing the pattern against the path relative to `ctx.rootDir` (project
root).

So `glob(path: "src", pattern: "*.ts")` compared `"src/index.ts"` against
`"*.ts"`, matched nothing, and returned `(no matches)` — even though `index.ts`
matches `*.ts` relative to `src`.

## Fix

Match against a path relative to `startAbs` (the walk root), while keeping the
**displayed** result relative to `rootDir` so output still shows `src/index.ts`:

```ts
const rel = displayRel(ctx.rootDir, full);      // for display (unchanged)
const matchRel = displayRel(startAbs, full);     // for matching (new)
if (!isMatch(matchRel)) continue;
```

When no `path` is given (`startAbs == rootDir`) behavior is unchanged.

## Test

Added a case: globbing `*.ts` under `path: "src"` finds `src/index.ts` and does
not return `(no matches)`. Verified it fails on the old code and passes on the
fix (123 filesystem-tool tests pass).

## Verification

`npx vitest run tests/filesystem-tools.test.ts` (123 passed); `biome check`
clean on the changed files.

### DIFF
--- a/src/tools/fs/glob.ts
+++ b/src/tools/fs/glob.ts
@@ -51,7 +51,8 @@ export async function globFiles(
       }
       if (!e.isFile() && !e.isSymbolicLink()) continue;
       const rel = displayRel(ctx.rootDir, full);
-      if (!isMatch(rel)) continue;
+      const matchRel = displayRel(startAbs, full);
+      if (!isMatch(matchRel)) continue;
       let mtimeMs = 0;
       if (sortBy === "mtime") {
         try {
--- a/tests/filesystem-tools.test.ts
+++ b/tests/filesystem-tools.test.ts
@@ -1315,6 +1315,27 @@ describe("filesystem tools (built-in, sandbox-enforced)", () => {
       expect(out).toContain("src/index.ts");
       expect(out).not.toContain("(no matches)");
     });
+
+    it("matches relative to an absolute path argument", async () => {
+      const absDir = join(root, "absdir");
+      await fs.mkdir(absDir, { recursive: true });
+      await fs.writeFile(join(absDir, "testfile.ts"), "export const a = 1;");
+      const out = await tools.dispatch("glob", JSON.stringify({ path: absDir, pattern: "*.ts" }));
+      expect(out).toContain("absdir/testfile.ts");
+      expect(out).not.toContain("(no matches)");
+    });
+
+    it("matches relative to dot path '.' and does not recurse unless requested", async () => {
+      await fs.writeFile(join(root, "top.ts"), "x");
+      await fs.mkdir(join(root, "sub"), { recursive: true });
+      await fs.writeFile(join(root, "sub", "a.ts"), "y");
+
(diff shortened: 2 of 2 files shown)


HUQIANTAO:
Thanks for the thorough investigation and clean fix — the diagnosis is spot-on for the TypeScript implementation.

Worth noting: the main codebase has since been rewritten in Go, and the Go `glob` tool (`internal/tool/builtin/glob.go`) doesn't have this bug. Its `matchGlobSuffix` already matches against the path relative to the walk root (not the project root), and it has no `path` parameter — the working directory is set at construction time via `resolveIn(workDir, pattern)`.

A few suggestions:

1. **Scope**: Since `v1` is the legacy TypeScript branch and the active development is on `main-v2` (Go), this fix only benefits users still on v1. If v1 is still shipping, it's worth merging; if it's been superseded, you might want to check with the maintainers whether v1 is still maintained before investing more time here.

2. **Edge case**: You may want to add a test for when `path` is an absolute path (e.g. `/tmp/test`) — `displayRel` with an absolute `startAbs` should still produce correct relative paths for matching. Also worth testing `path: "."` (current directory) to confirm it doesn't break.

Nice catch overall. The single-line fix is exactly right.

Tatlatat:
Thanks for the careful review! Added both edge-case tests in the same describe block:

- **Absolute `path`**: globs with an absolute directory arg and asserts the `.ts` file resolves (and not `(no matches)`). Confirmed it fails on the pre-fix code and passes after — so it exercises the fix, not just the happy path.
- **`path: "."`**: asserts a top-level `*.ts` is found while a nested `sub/a.ts` is *not* pulled in by the non-recursive pattern, guarding against a `.`-base regression.

`npx vitest run tests/filesystem-tools.test.ts` → 125 passed; `biome check` clean.

On scope: agreed this is v1-only — the Go rewrite's `glob` already does the right thing (no `path` param, matches relative to the walk root), so nothing to port there. I targeted the `v1` branch since v0.x is still the npm `latest` line and `v1` is labeled the maintenance branch; happy to close this if v1 is no longer shipping fixes — your call.

Bernardxu123:
🙏 **真诚感谢您的贡献！**

感谢您为 DeepSeek-Reasonix v1 分支提交的修复 PR。这些改进体现了您对项目的认真态度和专业技术能力。

**关于 v1 分支的说明：**

由于 v1 已不再接受代码变更，此 PR 将被关闭。

2. 基于 `main-v2` 分支创建新的 PR

再次感谢您的时间和精力！您的贡献对开源社区非常有价值。❤️

---

>
> Thank you for submitting these fix PRs to the DeepSeek-Reasonix v1 branch. Your work demonstrates professional expertise and dedication.
>
>
>
>
> Thank you again! ❤️

PR #3027, opened 2026-06-04
netclient: remove Client.Timeout from NewHTTPClient, delegate to context

### LINKED ISSUE #3028: netclient: 移除 NewHTTPClient 的 timeout 参数，截止时间由调用方通过 context 控制
关联 PR #3027

AtmoOmen:
http.Client.Timeout is an end-to-end deadline that includes reading the entire response body — a blunt instrument unsuitable for both large downloads and long-lived streaming requests. Remove both the field and the parameter from the shared factory; let each call site express its deadline via context.Context.

### Changes

| File | Change |
|------|--------|
| `internal/netclient/netclient.go` | Removed `timeout time.Duration` parameter and `Client.Timeout` field |
| `desktop/updater.go` | Dropped `httpTimeout` argument from call |
| `desktop/updater_app.go` | Wrapped manifest fetches with `context.WithTimeout(ctx, 15s)`; download/signature left unbounded |
| `internal/control/controller.go` | Wrapped Balance query with `context.WithTimeout(ctx, 12s)` |
| `internal/boot/boot.go` | Dropped timeout arguments from 2 call sites, removed unused `"time"` import |
| `internal/cli/codegraph.go` | Dropped `0` argument |
| `internal/provider/openai/openai.go` | Dropped `0` argument |
| `internal/provider/anthropic/anthropic.go` | Dropped `0` argument |

### Why this is the right fix

- **LLM providers are untouched in behaviour** — they already pass `0` (no Client.Timeout) and rely on context for lifecycle. Dropping the argument changes nothing.
- **The updater bug is fixed** — large installer downloads are no longer cut off at 15s.
- **Deadlines are explicit at the call site** — each caller expresses its own time budget, rather than inheriting a one-size-fits-all value from the shared factory.

Closes #3028

### DIFF
--- a/desktop/updater.go
+++ b/desktop/updater.go
@@ -65,7 +65,7 @@ func httpClient() (*http.Client, error) {
 	if err != nil {
 		return nil, err
 	}
-	return netclient.NewHTTPClient(cfg.NetworkProxySpec(), httpTimeout, netclient.TransportOptions{})
+	return netclient.NewHTTPClient(cfg.NetworkProxySpec(), netclient.TransportOptions{})
 }
 
 // canSelfUpdate reports whether in-place update is possible. macOS is excluded:
--- a/desktop/updater_app.go
+++ b/desktop/updater_app.go
@@ -34,7 +34,9 @@ func (a *App) CheckUpdate() (*UpdateInfo, error) {
 			Err:           err.Error(),
 		}, nil
 	}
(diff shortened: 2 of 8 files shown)


esengine:
C:/Program Files/Git/e2e

esengine:
Running the e2e suite to confirm this does not regress cache-hit % (watching the regression in #3091 before we merge any cache-sensitive change).

/e2e

PR #3030, opened 2026-06-04
feat(desktop): add close-to-tray option

### LINKED ISSUE #3014: [Feature]: Add close-to-tray option for desktop window close
﻿### Version line

The desktop app currently exits when the main window is closed. For users who keep Reasonix running while switching sessions, checking settings, or waiting for background work, closing the window should be able to hide the app instead of terminating it.

This is related to, but distinct from, #2573 / #2774. That work focuses on auto-start and silent start with a tray icon. This request is specifically about the normal window close behavior after the app is already running.

### Proposed solution

Add a desktop setting such as **Close window to tray** / **关闭窗口时最小化到托盘**.

Expected behavior:

- When the option is enabled, clicking the window close button hides/minimizes the app to the system tray instead of quitting.
- The tray icon should provide at least **Show Reasonix** and **Quit** actions.
- **Quit** must really terminate the app and bypass the close-to-tray behavior.
- Keep the option disabled by default if there is concern that users may think the app cannot be closed.
- Scope this to the desktop app only; it should not affect CLI behavior.

wade19990814-hue:
﻿## Summary
- add a desktop-only `closeToTray` setting stored in `UserConfigDir/reasonix/desktop.json`
- initialize a cross-platform system tray with `Open` / `Quit` actions only
- intercept window close only when the setting is enabled and a tray is available
- add the setting under Settings → Appearance → Desktop behavior

Closes #3014

## Notes
- The setting defaults to off, so existing close-to-exit behavior is unchanged unless users opt in.
- The tray menu intentionally does not expose the setting toggle; it is managed from the desktop Settings panel.
- Windows uses the bundled `.ico` for the tray icon, while macOS/Linux keep using the PNG app icon.
- Linux checks for a StatusNotifier watcher before enabling close-to-tray, so the app should not hide itself when no tray host is available.
- Windows manual verification covered the Settings entry, the tray icon, and the localized tray menu (`打开` / `退出`). Screenshots were captured locally by the tester.

## ScreenShot
屏幕截图 2026-06-05 152222
屏幕截图 2026-06-05 152208
屏幕截图 2026-06-05 152151

## Test plan
- `cd desktop && go test -count=1 ./...`
- `cd desktop/frontend && pnpm exec tsc --noEmit`
- `cd desktop && wails build`


wade19990814-hue:
Hi @SivanColal, could you take a look when you have a moment?

I rebased this PR onto the latest `main-v2` and added fresh real-machine desktop testing on Windows. Verified the close-to-tray flow manually: default close still quits, enabling the setting hides the window to tray, tray `Open` restores the app, `Quit` exits, and the setting persists after restart.

PR #3056, opened 2026-06-04
fix: show sent desktop messages immediately

YoungDan-hero:
Desktop chat waited for the backend's first real event before adding the user's submitted message to the transcript. This made short prompts look delayed whenever the bridge, provider, or model had startup latency.

Show the submitted message right away with a pending state so the UI responds immediately. Confirm that message when the first backend event arrives, remove it if the user cancels before confirmation, and mark it failed only when submission fails before the backend has accepted the turn.

Add reducer coverage for immediate display, confirmation without duplicates, cancel-before-confirmation, early submit failure, and late submit rejection.

### DIFF
--- a/desktop/frontend/package.json
+++ b/desktop/frontend/package.json
@@ -7,6 +7,7 @@
     "dev": "vite",
     "build": "tsc --noEmit && vite build",
     "preview": "vite preview",
+    "test": "vitest run",
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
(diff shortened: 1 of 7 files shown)


esengine (approved):

esengine:
Thanks for this — the change reviewed cleanly and CI was green. But #3603 just merged into main-v2 and touches the same files (`useController.ts`, `package.json`, and the lockfile), so this PR now conflicts with main-v2. Could you rebase onto the latest main-v2 and resolve the conflicts? It should be ready to merge right after. 🙏

PR #3057, opened 2026-06-04
feat(agent): cache prefix diagnostics for prompt-cache churn

esengine:
Rebase of #2405 (by @SivanCola) onto current main-v2. The branch had drifted badly; rather than replay four commits through the intermediate-state conflicts, I applied the net diff and re-merged by hand against the newer agent code.

## What it does
Each turn, hashes the prompt prefix (system / tools / combined) and, on a cache miss, attributes it to one of: system-prompt drift, tool-schema drift, a log rewrite (compaction), or an inherent tail miss. The reason is surfaced on the CLI/TUI usage line and forwarded through the serve and desktop wire so every frontend can show *why* the cache churned — directly in service of the cache-first positioning.

## Rebase notes (where main-v2 had moved on)
- Kept main-v2's mutex-safe `Session.Replace` / `Snapshot` and the `atomic.Int64` session cache counters, and threaded the new `rewriteVersion` / `RewriteVersion()` / `IncrementRewrite()` alongside them rather than reverting to the old direct-field writes.
- Left `stream()` on main-v2's signature (it still needs `turn` for the PostLLMCall hook and returns the reasoning `signature`); the prefix shape is captured from `a.tools.Schemas()` in the loop, which is what `stream` sends.
- Adapted `TestCompactRewriteVersionFeedsCacheDiagnostics` to the current 3-arg `compact(ctx, trigger, instructions)` signature.

All four review items from the original #2405 thread are already folded in (the `log_rewrite` wiring, the usage-line consumer, the `Rewite`→`Rewrite` typo, and the dead-field cleanup).

Local: `go build ./...`, `go vet`, and `go test ./internal/agent ./internal/serve ./internal/event` all green.

Closes #2405

### DIFF
--- a/desktop/frontend/src/lib/types.ts
+++ b/desktop/frontend/src/lib/types.ts
@@ -45,11 +45,24 @@ export interface WireUsage {
   cacheHitTokens: number;
   cacheMissTokens: number;
   reasoningTokens?: number;
+  cacheDiagnostics?: WireCacheDiagnostics;
   // Session-cumulative cache tokens — the status bar shows the aggregate
   // hit-rate (Σhit/Σ(hit+miss)), steadier than the single-turn cacheHitTokens.
   sessionCacheHitTokens: number;
   sessionCacheMissTokens: number;
   costUsd?: number;
+}
+
+export interface WireCacheDiagnostics {
+  prefixHash: string;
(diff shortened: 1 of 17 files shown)


esengine:
C:/Program Files/Git/e2e

PR #3073, opened 2026-06-04
feat(agent): structured, economical compaction (soft/force ratios + fold economics)

esengine:
Rebase of #2406 (by @SivanCola) onto current main-v2 (after #3050/#3057 landed). The branch had drifted; I applied the net diff and re-merged the compaction core by hand, since `compact()` had diverged on both sides.

## What it does
- **Soft ratio (0.5)** — between the soft mark and the trigger, emit a one-shot notice instead of compacting, so the cache-stable prefix is kept intact while the user is warned.
- **Force ratio (0.9)** — a high-water mark that compacts even a low-value region.
- **Fold economics** — skip the summarizer round-trip when the foldable region is too small to pay for the extra latency/cost (bypassed by `force` and by manual `/compact`).

## Rebase notes (signature reconciliation)
- main-v2 had `compact(ctx, trigger, instructions)`; this PR had `compact(ctx, force bool)`. Unified to **`compact(ctx, trigger, instructions, force)`** so the force flag coexists with the existing `/compact <focus>` text + PreCompact-hook path and the cache-diagnostics rewrite tracking from #3057.
- Kept main-v2's **token-budgeted tail** (`planCompaction`, parameterized with a min-message count for the single-large-message fallback) rather than reverting to the PR's message-count tail, and kept main-v2's more detailed summary prompt.
- `maybeCompact` now layers the soft-notice and force logic on top of main-v2's anti-re-compaction `compactStuck` latch.
- Adapted the threshold tests to the token-budgeted tail (a single large old message folds 1→1: the summary is installed but the message count is unchanged).

Local: `go build ./...`, `go vet`, and `go test ./internal/agent ./internal/config ./internal/boot ./internal/cli ./internal/control` all green.

Closes #2406

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -195,9 +195,13 @@ type Agent struct {
 	// tail verbatim (recentKeep is the message floor) and archiving the originals
 	// under archiveDir. compactStuck latches when compaction can't get the prompt
 	// under the window (consecutiveCompacts crosses the limit), so auto-compaction
-	// pauses instead of looping.
+	// pauses instead of looping. softCompactNoticed gates the one-shot soft-ratio
+	// notice so it fires once per approach, not every turn.
 	contextWindow       int
+	softCompactRatio    float64
 	compactRatio        float64
+	compactForceRatio   float64
+	softCompactNoticed  bool
 	recentKeep          int
 	archiveDir          string
 	compactStuck        bool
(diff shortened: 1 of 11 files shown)


esengine:
C:/Program Files/Git/e2e

esengine:
C:/Program Files/Git/e2e

PR #3083, opened 2026-06-04
fix(styles): adjust workspace tab flex properties for better layout

YoungDan-hero:
Updated the flex properties of the workspace tabs to improve layout consistency. The flex-grow value was changed to 0% to prevent tabs from expanding, and a new rule was added to prevent flex items from shrinking. This enhances the overall appearance and functionality of the workspace tabs.

### DIFF
--- a/desktop/frontend/src/components/WorkspacePanel.tsx
+++ b/desktop/frontend/src/components/WorkspacePanel.tsx
@@ -624,12 +624,12 @@ export function WorkspacePanel({
                 </button>
               </Tooltip>
             ))}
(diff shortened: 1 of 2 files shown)


esengine:
Thanks! This was a real layout fix at the time. The workspace tab strip has since been replaced by the new `TabBar` component, so the `.workspace-tabs` flex rules this touched are gone from main-v2. Closing as superseded by that rework — if the new tab strip shows the same squeeze, a fresh PR against `TabBar` would be very welcome 🙏

PR #3093, opened 2026-06-04
fix(e2ebench): surface JSON marshal + write errors instead of silently dropping them

HUQIANTAO:
## Summary

The bench's `-json` write path uses `json.MarshalIndent` + `os.WriteFile` and discards both errors. A marshal failure or a write failure is silently swallowed: the bench exits 0, the markdown report goes out, and the JSON file is either missing or contains `null`.

## Fix

Surface both errors on stderr and exit 1, matching the markdown-report write path.

## Test plan

* [x] `go build ./…` passes.

### DIFF
--- a/cmd/e2ebench/main.go
+++ b/cmd/e2ebench/main.go
@@ -102,8 +102,15 @@ func main() {
 		fmt.Print(report)
 	}
 	if *outJSON != "" {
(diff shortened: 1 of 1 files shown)


esengine:
Duplicate of #3092 — byte-identical diff to cmd/e2ebench/main.go. Closing this one; tracking the fix in #3092. Thanks!

Issue #3091, opened 2026-06-04
[Bug]: 升级后缓存命中率从之前99%降到78%，字 小得看不清，比雷布斯还小。不针对deepseek优化估计慢慢就没人用了。


qiyuanlicn:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.1.0

### What happened?

升级后缓存命中率从之前99%降到78%，字 小得看不清，比雷布斯还小。不针对deepseek优化估计慢慢就没人用了。

### Steps to reproduce

0

### OS / platform

windows

### Relevant logs or output

```shell

```

esengine:
Thanks for the report — and sorry the 1.1.0 upgrade landed worse for you on all three fronts. Let me take them one at a time, since they're really separate issues bundled into one.

### 1. Cache hit rate 99% → 78% (this is the one I care about most)

A drop this size is not expected and not something I'd shrug off — cache-first is the whole point of this project, so a regression here is a real bug, not a tuning preference. To track it down I need a bit more from you:

- The output of the in-session cache stats (the per-turn `cache hit` line, or the summary at session end). Even a screenshot is fine.
- Roughly what your workflow looks like: long single sessions, lots of `/clear` or branching, frequent file edits between turns? Anything that rewrites the prefix of the prompt will cost you hits.
- Which model/provider endpoint you're on (official DeepSeek API vs. a proxy), since cache behavior is endpoint-side.

My first suspicion is a prefix-stability regression — something in 1.1.0 that perturbs the immutable prefix of the request (system block, tool list ordering, or per-turn metadata) and silently invalidates the upstream cache. If you can drop a `cache hit` log line from a 1.0.x run and a 1.1.0 run side by side, that alone will probably point straight at it.

### 2. Font "too small to read"

The TUI doesn't ship its own font — it renders into your terminal at the terminal's own font size, so the size itself is controlled by your terminal (e.g. Windows Terminal: `Ctrl` + `+` to zoom, or set the font size in the profile). That said, if there's a *specific* element that got smaller or dimmer in 1.1.0 (dimmed/secondary text, the status line, a particular panel), that would be a real styling regression and I'd like to fix it. Can you share a screenshot pointing at the text you mean?

### 3. "Not optimizing for DeepSeek"

[...]

PR #3100, opened 2026-06-04
chore(e2ebench): v1.0.0 cache baseline for #3091 (do not merge)

esengine:
**Do not merge — investigation only.** The tree here is the `v1.0.0` release (an empty marker commit sits on top only so a PR can be opened; it changes nothing). The e2e bot builds reasonix from PR head, so this runs the **v1.0.0 agent** against the current suite, giving the "before upgrade" cache-hit numbers to quantify the 99%→78% regression reported in #3091.

Compare the per-task **`compaction`** cache hit against the main-v2 baseline (68%): if v1.0.0 is markedly higher, the regression lives in a compaction change between v1.0.0 and v1.1.0. (The huge diff and any red non-e2e CI are expected — only the `/e2e` report matters; v1.0.0 lacks the `compactions` metric field, so that column reads 0.)


esengine:
v1.0.0 agent baseline run for the #3091 cache regression.

/e2e

esengine:
Done. v1.0.0 agent: 79% total / 56% on the compaction task (task failed) / normal turns 98-99% — statistically the same as the main-v2 baseline (79% total, normal turns 98-99%). No version-level cache regression reproduced between v1.0.0 and v1.1.0; the cache drop in #3091 tracks compaction firing (workload), not a code change. Closing the marker.

PR #3114, opened 2026-06-04
perf(desktop): coalesce streaming text/reasoning deltas per animation frame

HUQIANTAO:
The agent stream pushes a text/reasoning delta per Webview event; at 200 tok/s that is one state update every ~5ms, while rAF is 16ms. The reducer was running 3-4 times per visible frame and the React tree was re-rendering the entire Message list each time.

Add `lib/rafBatch.ts`: a tiny rAF-coalescing queue with a synchronous `drain()` so non-text events (tool_dispatch, usage, notice, turn_started/done, message) flush any pending deltas first to keep causal ordering intact. Drain is also called on unmount so a turn-end that races the teardown doesn't strand the last few tokens in the buffer.

`tsc --noEmit` passes; `go build ./desktop/...` is clean. Frontend test spec is included as a contract test to be enabled when vitest is added.

### DIFF
--- a/desktop/frontend/src/lib/rafBatch.ts
+++ b/desktop/frontend/src/lib/rafBatch.ts
@@ -1,0 +1,69 @@
+// rafBatch coalesces repeated calls into one execution per animation frame. The
+// desktop agent stream pushes a text/reasoning delta per Webview event; at 200
+// tok/s that's a state update every ~5 ms, so a 16 ms rAF window can absorb 3–4
+// deltas into a single React render. Non-text events (tool dispatch, usage,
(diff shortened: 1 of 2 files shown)


esengine:
Thanks! This conflicts with the latest main-v2 in useController.ts (other desktop PRs have since landed there). Could you rebase onto the latest main-v2? Note your open desktop PRs also touch App.tsx (#3116, #3126) and useController.ts (#2948) — rebasing them as one ordered chain will avoid repeated conflicts. Happy to merge once green. Thank you!

PR #3116, opened 2026-06-04
perf(desktop): code-split the four drawable panels via React.lazy

HUQIANTAO:
Memory, History, Settings, and the Update banner are opened via state toggles and are mutually exclusive in practice. They each pull in their own transitive deps. Convert each import to `React.lazy()` with a Suspense boundary. Each panel's JS chunk is fetched on first render (the first time the user clicks the corresponding topbar chip), so the initial bundle loses those four trees and any of their imports.

The fallback is null because the panels themselves are modal drawers with their own loading state. The `.then(m => ({ default: m.X }))` adapter handles modules whose named export is the public surface (per CodeViewer's 'editor seam' pattern).

### DIFF
--- a/desktop/frontend/src/App.tsx
+++ b/desktop/frontend/src/App.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import { Suspense, lazy, useCallback, useEffect, useMemo, useRef, useState } from "react";
 import type { CSSProperties, KeyboardEvent, PointerEvent as ReactPointerEvent } from "react";
 import {
   SquarePen,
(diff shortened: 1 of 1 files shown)


esengine:
Thanks! This conflicts with the latest main-v2 in App.tsx. Could you rebase onto the latest main-v2? Several of your desktop PRs touch App.tsx (#3126, #2948) and useController.ts (#3114, #2948) — rebasing them as one ordered chain avoids repeated conflicts. Happy to merge once green. Thank you!

esengine:
Nice perf win — code-splitting these panels is worth doing. Could you rebase on the latest `main-v2`? It drifted under this PR: `HistoryPanel` and `SettingsPanel` changed props (trash support — `kind`/`onRestore`/`onPurge`, and `initialTab`/`settingsTarget`), and a new `ContextPanel` landed. Re-applying the `React.lazy`/`Suspense` wrapping on top of the new component calls (and deciding whether to lazy-load `ContextPanel` too) is cleaner from your side than a hand-merge from mine. Once rebased I'll get it in.

esengine:
﻿Covering the whole series (#3116, #3120, #3125, #3126, #3127, #2948) in one note: all six branches now conflict with main-v2 after today's merge batch. They're each still wanted — the React.lazy code-split and the batch delete in particular. If you rebase them onto the latest main-v2 I'll approve fork CI and review as they come green. If any of them no longer applies after the recent status-bar/overview redesign (#3922 reworked that area — the per-turn usage pill in #3127 likely needs rethinking against the new grouped metrics), feel free to close those and keep the rest.

PR #3125, opened 2026-06-04
feat(desktop): keyboard shortcuts cheatsheet + a11y landmarks / focus

HUQIANTAO:
Three related improvements to make the desktop app reachable from the keyboard and readable by screen readers.

1. **Shortcuts cheatsheet (press ?)**. New component renders a kbd+description table, grouped by section. Triggered by a '?' keypress from anywhere outside an editable element, or from a Help icon in the topbar. role=dialog + aria-modal; focus moves to close button on open, returns to trigger on close.
2. **ARIA landmarks + labels**. App.tsx's <header>/<main>/<footer> get role=banner / role=main / role=contentinfo, every icon-only topbar button gets an explicit aria-label, the error banner is role=alert, the composer textarea gets id+aria-label so a skip-link can target it.
3. **Skip-to-composer link**. The first Tab from the topbar lands on a visually-hidden link that jumps focus to the textarea.
4. **Focus rings**. `:focus-visible` paints a 2px accent outline with a 2px offset.

### DIFF
--- a/desktop/frontend/src/components/ShortcutsCheatsheet.tsx
+++ b/desktop/frontend/src/components/ShortcutsCheatsheet.tsx
@@ -1,0 +1,149 @@
+import { useEffect, useRef } from "react";
+
+// ShortcutsCheatsheet is the "?" overlay — a single column of every
+// keyboard shortcut the desktop app honors. The list is intentionally
+// declarative (an array of [keys, description] tuples) so adding a
+// shortcut is a one-line change here. The component is mounted by App and
+// toggled by a "?" keypress on the topbar; it also opens via the menu in
+// the topbar's chip cluster.
+//
+// Focus is trapped while open (the cheatsheet is a small dialog), Esc
+// closes it, and focus returns to the element that opened it. We use a
+// role="dialog" + aria-modal="true" + aria-labelledby so screen readers
+// announce the title and don't expose the page behind.
+//
+// The cheatsheet is NOT localized (shortcut keys are universal); the
+// descriptions are. Both en.ts and zh.ts have their own table that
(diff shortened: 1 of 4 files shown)


esengine:
Thanks! This conflicts with the latest main-v2 in styles.css (other desktop PRs have since landed there). Could you rebase onto the latest main-v2? Happy to merge once green. Thank you!

esengine (changes requested):
Blocking finding: **the component is never mounted** — the diff adds ShortcutsCheatsheet.tsx, locales, and CSS, but no App.tsx wiring, so the "?" keybind and help entry don't exist and this ships as dead code. Three more for the same revision:

1. Keep the `DictKey` typing instead of `t: (id: string) => string` — the dynamic `"shortcuts.desc." + id` lookups silently render raw keys on any typo.
2. The shortcut table predates the keybinding unification: Shift+Tab now only toggles Plan, Ctrl+Y toggles YOLO. Sync the content.
3. Rebase (styles.css moved substantially).

SivanCola:
Thanks @HUQIANTAO for the shortcuts cheatsheet and accessibility work here.

I integrated this direction into #4515 on top of the latest `main-v2`. The new PR mounts the shortcuts sheet, keeps typed locale keys, renders the sheet from the shared shortcut registry so it stays in sync with custom bindings, and includes the skip-to-composer accessibility link.

Authorship note: @HUQIANTAO remains the primary author of the shortcuts help/a11y direction. @SivanCola contributed as a collaborator by integrating it with the current shortcut registry, folding in the related settings direction from #4202, and verifying the final state.

Closing this PR as superseded by #4515.

PR #3127, opened 2026-06-04
feat(desktop): per-turn usage pill in the status bar

HUQIANTAO:
The status bar showed session-aggregate cache-hit % and the running turn's wall-clock + token count, but the actual per-turn prompt/completion/cache numbers were only visible by digging into the WireUsage payload in devtools. For a cost-aware user the billable numbers deserve a permanent place.

Add `components/UsageIndicator.tsx`: a 3-cell pill that mounts in the status bar:

  ↑ 1.2k   prompt tokens this turn
  ↓  280   completion tokens this turn
  ↻  850   cache-hit tokens (only rendered when > 0)

The values are k-rounded to keep the bar compact. The cache cell uses --add-fg so the green cue matches the rest of the UI.

### DIFF
--- a/desktop/frontend/src/components/StatusBar.tsx
+++ b/desktop/frontend/src/components/StatusBar.tsx
@@ -3,6 +3,7 @@ import { Coins, Cpu, Wallet } from "lucide-react";
 import { EffortSwitcher } from "./EffortSwitcher";
 import { ModelSwitcher } from "./ModelSwitcher";
 import { Tooltip } from "./Tooltip";
+import { UsageIndicator } from "./UsageIndicator";
 import { SPINNER_WORDS, useI18n } from "../lib/i18n";
 import type { BalanceInfo, ContextInfo, EffortInfo, JobView, Meta, Mode, WireUsage } from "../lib/types";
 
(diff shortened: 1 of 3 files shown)


esengine (changes requested):
Found a real one while reviewing: the entire `.usage` rule block is spliced *inside* `.onboarding__skip:disabled { … }` and reuses its closing brace (the last added rule has no `}` of its own), so the selectors nest under `.onboarding__skip:disabled` and the pill ships unstyled. Re-place the block at top level. Two smaller things for the same push: trim the 15-line header essay in UsageIndicator.tsx per the repo comment policy, and rebase (StatusBar/styles moved under this PR in today's batch). The feature itself is on-brand — cost visibility is the pitch — so happy to merge once those are in.

SivanCola:
Thanks @HUQIANTAO. We are closing this PR as superseded by #4329, which is now the canonical integration route for the cost/currency/status-bar work.

Your contribution is recorded in #4329: you introduced the compact per-turn usage indicator concept for prompt, completion, and cache-hit tokens in the desktop status bar.

You are acknowledged in #4329 as a repository co-contributor for this work.

PR #3180, opened 2026-06-05
fix(cli): clarify TUI status rows and welcome key prompts

SivanCola:
## Summary

- Move CLI TUI git/worktree identity (`repo@branch (+N -N ?N)`) from the live metrics row to the mode/session row.
- Keep the second status row focused on run data: model, effort, context, compact headroom, cache, jobs, and balance.
- Fix zero-arg `reasonix` welcome flow so built-in defaults alone do not trigger missing API key prompts such as `MIMO_API_KEY`; missing-key prompts now require a real project/user config source.

## Screenshot

CLI TUI status layout after fix

## Test plan

- `go test ./internal/cli`


SivanCola:
Superseded by the split PRs: #3182 for the CLI TUI status-row layout, and #3183 for the zero-arg welcome/key-prompt fix.

PR #3183, opened 2026-06-05
fix(cli): require config source before key prompt

SivanCola:
## Summary

- Fix zero-arg `reasonix` welcome flow so built-in defaults alone do not trigger missing API key prompts such as `MIMO_API_KEY`.
- Only run the inline missing-key prompt path when a real project/user config source exists.
- Add a regression test for the `SourcePath == "" && cfgErr == nil` case.

## Test plan

- `go test ./internal/cli`

Split from superseded PR #3180.


SivanCola:
@codex review

PR #3202, opened 2026-06-05
Fix repeated successful writer loops

### LINKED ISSUE #3157: [Bug]: 重复执行相似命令，陷入死循环
### Version line

### Exact version

1.1.0 

### What happened?

昨天晚上下载了最新的1.1.0，跑的时候，有一行with open 用python打开文件的命令，后面还跟了很多命令，总的来说是想读取之后再写入，但是一直执行，同一时间执行了几十次，发现的时候已经消耗了不少token，后来给AI反馈，他说应该使用edit file的skill，不过他改了一段时间后，又变成了之前的问题，重复执行

### Steps to reproduce

1. 让reasonix修改文本内容，比如提示词

### OS / platform

windows 11

### Relevant logs or output

```shell

```

paradoxSCH:
﻿## Summary

Fixes #3157.

This adds a narrow repeat-success loop guard for write-like tool calls. When the model repeats the same successful file-writing operation more than twice in one user turn, the agent blocks the next identical call and nudges the model to switch approach, use `edit_file`/`multi_edit`, verify with a read/test command, or explain the blocker.

## Root cause

The existing storm guard only handled repeated failures. In the reported loop, each repeated `bash` write succeeded, so the agent treated every call as progress and kept feeding successful results back to the model.

## Changes

- Track identical successful write-like calls within a single user turn.
- Scope `bash` tracking to commands that appear to write files, leaving repeated read/test commands alone.
- Return an explicit `[loop guard]` tool result on the third repeated write-like success.
- Make `write_file` return a no-op signal when the target already contains the exact same content.
- Add agent-level and desktop binding-level regression coverage.

## Validation

- `go test ./internal/agent ./internal/tool/builtin -count=1`
- `go test . -run TestDesktopE2EBlocksRepeatedSuccessfulBashFileWrite -count=1 -v` from `desktop/`
- `go test . -count=1` from `desktop/`
- Copied the new desktop E2E test into a temporary `origin/main-v2` worktree and confirmed the old code fails with `bash executed 3 times, want 2 before the repeat guard blocks`.

## Notes

The desktop E2E uses a scripted provider instead of a real model so the regression is deterministic and does not require API keys, network stability, model-version behavior, or real file writes outside the fake tool.

### DIFF
--- a/desktop/repeat_guard_e2e_test.go
+++ b/desktop/repeat_guard_e2e_test.go
@@ -1,0 +1,84 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"strings"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"reasonix/internal/agent"
+	agenttest "reasonix/internal/agent/testutil"
+	"reasonix/internal/control"
+	"reasonix/internal/event"
+	"reasonix/internal/provider"
+	"reasonix/internal/tool"
+)
+
+type desktopCountingTool struct {
+	name     string
+	readOnly bool
+	calls    *int32
+}
+
+func (t desktopCountingTool) Name() string        { return t.name }
+func (t desktopCountingTool) Description() string { return "test tool" }
+func (t desktopCountingTool) Schema() json.RawMessage {
+	return json.RawMessage(`{"type":"object","properties":{"command":{"type":"string"}}}`)
+}
+func (t desktopCountingTool) ReadOnly() bool { return t.readOnly }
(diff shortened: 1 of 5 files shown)


esengine (approved):
成功-loop 守卫:镜像 stormBreaker,每轮重置,执行前拦+成功后计数,阈值2。bash 写入判定收窄且只对规范化全等命令计数,误伤极低;软提示非硬失败。agent 单测+桌面 e2e 覆盖。修 #3157,护住省钱定位。合并。

PR #3231, opened 2026-06-05
Fix workspace builtin tool assembly

### LINKED ISSUE #3218: [Bug]: workspace-bound tool registry drops session-level builtins
﻿## Problem

When Reasonix builds a tool registry for a workspace-bound session, some session-level built-in tools are dropped from the runtime registry even though the system prompt and other tool schemas still tell the model to use them.

This shows up as `unknown tool` during normal agent execution. In #3197 screenshots the missing tools are:

- `todo_write`
- `bash_output`
- `wait`

This is also related to #3165, where `todo_write` appears unavailable despite being referenced by the prompt.

## Root cause

`boot.addBuiltins` has a workspace branch:

```go
if workDir != "" {
    ws := builtin.Workspace{...}
    for _, t := range ws.Tools(enabled...) {
        reg.Add(t)
    }
    return
}
```

`Workspace.Tools` only returns workspace/path-bound tools such as `read_file`, `write_file`, `edit_file`, `bash`, `grep`, `glob`, `ls`, and `web_fetch`.

It does not include session-level builtins such as:

- `todo_write`
- `complete_step`
- `bash_output`
- `kill_shell`
- `wait`
- `notebook_edit`

As a result, workspace-bound CLI/desktop sessions can prompt the model to use tools that are not actually registered.

ACP has a parallel assembly path in `internal/cli/acp.go` that directly calls `Workspace.Tools(...)`, so it can drift from `boot.Build` and has the same missing-tool risk.

## Expected behavior

Workspace-bound sessions should keep the full built-in tool surface. Tools that need workspace binding should be replaced with workspace-aware implementations, while session-level builtins should remain registered.

## Suggested fix

- Share one builtin assembly path between `boot.Build` and ACP where possible.
- Ensure workspace-bound assembly preserves all builtins, including session-level tools.
- Add regression tests for workspace-root sessions verifying that the registry/tool schemas include at least:
  - `todo_write`
  - `complete_step`
  - `bash_output`
  - `kill_shell`
  - `wait`
  - `notebook_edit`

Related: #3197, #3165

paradoxSCH:
﻿## Summary

Fixes #3218.

Workspace-bound sessions now preserve the full built-in tool surface instead of only exposing path-bound workspace tools. This prevents normal CLI/desktop/ACP sessions from prompting the model toward tools that are missing from the runtime registry.

Related: #3197, #3165.

## Root cause

`boot.addBuiltins` used `builtin.Workspace.Tools(...)` whenever a workspace root was set. `Workspace.Tools` only returned workspace/path-bound tools such as `read_file`, `write_file`, `bash`, `grep`, and `glob`, so session-level builtins were dropped even though the prompt and tool schemas mention them.

The missing tools included:

- `todo_write`
- `complete_step`
- `bash_output`
- `kill_shell`
- `wait`
- `notebook_edit`

ACP also called `Workspace.Tools(...)` directly, so it had the same missing-tool risk and could drift from the shared boot assembly.

## Changes

- Make `Workspace.Tools` start from the complete builtin set and override only workspace-aware tools.
- Preserve explicit `tools.enabled` filtering while allowing session-level builtins through when enabled.
- Add boot-level coverage for workspace-root builtin assembly.
- Add ACP-level coverage for workspace builtin assembly.

## Validation

- `go test ./internal/tool/builtin -run "TestWorkspace(PreservesSessionLevelBuiltins|EnabledFilter|BindsReadAndWrite|WriteConfinement|PreviewBinds)" -count=1 -v`
- `go test ./internal/boot -run TestAddBuiltinsWithWorkspaceRootKeepsSessionTools -count=1 -v`
- `go test ./internal/cli -run TestACPBuiltinToolsKeepSessionLevelBuiltins -count=1 -v`
- `go test ./internal/tool/builtin ./internal/boot -count=1`
- `git diff --check`

Note: a broader CLI run including the pre-existing `TestModelSwitchRefreshesCustomStatusline` still fails with `model switch did not refresh custom statusline with the new model`; this is unrelated to this change.

### DIFF
--- a/internal/boot/boot_test.go
+++ b/internal/boot/boot_test.go
@@ -20,12 +20,13 @@ import (
 	"reasonix/internal/event"
 	"reasonix/internal/plugin"
 	"reasonix/internal/provider"
+	"reasonix/internal/sandbox"
+	"reasonix/internal/tool"
+	"reasonix/internal/tool/builtin"
 
-	// Blank imports register the provider kind and built-in tools the same way
-	// cmd/reasonix's main does; without them Build sees an empty provider
-	// registry and a bare tool set.
+	// Blank import registers the provider kind the same way cmd/reasonix's main
+	// does; importing builtin above registers the built-in tools.
 	_ "reasonix/internal/provider/openai"
-	_ "reasonix/internal/tool/builtin"
 )
 
 // TestBuildFoldsProjectMemoryIntoSystemPrompt is the end-to-end proof of the
@@ -169,6 +170,24 @@ api_key_env = "REASONIX_TEST_KEY_UNSET"
 	}
 	if !strings.Contains(sys, "projskill") || !strings.Contains(sys, "explore") {
 		t.Fatalf("skill names missing from index:\n%s", sys)
+	}
+}
+
+func TestAddBuiltinsWithWorkspaceRootKeepsSessionTools(t *testing.T) {
+	reg := tool.NewRegistry()
+	var stderr bytes.Buffer
+	addBuiltins(reg, nil, []string{t.TempDir()}, sandbox.Spec{}, builtin.SearchSpec{}, &stderr, t.TempDir())
+	for _, name := range []string{
+		"todo_write",
+		"complete_step",
+		"bash_output",
+		"kill_shell",
+		"wait",
+		"notebook_edit",
+	} {
+		if _, ok := reg.Get(name); !ok {
+			t.Fatalf("workspace builtins missing %q; got %v", name, reg.Names())
+		}
 	}
 }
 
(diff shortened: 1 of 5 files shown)


SivanCola (approved):
Approved. This is a well-scoped and worthwhile fix. The PR correctly addresses the workspace-bound builtin assembly bug by preserving the full session-level builtin tool surface while still overriding the tools that need workspace-aware behavior. That matches the current multi-project, multi-session architecture: each session can keep its own cwd/write-root behavior without losing tools such as todo_write, complete_step, bash_output, kill_shell, wait, and notebook_edit.

The main thing I wanted to tighten was the cache-sensitive contract. The original coverage proved that the missing tools are restored, but it did not explicitly prove that different workspace roots keep the same provider-visible tool schema. I added a small regression test, TestWorkspaceToolSchemasStableAcrossRoots, that builds two workspace-bound registries under different roots and verifies that Registry.Schemas() is identical and does not leak workspace paths. That gives us a clearer guard that this fix is compatible with high prompt-cache hit rates across project sessions.

I also re-ran the focused builtin, boot, and ACP tests plus git diff --check after the added test. Everything passed.

PR #3235, opened 2026-06-05
fix(desktop): refresh checkpoints after resume

SivanCola:
## Summary
- refresh desktop checkpoint state after turn completion
- refresh checkpoint state after resuming a saved session
- guard checkpoint refreshes against stale async responses per tab

## Testing
- npm --prefix desktop/frontend run build
- cd desktop && go test ./...

## Notes
- Missing .ckpt directories still degrade to an empty checkpoint list and the existing disabled restore-point UI.

### DIFF
--- a/desktop/frontend/src/lib/useController.ts
+++ b/desktop/frontend/src/lib/useController.ts
@@ -338,6 +338,15 @@ export function useController() {
     }
   }, [bump]);
 
+  const checkpointRefreshSeq = useRef(new Map<string, number>());
+  const refreshCheckpoints = useCallback(async (tabId: string) => {
+    const seq = (checkpointRefreshSeq.current.get(tabId) ?? 0) + 1;
+    checkpointRefreshSeq.current.set(tabId, seq);
+    const checkpoints = await app.CheckpointsForTab(tabId).catch(() => undefined);
+    if (checkpointRefreshSeq.current.get(tabId) !== seq || checkpoints === undefined) return;
+    dispatchTo(tabId, { type: "checkpoints", checkpoints: asArray(checkpoints) });
+  }, [dispatchTo]);
+
   const loadSessionDataForTab = useCallback(async (tabId: string, reset = false) => {
     try {
       if (reset) dispatchTo(tabId, { type: "reset" });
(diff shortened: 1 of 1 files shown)


SivanCola:
@codex review

SivanCola:
Review finding: one stale-response gap should be fixed before approval.

The direction of the PR is good: refreshing checkpoints after turn completion and after resume is the right behavior, and the per-tab sequence guard is a useful way to prevent older `CheckpointsForTab` responses from winning over newer refreshes. I also checked the security/cache surfaces: this change is frontend state refresh only, with no provider-visible prompt/tool schema changes and no new network/dependency/command/secret-handling behavior.

Issue: `desktop/frontend/src/lib/useController.ts:487` resets the tab in `newSession()` without invalidating any in-flight `refreshCheckpoints()` call for the same tab. If a `turn_done` checkpoint refresh is still in flight and the user starts a new session before it resolves, the old response can still pass the sequence check and dispatch stale checkpoints into the freshly reset session state. That leaves the new empty session showing restore points from the previous session, and later rewind actions can be based on misleading checkpoint metadata.

Suggested fix: invalidate outstanding checkpoint refreshes when a tab's session is reset/rotated, for example by bumping the per-tab checkpoint generation before `dispatchTo(activeTabId, { type: "reset" })` in `newSession()` or by routing resets through a helper that cancels pending checkpoint refreshes.

Verification performed:
- Reviewed the diff against current `origin/main-v2`.
- Confirmed GitHub checks are passing, including desktop, lint, race, govulncheck, and platform tests.
- `git diff --check origin/main-v2...HEAD` passed locally.
- Local frontend build could not be reproduced in the isolated worktree without generated Wails bindings/dependencies; CI's desktop job covers that path.

SivanCola:
Re-reviewed the latest head `ff4651fe`.

The follow-up fix addresses the stale checkpoint refresh issue I flagged earlier. `newSession()` now captures the current tab id and bumps the checkpoint refresh generation before rotating/resetting the session, so any in-flight `CheckpointsForTab` response from the previous session will fail the sequence check and cannot repopulate the fresh session state with stale restore points.

I also re-checked the security and cache surfaces:
- Security: no new dependency, network, command execution, secret handling, permission, persistence, or sandbox behavior is introduced. This remains a frontend state-refresh change.
- Cache: no provider-visible prompt prefix, tool schema, system prompt, memory/skill prefix, or provider serialization path is touched, so I do not see prompt-cache risk.

Verification:
- GitHub checks are green on the latest head, including desktop, lint, race, govulncheck, and macOS/Ubuntu/Windows tests.
- `git diff --check origin/main-v2...HEAD` passed locally.
- Reviewed the latest diff against current `origin/main-v2`.

No further issues from my side. I cannot submit a formal approve review from this account because GitHub treats it as the PR author, but this follow-up resolves my previous review finding.

PR #3249, opened 2026-06-05
fix(desktop): auto-recover dangling model refs on provider removal

SivanCola:
## Summary

When a provider is deleted or replaced in desktop settings, tabs with stale model refs (`default_model`, `planner_model`, `tab.model`) would fail at startup with `unknown model "deepseek-flash" (configured: glm-4v)` errors. This adds automatic fallback to the first configured provider, fixes all open tab refs on provider deletion, and shows actionable recovery UI.

### Key changes

- **`ResolveModelWithFallback`** — resolves a model ref with automatic fallback to the first configured provider when the ref is stale. Only picks providers with a valid API key.
- **`RemoveProvider` auto-migration** — detects both `"name"` and `"name/model"` refs in `default_model` and `planner_model`. Auto-migrates to fallback instead of blocking. Only refuses when no other configured provider exists.
- **buildTabController / rebuild / currentProviderEntryForTab** — all use fallback resolution so stale refs don't crash the session.
- **`DeleteProvider` tab sync** — traverses all open tabs and fixes model refs targeting the deleted provider. Always calls `rebuild()` and propagates errors.
- **Settings health diagnostics** — `SettingsView` now exposes `currentModel`, `currentModelValid`, `defaultModelValid`, `plannerModelValid`, `modelWarning`.
- **Sync methods** — `SyncActiveTabModelToDefault()` and `SyncAllTabModelsToDefault()` Wails bindings with frontend buttons.
- **Startup error recovery UI** — structured error fields (`startupErrCode`, `startupErrModel`, `startupErrConfigured`, `startupRecoveryModel`). Banner shows "Switch to X" button, "Open model settings" shortcut, and expandable error details.
- **`SetModelForTab` fix** — now clears `tab.StartupErr`, sets `tab.Ready = true`, and emits `agent:ready` so the frontend refreshes Meta/Balance/Effort after model recovery.

### Risk mitigations verified

1. `LoadForRoot` loads global credentials via `loadDotEnvForRoot` before `ResolveModelWithFallback` is called — `Configured()` correctly detects keyed providers.
2. CLI does not call `config.RemoveProvider` — the auto-migration semantic is desktop-only in practice.
3. `DeleteProvider` correctly propagates `rebuild()` errors (no `_ = a.rebuild()`).
4. `StartupErrorBanner` uses `useController.setModel` which dispatches Meta/Context/Effort refreshes after recovery.

## Test plan

```bash
go test ./internal/config/...     # 15 new tests + all existing pass
cd desktop && go test ./...       # existing tests pass
cd desktop/frontend && npm run build  # type check passes (pre-existing Wails errors only)
```


esengine:
Thanks for this — the model-ref fallback logic looks right and the main-module ResolveModelWithFallback tests pass. Two things block merging right now:

1. Conflict with the latest main-v2 in desktop/settings_app.go: your SyncActiveTabModelToDefault / SyncAllTabModelsToDefault landed next to the SetSubagentModel / SetSubagentEffort methods from #3330. They combine cleanly, just needs a rebase.

2. The new test TestDeleteProviderRebuildsInactiveAffectedTab fails deterministically (not a flake): it sets tab.model to deepseek-flash/... in memory but never creates the deepseek-flash provider in the isolated user config, so DeleteProvider returns a no-provider error. Writing the providers to config.UserConfigPath (the way TestSettingsSubagentDefaultsRoundTrip does) did not take effect in my attempt — the app still resolved the built-in preset (error said: configured deepseek/deepseek-pro), so the test needs setup that the running app actually reads (setTestCtrl may be bypassing config load). The pre-existing TestSetEffortRebuildsController fails the same way on main-v2.

Could you rebase onto the latest main-v2 and fix the provider fixture in these two tests? The feature is valuable (it fixes #3295). Happy to merge once green. Thank you!

SivanCola:
Thanks for the review. I rebased this PR onto the latest `main-v2` and pushed the updated head at `dce464a4`.

What changed:
- Kept the new `SetSubagentModel` / `SetSubagentEffort` methods together with `SyncActiveTabModelToDefault` / `SyncAllTabModelsToDefault` during the rebase.
- Rechecked the provider-fixture path for `TestSetEffortRebuildsController` and `TestDeleteProviderRebuildsInactiveAffectedTab`; both now pass on the rebased branch.
- Fixed the frontend build blocker by removing duplicate history locale keys introduced by the rebase.

Verified locally:
- `cd desktop && go test . -run 'Test(DeleteProviderRebuildsInactiveAffectedTab|SetEffortRebuildsController)' -count=1`
- `cd desktop && go test . -count=1`
- `wails generate module`
- `go test ./internal/plugin -run TestStartPhaseAReturnsBeforePhaseB -count=1`

CI has been triggered on the updated PR head and is still in progress.

SivanCola:
@codex review

esengine:
Superseded by #3421, which is a strict superset of this PR's provider-removal work (same-author reimplementation): #3421's `RemoveProvider` also migrates `subagent_model` + per-skill `SubagentModels` and snapshots/closes controllers before rebuild, whereas this PR handles only default+planner. The two hard-conflict (16 merge-tree conflicts across shared files; identically-named `ResolveModelWithFallback` + `ModelRefsProvider`). Closing in favor of #3421.

If #3421 doesn't already include them, the salvage-worthy delta here is the `StartupErrorBanner` startup-recovery UI (+ structured `Meta.StartupErr*` fields) and the `SyncActiveTabModelToDefault`/`SyncAllTabModelsToDefault` buttons — worth porting in a follow-up.

PR #3259, opened 2026-06-05
fix(cli): prevent bang shell commands from grabbing the TUI

### LINKED ISSUE #3229: [Bug]:cli 界面输入"!sudo ls"后，会卡在思考中，esc 无效，control+c 也无效
### Version line

### Exact version

v1.2.0-rc.1

### What happened?

Image

### Steps to reproduce

1. 进入cli
2. 输入 "!sudo ls"

### OS / platform

macos 12.5

### Relevant logs or output

```shell

```

paradoxSCH:
## Summary
- Run user-invoked `!` shell commands in a new POSIX session so interactive tools such as `sudo` cannot grab the TUI controlling terminal.
- Keep cancellation killing the whole shell process group.
- Add coverage for shell cancellation and the POSIX session-detach contract.

## Root cause
`!sudo ls` can make sudo open `/dev/tty` for password input instead of reading from the captured stdin/stdout pipes. In the chat TUI that steals the terminal from Bubble Tea, so the turn appears stuck and Esc/Ctrl+C cannot be handled reliably until the shell timeout.

## Validation
- `go test ./internal/control -count=1`
- `GOOS=darwin GOARCH=amd64 go test -c ./internal/control`
- `go test ./internal/cli -run Test.*(Shell|Bang|Cancel|Esc|Ctrl|ToolProgress|ToolResult) -count=1`

Note: `go test ./internal/cli ./internal/control -count=1` still hits the existing unrelated `TestModelSwitchRefreshesCustomStatusline` failure in `internal/cli`.

Closes #3229

### DIFF
--- a/internal/control/shell_kill_other.go
+++ b/internal/control/shell_kill_other.go
@@ -7,12 +7,11 @@ import (
 	"syscall"
 )
 
-// setShellKillTree makes a cancelled shell command kill its whole process group,
-// not just the shell leader — otherwise a timed-out pipeline (e.g.
-// !find / -name foo | grep bar) orphans the children. The child gets its own
-// group (Setpgid) so the negative-pid signal reaches every descendant.
+// setShellKillTree makes cancellation kill the whole shell tree. Running the
+// child in a new session also keeps interactive prompts from grabbing the TUI's
+// controlling terminal.
 func setShellKillTree(cmd *exec.Cmd) {
-	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setsid: true}
 	cmd.Cancel = func() error {
 		if cmd.Process == nil {
 			return nil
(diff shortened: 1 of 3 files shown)


SivanCola (commented):
Thanks, this looks like the right fix for the bang-shell path. One small follow-up request: can you apply the same POSIX session-detach contract to the builtin bash tool path as well?

`internal/tool/builtin/bash_kill_other.go` still uses `Setpgid: true`. That preserves process-group cancellation, but it leaves model-invoked or user-approved bash commands in the TUI controlling-terminal session. A command like `sudo ...` can therefore hit the same `/dev/tty` behavior when run through the bash tool instead of the user-entered `!` path.

I think the same `Setsid: true` change is appropriate there too, with a matching non-Windows SysProcAttr regression test if possible. The negative-pid kill should continue to target the shell process group because `setsid` makes the child a new session and process-group leader.

Once this is addressed and CI stays green, I will consider merging this into `main-v2`.

PR #3336, opened 2026-06-06
feat(composer): add past chats reference flow

SuMuxi66:
Part of #3185
## Summary

Add `@ → past:chats` flow to the Composer, allowing users to select a historical chat session and inject its context into the current conversation.

### What's included

- **`@` menu**: New `past:chats` entry in the `@` context menu
- **Session list**: Browse recent chat sessions with title and turn count
- **Search**: Filter sessions by keyword
- **Hover preview**: Preview session content on hover before selecting
- **Context injection**: Selected session's conversation is appended to the prompt as context
- **Architecture doc**: `docs/SESSION_REFERENCE_ARCHITECTURE.md` documenting the data flow

### Files changed

| File | Change |
|---|---|
| `Composer.tsx` | `@` menu entry, session list, search, hover preview, reference card in Composer |
| `FileMenu.tsx` | Wire `past:chats` menu item |
| `types.ts` | `SessionReference` interface |
| `styles.css` | Composer reference card styles |
| `SESSION_REFERENCE_ARCHITECTURE.md` | Architecture documentation |

## Test checklist

- [ ] `@` menu shows `past:chats` option
- [ ] Clicking `past:chats` opens session list
- [ ] Search filters sessions by keyword
- [ ] Hover preview shows session content
- [ ] Selecting a session adds reference card above Composer input
- [ ] Reference card shows title, turn count, file name
- [ ] Hover on reference card shows full path
- [ ] Sending with reference: AI can reference the selected session content
- [ ] Sending without reference: normal behavior unchanged
- [ ] Deleting reference card before send: no reference injected
- [ ] `@` file references still work
- [ ] `/` command menu still works
- [ ] No regression in existing Composer behavior

### DIFF
--- a/desktop/frontend/src/components/Composer.tsx
+++ b/desktop/frontend/src/components/Composer.tsx
@@ -1080,6 +1080,15 @@ export function Composer({
     setText((prev) => removeAtToken(prev));
     setPastChatQuery("");
     setShowPastChats(false);
+    setActive(0);
+    // Return focus to textarea so the user can keep typing immediately.
+    requestAnimationFrame(() => {
+      taRef.current?.focus();
+      taRef.current?.setSelectionRange(
+        taRef.current.value.length,
+        taRef.current.value.length,
+      );
+    });
   };
 
   const removeSessionRef = (path: string) => {
--- a/desktop/frontend/src/components/Composer.tsx
+++ b/desktop/frontend/src/components/Composer.tsx
@@ -164,6 +164,7 @@ const SESSION_REF_MAX_MESSAGES = 30;
 const SESSION_REF_MAX_CHARS = 20_000;
 const SESSION_CONTEXT_HEADER = "以下是用户引用的历史会话上下文：";
 const SESSION_CONTEXT_FOOTER = "当前用户问题：";
+const PAST_CHATS_MENU_ITEM = "past:chats";
 
 // limitSessionMessages keeps the most recent useful messages within a char budget.
 // Walks from the end so the truncation is always "drop the oldest", which matches
@@ -325,6 +326,7 @@ export function Composer({
   const [loadingPastChats, setLoadingPastChats] = useState(false);
   const [submitting, setSubmitting] = useState(false);
   const taRef = useRef<HTMLTextAreaElement>(null);
+  const pastChatsItemRef = useRef<HTMLButtonElement>(null);
   const composerCardRef = useRef<HTMLDivElement>(null);
   const workspaceAnchorRef = useRef<HTMLDivElement>(null);
   const wasRunning = useRef(running);
@@ -486,7 +488,7 @@ export function Composer({
     | { kind: "pastChats" }
     | { kind: "file"; entry: DirEntry };
 
-  const includePastChatsItem = atRaw !== null && atDir === "";
+  const includePastChatsItem = atRaw !== null && atDir === "" && (atFrag === "" || PAST_CHATS_MENU_ITEM.startsWith(atFrag));
 
   const atMenuItems = useMemo<AtMenuItem[]>(
     () => [
@@ -1056,6 +1058,12 @@ export function Composer({
     setActive((prev) => (prev > maxIdx ? 0 : prev));
   }, [count]);
 
+  useEffect(() => {
+    if (menuMode === "at" && !showPastChats && includePastChatsItem && active === 0) {
+      pastChatsItemRef.current?.scrollIntoView({ block: "nearest" });
+    }
+  }, [active, includePastChatsItem, menuMode, showPastChats]);
+
   const removeAtToken = (value: string) => {
     return value.replace(/(?:^|\s)@[^\s]*$/, "").trimEnd();
   };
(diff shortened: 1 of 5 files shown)


SuMuxi66:
## Scope

This PR only adds the `past:chats` reference flow to the desktop Composer UI (`desktop/frontend`). It does not change the CLI/TUI input flow.

SivanCola (changes requested):
Thanks, this is directionally useful, but I don't think it is ready to merge yet.

Please address these before merging:

- The new `past:chats` entry is rendered inside the `@` menu, but the menu state machine still only counts and picks file entries. `menuMode` only opens when `atMatches.length > 0`, `count` is still `atMatches.length`, and `pickActive()` still routes Enter/Tab to `pickEntry(atMatches[active])`. That means keyboard selection cannot open `past:chats`, keyboard selection inside the past-chat list cannot choose a session, and the entry can disappear entirely when there are no file matches. Please make `past:chats` a real selectable item in the `@` menu state model, including keyboard navigation and the empty-file-list case.

- `submit()` now awaits `buildSessionContext(sessionRefs)` before calling `onSend`, but there is no local submitting guard while those `PreviewSession` calls are pending. During that window, the send button remains enabled and Enter can trigger another submit, so users can duplicate the same message if session preview reads are slow. Please add a pending-submit state or otherwise disable duplicate sends until the async context build finishes.

- `git diff --check` currently fails due to trailing whitespace in `docs/SESSION_REFERENCE_ARCHITECTURE.md` at lines 71, 144, 164-166, 173, 200, 209, and 219.

Notes from review:

- I did not see new dependency, network, command execution, or provider/tool-schema cache-stability risk.
- The PR is behind latest `main-v2`, but a merge-tree check against the latest base did not show a text conflict. Please still rebase or update before final verification.
- Frontend `typecheck` could not be completed in a temporary worktree because this checkout lacks generated Wails `wailsjs` types; that is not counted as a PR-specific failure here.

Once these are addressed and CI stays green, I will consider merging it into `main-v2`.

SuMuxi66:
Updated.

Changes addressed:
- Made `past:chats` a first-class selectable `@` menu item, including keyboard navigation and root-level empty-file-list behavior.
- Added pending-submit guard with ref + state so async `buildSessionContext(sessionRefs)` cannot duplicate sends.
- Cleaned trailing whitespace in `docs/SESSION_REFERENCE_ARCHITECTURE.md`.
- Rebased onto latest `main-v2`.
- Verified `git diff --check` passes and manually tested `@`, `past:chats`, file refs, `/` menu, and duplicate-submit prevention.

SuMuxi66:
Updated again.

Rebased onto the latest `main-v2` and resolved the new `styles.css` conflict.

Verified:
- `git diff --check origin/main-v2..HEAD` passes.
- `git merge-tree --write-tree origin/main-v2 HEAD` returns a clean tree with no conflicts.
- The PR still only touches the intended files:
  - `Composer.tsx`
  - `FileMenu.tsx`
  - `types.ts`
  - `styles.css`
  - `docs/SESSION_REFERENCE_ARCHITECTURE.md`

Manual checks still pass for root-level `@`, `past:chats`, file refs, `/` menu, session selection focus return, and duplicate-submit prevention.

SivanCola (changes requested):
Thanks for the updates. The previous blocking items are mostly addressed, but I found one remaining regression that should be fixed before merge:

- The root-level `@` menu now always inserts `past:chats` as the first item whenever `atDir === ""`. Because `active` resets to 0 as the query changes, typing something like `@README` or `@src` and pressing Enter/Tab opens the past-chats picker instead of selecting the first matching file entry. This regresses the existing `@file` keyboard flow and conflicts with the PR checklist item that file references still work. Please only show/select `past:chats` when the empty root `@` menu is open or when the current fragment actually matches `past:chats`, or otherwise preserve the first file match as the default active item for non-empty file queries.

Also, please avoid closing #3185 from this PR as-is. The issue asks for a broader `#past_chats` flow and is tagged for desktop/TUI/agent, while this PR explicitly scopes itself to the desktop Composer. `Part of #3185` or a follow-up issue for the remaining TUI/trigger behavior would be more accurate.

Once the `@file` keyboard regression is fixed and verification stays green, I will consider merging this into `main-v2`.

SivanCola (approved):
Approved. The latest update addresses the remaining blocker from my previous review: `past:chats` is no longer unconditionally inserted ahead of file matches for non-matching `@` fragments, so normal `@README` / `@src` style keyboard selection can keep resolving to file entries while `@` and matching `@past...` fragments can still reach the past-chats picker.

- The feature stays scoped to the desktop Composer and uses the existing `ListSessions` / `PreviewSession` bridge methods instead of adding new persistence, network, command execution, or permission surfaces.
- Referenced chat content is added only to the dynamic submit payload after an explicit user selection; it does not change provider-visible stable prefixes, tool schemas, provider serialization, or cache-sensitive prompt setup.
- The PR body now correctly says `Part of #3185`, which matches the desktop-only scope and leaves the broader TUI / trigger behavior open for follow-up.

Related work: #3338 is still the stacked follow-up for rendering selected references inside user message bubbles, so it should be rebased/reviewed after this core Composer flow lands.

Verified:
- `git diff --check origin/main-v2...origin/pr/3336` passed
- `git merge-tree --write-tree origin/main-v2 origin/pr/3336` produced a clean tree
- `pnpm --dir desktop/frontend check:css` passed
- `pnpm --dir desktop/frontend test:all` passed (85 tests)

Caveat: `pnpm --dir desktop/frontend typecheck` still cannot complete in a temporary checkout because the Wails-generated `wailsjs` bindings are absent there; this is the same environment limitation as before, not a PR-specific failure.

PR #3140, opened 2026-06-04
fix(tool-card): collapse sub-agent tree with parent chevron

HUQIANTAO:
## Summary

The parent task tool card's chevron only toggled args/output visibility; nested sub-agent calls (read_file / grep / ls from an /explore run) **always** rendered underneath.  A 50-step /explore flooded the transcript with its full step list, with no way to fold it back.

This is the ToolCard half of the original #3140.  The streaming-markdown half has been merged via #3444.

## Fix

The `open` state now also controls nested items.  A single chevron click collapses or expands the whole subagent tree:

- **Open by default** while the sub-agent is still `running` so the user can watch progress.
- **Closed by default** once it settles, so a finished explore folds to a single summary line ("12 steps") on first paint.
- Individual sub-calls can still be expanded/collapsed independently via their own ProcessCard chevrons.

### Diff

```diff
-  const [open, setOpen] = useState(item.isShell && hasArgsOrOutput);
+  const [open, setOpen] = useState(() => {
+    if (item.isShell) return hasArgsOrOutput;
+    if (hasNested) return item.status === "running";
+    return false;
+  });
```

## Scope

- **1 file changed**: `ToolCard.tsx` (+10 / −3)
- No Message.tsx changes (covered by #3444)
- No Go backend changes

## Verification

- `pnpm typecheck` ✅ clean
- `vite build` ✅ built in 28.58s
- Running sub-agent: nested items visible, chevron works
- Completed sub-agent: nested items hidden, chevron expands
- Shell commands: unchanged behavior (auto-open when has output)
- Regular tools: unchanged behavior (args/output folded by default)

PR #3337, opened 2026-06-06
fix(message): render markdown while streaming assistant output

SuMuxi66:
## Summary

This PR is now a focused follow-up on top of the merged streaming Markdown work in #3444.

It keeps the streaming cursor rendered inside the Markdown flow instead of outside the rendered block, so the cursor placement remains visually consistent while Markdown content is streaming.

## Scope

After rebasing onto the latest `main-v2`, this PR contains only:

- `Markdown.tsx`
- `Message.tsx`

The original broader streaming Markdown changes are no longer included because #3444 has already landed.

## Validation

- Rebased onto latest `origin/main-v2`
- `npx tsc --noEmit` passes with 0 errors
- Final diff contains 1 commit:
  - `fix(message): keep streaming cursor inside markdown`

## Notes

This PR also preserves inline Markdown elements such as `<strong>` and `<em>` via the `cloneElement` path.

### DIFF
--- a/desktop/frontend/src/components/Markdown.tsx
+++ b/desktop/frontend/src/components/Markdown.tsx
@@ -1,11 +1,4 @@
-import {
-  Children,
-  cloneElement,
-  isValidElement,
-  memo,
-  useDeferredValue,
-  type ReactNode,
-} from "react";
+import { memo, useDeferredValue, useLayoutEffect, useRef } from "react";
 import ReactMarkdown from "react-markdown";
 import type { Components } from "react-markdown";
 import remarkGfm from "remark-gfm";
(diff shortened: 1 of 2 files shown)


SivanCola (changes requested):
Thanks, this is the right direction and the scope is much cleaner than the broader streaming-markdown work in #3140. I think this PR should be the main path for the focused assistant streaming markdown fix.

Before merging, please fix the streaming cursor placement. `Markdown` renders a block-level `.md` wrapper, so rendering `<span className="cursor" />` as a sibling after `<Markdown />` places the cursor after the markdown block rather than at the end of the rendered text. For normal paragraphs this will show the cursor on the next line, not after the last token.

Please make the cursor render as part of the final rendered markdown content, then verify paragraphs, lists, tables, and code blocks while streaming.

Once this is addressed and verification stays green, I will consider merging it into `main-v2`.

SuMuxi66:
Updated.

Changes addressed:
- Kept the streaming cursor inside the rendered Markdown content instead of rendering it as a sibling after `<Markdown />`.
- Replaced the raw streaming branch with real-time Markdown rendering.
- Added a `showCursor` path in `Markdown.tsx` using a sentinel that is rendered as the cursor inside Markdown content.
- Covered paragraphs, lists, tables, and code blocks so the sentinel is not shown as raw text.

Verified:
- `git diff --check origin/main-v2..HEAD` passes.
- TypeScript check passes.
- Manually tested streaming paragraphs, lists, tables, and code blocks.
- Cursor disappears after the turn completes.

SuMuxi66:
更新一下当前 PR 状态：

因为 #3444 已经合并到 `main-v2`，所以这个 PR 已经重新 rebase 到最新 `origin/main-v2`，并且收窄为一个很小的 follow-up。

当前这个 PR 不再重复实现 streaming Markdown 主逻辑，只保留一个剩余问题修复：

- streaming 过程中，让光标继续保持在 Markdown 渲染流内部
- 避免光标跑到渲染块外面，导致视觉位置不一致
- 保留 `<strong>` / `<em>` 等 inline element 的正常渲染，已通过 `cloneElement` 路径处理

当前最终范围：

- 1 个 commit：
  - `fix(message): keep streaming cursor inside markdown`
- 只改 2 个文件：
  - `Markdown.tsx`
  - `Message.tsx`

验证结果：

- 已 rebase 到最新 `origin/main-v2`
- `npx tsc --noEmit` 通过，0 错误

所以现在这个 PR 可以理解为基于 #3444 之后的 cursor placement follow-up，不是和 #3444 重复的 streaming Markdown PR。

SivanCola (changes requested):
Thanks for narrowing this PR down. I still need changes before merge:

- In `Markdown.tsx`, `renderCursorChildren` and `stripSentinel` unwrap every React element they encounter by returning only `props.children`. Because the new `p`/`li`/`td`/`th` renderers run for all Markdown, this regresses normal rendered output too: emphasis, strong text, links, inline code wrappers, task-list checkboxes, and nested list/table structure can lose their element type and props even when `showCursor` is false. Please preserve the original element with transformed children, for example via `cloneElement`, instead of returning only the descendants.

- For an in-progress fenced code block, the appended sentinel is part of the code children. The `code` renderer calls `stripSentinel` and then renders `CodeViewer` without any cursor, so the streaming cursor disappears while the assistant is inside an unfinished code block. Please keep the copied/highlighted code text pure while still rendering a visible cursor in the code-block streaming state.

Reviewed/verified:
- `git diff --check origin/main-v2...origin/pr/3337` passed.
- `pnpm --dir desktop/frontend test` passed.
- `pnpm --dir desktop/frontend typecheck` could not complete in a clean review worktree because the generated Wails modules are absent there; the errors are outside this diff.

Once these are addressed and verification stays green, I will consider merging this into `main-v2`.

desktop/frontend/src/components/Markdown.tsx
```diff
@@ -43,16 +138,23 @@ const components: Components = {
...
-export const Markdown = memo(function Markdown({ text }: { text: string }) {
+export const Markdown = memo(function Markdown({
+  text,
+  showCursor,
+}: {
+  text: string;
+  showCursor?: boolean;
+}) {
   const deferred = useDeferredValue(text);
+  const withCursor = showCursor ? deferred + CURSOR_SENTINEL : deferred;
```
SivanCola: Appending the sentinel to the raw Markdown can change the Markdown grammar before ReactMarkdown parses it. For example, a just-closed fenced block like ` ```js\nconsole.log(1)\n``` ` becomes a closing fence with trailing `\uE000`, so CommonMark no longer treats it as the fence terminator. I verified react-markdown then renders the literal closing fence inside the code block and copy text. Please avoid changing raw Markdown syntax when placing the cursor, or special-case this so closed fenced blocks stay closed.

desktop/frontend/src/components/Markdown.tsx
```diff
@@ -19,16 +26,104 @@ import { openExternal } from "../lib/bridge";
...
+    }
+    return child;
+  });
+}
+
 const components: Components = {
+  p: ({ children }) => <p>{renderCursorChildren(children)}</p>,
+  li: ({ children }) => <li>{renderCursorChildren(children)}</li>,
+  td: ({ children }) => <td>{renderCursorChildren(children)}</td>,
+  th: ({ children }) => <th>{renderCursorChildren(children)}</th>,
```
SivanCola: These custom renderers still drop props that ReactMarkdown/GFM attaches to the original nodes. This regresses completed Markdown too: table alignment styles on `td`/`th` disappear, and task-list `li` class names are removed. Please preserve the original props while transforming children, e.g. by accepting the full props object and spreading it back onto `p`/`li`/`td`/`th`.

desktop/frontend/src/components/Markdown.tsx
```diff
@@ -19,16 +26,104 @@ import { openExternal } from "../lib/bridge";
...
+      const props = child.props as { children?: ReactNode };
+      if ((child.type as unknown) === CURSOR_SENTINEL) return null;
+      return cloneElement(child, undefined, stripSentinel(props.children));
+    }
+    return child;
+  });
+}
+
 const components: Components = {
+  p: ({ children }) => <p>{renderCursorChildren(children)}</p>,
```
SivanCola: The cursor replacement only runs for `p`/`li`/`td`/`th`/`code`, so other block renderers can still leak the raw sentinel. A streaming heading such as `# Heading` currently renders as `Heading` instead of showing the cursor span. Please cover headings and any other inline-containing Markdown nodes, or use a placement approach that does not require per-node renderer coverage.

SivanCola (changes requested):
Thanks for the update. I re-ran the cursor cases against the current head. The cloneElement fix moves this in the right direction, but there are still Markdown correctness regressions that need to be addressed before merge.

SuMuxi66:
已经做了最新的修改了，可以审核了

esengine (approved):
﻿Re-reviewed the latest head. This iteration replaces the sentinel + element-unwrapping approach wholesale with post-render DOM injection (`useLayoutEffect` + TreeWalker), which structurally removes the earlier correctness concerns: the markdown source string is never touched, the React component tree keeps its default renderers, so emphasis/links/task-list/table structure can no longer regress — there is simply no transform pass anymore.

Verified against main-v2:
- `CodeViewer`/`HljsCode` render real `<pre>` elements, so the PRE rejection keeps copied/highlighted code text pure.
- The `.cursor` class this relies on already ships in `styles.css` on main-v2, so the injected span gets the blinking caret styling.
- Cursor cleanup is idempotent (remove-all then re-inject each commit), and `showCursor=false` removes it on stream end.

One accepted trade-off: while the assistant is mid-way through an unfinished fenced code block, the cursor sits after the last text outside the block rather than inside the highlighted code. That keeps the code DOM untouched, and I prefer that purity over chasing the caret into the highlighter's output — fine to revisit in a follow-up if it bothers anyone in practice.

Thanks for the patience through three iterations — this landed in a much better place than where it started. Merging.

esengine:
﻿@SivanCola the latest head replaced the sentinel/cloneElement transform wholesale with post-render DOM injection — the element-unwrapping regressions your review flagged no longer apply (details in my review above). I've approved and armed auto-merge; please re-review or dismiss your earlier changes-requested when you get a chance, it's the last thing holding this back.

PR #3338, opened 2026-06-06
feat(message): show past chat references and files in user message bubble

SuMuxi66:
## 概述

本 PR 是 #3336 合并后的 follow-up。

#3336 已经完成 `past:chats` 的选择、预览和上下文注入流程。本 PR 只补充发送后的展示能力：当用户消息引用了历史会话时，在用户消息气泡中展示对应的引用卡片。

## 改动内容

- 将已选中的 `past:chats` references 随用户消息一起传入 transcript
- 用户消息支持携带并渲染 references
- 在用户消息气泡中展示历史会话引用卡片
- 引用卡片展示会话标题、消息数量、文件名等信息
- 调整用户消息气泡和引用卡片样式

## 范围说明

本 PR 只处理“发送后如何展示引用”的问题。

不修改：

- `past:chats` 菜单选择逻辑
- 历史会话预览逻辑
- 上下文注入逻辑
- `@` 文件引用逻辑

## 验证

- 基于已合并的 #3336
- 发送带 `past:chats` 引用的消息后，用户消息气泡中可以看到引用卡片
- 未引用历史会话的普通用户消息显示正常
- 文件引用流程不受影响

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -579,6 +579,21 @@ func (a *App) SubmitDisplayToTab(tabID, display, input string) {
 	ctrl.SubmitDisplay(display, input)
 }
 
+// SubmitDisplayToTabWithRefs submits a message with display text and persists
+// the user's past:chat references into the sidecar so they survive reload.
+func (a *App) SubmitDisplayToTabWithRefs(tabID, display, input string, references json.RawMessage) {
+	ctrl := a.ctrlByTabID(tabID)
+	if ctrl == nil {
+		return
+	}
+	ctrl.SubmitDisplay(display, input)
+	dir := ctrl.SessionDir()
+	if dir == "" {
+		dir = config.SessionDir()
+	}
+	_ = recordSessionReferences(dir, ctrl.SessionPath(), input, references)
+}
+
 func (a *App) bindControllerDisplayRecorder(ctrl *control.Controller) {
 	if ctrl == nil {
 		return
@@ -1505,6 +1520,7 @@ type HistoryMessage struct {
 	Messages   int               `json:"messages,omitempty"`
 	Summary    string            `json:"summary,omitempty"`
 	Archive    string            `json:"archive,omitempty"`
+	References json.RawMessage   `json:"references,omitempty"`
 }
 
 type HistoryToolCall struct {
@@ -1524,10 +1540,14 @@ func (a *App) HistoryForTab(tabID string) []HistoryMessage {
 		return []HistoryMessage{}
 	}
 	msgs := ctrl.History()
-	return historyMessages(msgs, sessionDisplayResolver(controllerSessionDir(ctrl), ctrl.SessionPath()))
+	dir := ctrl.SessionDir()
+	if dir == "" {
+		dir = config.SessionDir()
+	}
+	return historyMessages(msgs, sessionDisplayResolver(dir, ctrl.SessionPath()), sessionReferencesResolver(dir, ctrl.SessionPath()))
 }
 
-func historyMessages(msgs []provider.Message, resolveUserContent func(string) string) []HistoryMessage {
+func historyMessages(msgs []provider.Message, resolveUserContent func(string) string, resolveUserRefs func(string) json.RawMessage) []HistoryMessage {
 	out := make([]HistoryMessage, 0, len(msgs))
 	for _, m := range msgs {
 		content := m.Content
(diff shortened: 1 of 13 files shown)


SivanCola (changes requested):
Thanks, this is directionally useful, but I cannot approve it yet.

Required changes:

- The send path waits for `PreviewSession`/`buildSessionContext` before calling `onSend` or clearing the composer. While that await is pending, the input and send button remain enabled, so pressing Enter or clicking send again can submit duplicate turns with the same referenced sessions. It also breaks the PR's stated "display immediately after send" behavior for referenced sends. Please snapshot/lock/clear the composer before the async session reads, or add an explicit sending guard that makes duplicate submission impossible.

- The new `past:chats` UI is not wired into the existing keyboard menu state. `menuMode` only opens the `@` menu when `atMatches.length > 0`, `count` still excludes the special/session rows, and `pickActive()` always calls `pickEntry(atMatches[active])`. This means the past-chat option can disappear when there are no file matches, and Enter/Tab cannot select a highlighted past-chat session. Please make past-chat rows first-class menu items for open/active/count/pick handling.

- New visible UI strings are hard-coded in Chinese in `Composer.tsx` and `Message.tsx`. The desktop frontend has an existing en/zh dictionary contract, so English locale users will see mixed Chinese. Please add the required i18n keys in both locale files and use `t(...)`/`useT()` for visible text.

- `git diff --check origin/main-v2...origin/pr/3338` currently fails because `docs/SESSION_REFERENCE_ARCHITECTURE.md` contains trailing whitespace. Please clean that up before merge.

Integration notes:

- This PR is still draft and stacked on #3336; it should be rebased after the core past-chats PR is finalized.
- It also overlaps #3356 on user-message bubble styling, so one of the two PRs should become the integration source for that layout.

Verified locally:

- `wails generate module` passed after creating the expected temporary `frontend/dist` placeholder in a throwaway worktree.
- `npm run build` passed in `desktop/frontend`.
- `go test . -run 'TestPreviewSession|TestListSessions|TestHistory|TestSession'` passed in `desktop`.
- `git merge-tree --write-tree origin/main-v2 origin/pr/3338` produced an auto-merge tree with no textual conflicts.

SivanCola (changes requested):
Thanks for the follow-up changes. I re-reviewed the current head and the earlier blockers around duplicate sends, menu state, i18n, and `diff --check` look addressed now.

I still found a couple of issues that should be fixed before merge:

- The referenced-session cards are not persisted. `send()` only passes `references` into the frontend reducer, while the backend history/display sidecar still stores only strings, and `historyMessagesToItems()` rebuilds user messages without references. So the cards appear right after sending, but disappear after reload/resume/history replay. Please persist the structured session refs alongside the display content and restore them when mapping history back to `Item`s.

- `Message.tsx` wraps a block `<div className="msg__reference-item">` inside `Tooltip`, which defaults to a `<span>` trigger. That creates invalid DOM / React nesting warnings and can affect layout. Please use the Tooltip block mode here or change the child to an inline-compatible element.

- A selected past-chat reference alone cannot be sent: the submit guard and send-button disabled condition still only count text, attachments, and workspace refs. If a session ref is meant to be usable as context by itself, include `sessionRefs.length` in those checks; if not, the UI should make the “question required” state explicit instead of showing an attached chip with a disabled send button.

Verified: `git diff --check origin/main-v2...origin/pr/3338` passes on the current PR head.

SuMuxi66:
感谢 review，我已经按反馈修复了这三个 blocker。

本次更新：

- "past:chats" references 现在会通过 ".references.json" sidecar 持久化，模式和现有 ".display.json" sidecar 保持一致。
- "HistoryForTab()" 和 history-to-transcript 映射现在会恢复 references，所以刷新、继续会话、历史回放后，用户消息里的引用卡片不会丢失。
- references 卡片块现在位于 ".msg__body" 下，不再嵌套进 Markdown paragraph / "<p>" 内，避免非法 DOM 结构。
- submit guard 和 send button disabled 逻辑都已加入 "sessionRefs.length === 0" 判断，因此只选择 past chat reference、正文为空时也可以发送。

验证结果：

- reload 测试通过：浏览器刷新后，引用卡片仍然显示。
- DOM 检查通过：reference card 的父节点是 ".msg__body"。
- console 检查通过：没有前端错误。
- "git diff --check" 通过。
- Go 端 "go build"、"go vet" 和相关 history / preview 测试通过。
- full frontend "tsc" 目前仍被既有的 "VirtualMenu.tsx" / "@tanstack/react-virtual" 依赖问题阻塞；本 PR 没有在改动文件中引入新的 TypeScript 错误。

esengine (changes requested):
The sidecar design itself is right (atomic write, purge cleanup, history re-attach) — but two regressions smuggled in by refactors outside the stated scope block this:

1. `useController.send()` was rewritten to `if (!activeTabId) return` — that deletes the `activeTabIdRef.current` fallback and the `activeTabFromBackend()` recovery currently on main-v2, so messages sent before the tab id settles are silently dropped.
2. `App.handleSend` replaces the three-call sync with `syncModeToController(modeFromAxes(plan, yolo))` — that collapses the ask/auto/yolo tri-state to a boolean (loses "auto") and drops the goal sync entirely.

Revert both hunks to current main-v2 behavior (or move them to their own PR with the rationale). Then the smaller items: the hardcoded Chinese strings in Message.tsx (`引用的会话`, `· {n} 轮`, the aria-label) must go through i18n; drop the duplicate `.rewind` CSS block and the unused `references` prop on TurnActions; and rebase. With those done I'm happy to take it — the feature is a good follow-on to #3336.

PR #3345, opened 2026-06-06
fix(cli): prompt before reusing setup API keys

wade19990814-hue:
## Summary
- prompt during `reasonix setup` before reusing an already-set provider API key
- keep Enter as the safe default that reuses and re-pins the existing value
- allow `y` to re-enter the key so stale credentials can be replaced without manually editing the credentials file
- add localized English/Chinese prompt text and regression coverage for reuse and reset paths

## Tests
- `go test ./internal/cli -run 'TestConfigureKeys|TestAppendEnvUpsertReplacesExistingKey|TestAppendEnvUpsertHandlesExportPrefix' -count=1`
- `go test ./internal/i18n -count=1`
- `git diff --check`

## Notes
- `go test ./internal/cli -count=1` still has unrelated local failures on this machine, including locale-sensitive assertions from the user config and `TestModelSwitchRefreshesCustomStatusline`.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -1296,17 +1296,12 @@ func providersWithMissingKeys(cfg *config.Config) []config.ProviderEntry {
 }
 
 // configureKeys reconciles each enabled provider's API key with the
-// environment. For every distinct api_key_env: if the variable is already
-// set — either by loadDotEnv from .env, or by an earlier wizard step that
-// called os.Setenv (the URL-fetch flow asks for the key once so it can call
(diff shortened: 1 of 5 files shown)


wade19990814-hue:
Manual test passed on Windows.

I built a local test binary from `main-v2` with both setup UX changes applied (#3342 + this PR) and exercised `reasonix setup` against the real `%APPDATA%\reasonix\config.toml` / `credentials` flow. The overwrite prompt and existing API-key re-entry prompt both behaved as expected, with Enter keeping the existing value and `y` allowing re-entry.

PR #3360, opened 2026-06-06
feat(desktop): mid-turn steer injection

JesonChou:
## Summary

- Implement mid-turn steer injection with tab-scoped routing, addressing reviewer feedback from #3263 (wire mappings + `SteerForTab`).
- Align `historyMessagesToItems` tool-result lookup with `SanitizeToolPairing`'s turn-scoped pairing: position-aware scan instead of a session-global Map to avoid cross-turn ID collisions with synthesized `call_%d` IDs.

## Changes

**Steer (#3263)**
- `internal/event/event.go`: `Steer` event kind
- `internal/agent/agent.go`: `steerQueue` + Run-loop injection + persistence
- `internal/control/controller.go`: `Steer`/`SteerConsumed`
- `desktop/app.go`: `App.Steer`/`App.SteerForTab`
- Wire mappings in `desktop/wire.go` + `internal/serve/wire.go` with tests
- Frontend: bindings, `runningRef` + steer dispatch in `handleSend`

**Tool-result pairing alignment (#3286)**
- `useController.ts`: global `resultByID` Map → `collectToolResultsAfter`
- `internal/serve/index.html`: same

## Related

Closes #3263

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -587,6 +587,18 @@ func (a *App) CancelTab(tabID string) {
 	}
 }
 
+// Steer sends mid-turn guidance to the agent without interrupting the in-flight request.
+func (a *App) Steer(text string) {
+	a.SteerForTab("", text)
+}
+
+// SteerForTab sends mid-turn guidance to a specific tab's agent.
+func (a *App) SteerForTab(tabID, text string) {
+	if ctrl := a.ctrlByTabID(tabID); ctrl != nil {
+		ctrl.Steer(text)
+	}
+}
+
 // Approve answers a pending approval_request by ID: allow runs the call, session
 // also remembers the grant for the rest of the session.
 func (a *App) Approve(id string, allow, session, persist bool) {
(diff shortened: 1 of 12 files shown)


JesonChou:
@SivanCola @esengine 大佬请审核下，已经根据1.5 rebase过

esengine (approved):
﻿Reviewed end to end. The agent-side design is right: steers consume at the top of each loop iteration, persist to the session (so history replay keeps them), and the post-final-answer `steerQueueLen() > 0 → continue` covers the race where guidance arrives as the model finishes. The controller fallback — converting to a fresh turn when the frontend's runningRef was stale — closes the other race. Bonus points for catching the unmapped `MCPSurfaceReady` wire name and extending the completeness test bound to `Steer`.

One note: the PR description still mentions the tool-result pairing alignment, but that part is no longer in the diff after the rebase (main-v2's `historyMessagesToItems` is untouched here) — the landed scope is steer only, which is fine by me.

Cache cost is one miss per steer and the comment says so explicitly — acceptable for an explicit user action. An agent-level test for queue→inject→event would be a nice follow-up, but the wire tests cover the contract that bit us before. Merging, thanks!

PR #3379, opened 2026-06-06
Feat/steering queue messages while reponse streams

wviana:
## Summary

Adds **mid-turn user steering** — when the model is busy executing tool calls, the user can type messages that are queued and injected into the next model call, allowing course correction without having to wait for the current turn to finish.

### New: Message queue during busy turns

Previously, when the model was busy (e.g. running a chain of shell commands or file edits), the prompt was disabled and user input was blocked. Now:

- **Type while busy** — messages typed during a turn are queued and displayed in a `QUEUE` indicator above the prompt
- **Esc to undo** — pressing Esc pops the last queued message back into the input buffer
- **Auto-dismiss** — queued messages expire after 5 seconds if not consumed
- **Injected at next iteration** — the loop drains the queue before the next model call, feeding steering messages as user turns the model sees before deciding on further tool calls

### New: Skip pending tool calls on steer

When a steering message arrives mid-turn, remaining tool calls in the current dispatch batch are skipped and the model is called fresh with the steering input — avoiding stale tool prompts after the user already changed direction.

### New: `user.queued` event in TUI history

Queued messages appear as user cards in the conversation history, keeping a complete record of steering interactions.

### New: `QueueIndicator` component

A compact indicator above the prompt showing how many messages are queued and a preview of the most recent one.

## Files changed

| File | Change |
|---|---|
| `src/loop.ts` | Added `_messageQueue` field, `queueMessage()` method, drain logic before each model call, skip-tools check in dispatch loop |
| `src/loop/types.ts` | Added `user.queued` event role |
| `src/core/eventize.ts` | Added `user.queued` case to produce cards |
| `src/cli/ui/App.tsx` | Queue-on-busy in `handleSubmit`, `user.queued` → `log.pushUser`, `messageQueue` state + clear-on-idle |
| `src/cli/ui/ComposerArea.tsx` | `QueueIndicator` rendering, `queueMessages` prop |
| `src/cli/ui/PromptInput.tsx` | `QueueIndicator` component |
| `src/cli/ui/hooks/useMessageQueue.ts` | New — queue management hook with pure helpers |
| `src/server/api/submit.ts` | Return `queued` flag in submit response |
| `src/server/context.ts` | Added `queued` to `SubmitResult` |
| `src/i18n/EN.ts`, `zh-CN.ts`, `types.ts` | `queuedSteerPending` string |
| `tests/loop-user-queue.test.ts` | New — 11 tests covering drain, FIFO order, skip-tools, multi-tool chains, streaming, TUI contract |
| `tests/use-message-queue.test.ts` | New — 16 pure-function tests |
| `tests/prompt-input-queue.test.tsx` | New — 18 tests for `QueueIndicator` + disabled prop |

## Testing

- 36 passing tests for the feature (loop-user-queue, use-message-queue, prompt-input-queue)
- Tested manually with echo + sleep tool chains up to 200 calls
- Verified steering with mid-turn course changes (e.g. "go back to 1/50 instead of continuing to 50")


esengine:
This is genuinely well-built 鈥?good test coverage, and skip-pending-tools-on-steer is the right call. Closing only because it targets the v1 (TypeScript) line, which is feature-frozen: npm `latest` is pinned-deprecated at 0.53, all installs point at `reasonix@next`, and nothing ships from v1 anymore. main-v2 already has the equivalent (the steer queue in `internal/agent` plus queue navigation from #3186). One UX idea here that main-v2 *doesn't* have: Esc popping the last queued message back into the input 鈥?if you'd like to port that against main-v2, we'd take it gladly.

PR #3418, opened 2026-06-07
fix(css): add missing closing brace that broke esbuild CSS compilation

CVEngineer66:
The `.onboarding__skip:disabled` block was missing its closing `}`, causing esbuild (Vite dev mode CSS compiler) to abort at line 7049. All rules after that point — ~1900 lines including `.workbench-dock`, `.context-panel`, `.context-inspector` layout positioning — were silently discarded, breaking the right-side dock layout.

Introduced as a merge conflict artefact between #2752 (onboarding overlay) and #3129 (InlineDiff component).

**Verification:**
- `pnpm build` — no more `Expected "}" to go with "{"` warning
- Build output now includes `.context-inspector,.workbench-dock{grid-row:2;grid-column:4;...}`

### DIFF
--- a/desktop/frontend/src/styles.css
+++ b/desktop/frontend/src/styles.css
@@ -7048,7 +7048,7 @@ a[href] {
 .onboarding__skip:disabled {
   opacity: 0.4;
   cursor: not-allowed;
-
+}
 
 /* InlineDiff: a compact before/after diff card. The header strip carries
    the filename, the +/- counts, and a Copy button. The body is a monospace


SivanCola (approved):
Approved. This is a well-scoped CSS syntax fix.

The missing closing brace is real: the base stylesheet leaves `.onboarding__skip:disabled` open, and this patch balances the rule so the following InlineDiff, dock, and context inspector styles are parsed normally again. I also checked the integration path against the latest `main-v2`; the no-commit merge was clean.

Security/cache review: this only changes one CSS rule terminator, with no dependency, runtime privilege, provider request, prompt-prefix, tool schema, or serialization impact.

Verified:
- `git diff --check` passed
- `cd desktop && wails generate module && pnpm --dir frontend build` passed
- Built CSS contains the downstream `.workbench-dock` / `.context-inspector` layout selectors

PR #3423, opened 2026-06-07
Reduce desktop stream render churn / 降低桌面端流式渲染卡顿

SivanCola:
## Summary
- lazily generate full-session Markdown/JSON only when the user copies or exports the session
- memoize transcript item rendering so live token updates only refresh the active assistant message slot
- move the frontend pnpm override into pnpm-workspace.yaml so frozen installs read the same override config as the lockfile

## Verification
- pnpm --dir desktop/frontend install --frozen-lockfile
- wails generate module && pnpm --dir frontend build
- go test ./... (from desktop/)

### DIFF
--- a/desktop/frontend/pnpm-workspace.yaml
+++ b/desktop/frontend/pnpm-workspace.yaml
@@ -1,2 +1,5 @@
 allowBuilds:
   esbuild: true
+
+overrides:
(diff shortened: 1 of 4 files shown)


SivanCola:
@codex review

PR #3436, opened 2026-06-07
fix(desktop): keep desktop-workspace pointer in sync across all navigation paths

drafish:
## Problem

The `desktop-workspace` file records which project is currently active. Two bugs cause it to become stale:

### Bug 1: `RemoveWorkspace` leaves a stale pointer

`RemoveWorkspace()` cleans up `desktop-workspaces.json` and `desktop-projects.json` but never clears `desktop-workspace`. When the removed project was the active one, a stale reference remains — subsequent sessions attach to a deleted project's directory, causing attachments and memory to land in the wrong place.

### Bug 2: `OpenProjectTab` never updates the pointer

`SwitchWorkspace` correctly calls `saveWorkspace(dir)`, but clicking an existing project in the sidebar goes through `handleOpenTopic` → `OpenProjectTab`, which never calls `saveWorkspace`. This means that after switching projects via the sidebar, `desktop-workspace` still points to whichever project was active at app startup.

**Combined effect:** If the user deletes project A (the startup-active one) and then switches to project B, `desktop-workspace` still contains the path to the deleted project A. Images pasted into the session save to the wrong `.reasonix/attachments/` directory.

## Reproduction

1. Add two projects (`project-a` and `project-b`)
2. Start Reasonix — `project-a` becomes active → `desktop-workspace` = `project-a`
3. Delete `project-a` from the project list
4. Click `project-b` in the sidebar to open it
5. `desktop-workspace` still contains `project-a` (stale)
6. Paste a screenshot → it saves to `project-a/.reasonix/attachments/` instead of `project-b`

### `desktop/app.go` — `RemoveWorkspace`
After deleting the project, check whether the removed dir matches the active pointer. If so:
- Fall back to the first remaining project (if any)
- Or clear the pointer entirely when no projects remain

### `desktop/tabs.go` — `OpenProjectTab`
Add `saveWorkspace(workspaceRoot)` so the pointer updates when opening any project tab, matching what `SwitchWorkspace` already does.

### `desktop/workspace.go` — new `clearWorkspace()`
Removes the `desktop-workspace` file so no stale reference persists.

## Tests

- `TestRemoveWorkspaceClearsActivePointerWhenRemovingCurrentWorkspace`
- `TestRemoveWorkspaceFallsBackToRemainingProject`
- `TestClearWorkspace`
- `TestOpenProjectTabUpdatesActiveWorkspacePointer`

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -1127,6 +1127,17 @@ func (a *App) RemoveWorkspace(dir string) error {
 	if err := removeProject(dir); err != nil {
 		return err
 	}
+	// If the removed workspace was the active one, clear the pointer
+	// so we don't leave a stale reference to a deleted project.
+	if loadWorkspace() == dir {
+		if remaining := loadProjectsFile(); len(remaining.Projects) > 0 {
+			// Fall back to the first remaining project
+			saveWorkspace(remaining.Projects[0].Root)
+		} else {
+			// No projects left; clear the active pointer entirely
+			clearWorkspace()
+		}
+	}
 	a.emitProjectTreeChanged()
 	return nil
 }
--- a/desktop/tabs.go
+++ b/desktop/tabs.go
@@ -300,6 +300,7 @@ func (a *App) OpenProjectTab(workspaceRoot, topicID string) (TabMeta, error) {
 	if abs, err := filepath.Abs(workspaceRoot); err == nil {
 		workspaceRoot = abs
 	}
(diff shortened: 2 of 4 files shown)


esengine (approved):
修掉两个真 bug：删除活动项目残留的 desktop-workspace 指针、侧栏切项目漏更新指针。fix 正确，tool/指针清理逻辑到位。

PR #3444, opened 2026-06-07
fix(desktop): render markdown in real time during streaming

HUQIANTAO:
## Problem

The assistant bubble renders **raw plain text** while the response is streaming and only switches to rendered markdown (headings, code blocks, lists, tables) once the turn completes. This creates a jarring visual jump — the user reads plain text for seconds, then the entire bubble reflows into formatted markdown.

## Root Cause

In `Message.tsx`, the streaming branch explicitly renders raw text to avoid "markdown reflow jitter":

```tsx
{item.streaming ? (
  <div className="msg__stream">
    {item.text}           {/* ← raw text, no markdown */}
    <span className="cursor" />
  </div>
) : (
  <Markdown text={item.text} />  {/* ← markdown only on completion */}
)}
```

The comment says partial markdown "reflows the layout and makes the view jitter." This was true before `useDeferredValue` was added to `<Markdown>` (commit f059d712), but now the deferred render handles this gracefully — React prioritises the cursor and layout frame over the expensive parse.

## Fix

Render `<Markdown>` during streaming. The component already uses `useDeferredValue` internally, so:

1. New tokens paint immediately (cursor moves, text grows)
2. The markdown parse runs in idle frames via `useDeferredValue`
3. If new tokens arrive mid-parse, React aborts the stale render and starts fresh
4. No layout jitter — the deferred render is invisible to the user's interaction frame

### Changes

| File | Change |
|------|--------|
| `Message.tsx` | Replace `{item.text}` with `<Markdown text={item.text} />` in the streaming branch; rename `.cursor` → `.msg__cursor` for BEM consistency |
| `styles.css` | Update `.msg__stream` from `white-space: pre-wrap` to `display: flex; flex-direction: column` so the cursor sits below the markdown block; add `.msg__cursor` with `margin-top: 4px` |

### Before / After

| | Before | After |
|---|--------|-------|
| During streaming | Raw text, no formatting | **Markdown rendered in real time** — headings, code blocks, lists appear as tokens arrive |
| On completion | Sudden reflow to markdown | No visual change — already rendered |
| Performance | Fast (plain text) | Same — `useDeferredValue` defers the expensive parse to idle frames |

## Verification

- `pnpm typecheck` ✅ clean
- `pnpm build` ✅ `built in 5.13s`
- No new dependencies
- No i18n changes
- Diff: 2 files, +12 / −10

### DIFF
--- a/desktop/frontend/src/components/Message.tsx
+++ b/desktop/frontend/src/components/Message.tsx
@@ -240,12 +240,13 @@ export const AssistantMessage = memo(function AssistantMessage({
       {hasText && (
         <div className="msg__body">
           {item.streaming ? (
-            // While streaming, render raw text (stable, monospace-free) instead of
-            // re-parsing markdown on every token — partial markdown reflows the
-            // layout and makes the view jitter. Markdown renders once, on completion.
+            // Render markdown in real time while streaming.  useDeferredValue
+            // inside <Markdown> lets React prioritise the cursor + layout frame
+            // over the expensive markdown parse — new tokens paint immediately,
+            // the formatted catch-up runs in idle frames.
             <div className="msg__stream">
-              {item.text}
-              <span className="cursor" />
+              <Markdown text={item.text} />
+              <span className="msg__cursor" />
             </div>
           ) : (
             <Markdown text={item.text} />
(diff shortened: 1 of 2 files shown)


HUQIANTAO:
Closing in favor of #3140, which has been rebased on the latest `main-v2` (7fba4ab9) with the new `ProcessCard`/`TurnActions` structure.  #3140 is MERGEABLE and CI is green.

This PR was a duplicate of the streaming markdown fix — both change the same lines in `Message.tsx`.  #3140 is the canonical version.

PR #3454, opened 2026-06-07
Refresh installed skills in active sessions / 刷新当前会话中新安装的 skills

SivanCola:
## Summary
- Rescan the live skill stores for controller skill listings and slash skill invocation so newly installed skills are visible in the active session.
- Refetch desktop slash commands when a turn finishes so installer tools can update the composer menu without a reload.
- Add regression coverage for dynamic skill discovery after controller creation.

## Verification
- go test ./...
- cd desktop && go test ./...
- cd desktop/frontend && pnpm run check:css
- cd desktop && wails generate module, then cd frontend && pnpm run typecheck

### DIFF
--- a/desktop/frontend/src/components/Composer.tsx
+++ b/desktop/frontend/src/components/Composer.tsx
@@ -170,9 +170,10 @@ export function Composer({
   insertRequest?: ComposerInsertRequest | null;
   disabled?: boolean;
   decisionPending?: boolean;
-  // ready/cwd re-trigger the command fetch: Commands() returns only built-ins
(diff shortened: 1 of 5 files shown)


esengine (approved):
Reviewed: live-scans the skill Store on demand so skills installed mid-session show up in the slash menu and `RunSkill`, while the system-prompt skill index stays frozen at boot (cache-stable). Nil-store fallback preserves prior behavior; complements the install_source meta tool (#3292). CI green incl. race + desktop build + 3-OS tests. LGTM.

PR #3462, opened 2026-06-07
Fix skill detail expand button / 修复技能详情展开按钮

### LINKED ISSUE #3460: [Bug]: 桌面端设置页无法展开技能详细描述
### Version line

### Exact version

1.3.0

### What happened?

桌面端的设置页，技能模块中的“展开”按钮点击无响应。
Image

### Steps to reproduce

1. 打开Desktop版reasonix
2. 打开设置
3. 选择“技能”
4. 尝试点击“展开”按钮
5. 点击按钮没有反应

### OS / platform

Windows 11

### Relevant logs or output

```shell

```

SivanCola:
## Summary
- make the skill card's lower expand/collapse affordance a real button
- wire that button to the same detail toggle handler as the skill card header
- preserve the existing visual treatment while adding hover and keyboard focus states

## Root cause
The visible "Expand" text under long skill descriptions was rendered as a plain `div`, so clicking it did not call the skill detail toggle handler. Only the card header button actually toggled the expanded state.

## Validation
- `pnpm --dir desktop/frontend build` passed after generating local Wails bindings

Fixes #3460

### DIFF
--- a/desktop/frontend/src/components/CapabilitiesPanel.tsx
+++ b/desktop/frontend/src/components/CapabilitiesPanel.tsx
@@ -1219,7 +1219,11 @@ function SkillRow({
         </Tooltip>
       </div>
       <div className="cap-skill-card__desc">{expanded ? skill.description : summary}</div>
-      {canExpand && <div className="cap-skill-card__more">{expanded ? t("common.collapse") : t("common.expand")}</div>}
+      {canExpand && (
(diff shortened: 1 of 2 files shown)


esengine (approved):
把展开文字从 div 换成真正的 button 并接上 onToggle，补了 aria-expanded 和键盘 focus 态。小而准。

PR #3471, opened 2026-06-07
fix(config): preserve LSP settings on config save

### LINKED ISSUE #3464: RenderTOMLForScope silently drops [lsp] / [lsp.servers] config section on Save
## Problem

`RenderTOMLForScope` in `internal/config/render.go` omits the `[lsp]` and `[lsp.servers.<lang>]` sections when serialising the config to TOML. Any user-defined LSP server overrides are silently discarded on every `SaveTo()` / `Save()` call.

## Scope

Every write path hits this:

- `reasonix setup` interactive wizard (`cli.go:610` -> `SaveTo`)
- Desktop settings panel saving any change (multiple mutators in `edit.go` -> `SaveTo`)
- MCP server add/remove (`edit.go:450-456` -> `SaveTo`)
- Any settings change (permissions, theme, etc.) that triggers a save

A user may change an unrelated setting (e.g. the theme) and unwittingly lose custom `[lsp.servers.lua]` entries without noticing.

## Root cause

`internal/config/render.go`

`RenderTOMLForScope` (line 30) renders these sections: `[ui]`, `[desktop]`, `[notifications]`, `[network]` + `[network.proxy]`, `[agent]`, `[[providers]]`, `[tools]`, `[codegraph]`, `[skills]`, `[permissions]`, `[sandbox]`, `[statusline]`, `[[plugins]]`.

**The `[lsp]` section is completely absent from the render output** -- the struct field `Config.LSP` (`config.go:58`) is loaded correctly but never serialised back.

The relevant types (`config.go:173-195`):

```go
type LSPConfig struct {
    Enabled bool                 `toml:"enabled"`
    Servers map[string]LSPServer `toml:"servers"`
}
type LSPServer struct {
    Command     string            `toml:"command"`
    Args        []string          `toml:"args"`
    Env         map[string]string `toml:"env"`
    LanguageID  string            `toml:"language_id"`
    Extensions  []string          `toml:"extensions"`
    InstallHint string            `toml:"install_hint"`
}
```

`render_test.go` also has no round-trip test covering the LSP section.

## Suggested fix

1. **`internal/config/render.go`**: Add `[lsp]` section rendering in `RenderTOMLForScope` (suggest inserting between `[codegraph]` and `[skills]`, or another natural position). Follow the same annotated-comment style as the existing `[codegraph]` block (lines 259-267). The output should cover:
   - `[lsp]` header with the `enabled` field
   - Iterate `c.LSP.Servers` to emit each `[lsp.servers.<lang>]` sub-table

2. **`internal/config/render_test.go`**: Add a round-trip test that verifies a TOML with `[lsp.servers]` content survives Load -> RenderTOML -> Load unchanged.

## Workaround

LSP server config placed in a project-local `reasonix.toml` is still loaded and honoured correctly (the `Load()` path is unaffected). However, avoid any `SaveTo()`-triggering operation (`reasonix setup`, desktop settings panel) until the fix lands -- otherwise the project config's LSP section will also be dropped by `RenderTOMLForScope`.

## Reproduction

1. Add to `~/.config/reasonix/config.toml`:

```toml
[lsp]
enabled = true

[lsp.servers.lua]
extensions = [".lua", ".script", ".gui_script"]
```

2. Run `reasonix setup` (or any operation calling `SaveTo`)
3. Inspect the file -- the `[lsp]` section is gone
4. Verify that LSP tools fall back to built-in defaults

## Environment

- Version: latest (main)
- Platform: all
- Config version: 2

GTC2080:
## Summary
- Render `[lsp]` and `[lsp.servers.<lang>]` in `RenderTOMLForScope` so `SaveTo` preserves user LSP overrides.
- Sort LSP server keys and quote non-bare language ids such as `c++` for deterministic, valid TOML.
- Add round-trip coverage for full, user, and project scoped renders.

## Root cause
`Config.LSP` was loaded correctly, but the annotated TOML renderer never serialized it back out. Any save operation rewrote the config without the LSP section.

## Validation
- `go test ./internal/config -count=1`
- `go test ./internal/config -run ^$ -bench BenchmarkRenderTOMLWithLSPServers -benchmem -count=5`
  - 64-server render: 132-153 us/op on Windows amd64 / i9-12900H
- `go vet ./...`
- `go test ./... -p 1 -count=1`
  - Local Windows note: all reached non-CLI packages passed; `internal/cli/TestACPFactoryLoadsSessionCwdProjectConfig` still fails at TempDir cleanup with a file-in-use error outside this render path.
- `go test ./desktop/... -count=1`
  - Local Windows note: unrelated desktop failures were observed in TempDir cleanup, skill path restoration, and escaped filename handling.

Closes #3464

### DIFF
--- a/internal/config/render.go
+++ b/internal/config/render.go
@@ -273,6 +273,8 @@ func RenderTOMLForScope(c *Config, scope RenderScope) string {
 	}
 	b.WriteString("\n")
 
+	renderLSPConfig(&b, c.LSP)
+
 	b.WriteString("[skills]\n")
 	if len(c.Skills.Paths) > 0 {
 		fmt.Fprintf(&b, "paths = %s   # extra custom skill roots\n", renderStringArray(c.Skills.Paths))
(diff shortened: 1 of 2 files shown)


esengine (approved):
根因定位准确——注释版 TOML renderer 漏序列化 LSP 段。修法干净，c++ 这类非裸 key 的引号处理也对，round-trip 三 scope 覆盖到位。

PR #3530, opened 2026-06-08
fix(chooser): separate Enter (confirm) from Space (toggle) in multi-s…

LVGG:
…elect

Multi-select mode now uses Space only for toggling individual options, while Enter (and number keys) confirms current selections and advances to the next question or submits, matching the single-select flow. Previously both keys toggled, making it impossible to commit without toggling the highlighted option.

The desktop AskCard already has this correct via explicit confirm buttons and was unaffected.


LVGG:
Fixes #3504

esengine (approved):
Good call separating Space (toggle) from Enter (confirm) in multi-select — verified the toggle path in handleChooserKey and the full cli suite is green.

Issue #3490, opened 2026-06-08
[Bug]: 多模型只规划不执行


guoke-yf:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.3.0

### What happened?

<img width="1020" height="7863" alt="Image" src="https://github.com/user-attachments/assets/120a2ed0-5851-4665-9819-41132a4efd7d" />

<img width="1073" height="639" alt="Image" src="https://github.com/user-attachments/assets/edb53c65-50ae-4102-85d4-4cf08da181bd" />

<img width="992" height="2881" alt="Image" src="https://github.com/user-attachments/assets/68df1d4e-6013-4d77-94d9-1302dafdf3e5" />

### Steps to reproduce

<img width="1073" height="639" alt="Image" src="https://github.com/user-attachments/assets/c473d73d-271f-4bdb-8db5-7bdef638fbe8" />

### OS / platform

win11

### Relevant logs or output

```shell

```

TheShinning:
同感，还容易崩，接入gpt做planner，v4做执行，会报错
报错：planner: paused after 6 tool-call rounds (agent.max_steps) — the work so far is saved; send another message to continue, or set max_steps higher or to 0 for no limit

SivanCola:
这里评论中提到的这条报错：

`planner: paused after 6 tool-call rounds (agent.max_steps)`

和 #3525 / #3467 / #3500 是同一个 planner 轮数上限问题，已经提交修复 PR：#3532。

修复点是新增独立的 `agent.planner_max_steps`（默认 12，设为 0 表示 planner 不限轮数），并让桌面设置页支持配置“执行轮数上限 / 规划轮数上限”。同时 planner 暂停提示会改为指向 `agent.planner_max_steps`，不再误导到 `agent.max_steps`。

如果本 issue 里除了这个 6 轮暂停外，还有“planner 完成后仍然没有进入执行器”的其他复现路径，那可能还需要单独继续排查。

SivanCola:
> 同感，还容易崩，接入gpt做planner，v4做执行，会报错 报错：planner: paused after 6 tool-call rounds (agent.max_steps) — the work so far is saved; send another message to continue, or set max_steps higher or to 0 for no limit

这其实是当时一个友好的设计，防止plan无限plan下去...看来是误伤了，已经修复

SivanCola:
Follow-up: opened #3541 for the handoff-confusion part of this issue.

[...]

PR #3541, opened 2026-06-08
fix(agent): prevent executor handoff role confusion / 防止执行器交接角色混淆

SivanCola:
Refs #3490

## Summary
- strengthen the planner-to-executor handoff so executor models ignore planner-only limitations
- add an executor handoff guard that retries when the executor answers as the planner instead of using tools
- cover the Chinese planner-style confusion case with a coordinator regression test

## Testing
- `go test ./internal/agent ./internal/boot ./internal/config ./internal/cli`
- `cd desktop && go test . -run 'TestSettings|TestSetAgent|Test.*Settings|TestProviderViewFromEntry'`
- `npm run check:css && npm run test:typecheck`

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -31,6 +31,7 @@ const maxToolOutputBytes = 32 * 1024
 const maxFinalReadinessBlocks = 3
 const maxEmptyFinalBlocks = 3
 const maxStreamRecoveries = 1
+const maxExecutorHandoffConfusionBlocks = 2
 
 // Renderer redraws the assistant's final-answer text as styled output. It is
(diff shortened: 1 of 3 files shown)


esengine (approved):
Solid fix for the planner→executor confusion (#3490), with a regression test. I tightened the confusion matcher to role-identity phrases only — bare permission claims like "no write access" / "没有写入权限" were also matching legitimate blocker reports (the exact vocabulary the handoff prompt asks the executor to use), which could hard-error a genuinely-blocked task. Added a regression test for that. Full `go test ./internal/agent` is green.

Issue #3498, opened 2026-06-08
[Bug]: 为什么一直让我确认，我点击本会话允许或者 全部允许了，还是一直让我确认


mailes:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.3.0

### What happened?

这个没人遇到吗。一直让我确认。windows 和 mac 都存在

### Steps to reproduce

1.开启会话
2.遇到写操作，我点击同意或者本会话同意，一直还是讯问是否允许
3.好像无法保存这个状态一样。

### OS / platform

windows 和 Mac

### Relevant logs or output

```shell

```

yuehaotan-ship-it:
进入死循环了吧

mailes:
> 进入死循环了吧

好像也不是，这个状态好像无法保存一样。

flyskyxie-droid:
感觉新版的bug太多了

SivanCola:
Fixed in #3549.

The PR changes approval grants to visible permission rules (`Bash(...)`, `Bash(...:*)`, and `Edit(...)`) and stores session/persistent approvals as rules instead of ad-hoc labels. Safe Bash commands now offer `Bash(command:*)` session/persistent grants so similar commands in the same scope do not repeatedly prompt; file mutation tools share `Edit` session grants.

Verified with:
- `go test ./internal/permission ./internal/control ./internal/acp ./internal/i18n ./internal/serve ./internal/config ./internal/boot ./internal/cli`
- `npm --prefix desktop/frontend run check:css`
- `xmllint --noout docs/assets/approval-authorization-mocks.svg`
- `git diff --check`

PR #3549, opened 2026-06-08
Fix approval scope reuse / 修复审批授权范围复用

SivanCola:
## Summary
- Represent remembered approvals as explicit permission rules such as `Bash(...)`, `Bash(...:*)`, and `Edit(...)`, while keeping legacy rule compatibility.
- Add scoped session and persistent approval handling across desktop, CLI TUI, ACP, and serve UI so similar safe Bash commands and file mutations do not repeatedly prompt.
- Persist approval rules into project config with coverage/deduplication notices, and update docs plus approval mock artifacts.

## Testing
- `go test ./internal/permission ./internal/control ./internal/acp ./internal/i18n ./internal/serve ./internal/config ./internal/boot ./internal/cli`
- `npm --prefix desktop/frontend run check:css`
- `xmllint --noout docs/assets/approval-authorization-mocks.svg`
- `git diff --check`

Addresses #3498


esengine (approved):

PR #3563, opened 2026-06-08
fix:解决Termux下无法呼出手机键盘问题

nostalgia296:


esengine (approved):
Verified build; disabling cell-motion mouse mode to restore the Termux soft keyboard is the right tradeoff.

PR #3568, opened 2026-06-08
fix(desktop): restore maximized window from tray

### LINKED ISSUE #3475: [Bug]: windows, 最大化下关闭窗口，点击托盘图标，不能恢复最大化窗口
### Version line

### Exact version

1.3

### What happened?

[Bug]: windows, 最大化下关闭窗口，点击托盘图标，不能恢复最大化窗口

### Steps to reproduce

最大化
点击托盘图标

### OS / platform

win

### Relevant logs or output

```shell

```

wade19990814-hue:
## What

Restore a maximized desktop window to its maximized state after it is hidden to the system tray.

Fixes #3475.

## Why

The normal startup path restores `WindowState.Maximised`, but the tray restore path only called `WindowShow` and `WindowUnminimise`. On Windows, closing a maximized window to the tray and reopening it therefore restored a non-maximized window.

## How

- Record whether the window was maximized immediately before background hide.
- Consume that one-shot state when the window is reopened.
- Re-maximize after show/unminimise on Windows and Linux.
- Keep macOS application-level hide/show behavior unchanged.
- Route tray open through the existing main-window restore path.

## Testing

- `go test . -run 'TestBackground(CloseHideStrategyByPlatform|RestoreMaximiseStrategy)|TestTrayMenuLabelsFollowLocale|TestBeforeCloseAllowsSystemQuitWhenBackgroundCloseEnabled' -count=1`
- `wails build` (Windows amd64)

`go test ./...` still has three unrelated existing Windows failures: a locked TempDir cleanup, skill path state, and a test filename containing a double quote that Windows rejects.

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -318,15 +318,30 @@ func showFromBackground(ctx context.Context, wasMaximised bool) {
 	if backgroundCloseUsesApplicationHide(goruntime.GOOS) {
 		runtime.Show(ctx)
 	}
-	runtime.WindowShow(ctx)
-	runtime.WindowUnminimise(ctx)
-	if backgroundRestoreShouldMaximise(goruntime.GOOS, wasMaximised) {
+	plan := backgroundRestorePlanFor(goruntime.GOOS, wasMaximised)
+	if plan.maximiseBeforeShow {
 		runtime.WindowMaximise(ctx)
+	}
+	runtime.WindowShow(ctx)
+	if plan.unminimiseAfterShow {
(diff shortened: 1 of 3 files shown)


wade19990814-hue:
Windows manual verification passed after the follow-up fix: maximized window -> close to tray -> reopen from tray restores directly as maximized, without the normal-size window flash.

esengine (approved):
Verified: background-restore strategy tests pass; nice fix for the maximized-window flash on tray restore.

PR #3572, opened 2026-06-08
feat: add session PDF and image export

SivanCola:
## Summary
- Add export menu entries for PDF and PNG image alongside Markdown and JSON
- Render session Markdown into a dedicated export surface for PDF/image output
- Use native save dialogs so users can choose the export path before files are generated or written

## Testing
- go test -run '^$' . (desktop)

## Notes
- Full `go test ./...` under `desktop/` was not used as a gate because existing `TestSetEffortPersistsAndAutoClears` fails on the current branch with provider effort persistence output unrelated to this change.

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/rand"
+	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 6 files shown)


esengine (approved):
Really nice — rolling the PDF writer by hand (raster XObject + per-page CM offset for pagination) and rasterizing via SVG foreignObject instead of pulling in html2canvas + jspdf keeps us dependency-free, which is exactly the right call for this project. The AppBindings interface + mock are both updated, both locales are in sync, safeExportFilename guards path traversal, and the honest note about the unrelated pre-existing test failure is appreciated.

Merging. Two follow-ups worth a quick pass later (neither blocks):

1. **Surface export failures to the user.** The whole export is wrapped in `catch (err) { console.error(...) }`, so on a user-initiated action a failure is invisible — they click Export and nothing happens. A small notice/toast on failure would fit our "don't silently swallow" stance.

2. **Verify the foreignObject path on older WebKit + KaTeX.** Rendering a cloned surface through `<foreignObject>` to canvas is known to go blank or drop external resources on older Safari/WebKit (we've been bitten on macOS ≤12 before), and KaTeX's woff2 fonts referenced from the inlined CSS won't be fetched inside the SVG sandbox, so math may fall back to box glyphs. Worth a spot-check on a mac build and, if it's an issue, embedding the KaTeX font as a data URL.

Thanks!

esengine:
Approved and auto-merge is armed, but GitHub now shows a conflict — a couple of sibling desktop PRs (#3886, #3888) just landed and touched App.tsx and the locale files this PR also edits. Could you rebase onto the latest main-v2? It should merge automatically once the conflict is resolved. Thanks!

PR #3577, opened 2026-06-08
feat(desktop): support external file drops as attachment cards / 桌面端支持外部文件拖拽附件卡片

SivanCola:
## Summary
- support dropped external files in the chat composer by keeping the native Wails file-drop path wired into attachment handling
- render dropped/pasted attachments in user messages as compact file cards instead of inline `[file:...]` text
- include workspace file/folder references in the same display system while preserving raw `@...` refs for model submission
- dedupe pasted and dropped attachments in the composer, with removal restoring the ability to attach again

## Testing


esengine (approved):
Verified: frontend typecheck clean + check:css passes. Clean attachment-card feature — hash+source dedup with proper add/forget/clear lifecycle, CJK-aware ref parsing, image/file/folder classification. Drag-drop itself is best smoke-tested on canary but the logic is sound.

PR #3588, opened 2026-06-08
feat: generative background music & notification sound system

ttmouse:
### 生成式背景音乐
- Token 驱动的实时音乐引擎，每 4 个 AI Token 触发一个音符
- C Major 五声音阶，听感舒适不干扰
- 4 种音色预设：Classic (三角波)、Ethereal (正弦波, 默认)、Digital (方波)、Retro (锯齿波)
- 设置面板可切换预设、预览、关闭

### 通知音效
- Success 音效：AI 完成生成时播放
- Attention 音效：AI 提问时播放
- 各 5 种音效可选：合成音 + 4 种 WAV 音效素材
- 设置面板即时选择 + 预览

### 后端改动
- 新增 agent:turn-done 事件，前端据此触发成功音效
- 新增 onTurnDone() bridge 方法


esengine (changes requested):
The note engine is competent and this can land as an opt-in extra, but defaults and assets block it:

1. **Everything must default OFF**: `readPresetPref()` falls back to `"ethereal"` (music plays out of the box), and `SoundWavPref` has no `"off"` member — the turn-done and AskCard chimes can't be disabled at all.
2. `agent:turn-done` is emitted without a tab id and the handler plays for *any* tab — background tabs chime over the active one. Scope it; same for the per-token note subscription.
3. The 4 embedded WAVs are ~1.5 MB inside the binary for every user. Convert/trim to OGG (~100–150 KB) or go synth-only, and add the Mixkit license attribution.
4. Prefs belong in desktop settings (config) like every other setting, not localStorage — they currently don't sync and reset per WebView profile.
5. Comment policy: the Chinese file-header essays and `──` section banners need to go (see CLAUDE.md).

Plus a rebase onto the post-#3752 SettingsPanel.

PR #3603, opened 2026-06-08
fix(desktop): restore project to sidebar when resuming session from history, show toast on failure

CVEngineer66:
### Bug 1: Sessions from a removed project cannot restore the project back to the sidebar

After removing project B from the sidebar, its sessions still appear in the history panel. Clicking to open one -- B does not come back to the sidebar. The user thinks nothing happened and clicks again -- still nothing.

**Root cause**: OpenProjectTab never calls addProject() to write the project back into desktop-projects.json, nor does it notify the frontend to refresh the project tree.

### Bug 2: Session-open failure silently falls back to the wrong tab

Clicking a session whose project no longer exists -- the history panel closes, and the session appears on the current active tab (an unrelated project A). The user has no idea what happened.

**Root cause**: openProjectTab and openGlobalTab swallow all exceptions with catch { return undefined }. The caller passes undefined to resumeSession(path, undefined), which falls back to activeTabId -- loading the data onto the wrong tab with zero feedback.

## Fix

| File | Change |
|---|---|
| desktop/tabs.go | OpenProjectTab now calls addProject() + emitProjectTreeChanged() -- restores the project to the sidebar and refreshes the tree |
| useController.ts | openProjectTab/openGlobalTab no longer use catch { return undefined } -- exceptions propagate to the caller |
| App.tsx | onResumeSession wraps in try-catch: on failure, shows a toast |
| lib/toast.tsx | New ToastProvider + useToast() -- reusable notification system (3 levels, 2.5s auto-dismiss, stackable, click-to-dismiss) |


esengine (approved):

PR #3605, opened 2026-06-08
fix(desktop): honor user font in UI chrome

jiaxindeyang-a11y:
## What
Update desktop UI chrome text to follow the user's selected font family instead of forcing the monospace stack in ordinary interface copy.

This covers visible chrome such as:
- composer metadata controls
- status bar text and model/effort switcher controls
- todo bar title/count
- settings sidebar descriptions
- run status text

Technical text surfaces remain monospace: code, diffs, tool output, file paths, hashes, logs, and other identifier-heavy content.

## Why
The Appearance -> Font setting already updates the app's `--sans` stack, but several user-facing labels were pinned to `--mono`, making them visually inconsistent with composer placeholder/input text such as "给 Reasonix 发消息...".

## Testing
- `npm.cmd run check:css`
- `npm.cmd run test:all`
- `npx.cmd tsc --noEmit -p tsconfig.test.json`
- `git diff --check`

Note: I attempted to start the Vite dev server for visual QA, but this Windows environment has duplicate `Path`/`PATH` environment entries that caused PowerShell process launching to fail. The CSS and frontend test checks above passed.

### DIFF
--- a/desktop/frontend/src/styles.css
+++ b/desktop/frontend/src/styles.css
@@ -2440,7 +2440,7 @@ a[href] {
   padding: 5px 14px 6px;
   overflow: hidden;
   color: var(--fg-dim);
-  font-family: var(--mono);
+  font-family: var(--sans);
   font-size: 12px;
 }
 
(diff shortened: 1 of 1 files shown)


jiaxindeyang-a11y:
Small note on the CSS font-family choices:

Parent UI containers use `var(--sans)` to follow the selected app font, while nested controls inherit from their local container to avoid reintroducing monospace styling.

esengine (approved):
Reviewed: this is a genuine bug-fix (not a cosmetic UI change) and CI is green. Approving.

PR #3611, opened 2026-06-08
feat(CLI):优化Termux体验，解决在termux上无法移动聊天上下文等问题

nostalgia296:


esengine (approved):
Thanks for improving the Termux experience — the context-movement fix and TUI adjustments for that environment are appreciated, and the added chat_tui/cli tests keep it covered.

PR #3613, opened 2026-06-08
fix(desktop): preserve curated provider model lists

### LINKED ISSUE #3465: bug: auto-refresh models overwrites user-curated model list on every Settings open
## Problem

After using auto-fetch to get available models from a provider and then manually deleting unwanted models, the deleted models reappear every time the user opens the Settings panel (or restarts and opens Settings).

## Root Cause

**File:** `desktop/frontend/src/components/SettingsPanel.tsx`, lines 666–692

The auto-refresh `useEffect` fires **every time the "usage" subtab renders** and **unconditionally replaces** the user's curated model list with the full API response:

```tsx
useEffect(() => {
    if (subtab !== "usage") return;
    // ...
    void backgroundApply(async () => {
      for (const { provider } of candidates) {
        const fetched = await app.FetchProviderModels(provider);
        // ↓↓↓ ALWAYS overwrites with full fetched list ↓↓↓
        await app.SaveProvider({ ...provider, models: fetched, default: currentDefault });
      }
    });
}, [backgroundApply, s.providers, subtab, t]);
```

### Exact bug sequence

1. User adds a provider → auto-fetch saves all models ✅
2. User edits provider, removes unwanted models, saves → curated list written to `config.toml` ✅
3. User restarts app → config loads correctly with curated list (boot path does NOT fetch) ✅
4. **User opens Settings → "usage" subtab** → `useEffect` fires, `autoRefreshKeyRef` has reset after restart → fetches ALL models from API → **overwrites curated list via `SaveProvider`** ❌

The `autoRefreshKeyRef` guard only prevents re-fetching within the same session. After any restart or config reload, the ref resets and auto-fetch runs again.

### Additional auto-fetch paths with the same issue

- **`saveKeyEnvAndAutoRefresh`** (lines 1064–1097): After saving an API key, immediately fetches and overwrites.
- **`refreshModels`** (lines 1024–1056): Manual "Refresh models" button also does a full replacement (this one is arguably intentional).

## Suggested Fix

**Option A — Only merge new models (never remove):**
```tsx
const merged = [...new Set([...provider.models, ...fetched])];
// Only save if there are actually new models
if (!sameStringLists(provider.models, merged)) {
  await app.SaveProvider({ ...provider, models: merged, default: currentDefault });
}
```

**Option B — Track user curation with a flag:**
Add a `modelsUserEdited: boolean` field to `ProviderView`/`ProviderEntry`. When the user manually edits the model list, set it to `true`. Auto-refresh skips providers where `modelsUserEdited === true`.

**Option C — Remove auto-refresh, keep manual only:**
Remove the auto-refresh `useEffect` entirely. Users explicitly click "Refresh models" when they want to discover new ones. This is the simplest fix and avoids the issue entirely.

### LINKED ISSUE #3506: [Feature]: 接入第三方模型时不要自动获取模型，订阅第三方的coding plan设置好几个模型后，自动获取模型后会打乱现有模型列表
### Version line

Image

### Proposed solution

_No response_

### LINKED ISSUE #3529: [Bug]: 不是你们为什么总是刷新自定义接入服务商的模型列表呢？
### Version line

### Exact version

v1.3.0

### What happened?

我就需要自定义接入服务商中的所有模型列表中的两个，你们怎么每次都自动刷新更新显示全部的模型列表呢？

### Steps to reproduce

111

### OS / platform

macoS

### Relevant logs or output

```shell

```

closss:
## Summary
This fixes the desktop provider model refresh flow so fetched model lists no longer overwrite a user's curated model list.

Before this change:
- Background model discovery on the Settings usage tab replaced saved provider models with the full API response.
- Saving an API key and auto-refreshing could also replace the saved list.
- Manual refresh replaced the list instead of preserving existing curated entries.

After this change:
- Background and API-key auto-refresh preserve a non-empty saved model list, and only populate providers that do not have models yet.
- Manual refresh appends newly discovered models while keeping existing entries intact.
- The provider default model is preserved when still available, otherwise it falls back to the first saved model.

## Issues
Fixes #3465
Fixes #3506
Fixes #3529

## Implementation notes
- Extracted provider model merge/default selection logic into `desktop/frontend/src/lib/providerModels.ts`.
- Reused that helper in the three model-refresh paths in `SettingsPanel.tsx`.
- Added focused tests for manual merge behavior, background curation preservation, empty-list population, and default fallback.
- Kept the patch scoped to desktop settings model-refresh behavior.

## Tests
- `npm run test:all`
- `git diff --check`
- `gofmt -l`
- `go test ./...`

### DIFF
--- a/desktop/frontend/package.json
+++ b/desktop/frontend/package.json
@@ -10,9 +10,9 @@
     "check:css": "node scripts/check-css-syntax.mjs src/styles.css && node scripts/check-z-index-tokens.mjs src/styles.css",
     "test:todo-visibility": "node scripts/test-todo-visibility.mjs",
     "typecheck": "tsc --noEmit",
-    "test": "tsx src/__tests__/math-golden.test.ts && tsx src/__tests__/text-size.test.ts",
+    "test": "tsx src/__tests__/math-golden.test.ts && tsx src/__tests__/text-size.test.ts && tsx src/__tests__/provider-model-refresh.test.ts",
     "test:typecheck": "tsc --noEmit -p tsconfig.test.json",
-    "test:all": "tsc --noEmit -p tsconfig.test.json && tsx src/__tests__/math-golden.test.ts && tsx src/__tests__/text-size.test.ts"
+    "test:all": "tsc --noEmit -p tsconfig.test.json && tsx src/__tests__/math-golden.test.ts && tsx src/__tests__/text-size.test.ts && tsx src/__tests__/provider-model-refresh.test.ts"
   },
   "dependencies": {
     "@tanstack/react-virtual": "^3.14.2",
--- a/desktop/frontend/src/__tests__/provider-model-refresh.test.ts
+++ b/desktop/frontend/src/__tests__/provider-model-refresh.test.ts
@@ -1,0 +1,57 @@
+// Run: tsx src/__tests__/provider-model-refresh.test.ts
+
+import { mergedFetchedProviderModels, providerDefaultModel } from "../lib/providerModels";
+
+let passed = 0;
+let failed = 0;
+
+function eq(a: unknown, b: unknown, label: string) {
+  if (JSON.stringify(a) === JSON.stringify(b)) {
+    process.stdout.write(`  PASS  ${label}\n`);
+    passed += 1;
+  } else {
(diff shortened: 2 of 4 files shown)


esengine (approved):
Full CI matrix is green after the re-trigger, and the model-list merge logic reviewed cleanly (preserve-curated vs union merge, with a focused test). Approving.

PR #3637, opened 2026-06-09
fix(desktop): extend diff row background on horizontal scroll

### LINKED ISSUE #3599: [Bug]: 文本编辑显示框删除新增颜色渲染不全
### Version line

### Exact version

1.3.0

### What happened?

如图所示
Image

Image

### Steps to reproduce

这个应该比较长的代码修改的时候就会这样吧

### OS / platform

Windows 11

### Relevant logs or output

```shell

```

whale-fall-ouo:
Add min-width: min-content to .diff__row and .inline-diff__row so background colors cover the full line width when horizontal scrolling is active.

Fixes #3599

### DIFF
--- a/desktop/frontend/src/styles.css
+++ b/desktop/frontend/src/styles.css
@@ -1934,6 +1934,7 @@ a[href] {
   gap: 8px;
   padding: 0 12px;
   white-space: pre;
+  min-width: min-content;
 }
 .diff__row--add {
   background: var(--add-bg);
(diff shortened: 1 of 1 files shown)


whale-fall-ouo:
修改前修改前
修改后
修改后

esengine (approved):
Correct fix — in a horizontally scrolling container the flex row gets clamped to the container width, so the add/del background stopped short of the scrolled content. min-content is exactly the right tool. Thanks!

PR #3638, opened 2026-06-09
fix(control): don't auto-answer 'ask' in yolo/bypass mode

### LINKED ISSUE #3624: [Bug]: yolo模式下ask选项跳过
### Version line

### Exact version

v1.3.0

### What happened?

yolo模式下ask选项自动选择第一个选项，实际用户可能看都没看清它就自动执行第一个选项往下走了。

### Steps to reproduce

修复yolo模式下ask跳过问题，允许调动工具代码与ask选择代码隔开，单独呈现，不被yolo模式覆盖

### OS / platform

Windows11

### Relevant logs or output

```shell

```

warvyvr:
The bypass flag was designed to skip permission approvals (requestApproval), but was incorrectly applied to Controller.Ask() — where the model asks the user a structured question, not where it asks permission. This meant yolo mode silently auto-selected the first option without the user ever seeing the prompt.

Fix: remove both bypassEnabled() checks from Ask(). The function now always emits an AskRequest and waits for AnswerQuestion(), even under bypass. The second check (after promptMu.Lock()) was guarding a race where bypass was turned on while Ask was queued behind another prompt — that race no longer matters because bypass no longer suppresses Ask.

Keep recommendedAskAnswers() as dead code with a doc comment explaining its history, for reference.

Update tests:
- TestBypassAutoAnswersAskWithRecommendedOptions → TestBypassDoesNotAutoAnswerAsk verifies Ask emits AskRequest under bypass and returns the user's actual choice
- TestBypassRecheckedForAskAfterPromptLock → TestAskSerializesBehindPromptLockEvenWithBypass verifies Ask serializes behind promptMu and still emits AskRequest

Fixes #3624


SivanCola (changes requested):
Thanks for the fix — the production change matches the reported YOLO/ask bug, and the targeted tests pass locally.

Could you tighten `TestAskSerializesBehindPromptLockEvenWithBypass` before merge? Right now it locks `promptMu`, starts `Ask`, sleeps for 20ms, then releases the lock and waits for `AskRequest`. That verifies the request is eventually emitted under bypass, but it does not actually prove `Ask` serialized behind `promptMu`: if a future regression emitted `AskRequest` before taking the lock, this test could still pass because the buffered `askCh` would simply be read after unlock.

Please add a pre-unlock assertion that no `AskRequest` is observed while `promptMu` is still held, then unlock and assert the request is emitted. That would make the test cover the serialization behavior its name/comment describes.

Also noted while validating: `go test ./internal/control` currently fails locally in `TestRemoveMCPServerRemovesUnconnectedLazyPlaceholder` because my environment has configured MCP servers, but the changed yolo/ask tests and `internal/agent` ask tests pass.

warvyvr:
@SivanCola as your request, I change testcase, please check it out, for any further consideration, please let me know.

esengine:
Thanks @warvyvr — your behavior tests (TestBypassDoesNotAutoAnswerAsk / TestAskSerializesBehindPromptLockEvenWithBypass) are great and are carried into #3712. Closing in favor of that, which keeps your tests + the minimal fix but drops the unrelated files this PR picked up (wow.md, README.md, .gitignore). Credited via Co-authored-by. Closes the same #3624.

PR #3658, opened 2026-06-09
fix(serve): restore query helper

### LINKED ISSUE #3645: Bug: `$` is not defined — Reasonix Web UI (serve mode) 页面交互完全失效
根因：internal/serve/index.html 中定义了 const $$ = s => document.querySelectorAll(s)，但缺少了 const $ = s => document.querySelector(s)。第 643 行首次调用 $('#log') 即抛出 ReferenceError，后续所有 JS 中断执行。
影响版本：v1.3.0 ✅ / v1.4.0 ✅
修复方式：在 const $$ 前插入一行 const $ = s => document.querySelector(s);
diff
+ const $ = s => document.querySelector(s);
  const $$ = s => document.querySelectorAll(s);
受影响行（v1.4.0）：internal/serve/index.html 第 642-643 行

paradoxSCH:
## Summary
- restore the missing `$` query helper in the embedded serve-mode web UI
- add a regression test that ensures both `$` and `$$` helpers are defined before the page uses them

Fixes #3645

## Tests
- `go test ./internal/serve -run TestServeIndexDefinesQueryHelpers -count=1`
- `go test ./internal/serve -count=1`

## Notes
- local commit/push hooks currently fail because they try to run npm from the repository root where `package.json` is absent; the relevant Go tests above were run manually

### DIFF
--- a/internal/serve/index.html
+++ b/internal/serve/index.html
@@ -639,6 +639,7 @@ document.addEventListener('DOMContentLoaded', () => {
 
 
 
+const $ = s => document.querySelector(s);
 const $$ = s => document.querySelectorAll(s);
 const log = $('#log'), input = $('#in'), btnSend = $('#btn-send'), btnStop = $('#btn-stop');
 const statusDot = $('#status-dot-footer'), statusText = $('#status-text');
--- a/internal/serve/serve_test.go
+++ b/internal/serve/serve_test.go
@@ -208,6 +208,18 @@ func TestServeIndexPage(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


esengine (approved):
Verified end-to-end. Reproduced #3645: the embedded serve UI references $ 51 times starting at `const log = $('#log')` but never defines it (only $$ was defined), so the page throws `ReferenceError: $ is not defined` and the whole script dies. With the one-line helper restored, the regression test and the full serve package pass. Thanks for the fix and the test.

PR #3659, opened 2026-06-09
fix(agent): collapse newlines in session preview

umarahzamy:
## Problem

`reasonix chat --resume` shows a raw-terminal picker (`selectOne`). When a session's first message spans multiple lines (e.g. contains `\n`), the embedded newline breaks the single-row-per-item assumption of the picker's cursor tracking. Pressing ↑/↓ leaves stale rows visible — looks like row duplication.

## Root cause

`previewSession()` in `internal/agent/save.go` extracts the first user message as a preview string but never sanitizes internal newlines. The raw-mode CLI renderer (`selectOne`) uses `\033[NA` cursor-up escape, which only works when every item is exactly **one** physical terminal line. An embedded `\n` makes one item occupy 2+ lines, breaking cursor repositioning.

## Fix

`internal/agent/save.go` — 3 lines added in `previewSession()`:

```go
s = strings.ReplaceAll(s, "\r\n", " ")
s = strings.ReplaceAll(s, "\r", " ")
s = strings.ReplaceAll(s, "\n", " ")
```

Collapses all line breaks into spaces before the existing 80-rune truncation, so every preview is a clean single line. All consumers of `SessionInfo.Preview` inherit the fix — the CLI `--resume` picker, the TUI `/resume` overlay, branch tree titles, and the desktop app's history panel.

### DIFF
--- a/internal/cli/chat_tui.go
+++ b/internal/cli/chat_tui.go
@@ -1017,6 +1017,7 @@ func (m chatTUI) update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			if m.state == tuiRunning {
 				line := strings.TrimSpace(m.input.Value())
 				if line == "" {
+					m.viewport.GotoBottom()
 					return m, nil
 				}
 				if m.queueEditCursor >= 0 && m.queueEditCursor < len(m.pendingInterject) {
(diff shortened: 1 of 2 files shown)


umarahzamy:
reasonix chat --resume row duplicates when navigating with arrow keys 
image


expected behaviour, arrow navigation works witout causing row duplication
image

PR #3669, opened 2026-06-09
fix(setup): align custom provider base URL between wizard probe and chat client

CnsMaple:

**Problem**
The setup wizard probes custom providers with the user's `base_url`, but the
chat client later normalises the URL through the OpenAI/Anthropic shape
detection (adds `/v1` if missing). The two paths disagreed: a wizard probe
that succeeded against `https://api.example.com` could fail at chat time, or
vice versa.

**Fix**
The setup wizard now normalises `base_url` the same way the chat providers do
(`/v1` suffix added if missing, trailing slashes stripped) before using it
for the probe and persisting it. `internal/provider/anthropic` mirrors the
same normalisation in its `BaseURL` getter so both legs of the
wizard-to-chat pipeline agree on the URL shape.

**Changes**
- `internal/cli/cli.go` (+41/-1): normalise custom-provider base_url in the setup wizard.
- `internal/cli/cli_test.go` (+85): table-driven coverage for the normalisation cases.
- `internal/provider/anthropic/anthropic.go` (+17/-1): consistent BaseURL normalisation.
- `internal/provider/anthropic/anthropic_test.go` (+40): sub-tests for `/v1`, trailing slash, path prefix, empty fallback.

**Tests**
- `go test ./internal/cli/ ./internal/provider/anthropic/ -count=1 -run 'SetupBaseURL|CustomProvider|BaseURL'`
  → all new tests pass.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -12,6 +12,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"log/slog"
 	"net/url"
 	"os"
 	"os/signal"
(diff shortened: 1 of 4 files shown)


esengine (approved):
Verified end-to-end. Traced that the Anthropic client posts to `c.baseURL+"/v1/messages"`, so a base URL ending in /v1 produced /v1/v1/messages (404). Round-trip test against a mock server confirms: before the fix the request hits /v1/v1/messages, after it hits /v1/messages, and the default root is unaffected. anthropic + cli packages pass. The wizard probe now walking the same candidate URLs closes the gap with the chat client. Thanks.

PR #3670, opened 2026-06-09
fix(tui): complete PR #3051 — Esc must not exit plan mode; Ctrl+C keeps draft

CnsMaple:

**Follow-up to #3051**

PR #3051 (merged) removed the Esc-from-YOLO branch but left the
Esc-from-plan-mode branch in place. In practice that means once the user
entered plan mode (`/plan`, Shift+Tab), a single Esc would silently flip
them back to the less-permissive "normal" mode — exactly the surprise that
#3051 set out to eliminate for YOLO. Mode toggles are now exclusively
driven by Shift+Tab, matching how YOLO is treated.

While in the file, the idle Ctrl+C branch was also reordered: an active
text selection now takes precedence over the composer-clear / double-press
quit gestures, so Ctrl+C stops silently wiping a half-typed draft just
because the user happened to have a selection.

**Changes**
- `internal/cli/chat_tui.go` (+26/-20): drop the `m.planMode` arm from the
  Esc handler; comment updated to spell out the Shift+Tab-only mode rule.
- `internal/cli/chat_tui_test.go` (+59): regression coverage for Esc in
  plan mode and Ctrl+C-with-active-selection.

**Tests**
- `go test ./internal/cli/ -count=1 -run 'TestEsc|TestCtrl|TestPlan'`
  → all pass.

**Out of scope**
- YOLO behaviour (already in #3051).


CnsMaple:
Re-running CI via empty commit (gh run rerun requires admin rights).

esengine (approved):
Verified end-to-end. Reverting each hunk to base behavior fails the two new tests: Esc flips plan mode off (TestEscInPlanModeDoesNotExitPlan) and Ctrl+C wipes a draft when a selection exists (TestCtrlCCopyBeatsClearInput). With the fix both pass and the full cli package is green. This correctly completes the plan-mode half that #3051 missed and makes selection-copy take precedence over clear-input. The earlier Windows CI red was the unrelated TestRunShell_CancelStopsCommand timing flake, green on re-run after updating the base. Thanks.

PR #3671, opened 2026-06-09
fix(model): persist /model selection to user config.toml

CnsMaple:

**Problem**
`/model` picked a runtime model but the choice was lost on the next
`reasonix chat` — there was no plumbing from the slash command back to
`~/.config/reasonix/config.toml`, so users had to re-`/model` every
session.

**Fix**
The `/model` slash command now writes the chosen provider into the user
config via `config.LoadForEdit(...).SaveTo(...)` (the same edit primitive
used by the `/theme` PR, #3043). Persist failures stay on `slog` (operator
visibility) while user-visible status — including the "kept previous
model on save failure" path — goes through the TUI notice channel so the
chat line shows it, not the operator log.

**Changes**
- `internal/cli/model.go` (+33/-2): `/model` subcommand persists to user
  config and routes feedback through the TUI notice.
- `internal/cli/model_test.go` (+43): coverage for the persistence path.
- `internal/config/edit.go` (+17/-5): edit helper now exposes the field
  needed by `/model`.
- `internal/config/edit_test.go` (+15): edit-helper regression tests.

**Tests**
- `go test ./internal/cli/ -count=1 -run 'TestModel'`
- `go test ./internal/config/ -count=1 -run 'TestPersistModel|TestLoadForEdit|TestEdit'`
  → all pass.

**Mirrors**
- Persistence pattern from PR #3043 (`/theme`).

### DIFF
--- a/internal/cli/model.go
+++ b/internal/cli/model.go
@@ -2,7 +2,6 @@ package cli
 
 import (
 	"fmt"
-	"log/slog"
 
 	tea "charm.land/bubbletea/v2"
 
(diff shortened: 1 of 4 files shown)


esengine (approved):
Verified end-to-end. On base the /model path writes nothing to config, so the choice was lost on the next launch. With the fix, persistModel writes default_model and a fresh config.Load() (next-launch simulation) resolves the default to the persisted provider/model ref. SetDefaultModel broadening to the provider/model form is covered by the config tests, including empty and unknown-model cases. Thanks.

Issue #3655, opened 2026-06-09
[Bug]: TUI can be suspended by tty input and leave terminal modes dirty


lizhengwu:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

reasonix desktop-v1.3.0-44-gba094b9f

## Problem

  Reasonix TUI was stopped by Unix/macOS job control with:

  ```text
  [1] + 16300 suspended (tty input) reasonix

  After returning to the shell, the terminal showed many mouse/terminal escape fragments, for example:

  70m0;22;70M32;21;70M32;20;70M...

  This looks like the TUI was stopped while alt-screen / raw mode / mouse tracking were still active, leaving the shell terminal in a dirty state.

  Screenshot:

  Evidence

  Process state showed reasonix was stopped, not crashed:

  PID    PPID  PGID  TPGID STAT TTY      COMMAND
  14708 14707 14708  14708 S+   ttys005  -/bin/zsh
  16300 14708 16300  14708 T    ttys005  reasonix
  16305 16300 16300  14708 T    ttys005  ...codegraph.js serve --mcp

  Key points:

  - reasonix was in state T, meaning stopped/suspended.
  - reasonix PGID was 16300.
  - terminal foreground PGID was 14708, the shell.
  - This matches a background process group reading from TTY and receiving SIGTTIN.

  Code context

  The TUI enables alt-screen and mouse tracking in the non-Termux path:

  v.AltScreen = true
  v.MouseMode = tea.MouseModeCellMotion

  The controller cleanup currently runs only after tea.Program.Run() returns. If the OS stops the process inside Run(), final cleanup does not run.

  A targeted search found no explicit handling for:

  ctrl+z
  tea.Suspend
  SIGTTIN
  SIGTTOU
  SIGTSTP
  SIGCONT

  Current hypothesis

  Likely flow:

  1. Reasonix TUI enables alt-screen / raw mode / mouse tracking.
  2. The reasonix process group is no longer the terminal foreground process group.
  3. Reasonix still tries to read from the TTY.
  4. Unix/macOS job control sends SIGTTIN.
  5. The process is stopped inside the TUI runtime.
  6. Normal cleanup does not run.
  7. The shell becomes foreground again while terminal modes are still dirty.
  8. Mouse/terminal escape sequences become visible in the shell.

### Steps to reproduce

Proposed fix plan

  I plan to split this into three small PRs.

[...]

PR #3697, opened 2026-06-09
fix(tui): suspend cleanly on ctrl-z

lizhengwu:
Map Ctrl+Z to Bubble Tea's suspend path so explicit user suspend can release terminal state before stopping.## Summary

  - Map Ctrl+Z to Bubble Tea's suspend path so explicit user suspend can release terminal state before stopping.
  - Add focused coverage that verifies Ctrl+Z produces `tea.SuspendMsg`.

  ## Scope

  This is the first mitigation for #3655. It covers explicit Ctrl+Z suspend only; the Unix job-control guard for background tty-read / `SIGTTIN` is planned separately.

  Refs: #3655

  ## Test plan

  - [x] `go test ./internal/cli -run TestCtrlZSendsSuspend -count=1`
  - [x] `go test ./internal/cli -count=1`
  - [x] `go test -race ./internal/cli -run TestCtrlZSendsSuspend -count=1`


esengine (approved):
Verified end-to-end. Reproduced #3655's gap: without the mapping, TestCtrlZSendsSuspend fails because Ctrl+Z returns no command. With `case "ctrl+z": return m, tea.Suspend` it returns Bubble Tea's suspend command (tea.SuspendMsg), and the full cli package passes. The earlier lint red was the unrelated SA5011 staticcheck false positive that #3715 already suppressed — green after updating the base. Thanks.

PR #3701, opened 2026-06-09
fix(dev): suppress pnpm TTY prompts and lockfile age checks in dev script

CVEngineer66:
pnpm v11 aborts node_modules removal without a TTY and rejects lockfiles with packages published within the minimum release age window. Both checks are meant for CI safety and are irrelevant during local development.

Set `PNPM_CONFIG_CONFIRM_MODULES_PURGE=false` and `PNPM_CONFIG_MINIMUM_RELEASE_AGE=0` before launching wails dev.

### DIFF
--- a/dev
+++ b/dev
@@ -29,6 +29,9 @@ wails_port="${REASONIX_DESKTOP_WAILS_PORT:-$((vite_port + 29000))}"
 
 export REASONIX_DESKTOP_VITE_PORT="$vite_port"
 export REASONIX_DEV=1
+# pnpm v11: suppress TTY prompts and skip lockfile age checks in headless mode.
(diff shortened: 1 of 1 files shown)


esengine (approved):
Low-risk dev-tooling fix. Verified statically: `bash -n dev` passes, the two exports sit before `exec wails dev` so they're inherited by the pnpm child, and `confirm-modules-purge` / `minimum-release-age` are real pnpm settings addressed via pnpm's documented `PNPM_CONFIG_*` env convention. Couldn't reproduce the pnpm v11 headless abort locally (have v10.22), but the change is purely additive env vars in a dev launcher — no impact on CI or the shipped build. Thanks.

PR #3704, opened 2026-06-09
fix(agent): degrade gracefully instead of looping on unmet readiness (#3664)

### LINKED ISSUE #3664: [Bug]: 更新步骤失败，会触发最后一步，最终形成死循环
### Version line

### Exact version

1.4.0

### What happened?

被循环执行了几轮编译与测试才发现
Image

### Steps to reproduce

1、执行一个带步骤的任务
2、每完成一步更新步骤失败
3、最后一步完成会执行全部步骤更新操作，但是会全部失败。
4、自动触发最后一次编译与测试
....重新更新全部步骤

### OS / platform

Deepin 25.1.1

### Relevant logs or output

```shell

```

esengine:
## Problem (#3664)
The final-answer readiness gate could trap weaker planner/executor models in a costly cycle: `todo_write` hard-rejected a completion lacking a `complete_step` receipt, so the model thrashed re-running compile/test, and after the retry budget the turn **hard-errored — discarding its work**. Users reported "several rounds of compile+test, then it dies".

## Approach
Follow how agent harnesses bound verification (LangChain `early_stopping`, Claude Code's loop-guarded Stop hook): **nudge, then degrade** — never trap or discard.

- `todo_write` no longer fails mid-flow; it only **warns** on an unsigned completion (progress tracking must not block).
- The completion-evidence guarantee **moves to the single final-answer readiness gate** (`CompletedTodosMissingReceipt`), enforced only once files were changed so no-write flows aren't gated.
- When the **same block reason recurs** (thrashing) or the retry budget is spent, the turn **accepts the answer with a visible `proceeding without confirmed readiness` warning** + an errored audit, instead of erroring and losing work.

## Test plan
- Unit/loop E2E (`internal/agent`): graceful degrade on repeated reason; the moved guarantee still caught; **unbounded-input bound check** (20 premature answers → still capped). Passing under `-count=10`, no hang.
- **Real end-to-end** (`internal/boot`): drives `boot.Build` → `Controller.Run` → agent loop → builtin `write_file`/`todo_write` → evidence ledger → readiness gate with a scripted provider that never signs off; asserts the turn degrades (no error, terminates within timeout), emits the warning, and the real write lands. Passing under `-count=10`.
- `internal/control` + `internal/cli` green; gofmt/vet clean.

Closes #3664

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -414,6 +414,7 @@ func (a *Agent) Run(ctx context.Context, input string) error {
 	a.session.Add(provider.Message{Role: provider.RoleUser, Content: input})
 
 	finalReadinessBlocks := 0
+	lastReadinessReason := ""
 	emptyFinalBlocks := 0
 	handoffNudges := 0
 	usedAnyTool := false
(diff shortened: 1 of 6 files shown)


esengine:
Closing in favor of #3673, which is the better-targeted fix for #3664: it pinpoints the stale-todo-baseline lock (a rejected batch todo_write never becomes the latest baseline, so follow-up complete_step calls validate against stale pending steps) and fixes it surgically by letting complete_step sign off a matched pending todo — preserving the evidence guarantee, with no change to the readiness terminal contract. This PR's graceful-degradation approach was broader than #3664 needs and changed more behavior surface; if a 'never hard-error after the retry budget' property is wanted later, it can be revisited as a separate, smaller change.

PR #3707, opened 2026-06-09
fix(desktop): restore clipboard image paste / 修复桌面端剪贴板图片粘贴

SivanCola:
## Summary
- Restore desktop image paste when WebView exposes clipboard images through DataTransferItem instead of DataTransfer.files.
- Add a native clipboard fallback using the existing control.SaveClipboardImage path when JS clipboard data is empty or the paste event is not delivered.
- Keep pasted attachments under the active workspace and add a regression test for that contract.

## Root cause
The composer only inspected e.clipboardData.files. In some desktop WebView clipboard states, Ctrl/Cmd+V provides an image hint or item but no files, and plain text is empty, so the textarea paste becomes a no-op with no visible attachment.

## Verification
- pnpm typecheck
- pnpm test
- go test . (desktop)
- git diff --check


esengine (approved):
Solid fix — the three-layer fallback (items merge → native clipboard read → keypress timer) is well thought out, with dedup so the layers never double-attach. Nice catch promoting the silent os.Chdir failure to a real error, and good workspace-root test coverage. Thanks!

PR #3709, opened 2026-06-09
fix(agent): stop bypassing ask tool in YOLO mode

### LINKED ISSUE #3696: [Bug]: ask 工具在 yolo 模式下无弹窗 UI，自动返回假结果
### Version line

### Exact version

1.4.0 & 1.3.0

### What happened?

yolo模式下调用 ask 工具时，用户侧看不到任何选择弹窗，但 agent 端收到一个自动填充的返回结果（似乎自动选了第一个选项）。切换为 auto 模式后则无此bug。
预期行为：用户侧应弹出选择弹窗，用户选择后 agent 收到真实结果。

### Steps to reproduce

1.Reasonix 桌面端，editMode 设为 yolo
2.agent 调用 ask 工具
3.用户侧：无弹窗出现
4.agent 侧：收到返回结果，显示用户"选择了"第一个选项

### OS / platform

Windows 11

### Relevant logs or output

```shell

```

CVEngineer66:
## Bug

In YOLO mode, when the agent calls the `ask` tool to get a user decision, no prompt UI appears and the agent receives a fake answer (the first option is automatically selected). This defeats the purpose of the `ask` tool, which exists specifically for cases where the model needs genuine user input.

## Root Cause

`Controller.Ask()` (which implements `agent.Asker`) checked `bypassEnabled()` before emitting the `AskRequest` event:

```go
func (c *Controller) Ask(ctx context.Context, questions []event.AskQuestion) ([]event.AskAnswer, error) {
    if c.bypassEnabled() {
        return recommendedAskAnswers(questions), nil  // ← silently picks first option
    }
    // ... normal flow: emit AskRequest, block for user ...
}
```

The same check appeared twice — once before `promptMu.Lock()` and once after — both returning `recommendedAskAnswers()` which always selects `q.Options[0].Label`.

## Logic Error

YOLO mode (`c.bypass = true`) was designed to **skip tool-approval gates** (the `requestApproval()` method), allowing autonomous execution without manual approval of each tool call. The `ask` tool is semantically different:

| Operation | YOLO mode expected behavior | Previous behavior |
|---|---|---|
| Tool call approval (`requestApproval`) | ✅ Auto-allow (skip prompt) | ✅ Correct |
| `ask` tool (user decision) | ✅ Show prompt, wait for real input | ❌ Silently returned first option |

The `ask` tool exists precisely for cases where the model cannot decide on its own — treating it the same as a tool approval gate undermines its purpose.

## Fix

Removed both `bypassEnabled()` checks from `Controller.Ask()`, so the `ask` tool always emits an `AskRequest` event and waits for the user to respond, regardless of mode.

Also removed the now-unused `recommendedAskAnswers()` helper function.

## Testing

- `go build ./internal/...` — clean
- `go test ./internal/agent/ -run TestAsk` — all 4 ask tool tests pass
- Manual verification: YOLO mode now shows the ask prompt UI and waits for user input

Fixes #3696

### DIFF
--- a/internal/control/controller.go
+++ b/internal/control/controller.go
@@ -858,17 +858,11 @@ func (c *Controller) EnableInteractiveApproval() {
 // Ask implements agent.Asker: it emits an AskRequest and blocks until
 // AnswerQuestion(ID, …) answers or ctx is cancelled. promptMu serialises it
 // against tool-approval prompts so at most one user prompt is outstanding.
+// Unlike tool-approval gates, Ask is NOT bypassed in YOLO mode — the ask
+// tool exists to get a user decision, and YOLO only skips tool approvals.
 func (c *Controller) Ask(ctx context.Context, questions []event.AskQuestion) ([]event.AskAnswer, error) {
-	if c.bypassEnabled() {
-		return recommendedAskAnswers(questions), nil
-	}
-
 	c.promptMu.Lock()
 	defer c.promptMu.Unlock()
-
-	if c.bypassEnabled() {
-		return recommendedAskAnswers(questions), nil
-	}
 
 	c.mu.Lock()
 	c.nextID++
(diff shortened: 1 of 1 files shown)


esengine:
Thanks @CVEngineer66 — your minimal controller.go fix is exactly right and is the basis of #3712. Closing in favor of that, which also updates the existing yolo_test.go cases that asserted the old auto-answer behavior (those were why CI went red here). Credited via Co-authored-by. Closes the same #3624.

PR #3717, opened 2026-06-09
fix(desktop): keep Windows tray loop on locked thread

### LINKED ISSUE #3516: [Bug]: 后台模式下右键无响应
### Version line

### Exact version

1.3.0

### What happened?

当设置关闭窗口后，应用后台运行时，任务栏右键无法退出应用

### Steps to reproduce

1. 设置关闭窗口为后台运行。
2. 关闭窗口
3. 任务栏点击应用右键

### OS / platform

windows 11

### Relevant logs or output

```shell

```

wade19990814-hue:
## What

Keep the Windows systray event loop on a locked OS thread.

Fixes #3516.

Also addresses the tray right-click/menu responsiveness symptoms reported in #3394 and the tray portion of #3548. The MCP orphan-process note in #3548 is a separate issue and is intentionally out of scope for this PR.

## Why

We caught a live repro on Windows with desktop v1.4.0 where the tray icon stopped responding to right click while the app was hidden:

- `reasonix-desktop.exe` was still alive and `Responding=True`.
- The main Wails window was hidden but not hung.
- The systray hidden window (`SystrayClass`) still existed but `IsHungAppWindow(SystrayClass) == true`.
- The Wails window (`wailsWindow`) reported `IsHungAppWindow(...) == false`.

That points to the systray Win32 message loop becoming stuck independently of the main Wails window. `fyne.io/systray` relies on Win32 hidden-window messages on Windows, so the tray loop should stay pinned to one OS thread.

## How

- Wrap the Windows systray loop in a helper that calls `runtime.LockOSThread()` before `systray.Run(...)`.
- Add a Windows-only regression test that verifies the helper does not move OS threads across scheduler yields.

## Testing

- `go test . -run 'TestDesktopTrayLoopRunsOnLockedOSThread|TestTrayMenuLabelsFollowLocale|TestBeforeCloseAllowsSystemQuitWhenBackgroundCloseEnabled' -count=1`
- `wails build`
- Windows manual test passed locally: launch the built desktop app, enable close-to-tray/background behavior, close to tray, repeatedly use the tray right-click menu, and confirm Open/Quit remain responsive.

### DIFF
--- a/desktop/tray_loop_windows.go
+++ b/desktop/tray_loop_windows.go
@@ -2,9 +2,21 @@
 
 package main
 
-import "fyne.io/systray"
+import (
+	"runtime"
+
+	"fyne.io/systray"
+)
 
 func startDesktopTray(onReady, onExit func()) func() {
-	go systray.Run(onReady, onExit)
+	go runDesktopTrayLoop(func() {
+		systray.Run(onReady, onExit)
+	})
 	return systray.Quit
+}
+
+func runDesktopTrayLoop(run func()) {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+	run()
 }
--- a/desktop/tray_loop_windows_test.go
+++ b/desktop/tray_loop_windows_test.go
@@ -1,0 +1,30 @@
+//go:build windows
+
+package main
+
+import (
+	"runtime"
+	"testing"
+
+	"golang.org/x/sys/windows"
+)
+
+func TestDesktopTrayLoopRunsOnLockedOSThread(t *testing.T) {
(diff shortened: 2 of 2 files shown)


esengine (approved):
Verified on Windows: the desktop module vets clean and TestDesktopTrayLoopRunsOnLockedOSThread passes, confirming the systray loop stays pinned to one OS thread across scheduler yields. This is the correct remedy for #3516 — fyne.io/systray's Win32 message loop must pump messages on the same thread that created the hidden window, and base code ran it on a migratable goroutine. Thanks for the live repro write-up and the regression test.

PR #3723, opened 2026-06-09
feat(tui): add Ctrl+Home/Ctrl+End shortcuts to scroll the transcript viewport to top/bottom

alxbnct:
# Summary

Adds Ctrl+Home and Ctrl+End keyboard shortcuts that scroll the TUI transcript viewport directly to the top or bottom, so users can jump to either end of a long scrollback without repeated PgUp/PgDn.

Ctrl+Home calls m.viewport.GotoTop() and Ctrl+End calls m.viewport.GotoBottom(), wired in the same scroll-key switch that handles PgUp/PgDn (lines 809–814 of chat_tui.go). The shortcuts work in any TUI state because the scroll-key handler runs before all modal checks. The plan-approval hint is also updated to mention them.

# Why

The transcript viewport already supports PgUp/PgDn for page-at-a-time scrolling, but there is no way to jump straight to the beginning or end of a long session. Users who scroll up to read history must either hold PgUp or guess how many PgDn presses return to the latest output. Ctrl+Home/End is the standard convention in editors, pagers, and terminals for this operation, so it fits naturally into the existing keybinding set.

# Scope / notes

Only the TUI chat viewport; the HTTP/serve frontend has its own scrollDown() logic and is not affected. The new test exercises both shortcuts through the real tea.KeyPressMsg path against a populated viewport.

# Test plan

• TestCtrlHomeEndScrollKeyBindings sends a populated viewport tea.KeyPressMsg{Code: tea.KeyHome, Mod: tea.ModCtrl}, asserts m.viewport.AtTop(), then sends tea.KeyPressMsg{Code: tea.KeyEnd, Mod: tea.ModCtrl} and asserts m.viewport.AtBottom().
• go build ./... – green.
• go test ./internal/cli/ – all tests pass (including the new one).
• go test ./internal/i18n/ – all pass (updated hint strings match the expected catalog)


esengine (approved):
Clean — Ctrl+Home/End to jump the transcript viewport to top/bottom, with a test and the plan-approval hint updated. Thanks!

PR #3750, opened 2026-06-10
fix(desktop): convert \slashed to \not for KaTeX compatibility

lightfront:
# Convert \slashed to \not for KaTeX compatibility

## Problem
KaTeX doesn't support the `\slashed` command from the LaTeX 'slashed' package, which is commonly used in physics for Feynman slash notation (`\slashed{p}`, `\slashed{\partial}`).

When the model emits these commands, Reasonix shows raw LaTeX source as red error text instead of rendering the formula.

## Solution
Convert `\slashed{X}` to `\not{X}` in `latexNormalizeForKatex`. The `\not` command provides a similar visual effect (a slash through the character) and is supported by KaTeX.

**Example:**
- Input: `$\slashed{p}$`
- After normalization: `$\not{p}$`
- Rendered: p with a slash through it (Feynman slash notation)

## Changes
- `desktop/frontend/src/components/latexNormalize.ts`: Added regex conversion
- `desktop/frontend/src/__tests__/math-golden.test.ts`: Added 3 test cases

## Tests
- 3 new test cases for `\slashed` conversion
- All tests pass (88 math-golden + 8 text-size + 6 provider-model-refresh = 102 total)

## Related
- This is a standalone fix that works with or without PR #3666
- PR #3666 has the core math pipeline improvements (classifier, code block protection, etc.)

### DIFF
--- a/desktop/frontend/src/__tests__/math-golden.test.ts
+++ b/desktop/frontend/src/__tests__/math-golden.test.ts
@@ -145,6 +145,12 @@ check("\\\\slashed{\\\\partial} is converted to \\not{\\\\partial}", () => {
 check("\\\\slashed in prose context", () => {
   return normalizeMath("The momentum $\\\\slashed{p}$ is conserved") === "The momentum $\\\\not{p}$ is conserved";
 });
+check("\\\\slashed\\\\epsilon(0) → \\not{\\epsilon(0)} (Greek + function)", () => {
+  return normalizeMath("$\\slashed\\epsilon(0)$") === "$\\not{\\epsilon(0)}$";
+});
+check("\\\\slashed a → \\not a (single letter)", () => {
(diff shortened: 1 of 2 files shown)


esengine (approved):
Thanks @lightfront — clean, well-scoped fix.

`\slashed` is a real gap (KaTeX ships no `slashed` package), and routing it to `\not` is the right approximation for Feynman slash notation. Since it runs inside `latexNormalizeForKatex`, the rewrite only touches recognized math spans — code blocks and prose are untouched — and the golden tests cover both the braced and unbraced forms. CI is green across the board.

One tiny thing I'll tidy in a follow-up (not blocking): a couple of the braced test inputs use a double backslash (`\slashed{p}`) where real model output is a single one (`\slashed{p}`); the regex handles both, so the tests pass either way. Merging now — appreciate the contribution!

PR #3777, opened 2026-06-10
fix(diff): avoid input-derived Myers allocation arithmetic

SivanCola:
## Summary
- Follow up on #3773 and #3775 / https://github.com/esengine/DeepSeek-Reasonix/pull/3773#discussion_r3385329112.
- Compute the Myers `maxD` cap without directly adding the two input lengths.
- Compute the allocation width from the fixed edit budget instead of `2*maxD+1`, and cover huge-int inputs with a unit test.

## Testing
- `go test ./internal/diff/`
- `go test ./...` fails in `internal/control`: `TestRemoveMCPServerRemovesUnconnectedLazyPlaceholder` sees locally configured MCP names instead of an empty config.

### DIFF
--- a/internal/diff/diff.go
+++ b/internal/diff/diff.go
@@ -144,6 +144,30 @@ type op struct {
 // rewrite whose line-by-line diff is unreadable anyway, so we fall back to tallies.
 const maxDiffEdits = 2000
 
+// myersMaxD returns min(n+m, maxDiffEdits) without adding the two input
+// lengths. The result feeds an allocation below, so every arithmetic step stays
+// inside the fixed maxDiffEdits budget.
+func myersMaxD(n, m int) int {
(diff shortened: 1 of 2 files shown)


esengine (approved):
Nice — this is the approach that actually satisfies the analyzer. By computing `min(n+m, maxDiffEdits)` and `2*maxD+1` with bounded loops instead of arithmetic on the input lengths, there's no input-derived `+`/`*` flowing into the allocation at all, so CodeQL's `go/allocation-size-overflow` has nothing left to flag — which my earlier clamp couldn't achieve. Both loops are O(maxDiffEdits) and run once per diff, so the cost is nil, and the `TestMyersMaxD_CapsWithoutOverflow` cases (incl. maxInt) pin the bound. Merging.

PR #3780, opened 2026-06-10
Split /new and /clear session flows / 区分 /new 与 /clear 会话流程

SivanCola:
## Summary
- Split `/new` and `/clear` semantics: `/new` starts a new session and saves the transcript, while `/clear` discards the current context.
- Add two-step `/clear` confirmation in the CLI TUI and an ASK-style in-app confirmation card in desktop.
- Update desktop state reset, command help, docs, and regression coverage for the new session/context behavior.

## Verification
- isolated config run of `go test ./internal/control ./internal/cli`
- `git diff --check origin/main-v2...HEAD`

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,10 +220,11 @@ convenient.
 
 ### Slash commands
 
-In `reasonix chat`, built-in commands (`/compact`, `/new`/`/clear`, `/rewind`, `/tree`,
+In `reasonix chat`, built-in commands (`/compact`, `/new`, `/clear`, `/rewind`, `/tree`,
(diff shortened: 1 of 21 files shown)


esengine (approved):
Good correctness fix — /new and /clear are now genuinely distinct: /new saves the transcript and rotates, /clear confirms then discards the current context without saving. ClearSession is gated on !running, only removes the current session's artifacts (history is untouched), and the confirmation card guards against accidents. Tests on both the controller and the TUI. Thanks!

PR #3783, opened 2026-06-10
fix(tui): return pendingModelSwitch after skill delete to unblock Enter

LVGG:
return pendingModelSwitch after skill delete to unblock Enter (#3764)


esengine (approved):
Good fix — clean unblock. Thanks!

PR #3786, opened 2026-06-10
feat: improve attachment preview cards

Abysstsuki:
## Summary

- Add display-name-aware attachment references so newly sent desktop messages can preserve original filenames in history.
- Update composer attachment cards with image-first ordering, larger image thumbnails, aligned remove controls, and rectangular file cards.
- Update user-message attachment cards to match the composer preview layout, including image thumbnail loading for stored image attachments.

## Scope

This is a frontend display change only. The submitted model input still uses the existing `@.reasonix/attachments/...` references, and the backend attachment storage / reference resolution flow is unchanged.

## Testing

- `npx.cmd tsx src/__tests__/attachment-display.test.ts`
- `npm.cmd run test:typecheck`
- `npm.cmd run check:css`


esengine (approved):
﻿Reviewed. The display/submit split is handled correctly — the named `@[label](path)` form only ever exists in the display string, the model still receives the raw `@.reasonix/attachments/…` refs, and the parser keeps backward compatibility with unnamed refs. `baseName` growing backslash awareness and the escaped-bracket cleanup in labels are the right edge cases to have covered, and `sortDisplayAttachments` relying on stable sort keeps non-image relative order intact.

One small thing I'll fix in a follow-up rather than another round here: `attachment-display.test.ts` isn't wired into the `test`/`test:all` scripts in `package.json` (the suite enumerates files explicitly), so CI never runs it as-is.

Merging — the composer/message card alignment is a clear visual upgrade. Thanks!

PR #3814, opened 2026-06-10
fix: adapt logos for dark theme / 修复深色主题 logo 适配

SivanCola:
## Summary
- render sidebar and welcome logos as monochrome in dark theme
- keep the original blue logo in light theme
- support system dark mode when no explicit theme override is set

## Verification
- pnpm --dir desktop/frontend check:css
- verified in the in-app browser that light theme keeps both logo filters as none and the loaded stylesheet contains the dark-theme logo rules

### DIFF
--- a/desktop/frontend/src/styles.css
+++ b/desktop/frontend/src/styles.css
@@ -2335,6 +2335,20 @@ a[href] {
   height: var(--welcome-brand-height);
   object-fit: contain;
 }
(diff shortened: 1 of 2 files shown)


esengine (approved):
Dark-theme logo adaptation looks right (filter + system prefers-color-scheme fallback). Merging.

PR #3815, opened 2026-06-10
fix(permission): prefer bash command prefix for approval grants

### LINKED ISSUE #3759: [Bug]: 1.4版本，批准依然不对，重复提示需要批准 CLI
### Version line

### Exact version

1.4

### What happened?

1.4版本，批准依然不对，重复提示需要批准。
我发现记录到配置文档里的指令批准竟然是一个完整的指令批准。比如说搜索，它连搜索关键词都放进去了。那么当然，下一次搜索，它的这个指令跟这一次就不同，因为搜索的关键词不同。即使我们在一个任务当中会用多次搜索，会用同一个工具，但是它每次都请求批准。因为在它看来，每次的搜索都不同。正常的逻辑是，只要用这个工具搜索，我已经批准在这个任务当中可以自由使用，它就不需要再申请批准。可是这个逻辑上是有问题的，它一直在要求批准。反正我现在的体验就是，它每一项任务都要批准，无法解决。每次使用，为了节省麻烦，你只能选择Y O L O模式。

### Steps to reproduce

1.4版本，批准依然不对，重复提示需要批准。
我发现记录到配置文档里的指令批准竟然是一个完整的指令批准。比如说搜索，它连搜索关键词都放进去了。那么当然，下一次搜索，它的这个指令跟这一次就不同，因为搜索的关键词不同。即使我们在一个任务当中会用多次搜索，会用同一个工具，但是它每次都请求批准。因为在它看来，每次的搜索都不同。正常的逻辑是，只要用这个工具搜索，我已经批准在这个任务当中可以自由使用，它就不需要再申请批准。可是这个逻辑上是有问题的，它一直在要求批准。反正我现在的体验就是，它每一项任务都要批准，无法解决。每次使用，为了节省麻烦，你只能选择Y O L O模式。

### OS / platform

ubuntu24

### Relevant logs or output

```shell

```

JesonChou:
### Reviewer feedback

| # | Request | Status |
|---|---------|--------|
| 1 | Fully remove scope plumbing | ✅ Done |
| 2 | Collapse TUI exact/prefix choices | ✅ Done |
| 3 | Document edit widening + add test | ✅ Done |

**Edit persistence scope (point 3):** approving one `edit_file` call with "Always allow (save to config)" saves the rule as `Edit` — tool-wide, no path restriction. This covers all six file-mutation tools on any file across all future sessions. Deny rules still take precedence. This is intentional and matches the existing session-grant behaviour.

---
Fixes #3759 — the approval system remembered rules with full command arguments (e.g. "Bash(grep -r specificTerm .)"), causing every slightly different invocation to re-prompt within the same session.

Root cause: SessionGrantRuleForScope and RememberRuleForScope only applied BashCommandPrefix when scope==ApprovalScopePrefix, but the default "allow this session" path used ApprovalScopeExact — so the prefix was never used.  Every command was stored literally.

Fix:
- Both functions now always prefer BashCommandPrefix for bash when a safe prefix can be extracted, regardless of the scope parameter. Safe prefix means: >=2 words, no shell metacharacters, not dangerous. When no prefix is available (single word, shell syntax, dangerous command), the exact command is still stored as the fallback.
- File mutation tools remain tool-wide ("Edit") — matching existing session-grant behaviour.
- The now-unused scope parameter is removed from both function signatures; all callers updated.

Concrete example:
  Before: "grep -r TODO ." approves -> stored "Bash(grep -r TODO .)" "grep -r FIXME ." -> prompts again (different subject) After:  "grep -r TODO ." approves -> stored "Bash(grep -r:*)" "grep -r FIXME ." -> auto-allowed (matches prefix) "git log" -> still prompts (different command family)

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -590,21 +590,17 @@ func (a *App) CancelTab(tabID string) {
 // Approve answers a pending approval_request by ID: allow runs the call, session
 // also remembers the grant for the rest of the session.
 func (a *App) Approve(id string, allow, session, persist bool) {
-	a.ApproveWithScope(id, allow, session, persist, "")
+	ctrl := a.ctrlByTabID("")
+	if ctrl != nil {
+		ctrl.Approve(id, allow, session, persist)
+	}
 }
 
-func (a *App) ApproveWithScope(id string, allow, session, persist bool, scope string) {
-	a.ApproveTabWithScope("", id, allow, session, persist, scope)
-}
-
+// ApproveTab is like Approve but scoped to a specific tab.
 func (a *App) ApproveTab(tabID, id string, allow, session, persist bool) {
-	a.ApproveTabWithScope(tabID, id, allow, session, persist, "")
-}
-
-func (a *App) ApproveTabWithScope(tabID, id string, allow, session, persist bool, scope string) {
 	ctrl := a.ctrlByTabID(tabID)
 	if ctrl != nil {
-		ctrl.ApproveWithScope(id, allow, session, persist, scope)
+		ctrl.Approve(id, allow, session, persist)
 	}
 }
 
--- a/desktop/frontend/src/App.tsx
+++ b/desktop/frontend/src/App.tsx
@@ -1936,11 +1936,11 @@ export default function App() {
             {state.approval && (
               <ApprovalModal
                 approval={state.approval}
-                onAnswer={(allow, session, persist, scope) => {
+                onAnswer={(allow, session, persist) => {
                   // Approving an exit_plan_mode plan leaves plan mode; sync the
                   // tab-local indicator and persisted safe mode immediately.
                   if (state.approval!.tool === "exit_plan_mode" && allow) applyCollaborationMode("normal");
-                  approve(state.approval!.id, allow, session, persist, scope);
+                  approve(state.approval!.id, allow, session, persist);
                 }}
                 onRevisePlan={(text) => {
                   setPendingPlanRevision(text);
--- a/desktop/frontend/src/components/ApprovalModal.tsx
+++ b/desktop/frontend/src/components/ApprovalModal.tsx
@@ -10,7 +10,7 @@ export function ApprovalModal({
   onExitPlan,
 }: {
   approval: WireApproval;
-  onAnswer: (allow: boolean, session: boolean, persist: boolean, scope?: string) => void;
+  onAnswer: (allow: boolean, session: boolean, persist: boolean) => void;
   onRevisePlan?: (text: string) => void;
   onExitPlan?: () => void;
 }) {
@@ -21,10 +21,7 @@ export function ApprovalModal({
   const cardRef = useRef<HTMLDivElement | null>(null);
   const inputRef = useRef<HTMLTextAreaElement | null>(null);
   const isPlanApproval = approval.tool === "exit_plan_mode";
-  const isBashApproval = approval.tool === "bash";
   const subject = approval.subject.trim();
-  const bashPrefix = isBashApproval ? bashCommandPrefix(subject) : "";
-  const hasBashPrefix = bashPrefix !== "";
   const subjectSummary = subject.split("\n").find((line) => line.trim())?.trim() ?? "";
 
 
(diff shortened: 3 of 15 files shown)


JesonChou:
Other generic tools (search, web_fetch, grep, etc.) already returned the bare tool name in the old code and are unchanged. File mutation tools now use the same "Edit" grant for persistent rules that they already used for session grants — no new behavior, just alignment.

esengine (changes requested):
Thanks — the root-cause analysis is exactly right, and preferring a safe bash prefix is the behaviour I want for #3759. The danger/shell-syntax guard in `BashCommandPrefix` keeps the widening safe, and the prefix-match tests are good. A few things to sort before I merge:

**1. Finish removing the `scope` concept — right now it's only half-gone.** The description says the unused scope parameter was removed and all callers updated, but it was only dropped from `SessionGrantRuleForScope` / `RememberRuleForScope`. `approvalReply.scope`, the `scope` parameter on `Controller.ApproveWithScope`, and the `ApprovalScopeExact` / `ApprovalScopePrefix` constants are still written but no longer read — `requestApproval` no longer consumes `r.scope`. Either fully remove the scope plumbing or keep it honoured; leaving it inert is dead code in a security-sensitive path.

**2. The TUI still offers "exact" vs "prefix" keys that now produce the same rule.** In `chat_tui.go` the session/persist handlers still pass `ApprovalScopeExact` vs `ApprovalScopePrefix`, but since the rule builder ignores scope, choosing the exact option silently stores the wider prefix grant. If we're committing to prefix-by-default (I'm fine with that), collapse those into a single "allow for this session" / "always allow" pair so the UI matches what's actually stored.

**3. Confirm the persistent file-edit widening is intended.** `rememberRule("edit_file", "src/app.go")` now returns `Edit` instead of `Edit(src/app.go)`, so "Always allow (save to config)" on one edit persists a tool-wide Edit grant — every file edit in every future session auto-approves. That matches the in-session grant, but persisting a blanket edit-allow is a bigger step than the session one. I'm inclined to accept it, but please call it out in the description and add a test asserting the persisted rule is tool-wide, so it's a deliberate choice rather than a side effect.

Core direction is right. Get the scope cleanup done and the Edit widening documented + tested, and I'll run CI and merge.

JesonChou:
Thanks for the detailed review and the careful security analysis.

I’ll address all three points before the merge:

1. Scope removal – You're right, the cleanup was incomplete. I'll fully remove the unused scope plumbing: delete approvalReply.scope, the scope parameter from Controller.ApproveWithScope, and the constants ApprovalScopeExact/ApprovalScopePrefix. The approval path will then have no dead code.
2. TUI options – I'll collapse the "exact" vs "prefix" choices into a single "allow for this session" / "always allow" pair, since they now behave identically. The UI will accurately reflect the stored rule (prefix‑by‑default).
3. Edit widening – I've confirmed this is intentional, but I agree it needs to be explicit. I'll update the PR description to highlight that persisting an edit approval creates a tool‑wide Edit grant. I'll also add a test that asserts the persisted rule has no path restriction, so future readers understand the behaviour is deliberate.

I'll push the cleanup, UI changes, and the additional test shortly. Once CI passes, feel free to merge.

Thanks again for the thorough review.

JesonChou:
All three points addressed:

1. **Scope fully removed** — `approvalReply.scope`, `ApproveWithScope`, and the
   `ApprovalScopeExact` / `ApprovalScopePrefix` constants are deleted. Every
   caller uses the single `Approve`. 11 files, net -59 lines.

2. **TUI collapsed** — `handleApprovalKey` no longer branches on `hasBashPrefix`.
   All options produce identical behaviour for every tool.

3. **Edit widening documented + tested** — `TestPersistedEditRuleIsToolWide`
   asserts tool-wide `Edit` with no path restriction. Function and test comments
   call out the deliberate choice. PR description updated with the note.

Full regression passes, three-way merge against upstream main-v2 is clean. @esengine

esengine (approved):
﻿All three points addressed exactly as promised — the scope plumbing is fully gone end to end (controller, ACP dispatch/service, desktop bridge + dev mock, serve HTTP API), the TUI and modal now present a single session/persist pair that matches what actually gets stored, and `TestPersistedEditRuleIsToolWide` locks in the tool-wide Edit grant as a deliberate choice with both positive and negative matches. Nice, thorough cleanup.

Merging once CI is green.

PR #3822, opened 2026-06-10
fix(desktop): right panel git workspace enchance

kekexunxun:
## Summary
- Enhanced Workspace Git capabilities: Refactored and improved the Workspace Git workflow, enabling granular visibility into staged and unstaged file states and allowing direct preview of Git diffs within the panel.
- Added Git commit and branch management: Introduced a new Commit Modal that allows users to enter commit messages and submit directly; added support for searching, switching, and creating branches; and optimized the performance and user experience of the Git Push action.
- Fixed a layout issue where opening the right panel could cause a horizontal scrollbar (x-overflow) to appear in the application.

## Verification
- Frontend Tests - All Passed (105)
- Backend Tests - desktop go test All Passed

image
image
image


esengine:
﻿Heads-up before you spend more time here: #3922 (the context-overview/status-metrics redesign) landed today and reworked a large part of the right-dock/workspace-panel area this PR touches — including scoped file/change lists driven from the overview and the removal of the in-panel detail views. This branch now conflicts heavily, and some of its features may already be covered or need a different shape on top of the new layout.

Suggest taking a look at the current main-v2 panel first, then deciding what still applies: a rebased, narrowed version of the git-workspace enhancements that don't overlap is very welcome. Happy to point at the new extension points if useful.

esengine (changes requested):
This is a strong PR — working-tree staging view, per-file diffs, commit modal, fail-closed git env (`GIT_TERMINAL_PROMPT=0` + BatchMode is the right call), real tests, and bindings/mock/i18n all kept in sync. One correctness ask blocks merge:

- `WorkspaceGitCommit` silently checkouts the selected branch and **leaves the workspace on it**. If a turn is running against the tree, this rug-pulls the agent mid-edit. Guard it: refuse (or explicitly confirm) when the controller is busy, and say in the modal that committing to another branch switches the working branch.

Follow-ups we can take after merge (or fold in if you like): unstage uses `git reset HEAD --`, which fails on an unborn HEAD (first-commit repos) — prefer `git reset --` with a cached-removal fallback; the `globalThis.__reasonixI18nContext` stash needs a one-line justification anchored to the actual dual-instance cause, or removal; `WorkspacePanel.tsx` is now ~1.7k lines — per repo policy the description should carry a split plan. Plus a rebase onto current main-v2.

kekexunxun:
I pulled the latest code and performed some verification; I also fixed some existing issues. I will submit the latest changes shortly.

PR #3396, opened 2026-06-06
feat(desktop/cli): add expand_thinking config toggle and persist /verbose (#3303 #3312)

ashishexee:
Closes #3303
Closes #3312

## Problem

**#3303** — Every AI response that includes reasoning shows a collapsible "thinking" section in the desktop app. It was hardcoded to collapsed — `useState(false)` in Message.tsx. Users who always want to see the reasoning had to click expand every single time, on every single message.

**#3312** — The `/verbose` toggle (Ctrl+O) resets on every CLI session. Users who prefer seeing reasoning text must re-toggle it each time they start a chat.

## Solution

**#3303** — Adds a `[desktop].expand_thinking` config option that controls the default expand/collapse state of the reasoning section. Collapsed by default (no behavior change for existing users).

**#3312** — Adds a `[ui].show_reasoning` config option. The `/verbose` toggle now persists across sessions by writing to config.

### #3303 — Desktop expand_thinking toggle

**Go (config layer)**
- `DesktopConfig.ExpandThinking` — bool field with TOML tag `expand_thinking`, default false
- `Config.SetExpandThinking(bool)` — mutator in edit.go, follows the same pattern as SetDesktopCloseBehavior
- `render.go` — TOML output includes `expand_thinking` with inline comment

**Go (desktop bridge)**
- `SettingsView.ExpandThinking` — bool JSON field, populated from config in `Settings()`
- `App.SetExpandThinking(bool)` — Wails-bound method, writes via `applyConfigOnly` (no controller rebuild — display-only setting)

**Frontend (types + bridge)**
- `SettingsView` in types.ts — added `expandThinking: boolean`
- `AppBindings` in bridge.ts — added `SetExpandThinking(on: boolean): Promise<void>`
- Mock implementation in bridge.ts for browser dev mode

**Frontend (UI)**
- `AssistantMessage` in Message.tsx — accepts `defaultExpanded` prop → `ProcessCard defaultOpen={item.streaming || defaultExpanded}`
- `Transcript.tsx` — threads `defaultExpandThinking` through hot/warm/cold zones (`LiveAssistantMessage`, `WarmZone`, `WarmTurnItems`)
- `App.tsx` — fetches setting on startup, stores in `expandThinking` state, passes to Transcript
- `SettingsPanel.tsx` — two-button segmented control using `<SettingsField>` in General tab ("Collapsed" / "Expanded")

**i18n**
[...]

PR #3827, opened 2026-06-10
fix(cli): persist /verbose toggle across CLI sessions (#3312)

### LINKED ISSUE #3312: /verbose: toggle state not persisted across CLI sessions
## Description

Using `/verbose` to expand the reasoning/thinking text works only for the current session. Exiting and restarting the CLI resets it to off, requiring the user to manually toggle it again each time. The preference should be persisted in the config file so it survives across sessions.

## Steps to reproduce

1. Open rx CLI.
2. Type `/verbose` — the reasoning text expands (notice: "verbose on — thinking text will be shown").
3. Exit the CLI (`/quit`).
4. Reopen rx CLI.
5. Observe: thinking text is collapsed again, `/verbose` is back to off.

## Expected behavior

The `/verbose` toggle state should persist across CLI restarts. Users should be able to either set it manually in the config file or toggle it at runtime with auto-save.

## Technical background

- Handler: `internal/cli/chat_tui.go:2969` → `toggleVerboseReasoning()` (:2657) flips the in-memory bool `chatTUI.showReasoning` (defined at :119) with **no persistence**.
- `UIConfig` (`internal/config/config.go:62-66`) currently has only `theme`, `theme_style`, and `close_behavior` — **no field for show_thinking**.
- CLI init (`internal/cli/cli.go:410-414`) loads only `outputStyle` and `statuslineCmd` from config, nothing related to thinking/verbose.
- Config save path (`internal/config/edit.go:508-591`) already supports atomic TOML writes via `Config.Save()` — adding a new field and saving from the toggle is straightforward.

## Suggested fix

1. Add a `ShowThinking bool` field to `UIConfig` (toml tag: `show_thinking`), defaulting to `false` for backward compatibility.
2. In `newChatTUI` or CLI init, load that field into `m.showReasoning`.
3. In `toggleVerboseReasoning`, call `config.Set` + `config.Save()` (possibly debounced) to persist the new state.

## Scope

- Config change: new optional field `ui.show_thinking`; `false` by default — fully backward-compatible.
- CLI-only change; non-interactive mode (`--show-thinking` flag) is unaffected.
- Web frontend (`internal/serve/`) reasoning toggle is independent and unaffected.

## Environment

- rx version: latest main
- OS: all platforms

NkAntony777:
## Summary

The `/verbose` (and Ctrl+O) toggle previously reset to off on every CLI
launch because the in-memory `showReasoning` bool was never written
back to the config. This adds an optional `ui.show_thinking` field
(default `false`, fully backward-compatible) and persists every flip
off the bubbletea event loop so the next session opens with the
reasoning block in the same state.

Fixes #3312.

## Changes

- **`internal/config/config.go`** — new `UIConfig.ShowThinking bool` with `toml:"show_thinking"`, defaults to `false`.
- **`internal/config/render.go`** — `RenderTOMLForScope` emits `show_thinking = true` only when set, with a commented-out placeholder only in `RenderScopeFull` so unmodified project configs stay byte-identical.
- **`internal/cli/cli.go`** — `chatREPL` now calls a new `m.applyConfig(cfg)` helper.
- **`internal/cli/chat_tui.go`** —
  - new `applyConfig(cfg)` method: seeds the TUI from a loaded config, using OR semantics so a persisted `ShowThinking=true` does not stomp the Termux native-scrollback default.
  - `toggleVerboseReasoning` spawns a goroutine that calls `persistShowThinking(show)`, capturing `show` into a local first so the save goroutine reads a stable value (no race with the Update goroutine's write to `m.showReasoning`).
  - new `persistShowThinking(show bool) error` method: writes the preference to the user-level config via `SaveTo(config.UserConfigPath())`.

## Design notes

**Why user config, not project config.** `show_thinking` is a personal
viewing preference; writing it to a project-local `reasonix.toml` that
may be git-committed would leak the user's setting to every
collaborator on the repo. The same rationale as `NotificationsConfig`
et al. — viewing preferences belong in `~/.config/reasonix/config.toml`.
Note: a project `reasonix.toml` with `show_thinking = false` will
silently win on the next load (project > user priority in `Load`),
which matches the long-standing merge semantics.

**Why async save.** The bubbletea `Update()` goroutine must not block
on disk I/O. `SaveTo`'s atomic temp+rename makes racing writes safe on
disk (last-writer-wins, never a torn file); a coalescer/debouncer is
left as a possible follow-up.

**Scope of the fix.** Per the issue, the `--show-thinking` flag in
non-interactive `run` mode and the desktop app's reasoning toggle
(`internal/serve/`, `desktop/`) are out of scope and explicitly
unaffected. The web/desktop frontends have their own
`TextSink.showReasoning` field.

## Tests

```
go test ./internal/config/  → ok
go test ./internal/cli/     → ok
```

New tests (all passing):

| Test | Covers |
|---|---|
| `TestToggleVerboseReasoningPersistsShowThinking` | full on→off round-trip via real disk using `isolateUserConfig` + `LoadForEdit` |
| `TestPersistShowThinkingNoConfigIsNoop` | nil cfg early-return (production toggle relies on this) |
| `TestApplyConfigRespectsShowThinking` | **boot-time load path** (the half of the bug that the round-trip test alone wouldn't catch) |
| `TestApplyConfigKeepsTermuxVerboseOn` | OR semantics — Termux default-on is not stomped by an absent preference |
| `TestApplyConfigNilIsNoop` | defensive nil guard on `applyConfig` |
| `TestProjectRenderShowThinkingScoped` | `RenderScopeProject` emits the key only when set |

The `applyConfig` extraction is the testability-driven refactor that
closes the gap identified in review: previously, deleting the four
`applyConfig` lines would still pass every existing test. The new
`TestApplyConfigRespectsShowThinking` covers that regression.

## Manual verification

1. Build: `go build ./cmd/reasonix`
2. Launch: `./reasonix chat` (or `reasonix.exe` on Windows)
3. Type `/verbose` → notice "verbose on — thinking text will be shown"
4. `/quit`
5. Reopen: reasoning block is expanded by default
6. Inspect `~/.config/reasonix/config.toml` (or `%AppData%\reasonix\config.toml` on Windows):
   ```toml
   [ui]
   show_thinking = true
   ```
7. Type `/verbose` again → notice "verbose off"
8. Reopen → reasoning block is collapsed, and the config now shows the field with the default value (commented placeholder in the full-render path).

## Out of scope

- Debouncing rapid toggles (current behavior: 10 toggles = 10 atomic writes, all safe via temp+rename).
- Surfacing a user-visible notice when `Save` fails (currently `slog.Warn`; the in-TUI notice fires regardless of save success — matches the project's existing preference-persistence pattern, e.g. `/effort`).
- The web/desktop frontends' reasoning toggles.

### DIFF
--- a/internal/cli/chat_tui.go
+++ b/internal/cli/chat_tui.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"log/slog"
 	"net/url"
 	"os"
 	"path/filepath"
@@ -2863,8 +2864,41 @@ func (m *chatTUI) cycleMode() {
 	}
 }
 
+// applyConfig seeds the TUI from a loaded config. Extracted from chatREPL so
+// the "load cfg → propagate to chatTUI" path is unit-testable without going
+// through the full bubbletea bootstrap. New chatTUI has already picked a
+// default for showReasoning (Termux native scrollback) — we use OR semantics
+// so the persisted preference never silently disables a Termux user's
+// always-on reasoning (issue #3312).
+func (m *chatTUI) applyConfig(cfg *config.Config) {
+	if cfg == nil {
+		return
+	}
+	m.outputStyle = cfg.Agent.OutputStyle
+	m.statuslineCmd = cfg.Statusline.Command
+	m.cfg = cfg
+	if cfg.UI.ShowThinking {
+		m.showReasoning = true
+	}
+}
+
 func (m *chatTUI) toggleVerboseReasoning(notify bool) {
 	m.showReasoning = !m.showReasoning
+	// Persist the new preference so a future CLI session opens with the same
+	// state (issue #3312). The save runs off the bubbletea event loop to avoid
+	// stalling the UI on slow disks. We capture `show` into a local before
+	// spawning the goroutine: passing it as a function argument establishes
+	// the happens-before edge the Go memory model requires, so the save
+	// goroutine reads a stable value rather than racing the Update goroutine
+	//'s write to m.showReasoning. A missed or racing write to disk is harmless
+	// — the atomic temp+rename in SaveTo() means we either persist the new
+	// value or the previous one, never a torn file.
+	show := m.showReasoning
+	go func() {
+		if err := m.persistShowThinking(show); err != nil {
+			slog.Warn("persist show_thinking", "err", err)
+		}
+	}()
 	if !notify {
 		return
 	}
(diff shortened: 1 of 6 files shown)


esengine:
﻿Thanks for the careful work here, including the clean config rendering compatibility — the diagnosis of #3312 was exactly right. In the end the `/verbose` persistence landed via #3919 (re-land of #3396), which bundles it with the related `expand_thinking` toggle under the `ui.show_reasoning` key, so I'm closing this one as superseded rather than carrying two config keys for the same preference. Sorry the two efforts crossed — your write-up on the bubbletea event-loop persistence timing was useful when reviewing the landed version.

PR #3830, opened 2026-06-10
fix(desktop): harden Windows WebView startup

SivanCola:
## Summary
- embed the WebView2 bootstrapper in Windows desktop installer builds instead of relying on the default strategy
- add a Windows WebView2 GPU fallback, enabled by default for canary builds and overrideable with REASONIX_DESKTOP_DISABLE_WEBVIEW2_GPU
- install frontend crash handlers before early Wails runtime/theme calls so startup failures render an actionable overlay
- document the Windows fallback and add regression coverage for the GPU toggle

Refs #3820

## Tests
- `cd desktop/frontend && pnpm typecheck`
- `bash -n scripts/desktop-build.sh`

### DIFF
--- a/desktop/README.md
+++ b/desktop/README.md
@@ -177,7 +177,10 @@ handled here, and what to reach for if a target misbehaves:
   release; the CSS deliberately avoids `backdrop-filter`/blur (slow & inconsistent
   there).
 - **Windows / WebView2** — `Theme: SystemDefault` follows the OS light/dark
-  setting; the runtime must be installed (bundle it for distribution).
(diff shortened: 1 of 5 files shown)


esengine (approved):
Clean, well-scoped, and the test matrix on the GPU toggle is a nice touch. `-webview2 embed` is the right fix for the missing-runtime blank window, and installing the crash handlers before the early theme/runtime calls means we actually get a diagnosable overlay instead of a featureless webview. CI is green across all three platforms. Merging.

coderpang:
image
It still dosen't work.

PR #3833, opened 2026-06-10
feat(cli): add fuzzy matching to the slash-command completion menu

CnsMaple:

## Summary

Replace the strict `strings.HasPrefix` filter in the chat-TUI slash-command
completion popup with a case-insensitive subsequence matcher, so typing a
partial token like `/mdl` surfaces `/compact`, `/model`, `/memory` and any
other command whose letters appear in order — not just commands that begin
with the typed prefix.

## Why

Users naturally abbreviate or fat-finger slash commands (`/compt`,
`/mdl`, `/skll`, `/brnch`). The current matcher silently closes the menu on
anything that isn't an exact prefix, forcing the user to retype or look
the command up via `/help`. A subsequence match keeps the "type a few
letters, see the candidates" muscle memory from shells like fzf.

## Scope

Inentionally limited to the live TUI completion popup. Typed-input
resolution (`/name` + Enter) and the model's `slash_command` tool still
use exact-name lookup, so the routing layer stays predictable. Both are
one-liner follow-ups if reviewers want them in the same change.

## Implementation

- `internal/cli/complete.go`
  - New `fuzzyFilterSlash(items, query)`: returns the matched compItems
    with prefix hits ranked first, subsequence-only hits after, each
    group preserving the original `slashItems()` order. Empty query
    lists everything (same as before). Zero matches returns `nil` so
    `updateCompletion` falls through and closes the menu.
  - New `subsequenceMatch(target, query)`: rune-level subsequence test
    on already-case-folded strings. O(n+m), no allocations beyond the
    rune slice.
  - `updateCompletion` slash branch: `filterByPrefix(...)` →
    `fuzzyFilterSlash(...)`. No other call sites touched.

## Tests

8 new tests in `internal/cli/complete_test.go`:

- `TestFuzzyFilterSlashSubsequence` — `/mdl` hits `/model`, not `/mcp`
- `TestFuzzyFilterSlashPrefixFirst` — `/mo` ranks `/model` first
- `TestFuzzyFilterSlashCaseInsensitive` — `/COMP` hits `/compact`
- `TestFuzzyFilterSlashEmptyQueryMatchesAll` — bare `/` lists everything
- `TestFuzzyFilterSlashNoMatchClosesMenu` — `/xzqzqz` closes the menu
- `TestFuzzyFilterSlashAppliesToCustomCommands` — custom commands covered
- `TestFuzzyFilterSlashAcceptFillsInput` — Tab/accept end-to-end
- `TestSubsequenceMatchUnit` — direct matcher coverage incl. negatives

All 4 pre-existing slash-completion tests still pass. `gofmt -l` and
`go vet ./internal/cli/...` are clean. The 4 unrelated failures in
`internal/cli/...` (`TestApplyMCPModeDropsLegacyTier`,
`TestModelRefsSkipsUnconfigured`, `TestOneLine`, `TestRenderRewindSmoke`)
are pre-existing on `main-v2` and were verified with `git stash` to be
unrelated to this change.

## Risk

Low. The matcher is a strict superset of the previous prefix filter:
every input that matched before still matches, and the menu's
"open/close" contract is preserved (empty = list all, no match = close).
The accept/insert path is untouched, so Tab still fills the input with
the top-ranked hit's `insert` string.

## Checklist

- [x] Rebased onto current `upstream/main-v2` (was 51 commits behind)
- [x] `gofmt -l internal/cli/...` clean on the LF content git will push
- [x] `go vet ./internal/cli/...` clean
- [x] New + existing slash-completion tests pass
- [x] PR targets `main-v2` per CONTRIBUTING.md

### DIFF
--- a/internal/cli/complete.go
+++ b/internal/cli/complete.go
@@ -126,7 +126,7 @@ func (m *chatTUI) updateCompletion() {
 		}
 		if !strings.ContainsAny(val, " \t\n") {
 			// Still naming the command itself.
-			if items := filterByPrefix(m.slashItems(), val); len(items) > 0 {
+			if items := fuzzyFilterSlash(m.slashItems(), val); len(items) > 0 {
 				m.setCompletion(compSlash, items, 0)
 				return
 			}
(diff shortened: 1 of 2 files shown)


esengine (approved):
﻿Reviewed: the subsequence matcher is correct and minimal, prefix hits ranking ahead of subsequence-only hits is the right UX call, and the empty-query / no-match contracts deliberately mirror the old prefix filter so the menu open/close behavior is unchanged. Keeping typed-input resolution and the `slash_command` tool on exact lookup is exactly the scope I'd have asked for. Test coverage is thorough — the unit table on `subsequenceMatch` plus the end-to-end accept path is the right split. Merging, thanks!

PR #3846, opened 2026-06-10
feat(desktop): enable fuzzy @-search with directory support in Composer

### LINKED ISSUE #3769: [Feature]: @选择上下文无法模糊搜索的问题
### Version line

1.0版本之前选择上下文是可以@模糊搜索的，现在只能一层一层，

Image非常麻烦，像之前这种，我直接输入@planind,就可以吧这个目录下的index.tsx直接提示出来的 很方便

Image

### Proposed solution

和老版保持一致，恢复模糊搜索

SuMuxi66:
## Summary

Restore the v1-style fuzzy `@file` search behavior in the Composer so that
an intermediate directory segment in a query (e.g. `@planind`) surfaces
nested matches such as `src/planind/index.tsx`, and directory names like
`assets/` are selectable from the @-menu.

Previously the backend matched only on `d.Name()` (basename), the frontend
filter reduced `searchEntries` to basename-only, and `SearchFileRefs`
hardcoded `IsDir: false` for all results — so deep paths and directories
were invisible to the @-menu.

Closes #3769.

## What changed

- `internal/fileref/search.go`
  - `Search` now returns `[]SearchResult` (with `Path` and `IsDir`) instead of
    `[]string`.
  - Matches are collected into three tiers: basename hits (preferred), path-
    segment hits (fallback), and directory-name hits (lowest). A `dirQuota`
    of 5 ensures directories are never fully crowded out by file matches.
  - `pathSegmentContains(relSlash, queryLower)` checks any slash-separated
    path segment for a match.
  - Noise directories (`.git`, `node_modules`, `build`, `dist`, ...) are
    still skipped. The public API signature is unchanged.

- `desktop/app.go`
  - `SearchFileRefs` now propagates `r.IsDir` from `SearchResult` to
    `DirEntry.IsDir`, fixing a pre-existing bug where all results were
    hardcoded `IsDir: false`.

- `internal/cli/complete.go`
  - Adapted to the new `[]SearchResult` return type (CLI only uses `Path`).

- `desktop/frontend/src/lib/atMatches.ts` (new)
  - Pure helper `filterAtMatches(entries, searchEntries, atFrag)` that
    matches against the full slash-normalized relative path and de-duplicates
    by `e.name` with the local list taking precedence.

- `desktop/frontend/src/__tests__/at-matches.test.ts` (new)
  - 5 cases covering nested-path surfacing, basename match, de-duplication,
    local-before-fuzzy ordering, and the empty-fragment legacy behavior.

- `desktop/frontend/src/components/Composer.tsx`
  - The `atMatches` `useMemo` now delegates to `filterAtMatches`. The rest
    of the @-menu logic, the `atDir !== ""` guard, and the directory
    navigation UX are intentionally untouched.

- `internal/fileref/search_test.go` (new)
  - 6 cases covering path-segment match, basename match, directory match
    with `IsDir=true`, sort order between tiers, and noise-directory skip.

## Test results

- `go test ./internal/fileref -count=1` — 6/6 PASS
- `pnpm exec tsx src/__tests__/at-matches.test.ts` — 5/5 PASS
- `go build ./...` — clean
- Manual Wails dev test — directories and files both surface correctly

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -3532,10 +3532,10 @@ func (a *App) SearchFileRefs(query string) []DirEntry {
 	if err != nil {
 		return nil
 	}
-	paths := fileref.Search(base, query, fileRefSearchLimit)
-	out := make([]DirEntry, 0, len(paths))
-	for _, path := range paths {
-		out = append(out, DirEntry{Name: path, IsDir: false})
+	results := fileref.Search(base, query, fileRefSearchLimit)
+	out := make([]DirEntry, 0, len(results))
+	for _, r := range results {
+		out = append(out, DirEntry{Name: r.Path, IsDir: r.IsDir})
 	}
 	return out
 }
--- a/desktop/frontend/src/__tests__/at-matches.test.ts
+++ b/desktop/frontend/src/__tests__/at-matches.test.ts
@@ -1,0 +1,96 @@
+// Run: tsx src/__tests__/at-matches.test.ts
+//
+// Regression coverage for the Composer @-menu filter (issue #3769).
+// The frontend must mirror the backend fuzzy @-search contract:
+// a fragment like "planind" should surface "src/planind/index.tsx"
+// because the search results return full slash-normalized relative
+// paths, not basenames.
+
+import { filterAtMatches } from "../lib/atMatches";
+import type { DirEntry } from "../lib/types";
+
+let passed = 0;
+let failed = 0;
+
+function eq(a: unknown, b: unknown, label: string) {
+  if (JSON.stringify(a) === JSON.stringify(b)) {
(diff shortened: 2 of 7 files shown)


esengine (approved):
﻿Reviewed. The three-tier match (basename → path-segment → directory) with a small directory quota is the right ranking, and the switch in the walk callback means tier 2 is effectively "non-basename segments only" — no double counting. Dropping the early `SkipAll` is the necessary cost of tiering and the walk stays bounded by `maxWalkEntries`. Propagating `IsDir` out of `SearchFileRefs` fixes a genuine pre-existing bug, and the pure `filterAtMatches` helper with its own test file is exactly how I want Composer logic factored. Both test suites are meaningful. Merging, thanks!

PR #3847, opened 2026-06-10
fix(katex): convert aligned→align when \tag is present

lightfront:
## Problem

KaTeX`s `aligned`/`gathered` environments treat the entire block as one
equation and only permit a single `\tag` per block (parse error "Multiple
\tag"). When LLMs emit `\tag{...}` inside `aligned` — as in:

```latex
$$
\boxed{\begin{aligned}
r_A E_\pi(k;0) &= B(k^2) \tag{10}\\
F_R(k;0) + 2r_A F_\pi(k;0) &= A(k^2) \tag{11}
\end{aligned}}
$$
```

KaTeX throws `KaTeX parse error: Multiple \tag`, which `rehype-katex`
renders as `<span class="katex-error" style="color:#cc0000">` —
showing red error text in the web UI.

## Fix

Instead of stripping `\tag` commands (which loses equation numbers),
convert the environments when `\tag` is present:

- `\begin{aligned}` → `\begin{align}`
- `\begin{gathered}` → `\begin{gather}`
- `\begin{alignedat}{N}` → `\begin{alignat}{N}`

KaTeX's `align`/`gather`/`alignat` support `\tag` on every row
natively. Plain `aligned` blocks without `\tag` are left unchanged
so they keep their un-numbered behaviour. The conversion is only
applied when `\tag` is detected in the math source.

```ts
if (/\\tag\*?\s*\{/.test(source)) {
  source = source
    .replace(/\\begin\{alignedat\}\{(\d+)\}/g, "\\begin{alignat}{$1}")
    .replace(/\\end\{alignedat\}/g, "\\end{alignat}")
    .replace(/\\begin\{aligned\}/g, "\\begin{align}")
    .replace(/\\end\{aligned\}/g, "\\end{align}")
    .replace(/\\begin\{gathered\}/g, "\\begin{gather}")
    .replace(/\\end\{gathered\}/g, "\\end{gather}");
}
```

## Files changed

- `desktop/frontend/src/components/latexNormalize.ts` — aligned→align
  conversion when `\tag` is present
- `desktop/frontend/src/__tests__/math-golden.test.ts` — 6 new unit
  tests + 1 updated e2e test

## Testing

All 99 golden tests pass (`npx tsx src/__tests__/math-golden.test.ts`).

### DIFF
--- a/desktop/frontend/src/__tests__/math-golden.test.ts
+++ b/desktop/frontend/src/__tests__/math-golden.test.ts
@@ -63,6 +63,19 @@ eq(latexNormalizeForKatex("\\tfrac{a}{b}"), "\\tfrac{a}{b}", "nested braces in c
 eq(latexNormalizeForKatex("\\|x\\|"), "\\|x\\|", "\\| is left alone (readCommand handles \\|, not | branch)");
 eq(latexNormalizeForKatex("\\\\|x|"), "\\\\\\vert x\\vert", "\\\\| line break + pipe: both | → \\vert");
 
+// ── latexNormalizeForKatex — \tag → align conversion (regression for KaTeX "Multiple \tag") ──
+eq(latexNormalizeForKatex("a = b \\tag{10}"), "a = b \\tag{10}", "\\tag without aligned passes through");
+eq(latexNormalizeForKatex("\\begin{aligned} a &= b \\\\ \\end{aligned}"),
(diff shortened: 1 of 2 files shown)


esengine (approved):
Nice fix — converting to align/gather/alignat instead of stripping \tag keeps the equation numbers, and gating on \tag presence means plain aligned blocks keep their semantics. The e2e tests actually rendering through KaTeX (including the boxed nesting) is exactly the verification this needed. Thanks!

PR #3854, opened 2026-06-10
fix: ContextPanel token breakdown uses mismatched units (snapshot vs cumulative)

Diarica:
The workspace overview donut chart shows broken segment proportions after turn 1 because `UsedTokens` comes from `ContextSnapshot()` (per-turn snapshot) but `PromptTokens`/`CompletionTokens`/`ReasoningTokens` came from `sessionUsageStats` (cumulative across all turns). After the first turn, cumulative >> snapshot, making "other" always 0 and segment percentages wildly inflated.

Fix: read `PromptTokens`/`CompletionTokens`/`ReasoningTokens`/`CacheHitTokens`/`CacheMissTokens` from `ctrl.LastUsage()` — the same per-turn snapshot `UsedTokens` already uses. Cumulative fields (`RequestCount`, `ElapsedMs`, `SessionCost`) stay on `telemetry.Usage`.

- **Files:** `desktop/tabs.go` (+10/-5)

### DIFF
--- a/desktop/tabs.go
+++ b/desktop/tabs.go
@@ -2879,6 +2879,16 @@ func (a *App) ContextPanel(tabID string) ContextPanelInfo {
 		used, window := ctrl.ContextSnapshot()
 		info.UsedTokens = used
 		info.WindowTokens = window
+		// Per-turn token breakdown from LastUsage (same snapshot as UsedTokens)
+		// so the donut segments are proportional to the current context fill,
+		// not inflated by cumulative session totals.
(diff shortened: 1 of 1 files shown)


esengine (approved):
Good catch. The donut was mixing per-turn snapshot (UsedTokens) with cumulative session totals, so the breakdown segments went meaningless after the first turn. Reading the breakdown from the same LastUsage() snapshot lines them all up, and keeping RequestCount/ElapsedMs/SessionCost on the cumulative Usage is exactly right. Thanks!

PR #3856, opened 2026-06-10
fix(windows): add SafeJoin utility to prevent cross-drive path corruption

HHHEEEWWW:
## Problem

Go's filepath.Join on Windows does NOT reset the result when joining elements from different drives. For example:

The rightmost absolute path should anchor the result. This is the root cause of issue #3850 where a global memory path (C: drive) was concatenated onto a project path (E: drive), causing the corrupt directory tree .

ref: https://github.com/golang/go/issues/51619

### 1. New  utility

Walks elements right-to-left and anchors on the rightmost absolute element:

### 2. Applied to critical call sites
-  - snapshot restore path resolution
-  - git status path resolution

### 3. Documented as preferred pattern
-  (internal/tool/builtin/workspace.go) - already had manual IsAbs guard, now points to SafeJoin as the canonical approach
-  (internal/memory/doc.go) - same treatment

## Testing

All 5 unit tests pass across Windows and Unix (cross-drive tests are Windows-only, skipped on other platforms).

## Files changed
-  (new) - SafeJoin implementation
-  (new) - tests
-  - use SafeJoin in safePath
-  - use SafeJoin in workspaceRelPathFromGitStatus
-  - comment reference
-  - comment reference

### DIFF
--- a/desktop/workspace_changes.go
+++ b/desktop/workspace_changes.go
@@ -8,6 +8,7 @@ import (
 	"sort"
 	"strings"
 
+	"reasonix/internal/fileutil"
 	"reasonix/internal/proc"
 )
 
@@ -176,7 +177,7 @@ func workspaceRelPathFromGitStatus(repoRoot, base, path string) string {
 		return ""
 	}
 	if !filepath.IsAbs(path) {
-		path = filepath.Join(repoRoot, filepath.FromSlash(path))
+		path = fileutil.SafeJoin(repoRoot, filepath.FromSlash(path))
 	}
 	return normalizeWorkspaceRelPath(base, path)
 }
--- a/internal/checkpoint/checkpoint.go
+++ b/internal/checkpoint/checkpoint.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"reasonix/internal/diff"
+	"reasonix/internal/fileutil"
 	fileenc "reasonix/internal/fileutil/encoding"
 )
 
@@ -300,7 +301,7 @@ func detectCurrentEncoding(path string) *fileenc.Kind {
 func safePath(root, p string) (string, error) {
 	abs := p
 	if !filepath.IsAbs(abs) {
-		abs = filepath.Join(root, p)
+		abs = fileutil.SafeJoin(root, p)
 	}
 	abs = filepath.Clean(abs)
 	if root != "" {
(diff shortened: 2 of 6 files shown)


esengine (changes requested):
﻿Thanks for the clean utility and the cross-drive test matrix — `SafeJoin` itself is correct, and CI is green across all three platforms. But I traced #3850 to the bottom before merging, and as written this PR cannot change the observed behavior. Two findings:

**1. Both call-site changes are behavioral no-ops.** In `checkpoint.go:safePath` and `workspace_changes.go:workspaceRelPathFromGitStatus`, the join only runs inside a `!filepath.IsAbs(path)` branch. For a relative `path`, `SafeJoin(root, path)` and `filepath.Join(root, path)` return byte-identical results (the rightmost-absolute scan finds only `root`). So no input reaches the cross-drive logic at either site.

**2. The corruption in #3850 is not produced by the `C:\...` form at all.** Reproduced on Windows (Go 1.24):

```
filepath.Join(`E:\proj`, `C:\Users\59263`)  → E:\proj\C:\Users\59263   // mkdir fails with EINVAL — can't create "C:" mid-path
filepath.Join(`E:\proj`, `/c/Users/59263`)  → E:\proj\c\Users\59263    // exactly the tree from the issue: lowercase c, no colon
filepath.IsAbs(`/c/Users/59263`)            → false
```

The on-disk `c/Users/...` tree can only come from an **MSYS-style path** (`/c/Users/...`) entering a join — and since `filepath.IsAbs` returns false for that form, every existing `IsAbs` guard *and* `SafeJoin` pass it straight through. The likely entry point is `resolveIn` in `internal/tool/builtin/workspace.go` (model-supplied tool paths land there), which this PR only annotates with a comment.

**What I'd like to see:**
- Teach `SafeJoin` (or a `NormalizeWindowsPath` step applied in `resolveIn`) to recognize, on Windows, the MSYS drive form `/c/...` → `C:\...` and the drive-relative/rooted forms, before deciding whether an element is absolute.
- A regression test that feeds `/c/Users/...` through `resolveIn`/`SafeJoin` and asserts the result does NOT nest under the project root — that's the test that actually pins #3850.
- Keep the current `SafeJoin` + tests; they're a correct hardening even though they're not the fix.

The investigation in the issue thread was genuinely useful for narrowing this down — with the entry point covered this will be a solid, complete fix.

HHHEEEWWW:
让我再找找相关的问题出现情况，将会提交一个可以复现问题的脚本，后续将再次提交isu

PR #3859, opened 2026-06-10
feat(cli): add self-update command with tag namespace filtering

douglarek:
Introduces 'reasonix update [--check] [--force]' that checks GitHub Releases for a newer CLI version, downloads the platform-matched archive, verifies its SHA-256 checksum, and atomically replaces the running binary.

Uses /releases (list) instead of /releases/latest to filter by CLI tag namespace (v*), correctly skipping desktop-v* and npm-v* releases that share the same repository.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -100,6 +100,8 @@ func Run(args []string, version string) int {
 	case "bot":
 		configureCLIThemeFromConfigNoProbe()
 		return botCommand(rest, version)
+	case "update":
+		return updateCommand(rest, version)
 	case "version", "--version", "-v":
 		fmt.Println("reasonix", version)
 		return 0
@@ -115,7 +117,7 @@ func Run(args []string, version string) int {
 
 func shouldMigrateLegacyConfigForCLI(cmd string) bool {
 	switch cmd {
-	case "", "run", "chat", "code", "serve", "setup", "config", "init", "acp", "mcp", "codegraph", "doctor", "bot":
+	case "", "run", "chat", "code", "serve", "setup", "config", "init", "acp", "mcp", "codegraph", "doctor", "bot", "update":
 		return true
 	default:
 		return false
--- a/internal/cli/update.go
+++ b/internal/cli/update.go
@@ -1,0 +1,458 @@
+// Package cli implements the `reasonix update` subcommand: it checks GitHub
+// Releases for a newer version, downloads the platform-matched archive,
+// verifies its SHA-256 checksum, and atomically replaces the running binary.
+package cli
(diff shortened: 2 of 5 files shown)


esengine (changes requested):
﻿This is a well-built foundation — the tag-namespace filter is exactly right for our three-namespace release layout (`v*` / `npm-v*` / `desktop-v*`), the GoReleaser asset-name matching and SHA256SUMS parsing line up with what the release workflow actually publishes, and the helper coverage is thorough. Three things need fixing before this can ship, two of them Windows/integrity critical:

**1. The install step breaks on Windows.** `atomicReplace` renames the temp file over `exePath` — but on Windows the running executable is memory-mapped, and `os.Rename` onto it fails with `ERROR_ACCESS_DENIED`. The standard dance is: rename the running exe aside (`reasonix.exe` → `reasonix.exe.old`, allowed even while running), rename the new binary into place, then best-effort delete the `.old` (and clean up stale `.old` files on a later run, since the delete fails while the old image is still loaded). The desktop updater under `desktop/internal/update` already deals with this — worth borrowing its approach. As written, `reasonix update` will download, verify, and then error out on the platform with the most users.

**2. Checksum verification must fail closed.** Right now a failed SHA256SUMS download (or a missing entry) prints a warning and installs the unverified binary anyway. The checksum is the only integrity gate on something we're about to execute; a CDN hiccup shouldn't silently degrade into an unverified install. Make it a hard error, and if you want an escape hatch, an explicit `--skip-verify` flag is fine.

**3. Filter prereleases.** `fetchLatestRelease` takes the first `v*` entry from `/releases`, which includes prereleases — a published `v1.6.0-rc.1` would be offered to every stable user (and `compareSemver` strips the suffix, so it also reads as equal to `v1.6.0`). Skip releases with `"prerelease": true` and tags containing `-` unless an opt-in flag asks for them.

Two non-blocking notes: the i18n catalogues gained entries but `updateCommand` prints hardcoded English — fine to unify in a follow-up; and a one-line warning when the binary lives under an npm or Homebrew install root would save users from package-manager drift, but I won't hold the PR on it.

Happy to merge promptly once the three above are addressed — the structure is right, it's the platform edges that need finishing.

esengine:
﻿Heads-up: #3941 is a parallel implementation of the same command. I've reviewed both against the same checklist — your namespace filter is the part that PR got wrong, while it has the fail-closed checksum and i18n that this one is missing; the Windows running-exe replacement blocks both. First PR to address all points lands and the other gets closed with credit to both — or feel free to team up on one branch, the complete implementation already exists across your two diffs.

PR #3876, opened 2026-06-10
fix(desktop): use DesktopLanguage for i18n slash command descriptions

### LINKED ISSUE #3875: [Bug]: Desktop slash command descriptions not localized when language set in desktop settings
## Description

When the user selects Chinese (or any non-English language) in the desktop app settings, the `/` command descriptions in the chat input still show English text.

**Expected**: `/new` shows "开启新会话并保存历史"
**Actual**: `/new` shows "start new session; save transcript"

## Steps to Reproduce

1. Open Reasonix desktop app
2. Go to Settings → Language → Select "中文"
3. Open a chat session
4. Type `/` in the input field
5. Observe that command descriptions are still in English

## Root Cause

The desktop language setting is stored in `DesktopLanguage` field, but the i18n system initialization in `desktop/app.go` only reads from `cfg.Language` (the CLI setting):

```go
// Current code (desktop/app.go ~line 360)
if cfg, err := config.Load(); err == nil {
    i18n.DetectLanguage(cfg.Language)  // cfg.Language is empty when set via desktop settings
}
```

When the user sets language via desktop settings, `SetDesktopLanguage()` only updates `DesktopLanguage`, not `Language`. So `cfg.Language` remains empty, and i18n defaults to English.

## Proposed Fix

Fall back to `cfg.DesktopLanguage()` when `cfg.Language\* is empty:

```go
if cfg, err := config.Load(); err == nil {
    lang := cfg.Language
    if lang == "" {
        lang = cfg.DesktopLanguage()
    }
    i18n.DetectLanguage(lang)
}
```

## Environment

- **OS**: Windows 11
- **Reasonix version**: 1.5.0 (main-v2)
- **Desktop app**: Yes

## Related PR

- https://github.com/Bernardxu123/DeepSeek-Reasonix/pull/4

Bernardxu123:
## Summary

Fixes #3875

When the user selects a language in desktop settings, the `/` command descriptions in chat still show English because the i18n system only reads from `cfg.Language` (CLI setting), not `cfg.DesktopLanguage()` (desktop setting).

## Changes

**`desktop/app.go`**
- When initializing i18n, fall back to `cfg.DesktopLanguage()` when `cfg.Language` is empty
- This ensures the desktop language preference is respected for slash command descriptions

```go
// Before
if cfg, err := config.Load(); err == nil {
    i18n.DetectLanguage(cfg.Language)
}

// After
if cfg, err := config.Load(); err == nil {
    lang := cfg.Language
    if lang == "" {
        lang = cfg.DesktopLanguage()
    }
    i18n.DetectLanguage(lang)
}
```

## Testing

```bash
go build ./...
```

Build succeeds ✅

## Backward Compatibility

- Fully backward compatible
- Does not affect CLI users (they set language via `reasonix.toml` or env vars)
- Only affects desktop app users who set language in settings

Signed-off-by: Bernardxu123 <bernardxu123@users.noreply.github.com>

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -352,7 +352,7 @@ func backgroundRestoreShouldMaximise(goos string, wasMaximised bool) bool {
 // restoreOrBuildTabs restores the tabs from the last session, or creates a
 // default Global tab on first launch.
 func (a *App) restoreOrBuildTabs() {
-// Load i18n from the first available config.
+	// Load i18n from the first available config.
 	// Prefer DesktopLanguage (desktop UI setting) over Language (CLI setting),
 	// so the user's language choice in desktop settings takes effect.
 	if cfg, err := config.Load(); err == nil {
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -352,12 +352,15 @@ func backgroundRestoreShouldMaximise(goos string, wasMaximised bool) bool {
 // restoreOrBuildTabs restores the tabs from the last session, or creates a
 // default Global tab on first launch.
 func (a *App) restoreOrBuildTabs() {
-	ctx := a.ctx
-	ensureWorkspace()
-
-	// Load i18n from the first available config.
+// Load i18n from the first available config.
+	// Prefer DesktopLanguage (desktop UI setting) over Language (CLI setting),
+	// so the user's language choice in desktop settings takes effect.
 	if cfg, err := config.Load(); err == nil {
-		i18n.DetectLanguage(cfg.Language)
+		lang := cfg.Language
+		if lang == "" {
+			lang = cfg.DesktopLanguage()
+		}
+		i18n.DetectLanguage(lang)
 	}
 
 	f := loadTabsFile()
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -352,6 +352,9 @@ func backgroundRestoreShouldMaximise(goos string, wasMaximised bool) bool {
 // restoreOrBuildTabs restores the tabs from the last session, or creates a
 // default Global tab on first launch.
 func (a *App) restoreOrBuildTabs() {
+	ctx := a.ctx
+	ensureWorkspace()
+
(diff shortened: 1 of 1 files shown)


esengine (changes requested):
Good catch on the root cause — desktop i18n init only ever read `cfg.Language`, so the desktop language picker never reached the slash descriptions. Right place to fix it, too.

One substantive point before merging: the comment and the code disagree, and I think the comment has it right. The comment says "Prefer DesktopLanguage (desktop UI setting) over Language (CLI setting)", but the code makes `cfg.Language` win and only falls back to `DesktopLanguage()` when it's empty. For a user with `language = "en"` in their toml who then picks 中文 in desktop Settings, the frontend UI switches to Chinese but slash descriptions stay English — which is exactly the mixed-language state #3875 complains about.

In the desktop app the desktop-specific setting is the user's most direct intent (it's also what the rest of the desktop UI follows), so the precedence should be:

```go
lang := cfg.DesktopLanguage()
if lang == "" {
    lang = cfg.Language
}
i18n.DetectLanguage(lang)
```

That matches your comment as written, keeps CLI users unaffected (they have no desktop.language), and fully fixes the issue for users with both set. Happy to merge right after that flip.

Bernardxu123:
Fixed! Thanks for the clear explanation.

Changed the precedence to:

```go
lang := cfg.DesktopLanguage()
if lang == "" {
    lang = cfg.Language
}
i18n.DetectLanguage(lang)
```

Now DesktopLanguage takes priority, matching the comment and the rest of the desktop UI behavior. CLI users are unaffected since they have no desktop.language set.

esengine (approved):
﻿That's exactly the flip — `DesktopLanguage()` first, CLI `language` as fallback, matching the comment and the rest of the desktop UI. CLI users are untouched (no `desktop.language`), and the mixed-language state from #3875 is gone for users with both set. Merging, thanks for the quick turnaround.

PR #3881, opened 2026-06-10
fix(site): use relative base so CSS loads on custom domain

Learnmore-smart:
The hardcoded /DeepSeek-Reasonix base prefix caused asset 404s on
reasonix.io. Switch to relative paths so one build works on both
GitHub Pages and the custom domain.

Co-authored-by: Cursor <cursoragent@cursor.com>

### DIFF
--- a/site/astro.config.mjs
+++ b/site/astro.config.mjs
@@ -1,8 +1,14 @@
 import { defineConfig } from 'astro/config';
 
-// Served from GitHub Pages under the repo subpath.
+// `site` is the canonical origin for OG / canonical tags.
+// `base: './'` emits **relative** asset paths so the same build works at
+//   - https://esengine.github.io/DeepSeek-Reasonix/   (GitHub Pages project URL)
+//   - http://reasonix.io/                             (custom domain, served from root)
+// The previous `base: '/DeepSeek-Reasonix'` baked that prefix into every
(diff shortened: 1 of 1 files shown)


esengine:
Thanks a lot for this — and for catching the broken styling on the custom domain! 🙏

The same issue was just fixed and merged in #3880 (now deployed and live on https://reasonix.io), so I'm going to close this one as a duplicate.

A quick note on the approaches, since yours is a reasonable one:

- This PR keeps `site: 'https://esengine.github.io'` and switches to `base: './'` (relative asset paths) to make a single build work on both the project URL and the custom domain.
- #3880 instead points `site` at `https://reasonix.io`, drops `base`, and derives `canonical`/`og:url` from `Astro.site`.

We went with the #3880 approach because reasonix.io is now the canonical home: once the custom domain is set, GitHub 301-redirects the `esengine.github.io/DeepSeek-Reasonix/` URL to it, so the dual-path flexibility isn't needed — and leaving `site` on the old origin would keep the `canonical`/`og:url` tags pointing at `esengine.github.io/DeepSeek-Reasonix/`, which hurts SEO for the new domain. We also preferred an absolute `base` since Astro's `base` is documented for absolute paths.

Really appreciate the fix though — the diagnosis was spot on. Hope to see more PRs from you! 🚀

PR #3878, opened 2026-06-10
fix(desktop): keep workspace dock within viewport

lifu963:
## Summary
- Clamp the right Workspace dock to available viewport width so opening file previews no longer pushes content past the window edge.
- Replace the docked layout's summed min-width with shrinkable grid tracks and keep Workspace resize metadata valid under constrained widths.
- Clarify the Workspace open behavior comment to match the viewport-clamped dock strategy.

## Test plan
- `pnpm --dir desktop/frontend run build` after `wails generate module`
- Manual macOS Wails GUI acceptance at normal window sizes, including Workspace file preview with the sidebar collapsed and expanded

## Notes
- Fixes #3863

Made with [Cursor](https://cursor.com)

PR #3886, opened 2026-06-10
fix(desktop): keep composer usable beside workspace / 保持工作区旁输入区可用

SivanCola:
## Summary
- reserve a more comfortable chat width while the right workspace dock is open
- collapse the workspace file tree into a rail when file preview space is tight
- left-align composer controls so cramped layouts do not show a misleading blank slot
- move lower-priority composer controls into a responsive More menu at medium widths

## Why
#3878 fixed right-side workspace clipping, but very wide workspace layouts could still squeeze the chat composer until bottom controls became hard to read. This follow-up keeps the composer usable while still allowing the workspace to shrink, rail, or maximize.

## Validation
- `pnpm --dir desktop/frontend run check:css`
- `pnpm --dir desktop/frontend run test`
- Browser check at the local dev URL confirmed the composer controls are left-aligned

### DIFF
--- a/desktop/frontend/src/App.tsx
+++ b/desktop/frontend/src/App.tsx
@@ -85,6 +85,7 @@ const SIDEBAR_MIN_WIDTH = 264;
 const SIDEBAR_MAX_WIDTH = 300;
 const SIDEBAR_VIEWPORT_RATIO = 0.18;
 const CHAT_MIN_WIDTH = 400;
+const CHAT_COMFORT_MIN_WIDTH = 560;
 const WORKSPACE_RESIZER_WIDTH = 8;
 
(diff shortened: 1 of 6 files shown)


esengine (approved):
Good follow-up to #3878. The comfort-width reservation (560 when the dock is open, 400 otherwise) plus the 280px min-render threshold is a sensible way to keep the composer readable instead of rendering a useless sliver, and the More-menu overflow reuses the existing intent-menu open/close/reduced-motion pattern cleanly. A11y on the listbox is there, both locales are updated, and tightening hasEffort to require levels avoids an empty menu. Layout-only, no data path. Thanks!

PR #3708, opened 2026-06-09
fix(desktop): 为更新错误横幅添加关闭按钮

whale-fall-ouo:
## 问题描述

当桌面应用启动时如果没有网络连接，自动更新检查会失败并显示错误横幅。该横幅**无法关闭**——只有"重试"按钮，没有关闭/取消选项。横幅会持续显示在所有标签页和会话中，直到网络连接恢复。

## 复现步骤

1. 断开网络（或屏蔽 GitHub 访问）
2. 启动 Reasonix 桌面应用
3. 更新检查失败，显示错误横幅
4. 无法关闭横幅

## 修复方案

在 `UpdateBanner` 组件的 `error` case 中添加 dismiss 按钮，使用已有的 `reset()` 函数将状态重置为 idle。与 `available` case 中的关闭按钮行为一致。

## 改动

- `desktop/frontend/src/components/UpdateBanner.tsx`：解构时加入 `reset`，在 Retry 按钮后添加 dismiss 按钮

Fixes #3699
image

PR #3888, opened 2026-06-10
[codex] Add desktop update check controls / 增加桌面更新检查控制

### LINKED ISSUE #2822: [Bug]: 异常更新提示
### Version line

### Exact version

1.0.0

### What happened?

每次启动都要检查更新，github连接不上时。出现如下异常提醒，且无法关闭

### Steps to reproduce

Image

### OS / platform

windows11

### Relevant logs or output

```shell

```

### LINKED ISSUE #3104: [Bug]:更新失败也不能x掉
### Version line

### Exact version

1.1.0

### What happened?

Image

更新失败也不能x掉

### Steps to reproduce

更新失败也不能x掉

### OS / platform

Windows11

### Relevant logs or output

```shell

```

### LINKED ISSUE #3699: Update error banner has no dismiss button — stuck on screen when offline
## Description

When the desktop app starts without network connectivity, the auto-update check fails and shows an error banner at the top of the window. This banner **cannot be dismissed** — it only has a "Retry" button, with no close/dismiss option. The banner persists across all tabs and sessions until a network connection is available.

## Steps to Reproduce

1. Disconnect from the network (or block GitHub access)
2. Launch Reasonix desktop app
3. The update check fails: `update: fetch manifest: Get "https://github.com/esengine/reasonix/releases/latest/download/latest.json": EOF`
4. Error banner appears with no way to close it

## Expected Behavior

The error banner should have a dismiss/close button, like the "update available" banner already does. Network failures are transient and shouldn't force a persistent UI element on the user.

## Actual Behavior

Only a "Retry" button is shown. The user is forced to either retry indefinitely or stare at the error.

## Suggested Fix

In `desktop/frontend/src/components/UpdateBanner.tsx`, the `error` case should include a dismiss button similar to the `available` case:

```tsx
case "error":
  return (
    <div className="banner banner--error">
      <span className="banner__msg">{t("updater.failed", { msg: status.message })}</span>
      <span className="banner__spacer" />
      <button className="btn btn--small" onClick={() => void check()}>
        {t("updater.retry")}
      </button>
      <button className="btn btn--small" onClick={() => setStatus({ kind: "idle" })}>
        {t("updater.dismiss")}
      </button>
    </div>
  );
```

SivanCola:
## Summary

- Add a desktop setting to enable or disable startup update checks while keeping manual checks available in Settings.
- Add a dismiss action for update error banners so transient network failures do not pin the UI.
- Keep settings helper text on one line with ellipsis and tooltip fallback, including the update preference row.
- Restore the tracked `dist/.gitkeep` placeholder after production frontend builds so Vite can still clear stale assets without leaving a tracked deletion.

## Validation

- `go test ./internal/config`
- `go test .` from the desktop module

## Issues

Fixes #2822.
Fixes #3104.
Fixes #3699.

### DIFF
--- a/desktop/frontend/src/App.tsx
+++ b/desktop/frontend/src/App.tsx
@@ -52,6 +52,7 @@ import {
   type ProjectNode,
   type SessionMeta,
   type SettingsTab,
+  type SettingsView,
   type TabMeta,
   type ToolApprovalMode,
 } from "./lib/types";
@@ -428,6 +429,7 @@ export default function App() {
   // clearing the key mid-session is the Settings panel's job, not the gate's.
   const [needsOnboarding, setNeedsOnboarding] = useState<boolean | null>(null);
   const [settingsTarget, setSettingsTarget] = useState<SettingsTab | null>(null);
+  const [startupUpdateChecksEnabled, setStartupUpdateChecksEnabled] = useState<boolean | null>(null);
   const [histView, setHistView] = useState<HistoryViewState | null>(null);
   const [paletteOpen, setPaletteOpen] = useState(false);
   const [paletteSessions, setPaletteSessions] = useState<SessionMeta[]>([]);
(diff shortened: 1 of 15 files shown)


esengine (approved):
Well-rounded. The *bool default (nil → true via DesktopCheckUpdates) keeps existing users on auto-check, the AppBindings interface + mock are both updated so the bridge contract doesn't drift, and both locales are in sync. Gating the banner on startupUpdateChecksEnabled === true (initial null) is a nice touch — it avoids firing a check before the preference loads. Good Go test coverage on persistence + scoped render. The error-banner dismiss here supersedes #3708, which I'll close with thanks. Thanks!

PR #3891, opened 2026-06-10
Fix macOS top chrome drag rail / 修复 macOS 顶部窗口拖拽区域

SivanCola:
## Summary
- Add a macOS-only invisible top drag rail above the tab strip so the window remains draggable even when tabs fill the chrome.
- Keep tab controls and buttons as `no-drag`, while the rail exposes a subtle full-width hover indicator and accent-colored active state.
- Leave Windows and Linux behavior unchanged.

## Root Cause
Reasonix uses an inset/hidden macOS title bar, so the webview must explicitly provide draggable regions. The tab strip is interactive and opts out of Wails dragging, which can leave no reliable drag target when the top chrome is full of tabs.

## Validation
- Browser preview with `platform=darwin` verified that the rail is the hit target, the rail is draggable, tab controls remain `no-drag`, and hover/active indicators render at the same full width.

Related: #3843, #3852, #3853

### DIFF
--- a/desktop/frontend/src/components/AppChrome.tsx
+++ b/desktop/frontend/src/components/AppChrome.tsx
@@ -91,6 +91,7 @@ export function AppChrome({
           <span />
         </div>
       )}
+      {darwinChrome && <span className="app-chrome__drag-rail" aria-hidden="true" />}
       <button
         className={[
           "app-chrome__panel-toggle",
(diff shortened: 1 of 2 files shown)


esengine (approved):
Nice — a dedicated invisible drag rail scoped to .app--darwin .app-chrome--tabs is the clean way to restore a draggable target when the tab strip eats the chrome, and leaving the tab controls as no-drag keeps them clickable. Windows/Linux untouched. Thanks!

PR #3893, opened 2026-06-10
fix(desktop): close stuck thinking window after stream interruption (#3746)

### LINKED ISSUE #3746: [Bug]: 无法关闭思考窗口
### Version line

### Exact version

1.4.0

### What happened?

 Image
点击或者ESC无法关闭，这应该是思考的窗口吧，可以继续对话，如果新的对话键入可以正常继续

### Steps to reproduce

目测是思考过程冲，连接中断，如模型服务中断了，即使后续模型重启服务也还是这样

### OS / platform

windows11

### Relevant logs or output

```shell

```

ashishexee:
Closes #3746

## Problem

When the model service is interrupted during the thinking/reasoning phase, the "thinking window" (ProcessCard) gets stuck open and cannot be closed by clicking or pressing ESC. The UI shows a perpetual "running" spinner.

## Solution

5 changes across 5 files (+113, -1):

### Fix 1: Stale-stream watchdog (`useController.ts`)

Top-level `useEffect` watchdog that detects when the frontend has `running + live` set but no token events (`text`/`reasoning`) have arrived for 30 seconds. Uses a `lastTokenAt` ref (reset on `turn_started`/`text`/`reasoning`) and explicit deps `[activeTabId, reconcileTabRuntime, activeState.running, activeState.live]`. Calculates elapsed time and schedules a timeout for the remaining delay, with a double-check guard inside the callback. When triggered, calls `reconcileTabRuntime()` which queries the backend and dispatches `backend_status: {running: false}` if the backend is no longer running.

Addresses review feedback:
- **Hoisted** from inside the event-subscription `useEffect` (was a nested hook → Rules of Hooks violation / crash) to a **top-level sibling** effect.
- **Replaced** the no-deps "reset on every render" trick with a `lastTokenAt` ref + explicit deps, so the timer only re-arms when `running`/`live`/`activeTabId` change, and correctly measures time since the last token event.
- **Added** `turn_started` to the `lastTokenAt` reset events to prevent spurious reconcile on fresh turns where `lastTokenAt.current` was still 0 or stale from a previous turn.

### Fix 2: Panic recovery (`controller.go`)

Adds `defer recover()` to the `runGuarded` goroutine. On panic: sets `c.running = false`, emits `TurnDone` with error. On normal path: `recover()` returns nil, guard is a no-op — no double `TurnDone`.

### Fix 3: Controlled `open` on reasoning card (`Message.tsx`)

Replaces `defaultOpen={item.streaming}` (uncontrolled, one-time `useState` initializer that ignores prop changes) with **controlled mode**: `useState(item.streaming)` seeds `reasoningOpen`, and `open={reasoningOpen}` + `onOpenChange={setReasoningOpen}` gives full control. No `useEffect` auto-close — the card stays open after normal completion (original UX: user can review reasoning), and the stuck-spinner is resolved by the watchdog transitioning the status icon from "running" to "done".

Addresses review feedback:
- Completed turns from disk: `useState(false)` → card starts collapsed (matches original `defaultOpen={false}`).
- Streaming turns: `useState(true)` → card starts open, stays open after completion.
- Watchdog stuck case: when `streaming → false`, the status icon changes from spinner to checkmark, so the card no longer appears "stuck" even though it remains open.

### Fix 4: Scoped ESC handler (`ProcessCard.tsx`)

Replaces the global `document.addEventListener("keydown")` (which closed ALL open cards on a single ESC and collided with ESC elsewhere) with a scoped `onKeyDown` on the `<button>` element. ESC only closes the card whose header button currently has focus. Removes `useEffect` import entirely.

### Fix 5: Panic recovery tests (`controller_test.go`)

Two new tests:
- `TestRunGuardedPanicEmitsTurnDone` — verifies panic recovery emits `TurnDone` with error and sets `c.running = false`
- `TestRunGuardedPanicDoesNotDoubleEmitTurnDone` — verifies no double `TurnDone` emission on the panic path

## Testing

```
go vet ./internal/control/...  — clean
go test ./internal/control/... — 22/22 pass (including 2 new panic recovery tests)
git diff --check               — clean
```

### DIFF
--- a/desktop/frontend/src/components/Message.tsx
+++ b/desktop/frontend/src/components/Message.tsx
@@ -245,6 +245,7 @@ export const AssistantMessage = memo(function AssistantMessage({
   const hasText = item.streaming || item.text.trim() !== "";
   const processOnly = Boolean(item.reasoning) && !hasText;
   const processWithText = Boolean(item.reasoning) && hasText;
+  const [reasoningOpen, setReasoningOpen] = useState(item.streaming);
   return (
     <div className={`msg msg--assistant${processOnly ? " msg--process-only" : ""}${processWithText ? " msg--process-with-text" : ""}`}>
       {item.reasoning && (
@@ -259,7 +260,8 @@ export const AssistantMessage = memo(function AssistantMessage({
               <span>{item.streaming ? t("msg.thinkingRunning") : t("msg.thinkingDone")}</span>
             </>
           }
-          defaultOpen={item.streaming}
+          open={reasoningOpen}
+          onOpenChange={setReasoningOpen}
         >
           <div className="reasoning__body">{item.reasoning}</div>
         </ProcessCard>
--- a/desktop/frontend/src/components/ProcessCard.tsx
+++ b/desktop/frontend/src/components/ProcessCard.tsx
@@ -142,6 +142,7 @@ export function ProcessCard({
         type="button"
         className="process-card__head"
         onClick={toggle}
+        onKeyDown={(e) => { if (e.key === "Escape") { e.preventDefault(); toggle(); } }}
         aria-expanded={hasBody ? actualOpen : undefined}
       >
         <span className="process-card__icon">{icon}</span>
--- a/desktop/frontend/src/lib/useController.ts
+++ b/desktop/frontend/src/lib/useController.ts
@@ -477,6 +477,7 @@ async function refreshMetaForTab(tabId: string, dispatchTo: (tabId: string, acti
 
 export function useController() {
   const statesRef = useRef<TabStates>(new Map());
+  const lastTokenAt = useRef(0);
   const [activeTabId, setActiveTabId] = useState<string | undefined>();
   const activeTabIdRef = useRef<string | undefined>(undefined);
   // A render-triggering counter so that mutations to a non-active tab's state still
@@ -599,6 +600,9 @@ export function useController() {
     const off = onEvent((e) => {
       const targetTabId = e.tabId || activeTabIdRef.current;
       if (!targetTabId) return;
+      if (e.kind === "turn_started" || e.kind === "text" || e.kind === "reasoning") {
+        lastTokenAt.current = Date.now();
+      }
       if (e.kind === "text" || e.kind === "reasoning") {
         textBatch.push({ tabId: targetTabId, e });
       } else {
@@ -638,6 +642,28 @@ export function useController() {
 
     return () => { textBatch.drain(); off(); offReady(); };
   }, [dispatchTo, loadSessionDataForTab, refreshCheckpoints, syncActiveTabFromBackend]);
+
+  // Stale-stream watchdog: if the frontend thinks the agent is running but
+  // no token events have arrived for 30 seconds, reconcile with the backend.
+  // This catches the case where the Wails event channel silently drops the
+  // turn_done event after a model-service interruption (#3746).
+  useEffect(() => {
+    if (!activeTabId) return;
+    const s = statesRef.current.get(activeTabId);
+    if (!s?.running || !s.live) return;
+    const since = Date.now() - lastTokenAt.current;
+    if (since >= 30_000) {
+      void reconcileTabRuntime(activeTabId);
+      return;
+    }
+    const timer = window.setTimeout(() => {
+      const cur = statesRef.current.get(activeTabId);
+      if (cur?.running && cur.live && Date.now() - lastTokenAt.current >= 30_000) {
+        void reconcileTabRuntime(activeTabId);
+      }
+    }, 30_000 - since);
+    return () => window.clearTimeout(timer);
+  }, [activeTabId, reconcileTabRuntime, activeState.running, activeState.live]);
 
   const send = useCallback((displayText: string, submitText = displayText) => {
     const submitForTab = (tabId: string) => {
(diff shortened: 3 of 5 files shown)


esengine (changes requested):
Thanks for the careful write-up — the root-cause analysis on the lost `turn_done` is exactly right, and the Go-side panic recovery in `runGuarded` (plus the two tests covering emit-once-on-panic) is solid work. That part I'd take as-is.

The frontend changes need another pass before this can land, though — one of them is a hard crash:

**1. The watchdog `useEffect` is nested inside another effect (blocker).** In `useController.ts` the new `useEffect` is placed *inside* the body of the existing event-subscription `useEffect` (the one that opens with `const textBatch = ...` and closes with `return () => { textBatch.drain(); ... }`, `}, [dispatchTo, ...])`). Calling a hook from inside another hook's callback violates the Rules of Hooks — React throws `Invalid hook call` the moment that outer effect commits on mount, which takes the whole main view down. It needs to be hoisted to the top level of `useController`, as a sibling effect. (This is also why the symptom wouldn't show up in `go test` — it only fails when the desktop app actually renders.)

**2. The no-deps "reset on every render" trick is fragile.** Even once hoisted, an effect with no dependency array re-arms a 30s timer on *every* render for any reason, not just token events — so an unrelated re-render keeps resetting it and it may never fire, while a genuinely-idle-but-not-re-rendering tab fires it. Prefer tracking a `lastTokenAt` ref updated on token events and comparing against it, with an explicit deps array so `react-hooks/exhaustive-deps` stays happy.

**3. `open={!userClosedReasoning}` regresses completed cards.** With the initial state `false`, every reasoning card renders expanded — including finished turns and history loaded from disk, which previously collapsed via `defaultOpen={item.streaming}`. Derive the open state from streaming as well, e.g. keep it controlled but seed/close it when `item.streaming` goes false, so done turns stay collapsed.

**4. Global ESC handler toggles every open card.** The `document`-level `keydown` listener in `ProcessCard` is registered per instance, so with the reasoning card and a tool card both open, one ESC closes all of them at once — and it collides with ESC used elsewhere (cancel/close). Scope it to the focused/hovered card, or handle ESC at a single owner instead of per-card.

The diagnosis is good and the backend half is ready — fix the hook nesting first (that's the crash), then the open-state and ESC scoping, and I'll re-review.

ashishexee:
> Thanks for the careful write-up — the root-cause analysis on the lost `turn_done` is exactly right, and the Go-side panic recovery in `runGuarded` (plus the two tests covering emit-once-on-panic) is solid work. That part I'd take as-is.
> 
> The frontend changes need another pass before this can land, though — one of them is a hard crash:
> 
> **1. The watchdog `useEffect` is nested inside another effect (blocker).** In `useController.ts` the new `useEffect` is placed _inside_ the body of the existing event-subscription `useEffect` (the one that opens with `const textBatch = ...` and closes with `return () => { textBatch.drain(); ... }`, `}, [dispatchTo, ...])`). Calling a hook from inside another hook's callback violates the Rules of Hooks — React throws `Invalid hook call` the moment that outer effect commits on mount, which takes the whole main view down. It needs to be hoisted to the top level of `useController`, as a sibling effect. (This is also why the symptom wouldn't show up in `go test` — it only fails when the desktop app actually renders.)
> 
> **2. The no-deps "reset on every render" trick is fragile.** Even once hoisted, an effect with no dependency array re-arms a 30s timer on _every_ render for any reason, not just token events — so an unrelated re-render keeps resetting it and it may never fire, while a genuinely-idle-but-not-re-rendering tab fires it. Prefer tracking a `lastTokenAt` ref updated on token events and comparing against it, with an explicit deps array so `react-hooks/exhaustive-deps` stays happy.
> 
> **3. `open={!userClosedReasoning}` regresses completed cards.** With the initial state `false`, every reasoning card renders expanded — including finished turns and history loaded from disk, which previously collapsed via `defaultOpen={item.streaming}`. Derive the open state from streaming as well, e.g. keep it controlled but seed/close it when `item.streaming` goes false, so done turns stay collapsed.
> 
> **4. Global ESC handler toggles every open card.** The `document`-level `keydown` listener in `ProcessCard` is registered per instance, so with the reasoning card and a tool card both open, one ESC closes all of them at once — and it collides with ESC used elsewhere (cancel/close). Scope it to the focused/hovered card, or handle ESC at a single owner instead of per-card.
> 
> The diagnosis is good and the backend half is ready — fix the hook nesting first (that's the crash), then the open-state and ESC scoping, and I'll re-review.

Thanks for the careful review and detailed feedback. You're absolutely right about the hook nesting issue — that's a blocker and I missed it during implementation. I'll rework the watchdog logic by hoisting it to the top level and will also address the concerns around timer re-arming, reasoning card open state, and ESC handling before updating the PR.

Appreciate the thorough review. I'll push a revised version shortly.

esengine (approved):
﻿Re-reviewed the latest push — all four points addressed properly:

1. The watchdog is now a top-level sibling effect with a real dependency array; the `lastTokenAt` ref + remaining-time timer is exactly the shape I asked for, and re-arming off `activeState.live` means token flushes naturally reset it while a stalled stream still fires after the full 30s.
2. Covered by the same rework.
3. `useState(item.streaming)` as the controlled seed keeps history/finished cards collapsed on mount while letting the user close a stuck one — which is the original #3746 symptom.
4. ESC now lives on the header button's own `onKeyDown`, so it only acts on the focused card and can't collide with global ESC semantics.

The Go-side panic recovery in `runGuarded` plus the emit-once tests were already solid in the first round. Nice turnaround — merging.

PR #3895, opened 2026-06-10
Improve context overview navigation and status metrics / 优化会话概览跳转与底部状态指标

SivanCola:
## Summary
- Redesign the bottom status bar around grouped model/cache/token/context/cost metrics with shared tooltips, value coloring for cache hit rates, and warning states for the compaction threshold.
- Refresh the right-side context overview by removing duplicated header/project chips, tightening runtime/cost/session cards, and making non-default themes inherit the Graphite layout while keeping their own colors.
- Add overview-to-workspace navigation: dependency rows and View all open scoped Files lists, while session-change rows and View all open scoped Changes lists and file history.
- Carry total/session token accounting through desktop telemetry, frontend controller state, and browser development mocks.

## Related PRs
- Related: #3822 overlaps the right-side workspace panel files; this PR focuses on the accepted overview/status/navigation experience.
- Related: #3854 overlaps token telemetry plumbing in `desktop/tabs.go`; this PR keeps the total-token field needed by the UI metrics.

## Cache safety
- UI/status-panel only. This does not change provider request serialization, prompt prefixes, tool schemas, or cache-key material.

## Verification
- `npm run build`
- `go test . -run TestWorkspaceTabAggregatesSessionUsageTelemetry` from the desktop module
- Browser verification on the local development UI for dependency-file reveal, dependency View all, session-change reveal, and session-change View all


esengine:
The redesign looks good and the cache-safety note holds — this is UI/status/telemetry only, and the new TotalTokens plumbing is clean and additive (nice that the test asserts both the telemetry aggregate and ContextPanel().TotalTokens).

One coordination item before I approve, on `desktop/tabs.go`. #3854 just landed and moved the per-turn token breakdown off the cumulative `telemetry.Usage` onto `ctrl.LastUsage()` — that was the fix for the donut segments being wrong after turn 1 (snapshot vs cumulative unit mismatch). This branch predates that, so its ContextPanel still assigns the breakdown from `usage.PromptTokens/CompletionTokens/...`, which would re-break #3854 if it merges on top.

Could you rebase onto the latest main-v2 and resolve it so that:
- the per-turn breakdown (Prompt/Completion/Reasoning/CacheHit/CacheMiss) stays sourced from `ctrl.LastUsage()` (the #3854 form), and
- only your new `info.TotalTokens = usage.TotalTokens` is added from the cumulative `telemetry.Usage` (that one is genuinely a session total, so cumulative is correct).

They're complementary — TotalTokens is additive and doesn't conflict with the snapshot breakdown. The locale/App.tsx files will also pick up conflicts from a few sibling desktop merges, same as the others. Ping me after the rebase and I'll do the full pass + approve. Thanks!

SivanCola:
Rebased onto the latest `main-v2` and resolved the conflict drift from #3854.

What changed:
- Kept the ContextPanel per-turn breakdown sourced from `ctrl.LastUsage()`.
- Added only `info.TotalTokens = usage.TotalTokens` from cumulative `telemetry.Usage`.
- Resolved the sibling desktop/frontend and locale conflicts from the rebase.
- Added a regression test that keeps cumulative telemetry totals from overwriting the LastUsage breakdown.

Verified:
- `go test .` in the desktop module

Ready for another pass when you have time.

SivanCola:
Updated the PR again on top of the latest `main-v2`.

Additional fix:
- Fixed the workspace preview close action after opening a dependency file from the Context overview. The revealed file request is now marked as dismissed when the user closes that preview, so the same request does not immediately reopen the file.

Verified:
- `go test .` in the desktop module
- `wails generate module`
- Browser verification: Context overview → dependency file → close preview now hides the preview and restores the file tree.

PR #3925, opened 2026-06-11
docs: use canonical repo in README badges

CooperSheroy:
## Summary
- update English and Chinese README badge links to use the canonical `esengine/DeepSeek-Reasonix` repository path
- keep npm, AtomGit, and Discord badges unchanged

## Validation
- verified all four updated shields.io badge URLs return HTTP 200
- confirmed no remaining `esengine/reasonix` README badge references

Note: local Go tests were not run because `go` is not installed in this environment; this is a README-only change.


esengine (approved):
﻿Correct fix — shields.io does not follow GitHub repo redirects, so the old `esengine/reasonix` slug renders broken/stale counts. Both READMEs updated consistently. Merging, thanks!

PR #3928, opened 2026-06-11
feat: toggle Developer Tools (Cmd+I)

ttmouse:
Add a native menu item "Toggle Developer Tools" (Cmd+I) that opens the WebKit Inspector via `runtime.WindowExecJS` with the `wails:openInspector` protocol.

### DIFF
--- a/desktop/menu.go
+++ b/desktop/menu.go
@@ -27,6 +27,11 @@ func (a *App) createAppMenu() *menu.Menu {
 			runtime.EventsEmit(a.ctx, "app:open-settings")
 		}
 	})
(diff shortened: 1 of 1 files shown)


esengine (approved):
﻿Useful escape hatch — `wails:openInspector` via the `external` message handler is the known WKWebView trick for opening the inspector outside dev mode. Worth noting for the record: `window.webkit` only exists on macOS, so on Windows/Linux this menu entry is a silent no-op (WebView2 users still have their own F12 path in dev builds). Fine as-is for a macOS debugging aid. Merging, thanks!

PR #3941, opened 2026-06-11
feat(cli): add self-update command (reasonix upgrade/update)

lanshi17:
## Summary

Add `reasonix upgrade` (alias: `reasonix update`) to download and apply the latest CLI release from GitHub Releases in-place.

## Changes

- **New command**: `reasonix upgrade [--check]` — fetches the latest release, compares semver, downloads the platform binary, verifies SHA256, and replaces the running executable atomically.
- **`--check` flag**: version-only comparison without downloading — useful for scripts/CI.
- **Proxy support**: uses the project's existing `netclient` proxy infrastructure so corporate environments work transparently.
- **i18n**: all user-facing strings are bilingual (en/zh) via the i18n catalogue.
- **Tests**: unit tests for version normalization, SHA256 checksum verification, and tar.gz extraction.

## How it works

1. Calls the GitHub Releases API for `esengine/DeepSeek-Reasonix/releases/latest`
2. Compares the running version against the release tag using `golang.org/x/mod/semver`
3. Downloads the platform-appropriate archive (`.tar.gz` for Linux/macOS, `.zip` for Windows)
4. Verifies SHA256 against the release's `SHA256SUMS` file
5. Extracts the `reasonix` binary and replaces the running executable via temp-file + rename

## Safety

- Dev builds (`version = "dev"`) are skipped — they never self-update.
- SHA256 checksum is verified before any binary replacement.
- Atomic replace via temp-file + rename (no partial writes).
- Respects the existing proxy configuration for network access.

## Usage

```bash
# Check if an update is available
reasonix upgrade --check

# Download and apply the latest release
reasonix upgrade
# or equivalently:
reasonix update
```

## Files changed

| File | Change |
|------|--------|
| `internal/cli/upgrade.go` | New: upgrade command implementation |
| `internal/cli/upgrade_test.go` | New: unit tests |
| `internal/cli/cli.go` | Register `upgrade`/`update` in command router |
| `internal/i18n/i18n.go` | Add upgrade message fields to Messages struct |
| `internal/i18n/messages_en.go` | English strings + usage text |
| `internal/i18n/messages_zh.go` | Chinese strings + usage text |
| `go.mod` / `go.sum` | Add `golang.org/x/mod` (semver comparison) |

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,6 +17,7 @@ require (
 	github.com/sabhiram/go-gitignore v0.0.0-20210923224102-525f6e181f06
 	github.com/yuin/goldmark v1.8.2
 	go.uber.org/goleak v1.3.0
+	golang.org/x/mod v0.37.0
 	golang.org/x/net v0.55.0
 	golang.org/x/sys v0.46.0
 	golang.org/x/term v0.44.0
--- a/go.sum
+++ b/go.sum
@@ -82,6 +82,8 @@ golang.org/x/exp v0.0.0-20231006140011-7918f672742d h1:jtJma62tbqLibJ5sFQz8bKtEM
 golang.org/x/exp v0.0.0-20231006140011-7918f672742d/go.mod h1:ldy0pHrwJyGW56pPQzzkH36rKxoZW1tw7ZJpeKx+hdo=
 golang.org/x/mod v0.2.0/go.mod h1:s0Qsj1ACt9ePp/hMypM3fl4fZqREWJwdYDEqhRiZZUA=
 golang.org/x/mod v0.3.0/go.mod h1:s0Qsj1ACt9ePp/hMypM3fl4fZqREWJwdYDEqhRiZZUA=
+golang.org/x/mod v0.37.0 h1:vF1DjpVEshcIqoEaauuHebaLk1O1forxjxBaVn884JQ=
+golang.org/x/mod v0.37.0/go.mod h1:m8S8VeM9r4dzDwjrKO0a1sZP3YjeMamRRlD+fmR2Q/0=
 golang.org/x/net v0.0.0-20190404232315-eb5bcb51f2a3/go.mod h1:t9HGtf8HONx5eT2rtn7q6eTqICYqUVnKs3thJo3Qplg=
 golang.org/x/net v0.0.0-20190620200207-3b0461eec859/go.mod h1:z5CRVTTTmAJ677TzLLGU+0bjPO0LkuOLi4/5GtJWs/s=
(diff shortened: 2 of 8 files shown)


esengine (changes requested):
﻿Thanks — the structure is clean, and two things here are genuinely better than the parallel attempt in #3859: every user-facing string goes through the i18n catalogue, and checksum verification **fails closed** (mismatch/missing-entry abort the install instead of warning and proceeding). Both of those are the bar.

But two of the same blockers from the #3859 review apply here too, and one of them fires on every single run today:

**1. `/releases/latest` is the wrong endpoint for this repository (blocker — currently 100% broken).** This repo publishes three release namespaces: `v*` (CLI), `npm-v*`, and `desktop-v*`. `/releases/latest` returns whichever release is marked *Latest* — which right now is **`desktop-v1.5.0`**. `semver.IsValid("desktop-v1.5.0")` is false, so as merged this command would error out on the invalid-tag path every time (and if a desktop release ever passed validation, it would try to download CLI archives from a desktop tag). You need to list `/releases` and select the newest tag matching `^v[0-9]` — see the namespace filter in #3859, which got this part right. Skip `"prerelease": true` entries while you're there.

**2. Windows cannot rename over the running executable (blocker).** `os.Rename(tmpName, resolved)` fails with `ERROR_ACCESS_DENIED` while `reasonix.exe` is memory-mapped — the standard dance is rename the running exe aside to `.old` (allowed while running), rename the new binary into place, best-effort delete the `.old` and sweep stale ones on a later run. `desktop/internal/update` already implements this pattern.

**On the overlap with #3859:** two PRs are now solving the same feature; I've reviewed both against the same checklist (namespace filter ✅ there / ❌ here; fail-closed checksum ❌ there / ✅ here; Windows replace ❌ both; i18n ❌ there / ✅ here). Whichever lands first with all four points addressed gets merged, and I'll close the other crediting both. You two are also welcome to combine forces on one branch — between #3859's release-selection logic and this PR's verification/i18n, the complete implementation already exists across the two diffs.

esengine:
Closing in favor of #3859 as the base implementation 鈥?see the cross-review there. The deciding issues were structural: `/releases/latest` doesn't filter our three tag namespaces (whenever the newest release is `npm-v*`/`desktop-v*`, update breaks entirely), and the Windows zip path searches for `reasonix.exe.exe` (binName already carries `.exe`, extractFromZip appends another) so Windows update can't work as shipped. What we're carrying over from yours into #3859's revision list, with credit: `x/mod/semver` for prerelease-correct comparison, hard-fail on checksum mismatch, and the full en/zh message catalogue. Thanks for the thorough work 鈥?the polish here genuinely raised the bar for the surviving PR.

PR #3947, opened 2026-06-11
fix: Ctrl+C in running state copies selection instead of cancelling turn

CnsMaple:
## Problem

When the chat TUI is running (tuiRunning state), selecting text and pressing Ctrl+C should copy the selection to clipboard — same as the idle-state behavior. Instead, it cancels the in-flight turn.

## Fix

In the Ctrl+C handler for the tuiRunning state, check for an active text selection first:
- **Selection active** -> copy to clipboard (reuses the same copy logic as the idle state)
- **No selection** -> cancel the current turn (original behavior)


esengine (approved):
﻿Exactly right — same selection-first contract as the idle path (and it reuses that path's restore/copy/clear dance verbatim, so the two can't diverge in behavior). Losing an in-flight turn because you tried to copy output was a nasty trap. Merging, thanks!

PR #3952, opened 2026-06-11
feat(cli): add /diff-fold command to toggle diff line folding

wudiandy:
Add a slash command that lets users toggle whether diffs show all lines or are folded to a limit (40 lines). The toggle is session-scoped and affects subsequent diff rendering without modifying historical content.

- Replace diffScrollbackMaxLines constant with diffFoldLimit (40)
- Add diffMaxLines field to chatTUI (default 0 = show all)
- Add /diff-fold command with toggle logic using i18n messages
- Register command in help listing and tab completion
- Add i18n messages for en and zh locales

### DIFF
--- a/internal/cli/chat_tui.go
+++ b/internal/cli/chat_tui.go
@@ -266,6 +266,10 @@ type chatTUI struct {
 	// shown as the current entry in the /output-style listing. "" = default.
 	outputStyle string
 
+	// diffMaxLines controls the max lines shown in a diff view. 0 = show all
+	// (default); non-zero = fold at that many lines. Toggled by /diff-fold.
+	diffMaxLines int
+
 	// statuslineCmd is the user's custom status-line command (config
 	// [statusline].command); "" disables it. statuslineOut caches its latest
 	// one-line stdout, refreshed at startup and after each turn and rendered in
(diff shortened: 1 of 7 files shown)


esengine:
One behavioral question before merge: as written, the session field's zero value flips the default from folded-at-40 to show-all — so after this lands every user's diffs dump fully into scrollback and `/diff-fold` becomes the way to turn folding *on*. If that flip is intended, say so in the description; otherwise initialize the session value from the configured fold limit so the default stays folded and `/diff-fold` is the opt-out. One-line change either way. Kicking fork CI now.

wudiandy:
> One behavioral question before merge: as written, the session field's zero value flips the default from folded-at-40 to show-all — so after this lands every user's diffs dump fully into scrollback and `/diff-fold` becomes the way to turn folding _on_. If that flip is intended, say so in the description; otherwise initialize the session value from the configured fold limit so the default stays folded and `/diff-fold` is the opt-out. One-line change either way. Kicking fork CI now.

You're right. I've changed the default state to folded.

esengine:
Confirmed the default-state fix — initializing `diffMaxLines` to `diffFoldLimit` keeps diffs folded-at-40 by default and makes `/diff-fold` the expand toggle, which is exactly right. Rename is complete (both `diffScrollbackMaxLines` refs updated), i18n covers en/zh, and it's wired into help + completion. Thanks for the quick turnaround — merging!

PR #3962, opened 2026-06-11
fix(desktop): prevent command palette session text clipping

### LINKED ISSUE #3842: [Bug]: “搜索”功能“最近会话”开头显示被截断问题
### Version line

### Exact version

1.5.0

### What happened?

bug描述：“搜索”里面的“最近会话”，当prompt显示很长的时候，顶头会超出显示框，并且对应的工作目录开头也会显示不全；

Image

### Steps to reproduce

1. 任选一个会话，输入比较长的prompt；
例如我的prompt：在templates\drawer.html中第1078行“保留最近AI消息数”下的input标签，最小min="3"，我想让它变成1，但是我修改为1后，页面上会自动更正为3，所以应该还有其他地方会修改这个，我想知道要把这个input的最小值限定为1，应该怎么改。

2. 同时打开多个会话标签页；
3. 打开右上角的搜索，滚动到“最近会话”，发现存在“最近会话”项开头显示不全，和工作目录一起被截断了。

### OS / platform

win10 22H2

### Relevant logs or output

```shell

```

GTC2080:
## Problem

The command palette still had a legacy Command Palette CSS block later in `styles.css`. Because it came after the current palette styles, it changed `.palette__item` to a vertical layout. Recent-session rows then centered the icon above the title and let long session text or workspace paths overflow horizontally, clipping the leading characters (#3842).

## Fix

Removed the stale duplicate palette block so the current row layout and theme overrides remain the only palette styling. Added a frontend CSS regression test that checks palette rows keep a side-by-side icon/text layout and retain the title/path ellipsis constraints.

## Tests

- `npm run test:all`
- `npm run build`
- Browser smoke check on `http://127.0.0.1:5184/?mock=demo`: filtered the palette to a recent session and confirmed the row computes to `display: flex; flex-direction: row`, with no horizontal list overflow.

Closes #3842


esengine (approved):
Nice catch, and thanks for going the extra mile with a regression test — the cascade-order trap here (stale block later in the file silently winning) is exactly the kind of thing that needs one. The test wiring into test/test:all matches our harness style, LGTM.

Two tiny things the old block carried that I'll restore in a small follow-up (no action needed from you): the `.palette__item:focus-visible` outline, and the detached search button's hardcoded `⌘K` label on Windows/Linux.

PR #3969, opened 2026-06-11
fix(desktop): add instruction-file create action in memory settings

### LINKED ISSUE #3901: [Bug]:
### Version line

### Exact version

1.5.0

### What happened?

Image

### Steps to reproduce

想请问下，上方添加，在哪添加？我找不到添加规则的按钮。

### OS / platform

windows 10

### Relevant logs or output

```shell

```

GTC2080:
Repro (#3901): open Settings → Memory → Instruction files with no REASONIX/AGENTS doc. The empty state pointed users toward adding something above, but the docs tab did not render any add/create action, so there was no obvious way to create the instruction file from that screen.

Cause: the quick-add form only lived on the memory entries tab and writes a note through Remember. The instruction-files tab could edit discovered docs through SaveDoc, but did not expose a create path for one of the canonical scope docs.

Fix:
- Add an instruction-file create action to the docs tab, defaulting to the project scope when available.
- Save new doc content through the existing SaveDoc/WriteDoc path, so backend allowedDocPaths continues to bound writes to recognized memory files.
- Update empty-state copy/locales and override the manager grid styles so the new form stays vertical and readable.
- Add a frontend regression test covering the docs-tab create affordance, copy, and layout guard.

Tests:
- npm run test:all
- npm run build
- go test ./...
- git diff --check
- Browser smoke check on http://127.0.0.1:5185/?mock=demo: Settings → Memory → Instruction files, opened the add form and confirmed the scope selector, full-width textarea, target path, and save button layout.

Closes #3901


esengine:
Reviewed — well-scoped, the SaveDoc/allowedDocPaths boundary is respected, locales + regression test included. The only blocker is mechanical: #3962 (yours!) appended its test to the same package.json test-script lines, so this now conflicts there. Rebase onto current main-v2 — should be a one-line resolution — and I'll merge.

SivanCola (changes requested):
Requesting changes for two blockers:

1. The new docs-tab "Add instruction file" flow can overwrite an existing instruction file. The button is rendered regardless of whether docs already exist, defaults to the project scope, and `activeDocTarget` comes from `view.scopes`. For a scope that already has a REASONIX/AGENTS/CLAUDE doc, `DocPath` returns that existing file, then `submitDoc` calls `app.SaveDoc(activeDocTarget.path, trimmed)`. `SaveDoc` is overwrite semantics, so a user can click Add and replace an existing project or global instruction file with the new draft. Please only expose true creation targets for scopes without an existing doc, or make this an explicit edit/replace flow with the existing body prefilled or confirmed, and add behavior coverage for the existing-doc case.

2. The branch is stale against current `main-v2` and cannot merge cleanly. `merge-tree` reports conflicts in `desktop/frontend/package.json`, `MemoryPanel.tsx`, `en.ts`, `zh.ts`, and `styles.css`. Current `main-v2` has the newer Memory settings structure with saved/archived/docs/suggestions tabs, storage controls, and additional frontend tests, so this fix needs to be replayed onto that UI rather than keeping the old two-tab page/test list.

Verification run locally on PR head: `pnpm install --frozen-lockfile`, `wails generate module`, `pnpm exec tsx src/__tests__/memory-settings-docs.test.ts`, and `pnpm exec tsc --noEmit -p tsconfig.test.json`. The added string-based test passes, but it does not exercise the overwrite path above.

GTC2080:
Updated in 06a978b.

Changes:
- Rebased onto current main-v2 and replayed the fix on the newer Memory settings structure.
- The docs-tab add flow now builds `instructionDocCreateTargets` from writable scopes minus already-discovered doc scopes/paths. If project/user/local already has an instruction file, that scope is not offered as a create target, so the add flow cannot silently overwrite it.
- Added behavior coverage for the existing-doc case: user/project docs present -> only local remains available.

Verification:
- `corepack pnpm install --frozen-lockfile`
- `wails generate module` with Wails v2.12.0
- `corepack pnpm exec tsx src/__tests__/memory-settings-docs.test.ts`
- `corepack pnpm exec tsc --noEmit -p tsconfig.test.json`
- `corepack pnpm run test:all`
- `corepack pnpm run build`
- `git diff --check`
- Browser smoke check on `http://127.0.0.1:5186/?mock=demo`: existing project/user docs left only the local create target (`REASONIX.local.md`), with the form layout intact.

PR #3988, opened 2026-06-11
fix(desktop): workspace git actions enhancement

kekexunxun:
The previous PR was closed due to the handling of the branch; a new PR is now being submitted. A fallback mechanism has been implemented for the Git "unstage" operation, and code related to a horizontal scrolling overflow issue—which did not require a fix—has been removed.

#3822


SivanCola (changes requested):
Thanks for the update here. I found two workspace-scoping issues that should be fixed before this lands:

1. `WorkspaceGitCommit` currently runs `git commit -m` without a pathspec. If the user opens a repository subdirectory as the workspace, the UI only shows staged files inside that subdirectory, but Git still commits the entire repository index. I reproduced this with `git -C sub commit`, which committed both `sub/a.txt` and an already-staged `other/b.txt`. Please scope the commit to the active workspace, for example with `-- .` or an explicit repo-root-relative workspace pathspec.

2. `WorkspaceGitStageAll` and `WorkspaceGitUnstageAll` have the same subdirectory leak: `git -C sub add -A` and `git -C sub reset HEAD --` affect changes outside `sub/`. These bulk actions should also use the same workspace-scoped pathspec, including the unborn-HEAD fallback path.

The Go tests passed locally, and the frontend build passed after regenerating the ignored Wails bindings. A regression test that opens/uses a repo subdirectory would be a good guard for both cases.

kekexunxun:
Fixes an issue where Git batch operations "leaked" outside the workspace when the workspace was set to a repository subdirectory.

### Issue

When a user opens a repository subdirectory (e.g., `sub/`) as the workspace:
- `git commit -m` (without a pathspec) commits changes from the entire index, rather than just files within the workspace.
- `git add -A` and `git reset HEAD` behave similarly, affecting files outside the subdirectory.

### Fix

Added the `-- .` pathspec to three batch operations to restrict them to the workspace directory.

### Testing

Added `TestWorkspaceGitScopesToSubdirectory` to verify that `StageAll`, `UnstageAll`, and `Commit` do not leak into the `other/` directory when `sub/` is the workspace, and to cover the `rm --cached` fallback path for an unborn HEAD. All relevant tests passed.

PR #4013, opened 2026-06-11
fix(evidence): allow rephrased todos to pass complete_step verification

### LINKED ISSUE #3992: [Bug]: 生成的待办没有跟随完成进度勾选已完成的待办
### Version line

### Exact version

v1.5.0

### What happened?

Image
任务已完成,但是下面的待办没有勾选

### Steps to reproduce

创建会话,自动生成的todoList

### OS / platform

windows11

### Relevant logs or output

```shell

```

JesonChou:
Fixes #3992

## Problem

When the model rephrases a todo item's `content` between successive `todo_write` calls (e.g., "Analyze the bug" → "Analyze the bug and understand root cause"), `hasSuccessfulCompleteStepForTodo` rejects the completion. The `sameTodoMatch` exact-text guard fails, and the bare `continue` on L592 prevents the index-based fallback (`matchTodoStep`) from ever running. The `todo_write` call is rejected by `verifyTodoCompletionTransitions`, so the frontend **never sees the completed status** — checkmarks don't appear (#3992).

## Fix

Replace the `continue` with a **content-overlap guard** (`todoContentRelates` /
`textOverlaps`):

- When old and new text share a recognisable substring (case-folded, normalized, ≥6 runes on the shorter side — same discipline as `stepTextContains` from #4006), the receipt falls through to the index-based fallback.
- When they are unrelated ("Add parser" → "Ship parser"), the original block is preserved (`continue` still fires).

`textOverlaps` uses `normalizeStepText` (introduced in #4006) for consistency with the matching layer.

Two new helpers:
- `todoContentRelates` — gates the fallback on related task identity
- `textOverlaps` — normalized substring overlap with ≥6-rune minimum

## Test

- `TestLedgerNumericCompleteStepAuthorizesRephrasedTodo` — covers the happy path now enabled (rephrased-but-same-task authorized)
- All **11** existing `evidence` tests pass unchanged
- All **33** `builtin` todo/complete_step tests pass unchanged
- `TestLedgerNumericCompleteStepDoesNotAuthorizeReplacedTodo` still blocks genuinely different tasks

## Files

| File | Δ |
|------|---|
| `internal/evidence/evidence.go` | +35/−1 |
| `internal/evidence/evidence_test.go` | +39 |

### DIFF
--- a/internal/evidence/evidence.go
+++ b/internal/evidence/evidence.go
@@ -606,7 +606,14 @@ func hasSuccessfulCompleteStepForTodo(receipts []Receipt, index int, current []T
 			if index >= 1 && index <= len(current) && sameTodoMatch(current[index-1], *r.TodoStep) {
 				return true
 			}
-			continue
+			// sameTodoMatch failed — the todo content changed.
+			// Allow the fallback (matchTodoStep by index or text) only
+			// when the old and new content are recognisably related
+			// (substring overlap); otherwise block a replaced todo
+			// from reusing an old numeric complete_step receipt.
+			if !todoContentRelates(current[index-1], *r.TodoStep) {
+				continue
+			}
 		}
 		match := matchTodoStep(r.Step, current)
 		if match.Found && match.Index == index {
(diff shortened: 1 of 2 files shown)


esengine:
Thanks @JesonChou — clean diagnosis of the bare `continue` swallowing the index fallback, and the overlap guard keeps replaced todos from reusing an old receipt. Merged. I'll fold `textOverlaps` into the existing `stepTextContains` and tidy the bounds guard in a small follow-up.

PR #4020, opened 2026-06-11
fix(chat): clear accumulated shell output on re-run of !commands

CnsMaple:
## Problem

Repeated `!pwd` (or any `!` shell command) accumulates output lines across runs instead of showing only the current run's output.

**Before (3× `!pwd`):**
```
  › !pwd

  ● Bash(pwd)
  ⎿  /d/about_fish/ProgramProtable/DeepSeek-Reasonix

  › !pwd

  ● Bash(pwd)
  ⎿  /d/about_fish/ProgramProtable/DeepSeek-Reasonix
     /d/about_fish/ProgramProtable/DeepSeek-Reasonix

  › !pwd

  ● Bash(pwd)
  ⎿  /d/about_fish/ProgramProtable/DeepSeek-Reasonix
     /d/about_fish/ProgramProtable/DeepSeek-Reasonix
     /d/about_fish/ProgramProtable/DeepSeek-Reasonix
```

## Root Cause

`controller.RunShell` generates a fixed tool ID for each `!` command (e.g. `"shell-pwd"`). `streamToolOutput` appends output with `m.shellOutputs[id] += chunk`, but `beginToolRunning` never cleared the map entry, so every subsequent run added to the previous output. `collapseShellSlot` then rendered all accumulated lines.

## Fix

Clear `m.shellOutputs[id]` in `beginToolRunning` so each re-run of the same tool ID starts with a clean slate. This is a one-line `delete` with a comment.

**After (3× `!pwd`):**
```
  › !pwd

  ● Bash(pwd)
  ⎿  /d/about_fish/ProgramProtable/DeepSeek-Reasonix

  › !pwd

  ● Bash(pwd)
  ⎿  /d/about_fish/ProgramProtable/DeepSeek-Reasonix

  › !pwd

  ● Bash(pwd)
  ⎿  /d/about_fish/ProgramProtable/DeepSeek-Reasonix
```

### DIFF
--- a/internal/cli/chat_tui.go
+++ b/internal/cli/chat_tui.go
@@ -1940,6 +1940,9 @@ func (m *chatTUI) beginToolRunning(id string) {
 	m.toolTail = m.toolTail[:0]
 	m.toolPartial = ""
 	m.toolLineCount = 0
+	// Clear accumulated output for this tool ID so a re-run (e.g. repeated
+	// !pwd with the same "shell-pwd" id) doesn't append to old output.
+	delete(m.shellOutputs, id)
 	m.toolStreamStart = time.Now()
 	m.toolStreamFrame = 0
 	if m.nativeScrollback {


esengine:
Thanks @CnsMaple — clean catch and the fix is correct. I traced the chain: `RunShell` derives a stable `shell-<cmd>` id, and `streamToolOutput` accumulates into `shellOutputs[id]`, so clearing it in `beginToolRunning` is exactly the right spot (runs on each dispatch, before the new run accumulates).

I've re-landed your commit on top of the latest `main-v2` with a regression test (three identical `!pwd` runs must leave one run's output, not three — verified it fails without your `delete`) in #4025, which will close this. Your authorship is preserved on the fix commit. 🙏

PR #4023, opened 2026-06-11
fix(ui): restrict inverted user bubble style to graphite theme only

CVEngineer66:
`[data-theme-style]` (bare, no value) matched every theme direction — graphite, carbon, aurora, slate — but the rule body uses graphite-specific colors (`--fg` as background, `--bg` as text). This turned user chat bubbles inside-out on non-graphite themes:

| Theme | Before (wrong) | After (correct) |
|---|---|---|
| carbon light | dark bubble, light text | `#f8f6f2` bg, `#1a1714` text |
| carbon dark | light bubble, dark text | `#1e1c1a` bg, `#ede9e3` text |

Fixed by narrowing both selectors to `[data-theme-style="graphite"]`.

Also fixes the pnpm environment variable in `dev`: pnpm v11 uses `pnmp_confirm_modules_purge`, not `PNPM_CONFIG_CONFIRM_MODULES_PURGE` (which is npm's format), so Wails' headless pnpm install was always blocked by the TTY prompt.

### DIFF
--- a/desktop/frontend/src/styles.css
+++ b/desktop/frontend/src/styles.css
@@ -17179,7 +17179,7 @@ a[href] {
   animation: graphite-message-in var(--dur-slow) var(--ease-out) both;
 }
 
-:root[data-theme-style] .msg--user .msg__body {
+:root[data-theme-style="graphite"] .msg--user .msg__body {
   max-width: min(80%, 650px);
   padding: 11px 15px;
   border: 1px solid color-mix(in srgb, var(--accent) 22%, var(--border));
@@ -17189,7 +17189,7 @@ a[href] {
   box-shadow: var(--app-graphite-shadow);
 }
 
-:root[data-theme="dark"][data-theme-style] .msg--user .msg__body {
+:root[data-theme="dark"][data-theme-style="graphite"] .msg--user .msg__body {
   background: var(--fg);
   color: var(--bg);
 }


SivanCola (changes requested):
Thanks for the CSS fix here. Narrowing the user-bubble selector to `data-theme-style="graphite"` looks right to me; the non-graphite themes still have the base/shared `.msg--user .msg__body` styling as fallback.

The blocker is the `dev` change. This PR is now dirty against `main-v2`: `#4024` already changed the same pnpm guard block and added `desktop/frontend/.npmrc` for the Wails headless pnpm case. Please rebase and keep the current base behavior there.

In particular, please do not resolve the conflict by using `export pnpm_confirm_modules_purge=false`. That env var form is missing the `config` part and does not match the pnpm config env convention. The current base uses `npm_config_confirm_modules_purge=false` plus `npm_config_minimum_release_age=0`, with `.npmrc` covering `confirm-modules-purge=false` and `verify-deps-before-run=false`.

So I think the mergeable shape should be: keep the CSS selector fix, drop/reconcile the `dev` change with the latest `main-v2` pnpm guard.

esengine:
Thanks @CVEngineer66 — correct narrowing; the inverted user bubble now only applies under `graphite`, other theme-styles keep the shared bubble. Merged. (The earlier change-request was a mis-set state; the fix is right.)

PR #4031, opened 2026-06-11
fix(ui): batch completed read-only tools into ReadOnlyBatch in compact/minimal mode

CVEngineer66:
Three render paths were rendering read-only tools (read_file, grep, glob, web_fetch, ls) as individual tool cards even after they completed:

1. **Hot zone final step (compact/minimal)** — running tools render individually, completed tools batch into ReadOnlyBatch.
2. **TurnCollapse body (compact/minimal folded steps)** — same batching logic added to the pre-computed body array.
3. **WarmTurnItems (expanded history turns)** — new roBatch/flushRO loop added so warm read-only tools are also grouped.

This keeps running read-only tools visible (with shimmer animation) and hides completed ones behind a compact "Read N files · Search N files" fold line, consistent across all compact/minimal display modes.

### DIFF
--- a/desktop/frontend/src/components/Transcript.tsx
+++ b/desktop/frontend/src/components/Transcript.tsx
@@ -465,7 +465,14 @@ export function Transcript({
           readOnlyBatch.length = 0;
         };
         for (const it of group.items) {
+          // Running read-only tools render individually (visible in progress);
+          // only completed read-only tools go into the batch (hidden after done)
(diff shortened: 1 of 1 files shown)


esengine:
Thanks @CVEngineer66 — verified all three render paths: hot-zone running tools stay individual (shimmer preserved), TurnCollapse mirrors the existing `displayItems` filter (no items dropped), and WarmTurnItems batches without losing turn-action state. Merged.

PR #4046, opened 2026-06-11
fix(control): prevent MemoryQuickAddNote from swallowing multi-line Markdown headings

### LINKED ISSUE #4008: [Bug]: 使用COSTAR提示词，导致程序直接不运行了
### Version line

### Exact version

v1.5.0

### What happened?

当我使用以下提示词来运行时：
```
# Context

- iconDialog `packages\gis-business\src\components\pit-gis-vector\components\iconDialog.vue`
- vector-image `packages\gis-business\src\components\pit-gis-vector\vectorpoint\components\image.vue`
- layerManager: 提供相关API `src\api\gis-business\layerManage.js`

`vector-image`组件引用`iconDialog`，实现了表单中`选取图标`的功能。但是在编辑表单回显图标时出现了bug。

# Objective

为了实现回显表单图标，需要对`iconDialog`进行修改。
关注点在props中的value值。此value值是图标的主键id，而不是`pit-image`需要的`imageFileId`.
需要调用`layerManager`中的`getIconInfo`以获取图标对象中的`imageFileId`
```

程序提示:
```
remembered →D:\gis\pit-gis-frame\AGENTS.md
```

结果：写入以下内容:
```
# Project memory

## Notes

....
```

我期望的结果: 正常运行

### Steps to reproduce

1. 运行桌面端Reasonix
2. 新建对话
3. 粘贴上述提示词
4. 没有开始任何运行

以下为截图:

Image

### OS / platform

windows11

### Relevant logs or output

```shell

```

warvyvr:
#4008: pasting a COSTAR-style prompt starting with "# Context" (or any Markdown H1 heading) on the Windows Wails desktop GUI caused the agent to never start — the entire multi-line input was silently saved as a memory note instead of being sent to the agent loop.
    
Root cause: MemoryQuickAddNote only checked for a "# " prefix. When the desktop calls ctrl.Submit() → submit(), the full multi-line input goes through MemoryQuickAddNote first. A prompt beginning with "# Context" matched the quick-add pattern, so submit() returned early after saving a memory note — the agent never ran.
    
Why the TUI is unaffected: the terminal UI folds pasted multi-line content into a "[P1]" label before the quick-add check runs, so the "# " prefix is never visible to MemoryQuickAddNote. The expanded text goes through ctrl.SendWithRaw(), which bypasses submit() entirely.
    
Fix: reject multi-line input in MemoryQuickAddNote. Any input containing a newline is now treated as a normal prompt, not a quick-add note. Single-line "# note" still works as before.
    
Test: added 3 multi-line cases (COSTAR-style with blank line, simple two-line, indented multi-line) all expecting ok=false; single-line "# Heading" preserves ok=true backward compatibility.
    
Fixes #4008

### DIFF
--- a/internal/control/input.go
+++ b/internal/control/input.go
@@ -146,11 +146,17 @@ func activeGoalBlock(goal string) string {
 	return b.String()
 }
 
-// MemoryQuickAddNote parses the legacy "# <note>" memory shortcut. The space
-// after "#" is intentional: "#7", "#issue", and "#标题" are ordinary user
-// prompts, not memory writes.
+// MemoryQuickAddNote parses the "# <note>" memory shortcut. The space after
+// "#" is intentional: "#7", "#issue", and "#标题" are ordinary user prompts,
+// not memory writes. Multi-line input starting with "# " is NOT treated as a
+// quick-add note — it is almost certainly a Markdown heading in a structured
+// prompt (e.g. "# Context\n\n- file.go\n# Objective"). Only single-line input
+// may be a quick-add note.
 func MemoryQuickAddNote(input string) (note string, ok bool) {
 	trimmed := strings.TrimSpace(input)
+	if strings.Contains(trimmed, "\n") {
+		return "", false
+	}
 	if strings.HasPrefix(trimmed, "# ") || strings.HasPrefix(trimmed, "#\t") {
 		return strings.TrimSpace(trimmed[1:]), true
 	}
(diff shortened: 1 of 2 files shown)


esengine:
Great catch and a clear writeup — the COSTAR `# Context` paste silently becoming a memory note (so the agent never ran) is a nasty one, and pinning down why the TUI's `[P1]` paste-folding hid it is exactly the right analysis. Treating any newline-containing input as a normal prompt is the right call: a `#`-leading multi-line block is almost always Markdown, not a one-line note, and single-line `# note` is preserved. Fixing it at the control layer covers every `submit()` caller. Merging — thanks!

PR #4047, opened 2026-06-11
feat(tui): /clear executes immediately without confirmation

lanshi17:
## Summary

Make the TUI `/clear` command immediately create a new session and clear the screen, removing the Y/N confirmation overlay.

## Changes

- `/clear` now calls `ctrl.ClearSession()` + `resetFreshContextView(true)` + `tea.ClearScreen` directly
- Removed `clearConfirm` overlay: struct, modal key handler, renderer, all nil-checks
- Removed unused `SlashClearPrompt` i18n string
- New test `TestSlashClearExecutesImmediately` verifies immediate session rotation, old file deletion, context reset to system-only, and transcript clear

## Net impact

5 files, +21 / −120 lines

### DIFF
--- a/internal/cli/chat_tui.go
+++ b/internal/cli/chat_tui.go
@@ -3453,7 +3453,13 @@ func (m *chatTUI) runSlashCommand(input string) tea.Cmd {
 		m.notice(i18n.M.SlashNewDone)
 	case "/clear":
 		m.echoLocalCommand(input)
-		m.clearConfirm = &clearConfirm{confirm: 1}
+		if err := m.ctrl.ClearSession(); err != nil {
+			m.notice(fmt.Sprintf("%s: %v", i18n.M.SlashClearFailed, err))
+			return nil
+		}
+		m.resetFreshContextView(true)
+		m.notice(i18n.M.SlashClearDone)
+		return tea.ClearScreen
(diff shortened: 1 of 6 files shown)


esengine:
Thanks for this — the removal is clean (the overlay, key handler, renderer, nil-checks, and the unused i18n string all go, and the new test pins the immediate behavior). I'm going to hold off on merging it as-is, though, and want to be clear about why so it's not discouraging.

The blocker is what `ClearSession()` actually does: its first step is `removeSessionArtifacts(oldPath)` — it **permanently deletes** the old transcript, not just rotates away from it, so a cleared session can't be brought back with `/resume`. The Y/N overlay is currently the only guard against an accidental, unrecoverable wipe. Claude Code's `/clear` feels immediate *because the session stays resumable*; here it's immediate **and** irreversible, which is the combination I want to avoid.

If you're up for it, the version I'd happily take is "immediate **and** safe": make `/clear` rotate to a fresh session and clear the screen without confirmation, but stop deleting the old transcript (drop the `removeSessionArtifacts` call from this path) so it stays in history and is `/resume`-able. That gets the friction-free UX you're after without the data-loss footgun. Closing this one for now, but a follow-up along those lines is very welcome.

PR #4059, opened 2026-06-11
fix(ui): batch completed read-only tools into ReadOnlyBatch, unify card body styles

CVEngineer66:
## Summary

Three related UI improvements to the transcript, focused on compact/minimal display modes in the desktop app.

### 1. ReadOnly batch for completed read-only tools
- Compact/minimal modes: running read-only tools (read_file, grep, web_fetch, etc.) render individually so the user sees progress; completed ones fold into a "Read N files · Search N files" fold line (`ReadOnlyBatch` component)
- Extracted to standalone `ReadOnlyBatch.tsx` to break circular dependencies
- Same batching applied in `TurnCollapse` (folded steps), `WarmTurnItems` (expanded history), and final step rendering
- Uses `isReadOnlyTool(name)` for tool identification to work around `readOnly` being `false` in session restore data

### 2. Sub-agent card nesting visuals
- Added `tool--subagent` class: left border on `tool__body` for task/explore/research/review/run_skill cards with `⊞N` count
- `tool__nested` styling with recursive nesting border for deeper sub-agent levels
- Removed dead `.tool__summary` CSS (outcome line + "⎿ " prefix)

### 3. Code copy button placement
- Moved `CopyButton` inside `<pre>` element so it renders within the code block background, not outside it
- Replaced `Tooltip`/portal (broken by `<pre> overflow:auto`) and native `title` (1s browser delay) with CSS `max-width` inline label

### 4. Cleanup
- Unify `reasoning__body`, `notice__body`, `warm-turn__body` padding/font to match `tool__body`
- Remove dead graphite theme overrides for body sections
- Skip partial `tool_dispatch` events so tool name + args appear at once

## Files changed
| File | Change |
|---|---|
| `ReadOnlyBatch.tsx` | New — standalone batch fold component |
| `ToolCard.tsx` | Subagent class + nested count |
| `Transcript.tsx` | Read-only batching logic |
| `CopyButton.tsx` | Inline label, removed Tooltip |
| `HljsCode.tsx` | CopyButton inside <pre> |
| `styles.css` | Unified body styles, subagent border, copy hover |
| `locales/en.ts`, `zh.ts` | `tool.otherReadCount` i18n key |
| `App.tsx` | Removed unused `showLabel` prop |

## Test status
- `go test ./...` passes (one pre-existing failure `TestSaveToScopesUserAndProjectFiles` — permission issue on dev machine, unrelated to changes)
- `gofmt -l .` clean
- `pnpm build` succeeds

### DIFF
--- a/desktop/frontend/src/App.tsx
+++ b/desktop/frontend/src/App.tsx
@@ -2462,7 +2462,6 @@ export default function App() {
               <CopyButton
                 getText={getSessionMarkdown}
                 label={t("topicBar.copyAll")}
-                showLabel={false}
                 className="topicbar__action-btn topicbar__action-btn--icon topicbar__action-btn--utility"
               />
               <div className={`topicbar__export${topicExportOpen ? " topicbar__export--open" : ""}`}>
--- a/desktop/frontend/src/components/CodeViewer.tsx
+++ b/desktop/frontend/src/components/CodeViewer.tsx
@@ -1,5 +1,4 @@
 import { lazy, Suspense } from "react";
-import { CopyButton } from "./CopyButton";
 
 export interface EditorProps {
   value: string;
@@ -24,7 +23,6 @@ const Impl = lazy(() => import("./editors/HljsCode"));
 export function CodeViewer(props: EditorProps) {
   return (
     <div className="code-block">
-      <CopyButton text={props.value} className="code-block__copy" />
       <Suspense
         fallback={
           <pre className="code code--loading">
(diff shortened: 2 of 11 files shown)


esengine (approved):
Thanks — extracting ReadOnlyBatch into its own file (breaking the circular dep) and the sub-agent nesting visuals (⊞N + tool--subagent) are a clean improvement, and the session-restore readOnly fix is the right call. CI is green across the board. Merging.

Follow-up on our side: consolidate the duplicated isReadOnlyTool list (it lives in both useController.ts and Transcript.tsx and mirrors the Go ReadOnly() classification, so it'll drift), and improve live-thinking + writer-output visibility in minimal mode.

PR #4061, opened 2026-06-11
Fix cost currency display / 修复成本货币符号显示

SivanCola:
## Summary

- Pair context-panel cost values with the currency from the same data source so panel telemetry cannot be overwritten by the default session currency.
- Share frontend money formatting across the context panel and status bar, including normalization for common currency codes and symbols.
- Normalize provider pricing currency symbols on the Go side so unexpected text cannot leak into UI cost values.
- Add nowrap/ellipsis guards for compact metric values and restore command-palette hint ellipsis expectations.

## Root cause

The context panel could select `info.sessionCost` while still preferring the global session currency, whose initial value defaults to yuan. Separately, unknown currency text was passed through directly, so bad or stale currency values could appear before the numeric cost.

## Validation

- `npm --prefix desktop/frontend test`
- `npm --prefix desktop/frontend run build`
- `go test ./internal/provider`
- `go test ./...` from the desktop module

This change is display-only and does not modify provider request serialization, prompts, tool schemas, or other prompt-cache-sensitive paths.

### DIFF
--- a/desktop/frontend/src/__tests__/context-panel-breakdown.test.ts
+++ b/desktop/frontend/src/__tests__/context-panel-breakdown.test.ts
@@ -1,6 +1,7 @@
 // Run: tsx src/__tests__/context-panel-breakdown.test.ts
 
-import { contextBreakdown } from "../components/ContextPanel";
+import { contextBreakdown, contextCostDisplay } from "../components/ContextPanel";
+import { currencySymbol, formatMoney } from "../lib/money";
 
 let passed = 0;
 let failed = 0;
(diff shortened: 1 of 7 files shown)


esengine:
Thanks @SivanCola — clean fix for the currency mislabeling, and good call sharing the formatter into `lib/money.ts`.

I pushed one follow-up commit before merging: the normalizer only passed through *single-rune* currency signs, so compact multi-character symbols (`A$`, `CA$`, `HK$`, `R$`) fell back to `¥` and would mislabel those configs. It now keeps any string that contains a currency sign, on both the Go (`provider`) and frontend (`money.ts`) sides, with tests covering it. Merging once CI is green.

PR #4071, opened 2026-06-11
Add lightweight history and memory retrieval / 增加轻量历史与记忆检索

SivanCola:
## Summary

- Add a pure-Go BM25 retrieval layer shared by saved session history and saved memories.
- Add the read-only `history` tool for searching local session transcripts and reading bounded context around a hit.
- Add the read-only `memory` recall tool for active saved memory search/list/read.
- Archive forgotten memories instead of permanently deleting them, and expose archived entries in CLI/TUI/desktop management surfaces for traceability.
- Refresh the desktop Memory settings page with separate saved, suggestion, archived, and instruction-file views; search/filter controls; quick add; archive confirmation; localized Project/Global labels; and collapsed instruction-file cards.
- Add desktop candidate suggestions: users can manually scan recent local history, or enable an auto-generate toggle that scans when the Suggestions tab opens. Candidates draft memory entries and repeated-workflow skills with preview, evidence snippets, and explicit confirmation before anything is saved.
- Require fresh human approval for agent-initiated `remember` and `forget` calls, including a local preview of the memory content being saved or archived.
- Add detailed technical documentation in `docs/SESSION_MEMORY_RETRIEVAL.md` and link it from the guide/spec docs.

## Technical Notes

This keeps Reasonix cache-first: dynamic session history is retrieved on demand, while the stable system prompt prefix remains unchanged during a session. Saved memory acts as a synthesis cache for approved, reusable conclusions rather than as a raw transcript dump.

The implementation avoids SQLite, CGO, vector databases, and embedding models. Retrieval is lexical BM25 with a relative score floor to keep strong matches while trimming common-word noise.

The desktop management UI now treats active memories, candidate suggestions, archived memories, and instruction files as separate views. Candidate suggestions are drafts: scans do not write memory or skills, the auto-generate toggle only controls local candidate generation on the Suggestions tab, and accepting a candidate goes through the existing memory store or skill store validation paths. Active memories can be searched, filtered, expanded, quick-added, and archived; archived entries remain read-only for audit/recovery; instruction files use the same collapsed card interaction pattern as memory entries.

Memory archive file operations are confined through the standard library root-scoped file APIs, so archive moves operate on validated names inside the memory store rather than passing absolute paths between helpers.

## Validation

- `go test ./...` from the desktop module.
- Isolated-HOME `go test ./internal/control`.
- `go test ./internal/memory`.
- `npm run typecheck`.
- `npm run check:css`.
- `wails generate module`, followed by `npm run typecheck`.
- `git diff --check`.
- Browser verification on the local desktop frontend: opened Settings → Memory → Suggestions, scanned candidates, expanded memory and skill previews, confirmed save/create actions, verified the auto-generate toggle, and checked console errors.

Created as a draft PR for review.


SivanCola:
@codex  review

SivanCola:
approve

esengine:
Merged — thank you, this is genuinely excellent work.

What stood out: you didn't just avoid breaking the cache-first prefix, you fenced it with regression tests — provider-visible tool order + schema/description stability, and asserting retrieval content stays out of the cached system prompt. That invariant is what this project lives or dies on, and it's now protected against future drift. The BM25-only approach (no SQLite/CGO/embeddings) is exactly the right call here, and the path handling is careful and correct throughout — `slug` + `os.OpenRoot` in the memory store, and `allowedPath` confinement so the history `around` op can't be steered outside the session roots.

Two notes, neither blocking:
- The CodeQL "uncontrolled data in path expression" alerts on `internal/memory/store.go` are false positives — every name is slug-sanitized to `[a-z0-9-]` (no separators or `..` survive) and all FS ops go through `os.OpenRoot` plus `safeJoin`/`IsLocal` validation. Safe to dismiss as such; the taint tracker just can't model the `slug`+`os.Root` confinement.
- `remember`/`forget` now always prompt, even under YOLO. Keeping that as a deliberate carve-out — durable cross-session memory is stickier than a reversible edit — just noting it on the record so we revisit if it ever blocks an unattended run.

Thanks again for the care here — look forward to more.

PR #4073, opened 2026-06-11
fix(memory): add GlobalDir routing + tab-aware panel to fix empty memory view (#3756)

ashishexee:
### Problem
The desktop memory panel always shows empty when the Global tab (or any tab whose project-specific memory directory has no auto-memories) is active. The root cause is two-fold:

1. **Store only has `Dir` (project-specific path)** — `App.Memory()` reads from the active controller's `Store.Dir`, which for the Global tab points to an empty `projects/<global-workspace-slug>/memory`. There is no concept of a shared memory directory for cross-project memories (user preferences, feedback).

2. **Panel has no tab awareness** — The memory panel always reads from the active tab's controller. When viewing settings from the Global tab, it gets the Global tab's empty store.

### Fix (two layers)

**Layer 1 — Backend `Store` (internal/memory):**
- Add `GlobalDir` field to `Store`, populated by `StoreFor()` as `<userDir>/memory/global`
- `DirFor(t Type)` routes `TypeUser`/`TypeFeedback` → `GlobalDir` (shared), others → `Dir` (project-specific); falls back to `Dir` when `GlobalDir` is empty
- `List()`, `Index()` merge both dirs with deduplication (global first)
- `Delete()` removes from all dirs (handles migration where same memory name exists in both)
- `Save()` routes via `DirFor()` and updates only the target dir's `MEMORY.md`
- `Index()` reconstructs a single clean index from managed lines across both dirs (no duplicate headers)

**Layer 2 — Desktop `App` + Frontend:**
- Add `MemoryForTab`/`RememberForTab`/`ForgetForTab`/`SaveDocForTab` using `ctrlByTabID(tabID)` with `fallback=false` (no silent fallback to wrong tab)
- Empty `tabID` falls back to active tab for backward compatibility
- Add `StoreGlobalDir` to `MemoryView` JSON payload
- `MemorySettingsPage` gets a tab selector dropdown (shown when >1 tab) so user can browse any project's memories
- All mutation calls use `*ForTab` when a tab is selected
- "Memory unavailable" screen still shows the tab selector so user can recover

### Backward Compatibility
- `Store{Dir: x}` with empty `GlobalDir` — `DirFor()` falls back to `Dir`, all methods work with single directory
- Old `Memory()`/`Remember()`/`Forget()`/`SaveDoc()` unchanged (delegate with `fallback=true`)

### Migration Safety
- Existing memories in `Dir` are unaffected — `List()` merges both dirs
- Same memory name in both dirs (pre-routing copy + new routing): `List()`/`Index()` deduplicate (GlobalDir first), `Delete()` cleans both
- `Index()` produces exactly one `# Memory` header regardless of how many dirs contribute entries

### Tests (9 new, 65 total pass)
- `TestStoreGlobalAndProject` — routing + merge + delete + header dedup
- `TestStoreForInitializesGlobalDir` — `StoreFor` sets `GlobalDir`
- `TestDirForRoutesCorrectly` — type→dir routing
- `TestDirForFallsBackWhenNoGlobalDir` — empty `GlobalDir` → `Dir`
- `TestStoreDeleteRemovesFromAllDirs` — migration: both copies deleted
- `TestStoreIndexDeduplicatesAcrossDirs` — no duplicate index lines/headers
- `TestStoreSaveVerifiesIndexDir` — `MEMORY.md` written to correct dir per type
- `TestStoreDeleteFlushesIndexPerDir` — both dirs' indexes flushed, `Index()` returns `""`
- `TestStorePathWithGlobalDir` — GlobalDir-first lookup + Dir fallback

---

Hi maintainers — this is my first time tackling a larger issue in this codebase. I've tried to be thorough with the implementation, test coverage, and edge cases (migration scenarios, stale tab handling, nil controller guards). That said, I'm sure there are things I could improve, and I'd genuinely appreciate any feedback on the approach, the code style, or anything I may have missed. Happy to iterate and learn. Thank you for your time reviewing this.

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -4326,12 +4326,13 @@ type MemoryScope struct {
 // MemoryView is the whole memory panel payload: hierarchical docs, active saved
 // facts, archived facts, and the writable scopes for the quick-add selector.
 type MemoryView struct {
-	Docs      []MemoryDoc     `json:"docs"`
-	Facts     []MemoryFact    `json:"facts"`
-	Archives  []MemoryArchive `json:"archives"`
-	Scopes    []MemoryScope   `json:"scopes"`
-	StoreDir  string          `json:"storeDir"`
-	Available bool            `json:"available"`
+	Docs           []MemoryDoc     `json:"docs"`
+	Facts          []MemoryFact    `json:"facts"`
+	Archives       []MemoryArchive `json:"archives"`
+	Scopes         []MemoryScope   `json:"scopes"`
+	StoreDir       string          `json:"storeDir"`
+	StoreGlobalDir string          `json:"storeGlobalDir,omitempty"`
+	Available      bool            `json:"available"`
 }
 
 // writableScopes are the quick-add targets the panel offers, broad → specific.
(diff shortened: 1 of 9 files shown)


ashishexee:
Hi @esengine / @SivanCola — regarding the CodeQL **"Uncontrolled data used in path expression"** alert:

### This is a false positive

The `name` parameter flows through `slug()` before reaching any path construction:

```go
var slugRe = regexp.MustCompile(`[^a-z0-9]+`)

func slug(s string) string {
    return strings.Trim(slugRe.ReplaceAllString(strings.ToLower(strings.TrimSpace(s)), "-"), "-")
}
```

After `slug()`, the output can only contain `[a-z0-9-]` — no `/`, no `\`, no `..`, no null bytes. **Path traversal is structurally impossible.** CodeQL does not recognize `slug()` as a sanitizer because it treats custom helpers as opaque.

### What I have tried

I wrapped the path construction in a `safeJoin()` helper using `filepath.Rel` containment checks (matching the repo's own `confine.within` pattern in `internal/tool/builtin/confine.go`), but CodeQL does not follow custom sanitizer wrappers — the taint still flows through.

### Two options going forward

1. **Add `// codeql[go/path-injection]` suppression** on the flagged lines with a comment referencing `slug()`'s guarantee — standard approach for verified false positives
2. **Restructure to avoid the taint flow entirely** — use `os.ReadDir` to discover the file by matching `e.Name()` instead of constructing the path from user input; eliminates the false positive structurally but adds minor I/O overhead

I would lean toward **option 1** as it is cleaner and `slug()` is trivially verifiable. Happy to go either way — let me know your preference and I will update the PR.

ashishexee:
**CI fix note for reviewers:**

The `test (windows-latest)` failure was caused by global memory being loaded from the runner's user config directory because `TestBuildWithoutMemoryLeavesPromptUnchanged` did not isolate `os.UserConfigDir()`. The failure path matches the CI output exactly — the global memory file name (`synthesis-cache-policy.md`) appears in the log, and the duplicated `# Memory` header is visible in the failure output.

Two fixes applied:

1. **Test isolation** — redirect `HOME`/`XDG_CONFIG_HOME`/`AppData` to a temp directory so `os.UserConfigDir()` doesn't pick up the runner's real `reasonix\memory\global`. This matches the pattern already used in other boot tests (e.g., `boot_test.go:1496-1497`).

2. **Removed duplicate `# Memory` header from `Index()`** — `Block()` already provides `# Memory` + `## Saved memories` as the section structure, so `Index()` now returns just the index lines without its own header. The on-disk `MEMORY.md` files still get headers from `flushIndexIn()` — only the in-memory merge output is stripped.

This makes the test deterministic across platforms and removes a redundant header that would have appeared in the system prompt whenever both GlobalDir and Dir contributed entries.

SivanCola:
Superseded by #4250, which has now been merged. #4250 integrates this work with attribution in the PR body. Thank you again for the contribution.

PR #4074, opened 2026-06-11
fix(agent): preserve mid-turn steer messages across tab switches (#4044)

### LINKED ISSUE #4044: [Bug]: 任务中途注入的引导提示在切换对话后就消失了
### Version line

### Exact version

1.6.0

### What happened?

当我在模型执行工作的轮次中输入引导信息，在对话中间引导信息是可以展现的：

Image

但是切换对话后再切回来，这个信息就消失了。

Image

### Steps to reproduce

1.执行任务，任务中途引导；
2.切换对话，再切换回来，回到引导被注入的位置；
3.引导信息消失

### OS / platform

Windows 11

### Relevant logs or output

```shell

```

JesonChou:
Closes #4044

> **Note**: steer notices display correctly in **Standard** display mode.
> In Compact/Minimal mode the frontend folds prior assistant turns into a "Processed" section when a new assistant turn begins — this is existing UI behaviour and not addressed here.

---

### Problem

Mid-turn guidance injected via Steer disappeared from the conversation after switching to another tab and back. The steer was visible during the live turn (as a `↪ text` notice) but gone after tab switch.

### Root Cause

`internal/control/input.go` included the steer prefix in `syntheticPrefixes`. Desktop's `historyMessages()` called `IsSyntheticUserMessage()` on every user-role message during history replay and **skipped** synthetic ones. This contradicted the agent's stated intent (`agent.go:507-510`) that steer messages be persisted to "survive tab switches and history replay."

The CLI TUI and HTTP/SSE frontends had the same class of bug: they replayed steer messages as raw user bubbles with the full prefix visible.

### Changes

**Commit 1 — Core fix (desktop):**

1. **`internal/agent/agent.go`** — Export `MidTurnSteerPrefix`; add `SteerText(content) (string, bool)` to extract the original user text from a steer-wrapped message without trimming (preserving exact character fidelity with the live Steer event).

2. **`internal/control/input.go`** — Remove steer prefix from `syntheticPrefixes`. Steer messages are no longer classified as synthetic.

3. **`desktop/app.go`** — In `historyMessages()`, detect steer messages using raw `m.Content` (before `resolveUserContent` applies `StripComposePrefixes` which trims whitespace) and emit them as `role: "notice"` entries with the `↪ ` prefix — matching the live Steer event appearance.

4. **`internal/agent/steer_test.go`** — New table-driven tests for `SteerText` (11 cases) + round-trip test for `midTurnSteerMessage`.

5. **`internal/control/input_test.go`** — Updated test expectation.

**Commit 2 — CLI TUI + HTTP/SSE parity:**

6. **`internal/cli/chat_tui.go`** — `replaySectionsFor` now detects steer messages via `agent.SteerText()` before rendering user bubbles. Steer messages are rendered as `  ↪ text` notice lines instead of raw `› [Mid-turn steer...]` user bubbles.

7. **`internal/serve/serve.go`** — `historyMessages` now detects steer messages and emits them as `role: "notice"` with the `↪ ` prefix, consistent with the desktop behaviour. Web clients reconnecting see clean notice text instead of the raw prefix.

8. **`internal/control/input_test.go`** — Hardcoded prefix string replaced with `agent.MidTurnSteerPrefix` constant reference so the test automatically tracks prefix changes.

### Design

All three frontends (desktop, CLI TUI, HTTP/SSE) now share the same pattern: before processing a `RoleUser` message for display, call `agent.SteerText()` to detect steer messages and convert them to a notice-style rendering. The model still sees the raw user message (with the prefix) for context; only the display layer transforms it.

### Verification

- `go test ./internal/agent/` — 80+ tests pass
- `go test ./internal/control/` — all pass (1 pre-existing failure in `TestRemoveMCPServerRemovesUnconnectedLazyPlaceholder`, confirmed on clean `main-v2`)
- `go build ./internal/cli/ ./internal/serve/` — clean
- `go vet ./desktop/...` + `go build ./desktop/...` — clean

### Affected areas

| Path | Before | After |
|:--:|:--:|:--:|
| Desktop tab switch → back | Steer disappeared | `↪ text` notice |
| Desktop tab close → reopen | Steer disappeared | `↪ text` notice |
| Desktop history panel preview | Steer disappeared | `↪ text` notice |
| CLI TUI session resume | Raw prefix user bubble | `  ↪ text` notice line |
| HTTP/SSE `/history` endpoint | `role:"user"` with raw prefix | `role:"notice"` with `↪ text` |

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -1529,6 +1529,16 @@ func historyMessages(msgs []provider.Message, resolveUserContent func(string) st
 	for _, m := range msgs {
 		content := m.Content
 		if m.Role == provider.RoleUser {
+			// Mid-turn steer messages are persisted in the session so they
+			// survive tab switches. They are surfaced as a notice (↪ text)
+			// — matching the live Steer event look — rather than as a
+			// regular user bubble or being filtered as synthetic (#4044).
+			// Check against the raw m.Content: resolveUserContent applies
+			// StripComposePrefixes which trims trailing whitespace.
+			if steerText, isSteer := agent.SteerText(m.Content); isSteer {
+				out = append(out, HistoryMessage{Role: "notice", Content: "↪ " + steerText})
+				continue
+			}
 			content = resolveUserContent(m.Content)
 			if control.IsSyntheticUserMessage(content) {
 				continue
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -334,10 +334,29 @@ func (a *Agent) SessionCache() (hit, miss int) {
 func (a *Agent) ContextWindow() int { return a.contextWindow }
 
 // mid-turn steer marker.
-const midTurnSteerPrefix = "[Mid-turn steer queued by the user. Do not treat this as a new task; use it only as additional guidance for the current task after completing the current step.]"
+// MidTurnSteerPrefix marks user messages that were injected mid-turn as
+// guidance (via Steer). The model sees them as instructions; frontends
+// display them as a notice, not a regular user bubble.
+const MidTurnSteerPrefix = "[Mid-turn steer queued by the user. Do not treat this as a new task; use it only as additional guidance for the current task after completing the current step.]"
 
 func midTurnSteerMessage(text string) string {
-	return midTurnSteerPrefix + "\n" + text
+	return MidTurnSteerPrefix + "\n" + text
+}
+
+// SteerText checks whether content is a mid-turn steer message and, if so,
+// returns the original user text without the wrapper prefix. The returned
+// text preserves the user's exact input — it only strips the prefix and the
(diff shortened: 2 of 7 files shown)


esengine:
Excellent root-cause writeup. The fix is exactly right: `syntheticPrefixes` has a single consumer (desktop's `historyMessages`), so dropping the steer prefix from it only affects history replay — nothing in compaction or prompt construction — and you re-handle it explicitly in all three frontends. Detecting on the raw `m.Content` before `StripComposePrefixes` trims is the subtle bit you got right, and `SteerText` round-trips `midTurnSteerMessage` character-for-character. Bonus points for fixing the CLI/SSE replay (raw-prefix bubbles) in the same pass. Merging — thanks!

PR #4108, opened 2026-06-12
Fix system_prompt_file workspace root resolution

lifu963:
## Summary
- Resolve relative `[agent] system_prompt_file` paths against the active workspace root instead of the process cwd.
- Wire `boot.Build` to pass its resolved workspace root into system prompt loading.
- Add config tests for workspace-relative and absolute prompt file paths.

## Test plan
- `gofmt -l . | rg -v '^desktop/' || true`
- `go test ./internal/config/ -run 'TestResolveSystemPromptForRoot' -v`
- `go test ./internal/config/ -count=1`
- `go test ./internal/boot/ -count=1`

## Notes
- PR diff intentionally excludes local-only files such as `AGENTS.md`, `architecture/`, `context/`, and `proposal.md`.
- Manual Desktop validation was not run in this session.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/boot/boot.go
+++ b/internal/boot/boot.go
@@ -154,7 +154,7 @@ func Build(ctx context.Context, opts Options) (*control.Controller, error) {
 		return nil, err
 	}
 
-	sysPrompt, err := cfg.ResolveSystemPrompt()
+	sysPrompt, err := cfg.ResolveSystemPromptForRoot(root)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


esengine:
Clean fix. Resolving `system_prompt_file` against the resolved workspace root means it now follows the same path rules as skills, memory, and write-roots — exactly right for Desktop tabs where the process cwd isn't the project root. Nice touch keeping `ResolveSystemPrompt()` as a `ForRoot(".")` shim so the old cwd-relative behavior is byte-for-byte preserved, and the relative/absolute tests cover both branches. Merging — thanks!

PR #4118, opened 2026-06-12
Add desktop hooks settings UI / 添加桌面端 Hooks 设置界面

SivanCola:
## Summary
- Add a desktop Hooks settings page for global and project scopes, including settings path display, project trust state, and direct JSON copy/paste/format/save.
- Add Wails bindings for reading, saving, and trusting Hooks settings with an explicit workspace root.
- Add `/hooks trust` support in the CLI/controller path and regression coverage for the trust behavior.

## Safety
- Project hook trust remains stored in user-global state, not in the project settings file.
- The desktop trust action uses the workspace root shown in the settings view, avoiding ambiguity when multiple projects are open.

## Verification
- `wails generate module`
- `go test` from the desktop package
- `go test ./internal/hook`
- `go test ./internal/control` with an isolated HOME
- `git diff --check`


SivanCola:
Fixed the Windows CI failure in 211aa260.

Root cause:
- The new hook trust test isolated `HOME` but not `USERPROFILE`; Windows `os.UserHomeDir()` reads `USERPROFILE`, so the test wrote and read different trust stores.

What changed:
- The test now sets both `HOME` and `USERPROFILE` to the same temporary home before exercising `/hooks trust`.

Verified:
- `go test ./internal/control` with isolated home passed locally.
- `go test` from the desktop package passed locally.
- `pnpm --dir desktop/frontend typecheck` passed locally.
- `pnpm --dir desktop/frontend check:css` passed locally.
- GitHub CI is now green, including `test (windows-latest)`.

esengine:
Thanks @SivanCola — really solid hooks settings UI. The binding contract (interface + dev mock kept in sync) and the trust model (project trust in global state, no auto-trust on save) are exactly right, and the test coverage is great.

I pushed one follow-up commit before merging: `SaveHooksSettings` recomputed the *active* project root, so the codex note is right — switching the active tab while the settings modal stayed open could write the edited JSON into the wrong workspace. Added a root-aware `SaveHooksSettingsForRoot` mirroring your `TrustProjectHooksForRoot`, and the UI now passes the displayed `projectRoot` on save, with a regression test. Merging once CI is green.

PR #4120, opened 2026-06-12
fix: skip fade-in animation for history-restored messages

Diarica:
Session restore loads every message with `animation: graphite-message-in 340ms ease-out both`. The `both` fill-mode means ALL messages start at `opacity:0` and animate in simultaneously — on a session with dozens of turns this causes visible text flickering / flashing on startup.

**Diagnosis:** Historical messages are reloaded all at once during session restore. Each `.msg` element has a CSS `graphite-message-in` keyframe animation (opacity 0→1 + translateY 10px→0, 340ms). With `animation-fill-mode: both`, all elements start invisible, then simultaneously fade in — the user sees a flash.

Live-streamed messages should keep the entrance animation; only static history items need to skip it. History items are reliably identified by their `"h"`-prefixed IDs (set by `historyMessagesToItems` in useController).

**Fix:** Add a `data-history-restore` attribute on `.msg` elements whose id starts with `"h"`, and disable the animation via CSS:

```css
.msg[data-history-restore] { animation: none; }
```

- **Message.tsx**: accept `id` prop on UserMessage; set `data-history-restore` on both UserMessage and AssistantMessage when id starts with `"h"`
- **Transcript.tsx**: pass `id` to UserMessage (hot zone + warm zone)
- **styles.css**: add override rule after the `.msg` animation definition

### DIFF
--- a/desktop/frontend/src/components/Message.tsx
+++ b/desktop/frontend/src/components/Message.tsx
@@ -65,11 +65,13 @@ export function UserMessage({
   failed,
   turn,
   anchorId,
+  id,
 }: {
   text: string;
   failed?: boolean;
   turn?: number;
   anchorId?: string;
+  id?: string;
 }) {
   const t = useT();
   const imSource = parseImSourceMessage(text);
@@ -106,6 +108,7 @@ export function UserMessage({
       data-question-anchor={anchorId}
       data-turn={turn}
       data-im-source={imSource?.provider || undefined}
+      data-history-restore={id && id.startsWith("h") ? "" : undefined}
     >
       <div className="msg__body">
         {imSource ? (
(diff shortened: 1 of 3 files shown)


esengine:
Verified the fix where it matters: the modified `UserMessage` call site is the **hot zone** (`for i := hotStartIdx..items.length`), which is exactly what renders all of the recent turns at once on restore and produces the simultaneous fade-in flash — so the reported flicker is fixed. The `"h"`-prefix check is reliable too: history items are `h*`/`hp*`, live items are `a/u/n/p/c/s/e`, so it never catches a live message, and the `[data-history-restore]` override has higher specificity than the base `.msg` animation. Nice, targeted fix — merging.

Minor follow-up (not blocking): the warm-zone `UserMessage` (the second call site, for an expanded older turn) doesn't get `id`, so on manual expand the user bubble still animates while the assistant bubble doesn't (AssistantMessage reads `item.id` directly). One-line `id={it.id}` there would make it fully consistent — I'll tidy that up separately.

PR #4150, opened 2026-06-12
feat(desktop): add time filter to sidebar project tree

ttmouse:
Add a clock button in the project tree header that lets users filter sidebar topics by time.

- All / Latest 10 / Latest 20 / Last 1h / 3h / 5h / 24h
- Dropdown grouped into three sections (all / count / time)
- Pure frontend filtering using `topicActivityTime()`
- Persistent filter state via `localStorage`
- Empty-state message + clear filter button when no results match

### Notes
- Backend (Go) changes for `CreatedAt` persistence on `ProjectNode` were already present in upstream, no backend changes needed
- Only frontend changes (6 files, +290/-10 lines)
- Filter is purely client-side, no extra API calls

### DIFF
--- a/desktop/frontend/src/App.tsx
+++ b/desktop/frontend/src/App.tsx
@@ -695,11 +695,11 @@ export default function App() {
   const [sidebarImExpanded, setSidebarImExpanded] = useState(false);
   const [isDevBuild, setIsDevBuild] = useState(false);
   const [sidebarCollapsed, setSidebarCollapsed] = useState(loadSidebarCollapsed);
-  type TimeFilter = "all" | "10" | "1h" | "3h" | "5h" | "1d";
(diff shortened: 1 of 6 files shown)


ttmouse:
image

esengine (approved):
Nicely self-contained — i18n in both locales, theme tokens throughout, persisted filter state, and a real empty state with a clear-filter action. I read through the filter logic (the time + count cutoffs and the folder-visibility handling) and it holds up.

One real thing, which @chatgpt-codex-connector also caught: the dropdown is positioned `left: 0` inside the `overflow: hidden` sidebar, so on a narrow sidebar — and with the longer zh labels — it clips at the right edge. Merging this and right-aligning the menu in a quick follow-up so you don't have to round-trip. Thanks for the contribution!

PR #4165, opened 2026-06-12
feat:add custom font family support

myNinety:
## Summary

This PR adds a **"Custom" font family** option to the desktop app's Appearance settings, allowing users to specify any font stack they prefer (e.g., `"LXGW WenKai", "JetBrains Mono", monospace`).

Previously, users were limited to four hardcoded choices: System Default, YaHei, PingFang, and Noto Sans. With this change, users can type in an arbitrary CSS `font-family` value, which is persisted in `localStorage` and applied via a CSS custom property (`--font-family-custom`).

## Motivation

- Users with specific typographic preferences (e.g., open-source CJK fonts like LXGW WenKai, or monospace coding fonts) had no way to override the built-in options.
- The "Custom" + textarea pattern is a low-cost, high-flexibility approach that avoids bloating the preset list while giving full control to power users.

## Changes

| File | Change |
|---|---|
| `desktop/frontend/src/lib/fontFamily.ts` | Added `"custom"` to `FONT_FAMILIES`; added `getCustomFontName()` / `setCustomFontName()` helpers; `applyFontFamily()` now sets `--font-family-custom` CSS custom property when `"custom"` is selected. |
| `desktop/frontend/src/components/SettingsPanel.tsx` | Added `customFontName` state + textarea input, wired into `AppearanceSection`; the textarea only renders when `fontFamily === "custom"`. |
| `desktop/frontend/src/styles.css` | Added `:root[data-font-family="custom"]` rule that uses `var(--font-family-custom)` as primary font with PingFang SC → Microsoft YaHei → Noto Sans SC → system sans-serif fallback chain. |
| `desktop/frontend/src/locales/en.ts` | Added i18n keys: `settings.fontFamilyCustom`, `settings.fontFamilyCustomName`, `settings.fontFamilyCustomPlaceholder`. |
| `desktop/frontend/src/locales/zh.ts` | Added Chinese i18n keys for the above. |

## How It Works

1. User selects **"Custom"** from the font family dropdown in **Settings → Appearance**.
2. A textarea appears below, accepting a CSS `font-family` value (e.g. `"LXGW WenKai", "JetBrains Mono", monospace`).
3. On change, the value is stored in `localStorage` under `reasonix-font-family-custom` and injected as `--font-family-custom` on `document.documentElement`.
4. The CSS rule `:root[data-font-family="custom"]` reads `var(--font-family-custom)` as the primary `--sans` value, with the existing system font stack as fallback.
5. Switching back to "System Default" clears the custom property; switching to other presets does not interfere.


SivanCola:
Thanks for adding this. I found one behavior issue that should be fixed before merge:

When `Custom` is selected before the user enters a font stack, or after the textarea is cleared, `applyFontFamily("custom")` still writes `--font-family-custom` with an empty string. The CSS then builds `--sans` as `var(--font-family-custom), ...`, so the computed font stack can start with a leading comma instead of reliably falling back to PingFang / YaHei / Noto / system fonts.

Could you guard empty custom values? For example, trim the custom value and remove/avoid `--font-family-custom` when it is blank, or structure the CSS/JS so an empty custom value falls back to the existing default stack until the user provides a non-empty font family.

I verified the PR merges cleanly with current `main-v2`. After generating Wails bindings, these passed locally: `typecheck`, `test:all`, `check:css`, and `build`.

esengine:
Thanks for this — landed it as #4194 on current `main-v2`. I fixed the empty-value case @SivanCola flagged (blank custom name no longer writes an empty `--font-family-custom`; the CSS now falls back via `var(--font-family-custom, "PingFang SC")`, so the stack never starts with a leading comma) and dropped a redundant double `applyFontFamily` call. Credit kept to you, with @SivanCola's review folded in. Merging #4194 once CI is green and closing this in favour of it.

Issue #4148, opened 2026-06-12
macOS: os.UserConfigDir() != ~/.config/ — config write path broken, sandbox prevents fixing it


SvenZhao:
## 问题描述

**一句话：macOS 上 `os.UserConfigDir()` 返回 `~/Library/Application Support/`，但文档和渲染注释全部硬编码为 `~/.config/reasonix/config.toml`，导致用户把配置放在 `~/.config/reasonix/` 后完全不生效。**

并且沙箱 `bash = "enforce"` 阻止了从内部修复——形成鸡生蛋死锁。

---

## 根因

代码 `internal/config/config.go` 中：
```go
func userConfigPath() string {
    dir, err := os.UserConfigDir()  // macOS → ~/Library/Application Support
    ...
    return filepath.Join(dir, "reasonix", "config.toml")
}
```

macOS 上 `os.UserConfigDir()` 返回 `~/Library/Application Support/`，而非 `~/.config/`。参见 Go 标准库文档。

### 1. 文档全部错误

所有以下位置硬编码了 `~/.config/reasonix/config.toml`，macOS 上均为错误信息：

- `internal/config/config.go:2` — 注释
- `internal/config/render.go:43` — 渲染到用户配置文件顶部的注释
- `internal/config/config.go:1543` — `UserConfigPath` 文档注释
- `internal/config/config.go:1548` — `UserCredentialsPath` 文档注释
- `README.md` / `README.zh-CN.md` / `docs/MIGRATING.md` / `docs/SPEC.md`

渲染出的 `config.toml` 顶部注释：
```
# Resolution order: flag > ./reasonix.toml > ~/.config/reasonix/config.toml > built-in defaults.
```
macOS 用户看到这个会以为配置放 `~/.config/reasonix/` 就行，但实际上放那里完全不会被读取。

### 2. 沙箱鸡生蛋死锁

- 用户把配置放到 `~/.config/reasonix/config.toml`（文档说的路径）—— reasonix 不读
- reasonix 实际读的是 `~/Library/Application Support/reasonix/config.toml`——但这个文件是自动生成的，**没有 `allow_write`**
- 没有 `allow_write` → 沙箱阻止写入任何非 workspace 路径
- 无法写入 → 无法添加 `allow_write` 到正确的配置文件

结果：**用户根本无法通过自身工具修复沙箱写入权限。**

### 3. 用户实际状态

macOS 上最终出现**两个并行目录**：

```
[...]

PR #4172, opened 2026-06-12
Add macOS home migration command / 新增 macOS 用户目录迁移命令

SivanCola:
## Summary
- add a shared home migration core plus `reasonix migrate-home` dry-run/apply CLI flow
- preserve legacy macOS Application Support compatibility until a completed migration marker switches CLI and desktop state to the documented root
- merge config/credentials safely, preserve session conflicts, and document the migration in English and Chinese

Addresses #4148.

## Testing
- `go test ./...`
- `git diff --check origin/main-v2..HEAD`


esengine:
Thanks for the careful work here — the dry-run/apply flow, the backup + conflict isolation, and the bilingual docs are all genuinely well done, and CI is green across the board.

I'm going to close this one, and I want to be transparent that it's a direction call, not a craft issue.

#4148 was already resolved by #4157, which I merged earlier today. That fix deliberately keeps `os.UserConfigDir()` (i.e. `~/Library/Application Support/reasonix` on macOS) as the single source of truth and only realigns the docs and rendered output to show the real per-OS path. Moving the canonical root to `~/.config/reasonix` — even behind a migration command — reintroduces exactly the dual-tree state I was trying to avoid: config and app state split across two locations, with precedence/merge logic to reconcile them, plus the migration machinery to manage it. That cost is the reason I went the docs-alignment route instead.

One piece here I do think is worth keeping: the `REASONIX_HOME` override. An explicit env var for advanced users to point the root wherever they want is clean and doesn't create the dual-directory problem. If you'd like to spin just that out into a small standalone PR (env override only — no default-root change, no migration), I'd be glad to take it.

Really appreciate the effort, and sorry this one landed right as the issue was being closed from the other side.

PR #4182, opened 2026-06-12
fix(desktop): reconcile stale completed turns

### LINKED ISSUE #4167: [Bug]: 已经执行完了，不停止
### Version line

### Exact version

desktop 1.5.0

### What happened?

已经执行完了，不停止

Image

### Steps to reproduce

已经执行完了，不停止

### OS / platform

macos

### Relevant logs or output

```shell

```

GTC2080:
## Summary
- Fix stale desktop running state when the final `turn_done` event is missed after the message stream has already closed.
- Rework the stale-turn watchdog to reconcile any quiet active turn, not only turns with a live text/reasoning stream.
- Track turn activity per tab so background-tab events do not delay recovery for the active tab.

Fixes #4167.

## Root cause
The desktop frontend only reconciled a missed `turn_done` when `state.live` was still present. After a final message or synthetic todo update, `live` is cleared before `turn_done`; if that last event is dropped, the UI can show completed todos while still displaying the running/stop state.

## Verification
- `npx tsx src/__tests__/use-controller-meta.test.ts` (red before the fix, green after)
- `npm run test:all` in `desktop/frontend`
- `npm run build` in `desktop/frontend`
- `go test ./...`
- `go test ./...` in `desktop`
- `gofmt -l .`
- `git diff --check`
- Browser smoke: `http://127.0.0.1:5173/` loads `Reasonix` with no console errors

### DIFF
--- a/desktop/frontend/src/__tests__/use-controller-meta.test.ts
+++ b/desktop/frontend/src/__tests__/use-controller-meta.test.ts
@@ -1,6 +1,6 @@
 // Run: tsx src/__tests__/use-controller-meta.test.ts
 
-import { sameMeta } from "../lib/useController";
+import { initialState, reducer, sameMeta, shouldReconcileStaleTurn } from "../lib/useController";
 import type { Meta } from "../lib/types";
 
 let passed = 0;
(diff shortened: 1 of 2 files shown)


esengine (approved):
Nice fix — the root cause is exactly right: gating the watchdog on `s.live` meant a dropped `turn_done` after the final message (which clears `live` first) could never be reconciled, leaving the UI stuck "running". Switching the gate to `turnActive`, tracking activity per tab, and broadening the activity events so tool-heavy turns don't false-trip are all the right calls, and `reconcileTabRuntime` keeps it safe (it syncs from the backend rather than force-stopping). The red→green helper test is a good touch. CI is green across the board. Merging.

PR #4189, opened 2026-06-12
fix(desktop): Unify the stylistic consistency of Chinese translations.

haoyuenan:
### 改动内容
将工作正在进行时随机显示的有趣状态的中文翻译改为统一风格，即“动词”+“中”的形式。
比如，原来的“文火慢炖”改为“慢炖中”，“来感觉了”改为“共鸣中”。
共改动了四个词组，改完后所有中文翻译均保持“动词”+“中”的统一风格。


esengine:
Thanks for the careful pass here, @haoyuenan — the consistency observation is fair, and most of the list is already in the 「X中」 shape.

We're going to keep the current set, though. For these spinner words the whimsy is the whole point, and the few four-character outliers (文火慢炖, 嘎吱运算, 来感觉了) are deliberately the flavorful ones — they carry more character than a uniform 「动词+中」, and the English list leans on varied, playful vocabulary (Simmering / Crunching / Vibing) rather than a rigid template. So I'd rather not flatten them.

Genuinely appreciate the eye for detail on the copy — please don't let this put you off; this kind of attention is welcome, and I'd be glad to see other contributions from you.

PR #4221, opened 2026-06-12
fix(migrate): import legacy .jsonl sessions alongside .events.jsonl

### LINKED ISSUE #4156: [Bug]: 无法将0.53.0聊天记录迁移至1.6.0 Cannot migrate chat history from 0.53.0 to 1.6.0
### Version line

### Exact version

1.6.0

### What happened?

我下载了1.6.0 desktop，打开后mcp服务器和deepseek api均自动同步，但聊天记录没有。我尝试删除%APPDATA%/reasonix和%APPDATA%/Reasonix.exe并重新加载，依然无法同步。如有需要，可以向我索要~/.reasonix/sessions中的文件。
I downloaded 1.6.0 desktop. When I open it, I found mcp servers and deepseek api were migrated successfully, but I cannot saw any chat history. I tried to remove %APPDATA%/reasonix and %APPDATA%/Reasonix.exe to restart, but seems useless. I can provide files in ~/.reasonix/sessions if the developers need them.

### Steps to reproduce

1. 下载1.6.0 desktop
2. 打开后无聊天记录

1. Download 1.6.0 desktop
2. Open it but found no chat history

### OS / platform

Windows 11

### Relevant logs or output

```shell

```

heshuimu:
The v0.x session migration only processed *.events.jsonl files. .jsonl-only sessions (ACP, desktop, subagent — later TS versions wrote directly in message format) were silently skipped.

**Changes:**
- Pass 1 (events): prefer .jsonl sidecar when newer than event log
- Pass 2 (.jsonl-only): import .jsonl files without events counterpart; skip subagent; .bak recovery
- Pass 3 (subdirectories): recurse into project-scoped subdirs
- transformAndCopyJsonl: flatten legacy nested-function tool calls to flat format. This heals the Desktop version displaying the session history where all the tool call names and input from legacy version are lost due to format differences. 
- Desktop: topic migration handles both global and project dirs, runs for all tabs
- CLI: resolveCLISessionDir uses project dir so /resume shows project history

**Tests:** 9 new migration tests; all 13 original unchanged. Desktop + CLI tests pass.
Testing is also done against my own data from the legacy TypeScript version. 

The only unavoidable defect of the current approach is that ACP-related legacy sessions have no workspace specified in its metadata, so they have to be lumped into the global session. 

This aims to resolve #4156 which #3944 did not fully address. 
This may have overlaps with #4215 as my change also migrates the subagent sessions of legacy version into the designated directory.

### DIFF
--- a/desktop/tabs.go
+++ b/desktop/tabs.go
@@ -949,24 +949,26 @@ func (a *App) buildTabController(tab *WorkspaceTab) {
 
 	sessionDir := desktopSessionDir(root)
 	topicID := strings.TrimSpace(tab.TopicID)
-	if tab.Scope == "global" {
-		migratedTopics := migrateLegacySessionsIntoGlobalTopics(config.SessionDir())
-		if len(migratedTopics) > 0 {
-			a.emitProjectTreeChanged()
+
+	// Assign Global topics to legacy sessions in the global session dir so
+	// imported history appears in the project tree regardless of which tab
+	// triggered the build (the migration now sends everything to global).
+	migratedGlobalTopics := migrateLegacySessionsIntoGlobalTopics(config.SessionDir())
+	if len(migratedGlobalTopics) > 0 {
+		a.emitProjectTreeChanged()
+	}
+	if tab.Scope == "global" && topicID == "" && len(migratedGlobalTopics) > 0 {
+		topicID = migratedGlobalTopics[0]
+		topicTitle := topicTitleForTab("global", "", topicID)
+		a.mu.Lock()
+		if strings.TrimSpace(tab.TopicID) == "" {
+			tab.TopicID = topicID
+			tab.TopicTitle = topicTitle
+			a.saveTabsLocked()
+		} else {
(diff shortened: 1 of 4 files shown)


esengine:
Thanks for this — the three-pass approach (events / `.jsonl`-only / subdirs) plus the nested-function tool-call flattening is exactly what the `.jsonl`-only sessions needed, and testing it against your own legacy data is appreciated.

I picked up two follow-ups in #4260 so it lands cleanly and reaches everyone it should:

- The early return in `migrateLegacySessions` bailed on the routed marker *before* the new `v3-jsonl` marker was checked, so users whose events pass already completed in a prior version (routed marker already stamped) never ran Pass 2/3 — the very upgraders the new marker was meant to serve. Gated the return on both markers; Pass 1 is idempotent via its dest checks. Added a regression test for that path.
- Dropped the unused `copyFile` — golangci-lint's `unused` linter would have failed CI on it.

#4260 builds on your commit with credit and closes this one. Nicely done.

PR #4247, opened 2026-06-13
style: unify expanded left-padding for foldable sections

CVEngineer66:
## Summary

Change the expanded-state  of ReadOnlyBatch, TurnCollapse, and ToolCard from 30px to 12px, matching the collapsed state and keeping the visual border indent consistent with the Reasoning section.

All four collapsible sections now share the same 12px + 2px border indent.


SivanCola (approved):
approve

PR #4253, opened 2026-06-13
feat(desktop): add workspace status bar items

### LINKED ISSUE #4063: [Feature]: Add project path, Git branch, and localized cost information to the status bar
### Version line

When working with multiple repositories, it can be difficult to identify the currently active workspace at a glance, especially when different projects share the same directory name.

The status bar already displays model and usage statistics, but it does not provide project-specific context such as the active project path or Git branch. This increases the risk of working in the wrong repository or branch.

Additionally, session cost is currently displayed only in CNY (¥), which may be less intuitive for users outside China.  Same for the  cost display in the Overview.

### Proposed solution

Extend the status bar area to display additional workspace information:

* Active project path (or sandbox path)
* Active Git branch
* Localized cost display based on user locale or preferred currency (or show both local currency and CNY)
* Same for the  cost display in the Overview 

This would make it easier to identify the current workspace, reduce mistakes when switching between repositories, and improve cost transparency for international users.

Screenshots/mockups attached below show a possible layout.

Image

Image

lizhengwu:
## Summary

  Fixes #4063.

  This PR adds workspace context to the desktop status bar by
  integrating workspace and Git branch as configurable status
  bar items.

  - Add `workspace` and `git_branch` to the existing
  configurable status bar item framework.
  - Show the active tab workspace path and Git branch in the
  status bar.
  - Keep workspace/Git items hideable and reorderable through
  the existing status bar settings.
  - Avoid exposing or displaying configured sandbox path as
  workspace context, since it is not confirmed to be the tab-
  effective sandbox path.

  ## Why

  The desktop status bar was missing stable workspace context,
  making it harder to identify which workspace/session is
  currently active. This is especially noticeable when multiple
  tabs or projects are open.

  ## Test Plan

  - `npm test`
  - `go test . -run 'TestMetaForTabIncludesWorkspaceContext|
  TestListTabsDoesNotExposeConfiguredSandboxPath|
  TestMetaForTabLeavesGitBranchEmptyOutsideGit|
  TestWorkspaceGitBranch|TestWorkspaceGitDisablesDaemonSpawns'
  -count=1`
  - `git diff --check`


esengine:
This landed. `a6aae98f8` (`feat(desktop): switchable workbench layout style`, #4369) states it integrates #4253, and you are recorded as its author.

The status bar items you built became part of a larger layout change rather than merging on their own — the surface they live on was being reworked at the same time, and shipping them separately would have meant building them twice.

Leaving this open afterwards was our mistake; from your side it looks like 546 lines that went nowhere, when they are in the product. Closing it now with that said.

Thanks for it.

PR #4286, opened 2026-06-13
fix(desktop): replay pending prompts on tab switch to preserve plan m…

myipanta:
…odal (#4275)

When switching back to a plan-mode tab, the 3-button modal (revise/execute/exit) was lost because ReplayPendingPrompts() was only called on initial mount, not on tab change. The backend still held the pending exit_plan_mode approval event but the frontend never re-emitted it.

Root cause: ReplayPendingPrompts() was only in the mount-effect, never triggered when activeTabId changes.

Fix: add a useEffect that calls app.ReplayPendingPrompts() on every activeTabId change, so switching back to a plan-mode session rebuilds the approval modal.

### DIFF
--- a/desktop/frontend/src/__tests__/send-failed.test.ts
+++ b/desktop/frontend/src/__tests__/send-failed.test.ts
@@ -1,6 +1,6 @@
 // Run: tsx src/__tests__/send-failed.test.ts
 
-import { initialState, reducer } from "../lib/useController";
+import { initialState, reducer, replayPendingPromptsForActiveTab } from "../lib/useController";
 import type { WireEvent } from "../lib/types";
 
 let passed = 0;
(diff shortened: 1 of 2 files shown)


SivanCola:
Added a supplemental regression commit `fd2967b2` for the tab-switch prompt replay behavior.

What changed:
- Extracted `replayPendingPromptsForActiveTab` so the tab-switch replay contract is directly testable.
- Extended the existing frontend test to verify that no active tab does not replay, an active tab does replay, and replay bridge failures are swallowed instead of breaking the effect.

Verified:
- `pnpm --dir desktop/frontend test` passed
- `go test ./internal/control -run TestReplayPendingPrompts -count=1` passed
- `git diff --check HEAD~1..HEAD` passed

Note: `pnpm --dir desktop/frontend exec tsc --noEmit -p tsconfig.test.json` is still blocked in this checkout because the Wails generated module types are absent (`desktop/wailsjs/go/main/App`).

SivanCola (approved):
Approved. This fixes the reported tab-switch plan approval loss by replaying pending approval/ask prompts whenever the active tab changes, while preserving the existing mount/reconnect replay path.

Why this solves #4275:
- The backend already retains the pending `exit_plan_mode` approval while the run loop waits.
- `ReplayPendingPrompts()` re-emits that pending prompt, and desktop tab events carry `tabId`, so the frontend reducer can rebuild the modal for the correct tab.
- The supplemental regression now covers the active-tab replay trigger and the failure-swallowing behavior expected from the effect.

Verified:
- `pnpm --dir desktop/frontend test` passed
- `go test ./internal/control -run TestReplayPendingPrompts -count=1` passed
- `git diff --check HEAD~1..HEAD` passed

PR #4314, opened 2026-06-13
fix: ensure todo panel syncs after complete_step sign-off

myipanta:
Backend: advanceCanonicalTodo now emits current todoState snapshot even when MatchStep fails or the step is already completed, preventing silent UI staleness.

Frontend: add state.items.length to todoEntry useMemo deps for reliable re-computation when synthetic todo_write items are appended.

### DIFF
--- a/desktop/frontend/src/App.tsx
+++ b/desktop/frontend/src/App.tsx
@@ -1289,7 +1289,7 @@ export default function App() {
       }
     }
     return null;
-  }, [state.items]);
+  }, [state.items, state.items.length]);
   const todoItem = todoEntry?.item ?? null;
   const todos = useMemo(() => (todoItem ? parseTodos(todoItem.args) : []), [todoItem]);
   const [dismissedTodo, setDismissedTodo] = useState<string | null>(null);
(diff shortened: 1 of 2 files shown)


esengine:
The bug you found is still there — this is worth rebasing rather than closing.

**Backend: still broken, and your diagnosis holds.** `advanceCanonicalTodo` today reads:

```go
m, ok := evidence.MatchStep(step, a.todoState)
if !ok || !evidence.AdvanceSerialTodo(a.todoState, m.Index-1) {
    a.todoMu.Unlock()
    return          // no emit — the panel keeps rendering the previous state
}
```

Same silent staleness you described. The reason the branch conflicts is that the condition itself was rewritten since June: the `canonicalTodoStatus(...) == "completed"` check your patch targets is now `!evidence.AdvanceSerialTodo(...)`. The failure path moved, the missing emit did not.

**Frontend: worth reconsidering.** `[state.items, state.items.length]` only changes behaviour when `state.items` keeps its identity while its contents change — i.e. when the array is being mutated in place. If that is happening, the dependency array is compensating for it rather than fixing it, and the same staleness will resurface anywhere else that reads `state.items`. If you can point at where the mutation happens, that is the more durable fix; if it turns out the array is genuinely replaced and the memo was fine, the backend half stands on its own.

Happy to take the backend change by itself if you would rather split it — a one-line emit on the failure path is easy to review and needs no frontend decision.

Sorry this sat for two months. It is a real bug and the report was precise.

PR #4325, opened 2026-06-13
Improve crash stats dashboard UI / 优化崩溃统计面板 UI

SivanCola:
## Summary
- Add a header language switcher for Chinese and English, persisted locally in the browser.
- Add y-axis tick labels, gridlines, and wider hover targets to the daily active installs chart.
- Redesign crash filters into grouped status and facet chips that handle long version labels cleanly.
- Redesign crash groups as row-level clickable diagnostic cards with clearer scope, health, and count columns.
- Redesign crash group detail pages with a compact summary header, key metrics, collapsible samples, copy actions, and a tighter admin management layout.

## Verification
- Worker TypeScript typecheck passed.
- `git diff --check` passed.
- Verified the stats dashboard and crash group detail page in the in-app browser with local previews.
- Verified Chinese/English switching, sample expansion, row-level navigation targets, and copy-button success/failure states.

## Cache impact
- No prompt-cache or provider-visible request surface changes; this only affects the crash stats Worker UI.


esengine:
Reviewed — the dashboard redesign reads well and the escaping is consistent: every report-derived field is `esc()`'d (versions, titles, samples, copy-button payloads) and breadcrumbs reuse the existing helper, so no XSS regression.

One thing before it can land: this conflicts with `main-v2` because #4322 relabeled the dashboard from "Crash / 崩溃" to "Report / 诊断", and this branch predates that. Could you merge `main-v2` and reconcile? Since you authored both, it's your call whether the redesign keeps the "diagnostic / 诊断" framing. Once it's green I'll merge.

SivanCola:
Addressed in b1b6e650.

What changed:
- Merged the latest main-v2 into this PR branch.
- Kept the stats dashboard on the Report / diagnostic framing and cleaned the remaining dashboard-facing copy to say diagnostic reports / report data instead of crash dashboard / crash data.

Verified:
- npm --prefix workers/crash-report run typecheck
- go test -race ./internal/control
- GitHub CI is green for this head, including race.

PR #4331, opened 2026-06-13
perf(desktop): wrap Transcript in React.memo to prevent IME input lag

weikejia123:
## Problem

When renaming a session topic or performing any action that triggers an App-level `setState` (e.g. typing in the rename input), the Transcript component re-renders even though none of its props changed. This causes all ~100+ child `useGSAPCollapse` `useLayoutEffect` hooks to execute synchronously, blocking the main thread and causing visible IME input lag.

## Root Cause

`Transcript` was exported as a plain function component without `React.memo`. Since Transcript is purely prop-driven (no internal context dependencies), it should not re-render when parent state changes that don't affect its props.

## Fix

Wrap Transcript in `React.memo`:

```diff
-export function Transcript({
+export const Transcript = memo(function Transcript({
   ...
-}
+});
```

`memo` was already imported at the top of the file — zero new imports needed.

## Change

- 1 file, 2 lines changed

### DIFF
--- a/desktop/frontend/src/components/Transcript.tsx
+++ b/desktop/frontend/src/components/Transcript.tsx
@@ -144,7 +144,7 @@ function buildTurnGroups(items: Item[], questions: QuestionAnchor[]): TurnGroup[
 
 // ── Transcript component ──────────────────────────────────────────────────────
 
-export function Transcript({
+export const Transcript = memo(function Transcript({
   items,
   live,
   footerHeight = 0,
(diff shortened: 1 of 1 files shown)


SivanCola:
Closing this after the merged integration PR #5512.

Resolution:
- This memoization approach was reviewed while preparing #5512, but was not adopted because the merged fix focuses on reducing the long-history hydrate payload and reducer work rather than parent re-render memoization.
- The investigation is still credited in #5512 as a reviewed-but-not-adopted performance approach.

Attribution:
- @weikejia123 is credited in #5512's consolidation notes and repository contributor credits.

Verification:
- #5512 is merged and its GitHub checks are green, including cache-impact, desktop, lint, race, coverage, CodeQL, and ubuntu/macos/windows tests.

PR #4346, opened 2026-06-14
perf: UserMessage memo + normalizeMath cache + Transcript hook stabilization

CVEngineer66:
## Summary

Three independent performance optimizations for the chat transcript during streaming. No behavioral changes.

1. **UserMessage memo** — UserMessage was the only top-level message component not wrapped in React.memo. Every streaming token caused reconciliation of all completed user messages (photo preview, mentions, attachments — entire subtree diffed for zero gain).

2. **normalizeMath cache** — The math pre-processing pipeline (multi-pass regex for LaTeX delimiter conversion, code-block protection, inline math classification) was re-running on every render. Now wrapped in useMemo.

3. **Transcript hook deps stabilization** — 3 hot-path hooks depended on the `items` array reference, which changes on every streaming token:
   - `contentVersion` — now depends on a stable key based on items.length + item kind count
   - `hotStartIdx` — now depends on user-message count instead of full items array
   - `stepGroups` — now depends on items.length + streaming flag
   - Removed the unused `scrollVersion` function

## Files

- `desktop/frontend/src/components/Message.tsx` — UserMessage memo
- `desktop/frontend/src/components/Markdown.tsx` — normalizeMath useMemo
- `desktop/frontend/src/components/Transcript.tsx` — hook deps stabilization

### DIFF
--- a/desktop/frontend/src/App.tsx
+++ b/desktop/frontend/src/App.tsx
@@ -1933,14 +1933,15 @@ export default function App() {
   // Display items: truncated when an optimistic rewind is pending.
   const displayItems = useMemo(() => {
     if (!rewindState) return state.items;
-    return state.items.slice(0, rewindState.boundaryIdx);
+    // Filter out compaction cards — they're legitimate history but
+    // confusing when shown in a rewound (truncated) transcript.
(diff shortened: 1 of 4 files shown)


CVEngineer66:
Superseded by #4354 which covers all performance optimizations plus additional memory and UX fixes

PR #4352, opened 2026-06-14
fix(npm): auto-publish stable 1.x releases under latest tag

Lilyandlucy:
## Problem

`npm i -g reasonix` currently installs version **0.53.2** (legacy TypeScript build) instead of the latest Go rewrite. This is because the dist-tag logic in `npm/build.mjs` only promoted 0.x stable releases to the `latest` tag, while all 1.x releases (1.0.0, 1.1.0, 1.2.0) were published under `next`.

The README explicitly says `npm i -g reasonix` is the install command for the Go binary, but users who follow it end up with the old TypeScript version — a confusing first experience.

## Root Cause

`npm/build.mjs` lines 107-111:
```js
const distTag = version.includes("-canary.")
  ? "canary"
  : version.startsWith("0.") && !version.includes("-")
    ? "latest"
    : "next";
```

The comment said "Promote a 1.x stable to default with a manual `npm dist-tag add reasonix@<ver> latest`" — but this manual step was never performed after any release.

## Fix

Any stable release (no prerelease suffix like `-rc` or `-canary`) now automatically publishes under `latest`. The CI `release` environment approval already gates stable publishing, so the separate manual promotion step is redundant.

New logic:
- **`-canary.`** → `canary` tag
- **`-rc` etc. (any prerelease)** → `next` tag
- **Stable (no `-` in version)** → `latest` tag

## Files Changed
- `npm/build.mjs` — updated dist-tag selection logic and comment
- `.github/workflows/release-npm.yml` — updated workflow comment to reflect new behavior

### DIFF
--- a/.github/workflows/release-npm.yml
+++ b/.github/workflows/release-npm.yml
@@ -6,15 +6,14 @@ name: Release npm
 # own tag lets a prerelease ship to npm's `next` channel without touching the
 # stable Homebrew cask (and lets a stable cask ship without forcing npm to leave
 # its rc). Bump with:
-#   git tag npm-vX.Y.Z          # stable  -> publishes under `next`
+#   git tag npm-vX.Y.Z          # stable  -> publishes under `latest`
 #   git tag npm-vX.Y.Z-rc.N     # prerelease -> publishes under `next`
 #   git push origin <tag>
 #
 # A manual workflow_dispatch with base_version publishes an opt-in canary to the
 # `canary` dist-tag (npm i reasonix@canary) from the current ref; it never moves
-# `next`/`latest`. build.mjs decides the dist-tag: 0.x stable is `latest`; a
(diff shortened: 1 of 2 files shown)


esengine:
Closing, but not because the change is wrong — it conflicts with a deliberate choice about what `latest` means here.

`latest` is pinned to 0.53, the final TypeScript release. The 1.x Go rewrite publishes under `@next`, and every install instruction points there. The effect you are describing — `npm i reasonix` giving people 0.53 — is the intended behaviour of that split, not a gap in the release workflow.

The reasoning is that `latest` is what an unattended `npm install` resolves to, including in scripts and Dockerfiles written months ago, so moving it across a rewrite boundary changes what those get without anyone opting in.

That said, this is a policy that could reasonably change once 1.x stops moving as fast, and if it does, this is the change that implements it. Worth knowing it exists.

Thanks for it, and sorry the policy was not written down anywhere you could have found it — that is the actual defect here.

PR #4358, opened 2026-06-14
fix(desktop): resolve session-switch freeze on Windows 11 (#4337)

### LINKED ISSUE #4337: [Bug]: 最新1.7.0版切换会话十分卡顿, 在1.6.0版本不会
### Version line

### Exact version

1.7.0

### What happened?

桌面端从1.6.0升到1.7.0后, 整个窗口都显得十分卡顿, 在已有的会话上切换需要十几秒时间.
在1.6.0的时候没有这情况

### Steps to reproduce

1. 点击已开的会话,
2. 切换到另一个已打开的会话
3. 需要卡顿十来秒才可以, 这时候整个窗口无响应

### OS / platform

Windows11

### Relevant logs or output

```shell

```

ferstar:
## Summary

Fixes #4337 — switching chat sessions in v1.7.0 caused the entire window to freeze for many seconds on Windows 11.

## Root cause

Three compounding factors, all introduced or worsened in v1.7.0:

1. **`ListTabs` used a write lock (`Lock()`) for a read-heavy operation** — blocked the 2-second polling timer and user tab switches from proceeding concurrently.

2. **`SetActiveTab` held the write lock across disk I/O** — `saveTabsLocked()` called `os.MkdirAll` + `os.WriteFile` + `fileutil.ReplaceFile` (which can retry transient Windows locks) all while holding `a.mu.Lock()`.

3. **`handleTabChange` called `ListTabs` twice in sequence** — once inside `switchTab` → `reconcileTabRuntime`, and once more via `refreshTabMetas`.

## Changes

| # | Change | File | Impact |
|---|--------|------|--------|
| 1 | `ListTabs` now uses an `RLock` read-only fast path and reacquires `Lock` only when stale tab order needs repair | `desktop/tabs.go` | Reduces polling/switch contention without writing shared state under a read lock |
| 2 | Split `saveTabsLocked` into collect-phase (under `mu`) + write-phase (outside `mu`); `SetActiveTab` unlocks before I/O | `desktop/tabs.go` | Disk I/O no longer blocks UI; added `tabsSaveMu` + version stamping to prevent stale snapshot races |
| 3 | `reconcileTabRuntime` returns `TabMeta[]`, `switchTab` propagates it, `handleTabChange` reuses it for `setTabMetas` | `useController.ts`, `App.tsx` | One `ListTabs` call instead of two per tab switch |
| 4 | Added stale-order and stale-snapshot regression tests | `tabs_order_test.go` | Verifies concurrent `ListTabs` does not race and older snapshots do not overwrite newer saves |

## Verification

- `go test . -run 'Test(ListTabsKeepsExplicitOrderWhenActiveChanges|ListTabsRepairsStaleOrderWithoutRacing|SaveTabsSkipsOlderSnapshot)' -count=1` — clean
- `go test -race . -run TestListTabsRepairsStaleOrderWithoutRacing -count=1` — clean
- `go test -race . -run TestSaveTabsSkipsOlderSnapshot -count=1` — clean
- `go test . -count=1 -timeout=120s` — clean
- `git diff --check` — clean

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -71,6 +71,11 @@ type App struct {
 	activeTabID string
 	readyHook   func()
 
+	// tabsSaveMu serializes writes to desktop-tabs.json and its fixed .tmp path.
+	tabsSaveMu             sync.Mutex
+	tabsSaveVersion        uint64 // protected by mu; assigned when collecting a snapshot
+	tabsLastWrittenVersion uint64 // protected by tabsSaveMu
+
 	forceQuit           atomic.Bool
 	backgroundMaximised atomic.Bool
 	trayReady           bool
--- a/desktop/frontend/src/App.tsx
+++ b/desktop/frontend/src/App.tsx
@@ -1848,10 +1848,10 @@ export default function App() {
 
   const handleTabChange = useCallback(async (id: string) => {
     closeTransientOverlays();
-    await switchTab(id);
-    await refreshTabMetas();
+    const tabs = await switchTab(id);
+    if (tabs) setTabMetas(tabs);
     setTabRevealSignal((signal) => signal + 1);
-  }, [closeTransientOverlays, refreshTabMetas, switchTab]);
+  }, [closeTransientOverlays, setTabMetas, switchTab]);
 
   const handleTabClose = useCallback(async (id: string) => {
     closeTransientOverlays();
--- a/desktop/frontend/src/lib/useController.ts
+++ b/desktop/frontend/src/lib/useController.ts
@@ -663,17 +663,17 @@ export function useController() {
     return active.id;
   }, [activeTabFromBackend, loadSessionDataForTab]);
 
-  const reconcileTabRuntime = useCallback(async (tabId: string) => {
+  const reconcileTabRuntime = useCallback(async (tabId: string): Promise<TabMeta[] | undefined> => {
     const tabs = asArray(await app.ListTabs().catch(() => [] as TabMeta[]));
     const tab = tabs.find((candidate) => candidate.id === tabId);
-    if (!tab) return;
+    if (!tab) return undefined;
     const local = statesRef.current.get(tabId);
     const needsInitialLoad = !local?.meta;
     const missedTurnDone = Boolean(local?.running && !tab.running);
     dispatchTo(tabId, { type: "backend_status", running: Boolean(tab.running) });
     if (needsInitialLoad || missedTurnDone) {
       await loadSessionDataForTab(tabId, missedTurnDone);
-      return;
+      return tabs;
     }
     const [jobs, effort, balance] = await Promise.all([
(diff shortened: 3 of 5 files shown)


SivanCola (changes requested):
Thanks for chasing #4337. I found one blocker before this can merge.

- `ListTabs` now takes `RLock`, but it still calls `orderedTabIDsLocked`, and that helper can assign to `a.tabOrder` when the saved/order state is stale or missing IDs. With the new read lock, two concurrent `ListTabs` callers can read/write `tabOrder` under shared readers. That is directly on the path this PR is trying to make concurrent: the polling refresh and tab-switch reconciliation can overlap.

  I reproduced this with a local race regression that creates a stale `tabOrder` and calls `ListTabs` concurrently; `go test . -race -run TestListTabsDoesNotRaceWhenTabOrderNeedsRepair -count=1` reports a race at `desktop/tabs.go:1380` / `desktop/tabs.go:1395`.

  Please either keep `ListTabs` under the exclusive lock, or split the ordering helper so the `RLock` path is purely read-only and any order repair happens under `a.mu.Lock()`.

Non-blocking: the PR body says `maxReplaceRetries` was reduced to 3, but the actual diff leaves `internal/fileutil/atomicwrite.go` at 8. Please either update the description or include the intended change.

Verified while reviewing:
- `go test . -run Test(SaveTabsSkipsOlderSnapshot|ListTabsKeepsExplicitOrderWhenActiveChanges) -count=1` passed
- `go test -race . -run TestSaveTabsSkipsOlderSnapshot -count=1` passed
- `go test . -race -run TestListTabsKeepsExplicitOrderWhenActiveChanges -count=1` passed
- CI checks are currently green

SivanCola (approved):
Approved. I pushed `f2198b3d` to address the locking blocker I found earlier.

What changed:
- `ListTabs` keeps the normal path under `RLock`, but now uses a read-only order snapshot there.
- When the saved tab order is stale and needs repair, `ListTabs` reacquires `a.mu.Lock()` and repairs/rebuilds under the write lock.
- Added `TestListTabsRepairsStaleOrderWithoutRacing` to cover concurrent `ListTabs` calls with stale `tabOrder`.
- Updated the PR body to remove the stale `maxReplaceRetries` claim.

Verified:
- `go test . -run 'Test(ListTabsKeepsExplicitOrderWhenActiveChanges|ListTabsRepairsStaleOrderWithoutRacing|SaveTabsSkipsOlderSnapshot)' -count=1` passed
- `go test -race . -run TestListTabsRepairsStaleOrderWithoutRacing -count=1` passed
- `go test -race . -run TestSaveTabsSkipsOlderSnapshot -count=1` passed
- `go test . -count=1 -timeout=120s` passed
- `git diff --check` passed
- GitHub CI is green, including `race` and `test (windows-latest)`

SivanCola:
@codex review

PR #4367, opened 2026-06-14
Improve runtime metric token layout / 优化运行指标 token 布局

SivanCola:
## Summary
- Move elapsed time and request count into the compact runtime metrics row.
- Give session tokens an explicit full-width metric card so large token totals stay readable.
- Keep the layout intent durable with a reusable wide metric card class.

## Verification
- `pnpm --dir desktop/frontend exec tsx src/__tests__/context-panel-breakdown.test.ts`
- `git diff --check`
- Attempted `pnpm --dir desktop/frontend typecheck`; it is blocked in this local checkout by missing generated Wails bindings and an existing bridge type constraint.

Authorship note: @SivanCola contributed the layout decision and review direction.


SivanCola:
approve

PR #4377, opened 2026-06-14
perf: reduce layer paints and remove streaming cursor

CVEngineer66:
### Changes

1. **content-visibility: auto** on .msg / .tool / .turn-collapse / .process-card
   - Off-screen elements skip paint → layer texture memory drops from ~97 MB to only the visible viewport
   - Paint count no longer grows linearly with session length

2. **Remove streaming cursor**
   - Remove the blink-animated cursor injected by Markdown.tsx into streaming output
   - Each blink triggered a full document paint — root cause of frequent flicker and high layer memory

### Files changed
- desktop/frontend/src/styles.css — content-visibility rules + cursor/blink keyframe cleanup
- desktop/frontend/src/components/Markdown.tsx — removed cursor injection/removal functions, showCursor prop, useLayoutEffect
- desktop/frontend/src/components/Message.tsx — removed showCursor={item.streaming}

### DIFF
--- a/desktop/frontend/src/components/Markdown.tsx
+++ b/desktop/frontend/src/components/Markdown.tsx
@@ -1,4 +1,4 @@
-import { memo, useLayoutEffect, useMemo, useRef } from "react";
+import { memo, useMemo, useRef } from "react";
 import ReactMarkdown from "react-markdown";
 import type { Components } from "react-markdown";
 import remarkGfm from "remark-gfm";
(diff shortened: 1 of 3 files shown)


SivanCola (approved):
Reviewed the small frontend perf patch. I did not find blocking issues in the Markdown cursor removal or the new content-visibility rules.

Local validation:
- npm run check:css passed
- npm run test passed
- npm run typecheck / npm run build are blocked locally by missing generated Wails modules (src/lib/bridge.ts, src/lib/theme.ts, src/lib/windowState.ts), not by this diff

GitHub checks are green.

Issue #4390, opened 2026-06-14
[Bug]:  为什么每次打开都要重新配置api key？


T1-ong:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.7.0 windows最新版，从之前三四个版本开始就这样

### What happened?

<img width="525" height="215" alt="Image" src="https://github.com/user-attachments/assets/98135fc4-c84a-4449-bdd9-3f86ca55d170" /> 每次都要重新配置api

### Steps to reproduce

1

### OS / platform

windows10

### Relevant logs or output

```shell

```

esengine:
Fixed: API keys are now persisted in the OS credential store — the config layer has a keyring-backed credentials store (Windows Credential Manager / macOS Keychain / Secret Service on Linux), so a key entered once survives restarts without living in a plaintext env var you have to re-export.

Please retest on the latest release; if keys still don't stick there, open a fresh issue with how you originally entered the key.

PR #4401, opened 2026-06-14
feat: add diagnostic logging for credential persistence

myipanta:
Add slog logging at critical points in the credential save/load path to help diagnose #4390 (API key lost on Windows restart):

- config.loadDotEnvFile: log loaded path + key count
- desktop.upsertEnvFile: log write failures
- desktop.promoteProviderKeysToCredentials: log migration activity

### DIFF
--- a/desktop/dotenv.go
+++ b/desktop/dotenv.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"log/slog"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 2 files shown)


esengine:
Thanks for this. Since it was opened, credential persistence moved to an OS keyring-backed store (Windows Credential Manager / Keychain / Secret Service), which resolved the underlying #4390 — so the diagnostic logging targeting the old dotenv promotion path no longer maps to the current code.

Closing as superseded by the keyring work. If you see credential loss on a current release, a fresh issue with repro details would be very welcome.

PR #4482, opened 2026-06-15
[codex] Fix short-window layout overflow

jiaxindeyang-a11y:
## Summary
- Let the sidebar and chat pane scroll vertically when the desktop window is shorter than the fixed chrome/footer stack.
- Change the status bar from clipping overflow to horizontal scrolling so right-side metrics remain reachable in windowed layouts.
- Update the typography overflow contract test to lock in the new behavior.

## Root Cause
The desktop shell used fixed-height chrome/footer regions inside a `body`/layout that hides overflow. At very short heights, content outside the available column height was clipped. The status bar also forced a single line while using `overflow: hidden`, so metrics at the right edge disappeared when the window was not wide enough.

## Validation
- `npm.cmd run check:css`
- `npx.cmd tsx src/__tests__/typography-overflow-contract.test.ts`
- `npm.cmd test`
- `npm.cmd run test:typecheck` was attempted but still fails locally because `desktop/frontend/wailsjs` is not generated in this checkout (`Cannot find module '../../wailsjs/go/main/App'`).
- Vite served `http://127.0.0.1:5174/` with HTTP 200, but screenshot-level Browser verification was blocked by the local Windows sandbox (`CreateProcessAsUserW failed: 5`).

### DIFF
--- a/desktop/frontend/src/__tests__/typography-overflow-contract.test.ts
+++ b/desktop/frontend/src/__tests__/typography-overflow-contract.test.ts
@@ -100,7 +100,10 @@ eq(
 );
 
 eq(finalDeclaration(".statusbar", "white-space"), "nowrap", "status bar keeps metrics on one row");
-eq(finalDeclaration(".statusbar", "overflow"), "hidden", "status bar clips instead of overflowing");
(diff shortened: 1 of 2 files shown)


jiaxindeyang-a11y:
Hi, could you please take a look at this PR when you have time?
CI is passing and the change is limited to the short-window overflow behavior.
Thanks!

PR #4485, opened 2026-06-15
[codex] Show status bar amounts in RMB

jiaxindeyang-a11y:
## Summary
- Format the desktop status bar's turn cost and session cost with the RMB symbol.
- Normalize the status bar balance display to the same RMB prefix so balance and spend readouts match.
- Format the right workbench/context panel cost cards and cost breakdown rows with the same RMB display.
- Keep the shared `formatMoney` behavior unchanged for places that intentionally need source-currency formatting.

## Root Cause
The desktop UI mixed display paths: status bar cost readouts used the session currency from usage events, wallet balance used the backend-provided display string, and the right workbench context panel still formatted its session cost with the original panel currency. That allowed spend to show as `$...` while balance showed as `¥...`.

## Validation
- `npx.cmd tsx src/__tests__/context-panel-breakdown.test.ts`
- `npm.cmd test`
- `npm.cmd run test:typecheck` was attempted earlier on this branch but still fails locally because `desktop/frontend/wailsjs` is not generated in this checkout (`Cannot find module '../../wailsjs/go/main/App'`).

### DIFF
--- a/desktop/frontend/src/App.tsx
+++ b/desktop/frontend/src/App.tsx
@@ -3001,7 +3001,6 @@ export default function App() {
               turnTokens={state.turnTotalTokens}
               turnCost={state.turnCost}
               cost={state.sessionCost}
-              currency={state.sessionCurrency}
               modelLabel={state.meta?.label}
               labelStyle={statusBarStyle}
               items={statusBarItems}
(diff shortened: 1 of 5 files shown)


SivanCola:
Thanks @jiaxindeyang-a11y. We are closing this PR as superseded by #4329, which is now the canonical integration route for the cost/currency/status-bar work.

Your contribution is recorded in #4329: you normalized the desktop status bar, wallet balance, ContextPanel, and cost-breakdown display around RMB so spend and balance readouts stay consistent.

You are acknowledged in #4329 as a repository co-contributor for this work.

PR #3986, opened 2026-06-11
fix(desktop): remove duplicate '+' button in TabBar

NkAntony777:
## Summary
- Remove the redundant '+' (new session) button from the TabBar component
- The sidebar already has a prominent 'New Session' button, making the TabBar's small '+' button unnecessary
- This simplifies the UI and eliminates user confusion about duplicate functionality

## Changes
- **TabBar.tsx**: Removed the '+' button, its Tooltip wrapper, and the onNewTab prop
- **App.tsx**: Removed the onNewTab prop from TabBar usage
- **styles.css**: Removed the .tabbar__new and .tabbar__new:hover CSS rules
- **en.ts** / **zh.ts**: Cleaned up the unused tabBar.newSession i18n key

## Related Issue
Fixes #3933

## Testing
- TypeScript type check passes (npm run typecheck)
- No new type errors introduced
- The sidebar's 'New Session' button remains functional for creating new sessions

PR #4493, opened 2026-06-15
Keep new tab button beside tabs / 固定新建会话按钮贴近标签页

### LINKED ISSUE #4483: [Feature]: 新建会话和搜索框会随着右边侧栏内容打开与合并进而发生位置变化，这个能固定住吗
### Version line

Image
Image

### Proposed solution

non

SivanCola:
## Summary
- Keep the themed desktop tab list sized to its tab content before it shrinks.
- Preserve a stable new-session button slot immediately after the last visible tab.
- Add AppChrome CSS contract coverage so the themed tab list cannot grow into empty chrome space again.

## Root cause
The themed AppChrome tab list used `flex: 1 1 auto`, so it expanded across the remaining titlebar space and pushed the `+` new-session button toward the right chrome controls when the right dock changed the available width.

## Verification
- `npm run check:css`
- `tsx src/__tests__/app-chrome-tabs.test.ts`
- Browser DOM check confirmed the tab list no longer flex-grows and the `+` button sits beside the last tab.

## Related
- Fixes #4483
- Related to #3986, which proposes removing the top `+` button entirely. This PR keeps the button and fixes its placement for the current UI direction.

### DIFF
--- a/desktop/frontend/src/__tests__/app-chrome-tabs.test.ts
+++ b/desktop/frontend/src/__tests__/app-chrome-tabs.test.ts
@@ -79,6 +79,16 @@ ok(
 );
 
 ok(
+  finalDeclaration(":root[data-theme-style] .app-chrome--tabs .tabbar__tabs", "flex") === "0 1 auto",
+  "themed AppChrome tab lists size to tab content before shrinking",
+);
+
+ok(
+  finalDeclaration(":root[data-theme-style] .app-chrome--tabs .tabbar__tabs", "width") === "max-content",
+  "themed AppChrome tab lists keep the new-tab button next to the last tab",
+);
+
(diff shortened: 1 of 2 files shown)


SivanCola:
Supersession note: #3986 explored the alternate product direction of removing the duplicate top `+` new-session button. For the current UI direction, this PR keeps the top `+` affordance and fixes its placement instead.

Thank you @NkAntony777 for the work in #3986: it clarified the duplicate-entry concern and helped frame the final product decision here.

Authorship note: @SivanCola is the primary author of this PR. @NkAntony777 contributed as a repository collaborator through #3986; that contribution is acknowledged as part of the shared repository work behind this final tab/new-session direction.

Issue #4464, opened 2026-06-15
[Bug]: agent模式更改的注入提示词泄漏到聊天框内


why37281:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.8.1

### What happened?

#3720 的问题，之前修过了，并且修好了，咋又冒出来了
不再描述情况，看那个issue

### Steps to reproduce

。

### OS / platform

 。

### Relevant logs or output

```shell

```

SivanCola:
Fixed by #5892, merged in 53086911.

Root cause: the goal/collaboration mode `<active-goal>` context was injected as transient user-turn context, but the display/preview cleanup path did not strip that block, so it could leak into visible chat/history text.

Fix: `main-v2` now strips `<active-goal>` alongside the other transient user blocks, while preserving the existing Memory v5 contract unwrapping and newer transient blocks such as `capability-route`.

Verified:
- `go test ./internal/agent ./internal/control -count=1`
- `cd desktop && go test . -run 'TestHistoryMessages' -count=1`

PR #4513, opened 2026-06-15
fix(agent): add active-goal to transient user block regex (#4464)

myipanta:
The StripTransientUserBlocks regex stripped reasoning-language, memory-update, and background-jobs XML blocks from persisted user messages before display, but missed active-goal blocks injected by goal/collaboration mode.  This caused the goal-mode injection prompt to leak into the chat transcript as visible user-bubble text.

Add active-goal to both alternation groups in the regex so the block is stripped alongside the other three transient blocks.

### DIFF
--- a/internal/agent/preview.go
+++ b/internal/agent/preview.go
@@ -5,7 +5,7 @@ import (
 	"strings"
 )
 
-var reTransientUserBlock = regexp.MustCompile(`(?s)^\s*<(?:reasoning-language|memory-update|background-jobs)>.*?</(?:reasoning-language|memory-update|background-jobs)>\s*\n?`)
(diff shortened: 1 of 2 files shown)


SivanCola:
Thanks @myipanta. This was the right root cause: `<active-goal>` was missing from transient user block stripping, so goal/collaboration context could leak into visible history/chat text.

I adapted the fix onto the current `main-v2` in #5892 because `preview.go` has since gained response-language, hook-context, capability-route, and Memory v5 contract unwrapping. #5892 is now merged in 53086911. It preserves the active-goal stripping, adds coverage for stacked transient prefixes and Desktop history-visible user content, and credits this PR in the PR body plus the co-author trailer in commit 91035ddd.

Closing this PR as superseded by #5892.

PR #4543, opened 2026-06-15
fix(desktop): inline-math classifier gaps + display-math/blockquote rendering bugs

lightfront:
# Fix: inline-math classifier gaps + display-math/blockquote rendering bugs

## Overview

This PR fixes **ten** rendering bugs in the math pipeline — eight in the inline-math classifier (`mathClassify.ts`) and two in the display-math pre-pass (`mathNormalize.ts`). All ten produce visible KaTeX failures (red error blocks or raw `$...$` text / literal `$` signs) in LLM chat output.

Each classifier gap has the same root cause: `isLikelyInlineMath()` returns `false` for a legitimate math token, so `normalizeMath` rewrites `$token$` → `&#36;token&#36;` (literal dollar HTML entities) and the token never reaches KaTeX.

---

### 1. LaTeX command immediately followed by a digit — `$\tfrac12$`, `$\sqrt2$`, `$\log3$`

```js
// Before (bug): \b is a word boundary, but \tfrac12 has no boundary
// between "tfrac" and "12" — both are word characters.
if (/\\[A-Za-z]+\b/.test(math)) return true;
// After (fix):
if (/\\[A-Za-z]+/.test(math)) return true;
```

### 2. Multi-letter group notation — `$SO(3,1)$`, `$SU(2)$`, `$SL(2)$`, `$GL(n)$`

```js
// Before (bug): only single-letter names accepted (f(x), g(x))
if (/^[A-Za-z]\s*\([^)]{1,80}\)$/.test(math)) return true;
// After (fix): 1–6 letter identifiers
if (/^[A-Za-z]{1,6}\s*\([^)]{1,80}\)$/.test(math)) return true;
```

### 3. Permutation cycle notation — `$(12)$`, `$(123)$`, `$(12)(34)$`

```js
// New rule: one or more parenthesised all-digit groups
if (/^(?:\(\d+\))+$/.test(math)) return true;
```

### 4. Pure numbers rejected — `$0$`, `$1$`, `$42$`

The classifier rejected all pure numbers (`return false`) to avoid `$5` currency false-positives. But `$0$` (mass eigenvalue), `$1$`, `$42$` are extremely common in math/physics and rendered as literal dollar signs.

```js
// Before (bug): rejects ALL pure numbers
if (/^\d+(?:\.\d+)?%?$/.test(math)) return false;
// After (fix): currency is almost always written WITHOUT a closing $
// ("costs $5", not "costs $5$"), so the $N$ form almost always means math.
if (/^\d+(?:\.\d+)?%?$/.test(math)) return true;
```

Unpaired currency (`$5` with no closing `$`) is still correctly treated as literal text by the `$...$` pairing logic, so "costs $5 and $6" is unaffected.

### 5. Binary operator with signed RHS — `$K = -iJ$`, `$p = +\alpha$`

The binary-operator regex required an operand character (`[A-Za-z0-9...]`) immediately after the operator, but `-` was not in that class — so `= -` (operator followed by a unary sign) failed to match. This rejected common physics expressions like `$K = -iJ$`, `$a = -b$`, `$p = +\alpha$`.

```js
// Before (bug): K = -iJ fails (no operand right after =)
/[A-Za-z0-9)\]}]\s*[+\-*/=<>]\s*[A-Za-z0-9([{\\]/
// After (fix):  allows an optional sign before the RHS operand
/[A-Za-z0-9)\]}]\s*[+\-*/=<>]\s*[+\-]?\s*[A-Za-z0-9([{\\]/
```

### 6. Lone operators — `$+$`, `$-$`, `$=$`, `$<$`

Single-character operator tokens (`+`, `-`, `=`, `<`, `>`, `±`, `∓`) matched no accept rule and were rendered as literal text. These are common in physics prose ("the sign of `$+$`", "the `$<$` relation").

```js
if (/^[+\-=<>±∓]$/.test(math)) return true;
```

### 7. Primed letters — `$S'$`, `$x'$`, `$y''$`, `$f'(x)$`

A letter or Greek command followed by one or more primes matched no rule and fell through to the final `/^[A-Za-z]$/` check (which requires exactly one letter), so primed tokens were escaped to literal dollars. The prime (`'`) is the LaTeX prime symbol — ubiquitous for derivatives, transformed quantities, and labelled states (e.g. the **S′** mixed-symmetry baryon in SU(6) physics).

```js
// New rules:
/^[A-Za-z]'{1,}\s*\([^)]{1,80}\)$/          // primed function call: f'(x), g''(t)
/^(?:[A-Za-z]|\\[A-Za-z]+)'{1,}$/           // letter/Greek + primes: x', S', y'', \psi'
```

### 8. Bracketed irrep labels — `$[56]$`, `$[8]$`, `$[56,0^+]$`

Square brackets in a `$...$` context denote group-theory irrep labels (`[56]` is the SU(6) multiplet), array subscripts, and interval notation. Pure bracketed numbers/letters matched no rule and were escaped. Markdown-link syntax `[text](url)` is already rejected upstream by the `](` guard (top of the function), so a bare `[content]` token is unambiguous.

```js
// New rule: [56], [8], [N], [\mathbf{56}], [56,0^+]
/^\[[A-Za-z0-9^_+\-,.\\\s{}]+\]$/.test(math)
```

---

### 9. Repair-before-extraction ordering — closing `$$` split off

The repair regex (which inserts `\n\n` before glued `$$`) ran **before** the display-pair extraction. So it split the closing `$$` of well-formed pairs like `\end{pmatrix}$$` or `> $$x$$`, breaking display math.

**Fix:** extract `$$…$$` pairs as a unit **first**, then repair only remaining orphaned `$$`. This is the `d450aec1` ordering fix applied to current main-v2.

### 10. Blockquote display-math fence — `remark-math` upstream limitation

`remark-math` has a known limitation: multi-line `$$...$$` display math inside a Markdown blockquote (`>`) breaks inline math parsing on subsequent blockquote lines. The closing fence isn't recognised and inline `$...$` after the display block gets swallowed.

**Fix:** when display math is detected inside a blockquote, `normalizeMath` closes the blockquote before the display math (`\n\n$$\n...\n$$\n\n> `) and reopens it after. This puts the math outside the quote, avoiding the remark-math fence bug entirely while preserving the visual blockquote structure.

---

## Verification

| Check | Result |
|---|---|
| math-golden suite | **242/242** (was 189, +53 regression tests across all fixes) |
| Typecheck | 0 errors in touched files (`bridge.ts` error is pre-existing on base) |
| `$\tfrac12$`, `$\sqrt2$`, `$\log3$` | ✅ now renders |
| `$SO(3,1)$`, `$SU(2)$`, `$SL(2)$`, `$GL(n)$` | ✅ now renders |
| `$(12)$`, `$(123)$`, `$(12)(34)$` | ✅ now renders |
| `$0$`, `$1$`, `$42$` | ✅ now renders |
| `$K = -iJ$`, `$p = +\alpha$` | ✅ now renders |
| `$+$`, `$-$`, `$=$`, `$<$` | ✅ now renders |
| `$S'$`, `$x'$`, `$y''$`, `$f'(x)$`, `$\psi'$` | ✅ now renders |
| `$[56]$`, `$[8]$`, `$[70]$`, `$[56,0^+]$`, `$[\mathbf{56}]$` | ✅ now renders |
| `the apples cost $5 and $6.` (unpaired currency) | ✅ stays literal (no false-positive) |
| markdown link `[text](url)` | ✅ still a link, not math |
| Haar theorem blockquote (display + inline) | ✅ 0 errors |
| Mackey theorem blockquote (display + inline after) | ✅ 0 errors |
| `$$\end{pmatrix}$$` (closing brace) | ✅ closing `$$` preserved |
| `$PATH$`, `costs $5` (unpaired) | ✅ still literal |

## Files changed

| File | Change |
|---|---|
| `desktop/frontend/src/components/mathClassify.ts` | Drop `\b`; broaden function-call rule to 1–6 letters; add cycle-notation, primed-letter, primed-function-call, and bracketed-label rules; accept pure numbers; signed RHS in binary-op rule; lone operators |
| `desktop/frontend/src/components/mathNormalize.ts` | Extract `$$…$$` before repair regex; newline-delimited DM restore; blockquote-aware closing (break quote around display math) |
| `desktop/frontend/src/__tests__/math-golden.test.ts` | +53 regression tests across all 10 fixes |

## Commits

- `792df63c` — multi-letter group notation (`SO(3,1)`, `SU(2)`, …) classified as inline math
- `6c64bc91` — classify LaTeX commands followed by a digit (`\tfrac12`, `\sqrt2`) as math
- `e7c0254f` — render permutation cycle notation like `(12)` as inline math
- `2b050155` — classify pure numbers (`$0$`, `$1$`, `$42$`) as inline math
- `7ceeebfa` — classify operators with signed RHS (`$K = -iJ$`) and lone operators (`$+$`)
- `6f7ef086` — repair display-math extraction ordering + blockquote `$$` closing
- `84e6ac81` — break blockquote around display math to avoid remark-math fence bug
- `d29b963b` — classify primed letters (`$S'$`) and bracketed irrep labels (`$[56]$`) as math

### DIFF
--- a/desktop/frontend/src/__tests__/math-golden.test.ts
+++ b/desktop/frontend/src/__tests__/math-golden.test.ts
@@ -133,6 +133,13 @@ check("$\\frac{a}{b}$", () => isLikelyInlineMath("\\frac{a}{b}") === true);
 check("$f(x)$", () => isLikelyInlineMath("f(x)") === true);
 check("$x+1$", () => isLikelyInlineMath("x+1") === true);
 
+console.log("\nisLikelyInlineMath — LaTeX command followed by a digit (math)");
+check("$\\tfrac12$", () => isLikelyInlineMath("\\tfrac12") === true);
+check("$\\frac12$", () => isLikelyInlineMath("\\frac12") === true);
+check("$\\sqrt2$", () => isLikelyInlineMath("\\sqrt2") === true);
+check("$\\log3$", () => isLikelyInlineMath("\\log3") === true);
+check("$\\overline3$", () => isLikelyInlineMath("\\overline3") === true);
+
 console.log("\nisLikelyInlineMath — multi-letter group notation (math)");
 check("$SO(3,1)$", () => isLikelyInlineMath("SO(3,1)") === true);
 check("$SU(2)$", () => isLikelyInlineMath("SU(2)") === true);
--- a/desktop/frontend/src/__tests__/math-golden.test.ts
+++ b/desktop/frontend/src/__tests__/math-golden.test.ts
@@ -241,7 +241,7 @@ check("\\|x\\| renders as double bars", () => {
 
 console.log("\nnormalizeMath — LLM delimiter conversion");
 eq(normalizeMath("\\(x^2\\)"), "$x^2$", "\\(…\\) → $…$");
-eq(normalizeMath("\\[E=mc^2\\]"), "$$E=mc^2$$", "\\[…\\] → $$…$$");
+eq(normalizeMath("\\[E=mc^2\\]"), "$$\nE=mc^2\n$$", "\\[…\\] → $$…$$");
 eq(normalizeMath("\\\\[4pt]"), "\\\\[4pt]", "\\\\[ line-break spacing protected");
 
 console.log("\nnormalizeMath — \\slashed conversion (regression)");
@@ -266,12 +266,12 @@ check("inline $$ after closing bracket", () => {
   return out.startsWith("(octet)\n\n$$");
 });
 check("inline $$ after closing brace (\\end{...}$$)", () => {
-  // A model that writes `\end{array}$$` or `\frac{a}{b}$$` on one line
-  // has the same micromark-fence problem as the comma case. The
-  // closing brace is the most common end-of-content marker in LaTeX
-  // math, so the repair-regex character class includes it.
+  // A display equation ending with }$$ must be extracted as a unit.
+  // The closing $$ must NOT be split off (the old bug inserted \n\n
+  // before it, emptying the equation). The extraction regex catches
+  // $$...$$ as a pair before the repair regex can split it.
   const out = normalizeMath("$$\\begin{pmatrix}a&b\\\\c&d\\end{pmatrix}$$");
-  return out.includes("\\end{pmatrix},\n\n$$") || out.includes("\\end{pmatrix}\n\n$$");
+  return out.includes("$$") && out.includes("\\begin{pmatrix}") && !out.includes("}\\n\\n$$") && !out.includes("$$\\n\\n");
 });
 check("inline $$ after comma on same line as content", () => {
   // User-reported (2026-06-12, soft-pion chat): the model wrote the
@@ -289,20 +289,22 @@ check("inline $$ after comma on same line as content", () => {
 });
 check("well-formed $$ already on own line is normalised consistently", () => {
   // Whether the model writes `decomposes as$$\n\mathbf{6}.$$` or
-  // `decomposes as\n\n$$\n\mathbf{6}.$$`, both must produce the same
-  // remark-math-parseable form: opening $$ on its own line, body, blank
-  // line, closing $$ on its own line.
+  // `decomposes as\n\n$$\n\mathbf{6}.$$`, the opening $$ must be
+  // paragraph-separated (blank line before it). The display pair is
+  // extracted as a unit, so the closing $$ stays attached to the body.
   const inline = normalizeMath("decomposes as$$\n\\mathbf{6}.$$");
   const block = normalizeMath("decomposes as\n\n$$\n\\mathbf{6}.$$");
-  const expected = "decomposes as\n\n$$\n\\mathbf{6}.\n\n$$";
-  return inline === expected && block === expected;
+  // Both must have the opening $$ on its own line (blank line before it)
+  return inline.startsWith("decomposes as\n\n$$") && block.startsWith("decomposes as\n\n$$")
+    // And the math body must be present (not emptied)
+    && inline.includes("\\mathbf{6}") && block.includes("\\mathbf{6}");
 });
 check("\\[…\\] → $$…$$ still works (no spurious blank line)", () => {
-  return normalizeMath("\\[E=mc^2\\]") === "$$E=mc^2$$";
+  return normalizeMath("\\[E=mc^2\\]") === "$$\nE=mc^2\n$$";
 });
 check("digit before $$ is NOT a prose boundary (preserves c^2$$)", () => {
   const out = normalizeMath("c^2$$ x $$");
-  return out === "c^2$$ x $$";
+  return out === "c^2$$\n x \n$$";
 });
 
 console.log("\nnormalizeMath — non-math dollar filtering");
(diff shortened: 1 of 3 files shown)


lightfront:
## Added: fix for display math inside blockquotes + closing-delimiter splitting

While testing, a user reported that a blockquote (`>`) containing `$$...$$` display math caused a cascading KaTeX failure that rendered the **entire following paragraph** as broken red math. Investigation found two related bugs in `mathNormalize.ts`:

### 1. Repair-before-extraction ordering
The repair regex (which inserts `\n\n` before glued `$$`) ran *before* the display-pair extraction. So it split the closing `$$` of well-formed pairs like `\end{pmatrix}$$` or `> $$x$$`, breaking display math. 

**Fix:** extract `$$…$$` pairs as a unit *first*, then repair only the remaining orphaned `$$`. This is the `d450aec1` ordering fix applied to current main-v2.

### 2. Blockquote display-math fence
When `$$...$$` appears inside a Markdown blockquote (`>`), `remark-math` requires the closing `$$` to also carry the `>` prefix. Without it, the closing fence isn't recognised and remark-math swallows the following paragraph as math content.

**Fix:** the DM-restore step now detects whether the opening `$$` was inside a blockquote and prefixes the closing `$$` with `>` accordingly.

### Test status
- **207/207 pass** (was 203, +4 updated expectations for the corrected display-math behavior)
- Typecheck: 0 errors
- The full "locally compact / Haar's theorem" text that triggered the bug now renders with 0 errors

myriad2:
同样受到 display-math 渲染 bug 影响：行内公式 $...$ 正常显示，但行间公式 $$...$$ 完全不显示。

经排查确认是修复 #9（repair‑before‑extraction ordering）所描述的问题 — Step 3 的修复正则错误地在闭合 $$ 前也插入了 \n\n，导致 $$...$$ 对无法正确配对，整个 display‑math 块丢失。

希望这个 PR 能尽快合并，display‑math 的缺失对数学/物理方向的用户影响很大。感谢贡献！

SivanCola:
Maintainer update: #5649 now carries the narrow display-math parser fix, including the follow-up commit for orphan opening `$$` and multiple display blocks on the same physical line.

I am not closing this PR because it still contains additional classifier and blockquote-related work that #5649 does not cover. To keep the review path clean, please rebase this PR after #5649 lands and drop the overlapping `mathNormalize.ts` display-normalization changes, so the remaining classifier/blockquote behavior can be reviewed independently.

SivanCola:
Maintainer update: I opened #5662 as the integration PR for this PR and #5649.

#5662 keeps/adapts the explicit classifier fixes from this PR and the blockquote display-math concern, while using the line-by-line display parser from #5649 as the base. One part was intentionally not adopted: treating pure numeric `$1$`, `$42$`, and `$100%$` as math, because that regresses the existing currency/prose-dollar guard. The new PR body credits this source PR and calls out that decision explicitly.

I am leaving this PR open for traceability until the integration PR lands.

SivanCola:
Closing as superseded by #5662, which has already landed on `main-v2`.

What was kept/adapted from this PR:
- The explicit inline-math classifier fixes for LaTeX commands followed by digits, group/function notation, permutation cycles, signed RHS operands, lone operators, primed symbols/functions, and bracketed irrep labels.
- The blockquote display-math regression concern, adapted onto the line-by-line display parser used by #5662.

What was not adopted:
- Treating pure numeric `$1$`, `$42$`, and `$100%$` as inline math, because that regresses the existing currency/prose-dollar guard.

Follow-up #5884 records commit-level GitHub attribution for the source PR contributors, including `lightfront`, so the contribution is reflected through `Co-authored-by:` trailers rather than PR-body credit only.

Thank you again for the thorough bug report, implementation, and regression coverage.

PR #4547, opened 2026-06-15
fix(plugin): add per-call timeout to stdio MCP transport (#4299)

### LINKED ISSUE #4299: Bug: MCP stdio transport hangs indefinitely when codegraph server responds slowly
# Bug Report: MCP stdio transport hangs indefinitely when codegraph server responds slowly

## Description

When using `reasonix run` with codegraph MCP enabled, the agent sometimes hangs indefinitely with no output (4+ minutes). The process becomes completely unresponsive and must be killed.

## Steps to Reproduce

1. Enable codegraph in `reasonix.toml`:
   ```toml
   [codegraph]
   enabled      = true
   auto_install = true
   ```

2. Run a task that triggers codegraph tools:
   ```bash
   reasonix run "用 codegraph 搜索 main 函数" --max-steps 3
   ```

3. Observe that the process hangs with no output (total_lines: 0)

4. Kill the process and retry with `--no-tools`:
   ```bash
   reasonix run "搜索 main 函数" --no-tools --max-steps 3
   ```
   This completes successfully in ~7 seconds.

## Expected Behavior

The MCP client should have a timeout mechanism. If the codegraph server doesn't respond within a reasonable time (e.g., 30-60 seconds), the client should:
1. Timeout the call
2. Return an error to the agent
3. Allow the agent to continue with other tools

## Actual Behavior

The MCP client hangs indefinitely waiting for a response from the codegraph server. The process produces no output and must be killed manually.

## Root Cause Analysis

Based on deep code review of `internal/plugin/transport_stdio.go`:

### 1. `stdioTransport.call()` has no timeout

```go
func (t *stdioTransport) call(ctx context.Context, method string, params any) (json.RawMessage, error) {
    // ... acquires callMu lock ...
    // ... writes request to stdin ...
    // ... waits for response via ch ...
    // ⚠️ NO TIMEOUT - waits indefinitely
    select {
    case resp := <-ch:
        // ...
    case <-ctx.Done():
        // ...
    }
}
```

The `ctx` passed in is typically `context.Background()` without any timeout, so the select blocks forever.

### 2. `callMu` serializes all MCP calls

```go
t.callMu.Lock()
defer t.callMu.Unlock()
```

All MCP calls must acquire this mutex. If one call hangs, all subsequent calls are blocked.

### 3. `readLoop` is not interruptible

```go
func (t *stdioTransport) readLoop() {
    scanner := bufio.NewScanner(t.stdout)
    for scanner.Scan() {
        // ...
    }
}
```

If the child process doesn't exit, `readLoop` blocks forever on `scanner.Scan()`.

### 4. `executeBatch` delays output

```go
func (a *Agent) executeBatch(ctx context.Context, calls []provider.ToolCall) []string {
    results := make([]string, len(calls))
    for i := range calls {
        results[i] = executeOne(...)
    }
    return results  // All tools must complete before returning
}
```

Users see no progress until all tools complete.

## Environment

- **OS**: Windows 10 (amd64)
- **Go**: 1.26.4
- **Reasonix**: v1.6.0 (commit 62645d1c)
- **CodeGraph**: v1.0.0
- **Database**: 42MB SQLite (779 files, 14K nodes)

### Priority 0: Add timeout to `stdioTransport.call()`

```go
func (t *stdioTransport) call(ctx context.Context, method string, params any) (json.RawMessage, error) {
    t.callMu.Lock()
    defer t.callMu.Unlock()

    // Add timeout if not already set
    if _, ok := ctx.Deadline(); !ok {
        var cancel context.CancelFunc
        ctx, cancel = context.WithTimeout(ctx, 30*time.Second)
        defer cancel()
    }

    // ... existing code ...

    select {
    case resp := <-ch:
        // ...
    case <-ctx.Done():
        // Close stdin to interrupt readLoop
        t.stdin.Close()
        return nil, fmt.Errorf("MCP call %q timed out: %w", method, ctx.Err())
    }
}
```

### Priority 0: Make `readLoop` interruptible

```go
func (t *stdioTransport) readLoop() {
    scanner := bufio.NewScanner(t.stdout)
    for scanner.Scan() {
        // ...
    }
    // When stdin is closed, scanner.Scan() returns false
    // and readLoop exits gracefully
}
```

### Priority 1: Emit ToolResult progressively

```go
func (a *Agent) executeBatch(ctx context.Context, calls []provider.ToolCall) []string {
    results := make([]string, len(calls))
    for i := range calls {
        results[i] = executeOne(...)
        // Emit result immediately
        a.sink.Emit(event.Event{Kind: event.ToolResult, ...})
    }
    return results
}
```

## Workaround

Use `--no-tools` flag to skip all MCP initialization:

```bash
reasonix run "任务描述" --no-tools --max-steps 3
```

This bypasses the codegraph MCP entirely and completes in ~7 seconds instead of hanging.

## Additional Context

- The codegraph MCP server itself starts quickly (0.3 seconds)
- The database is healthy (42MB, 779 files, 14K nodes)
- The issue is specifically in Reasonix's MCP client, not the codegraph server
- The codegraph server has proper timeout mechanisms (5s busy_timeout, 6s daemon connect timeout)

ashishexee:
## Problem

#4299 reports that `reasonix run` with codegraph MCP enabled hangs indefinitely (4+ minutes) when the codegraph server responds slowly. The process becomes completely unresponsive and must be killed.

## Root Cause

`stdioTransport.call()` in `internal/plugin/transport_stdio.go` has no timeout. The agent's turn context is only cancelled by explicit user action (Ctrl+C) — it carries no deadline. The `select` in `call()` parks on `<-ctx.Done()` which never fires, so a slow or hung MCP server blocks the agent's turn forever. Additionally, `callMu` serializes all calls to the same server, so one hung call blocks every subsequent tool from that server.

## Fix

Add a per-call timeout (`defaultCallTimeout = 60s`) applied via `context.WithTimeout` when the caller's context carries no deadline. If the server doesn't respond within the timeout, the call returns `context.DeadlineExceeded` and the agent can continue with other tools.

- Single timeout, no retry — matches the industry pattern (Claude Code, the official Python and TypeScript MCP SDKs all treat timeout as a hard failure; retries are unsafe for side-effecting `tools/call` and pointless when the server is truly hung)
- Caller's existing deadline is respected and not overridden
- `slog.Warn` fires on timeout (not on user cancellation) with server name, method, and timeout duration for debugging
- `callTimeout` field on `stdioTransport` allows per-transport configuration (tests use it; config wiring is a follow-up)
- `callMu` serialization is kept — removing it would allow interleaved JSON-RPC on the shared stdin pipe, which is correct in theory (the `pending` map already demuxes by id) but risky without upstream MCP conformance tests. The timeout cap (infinity to 60s) is the critical fix.

## Verification

```
go test ./internal/plugin/ -run "TestStdio" -v -count=1 -race
go vet ./internal/plugin/...
gofmt -l internal/plugin/transport_stdio.go internal/plugin/stdio_cancel_test.go
```

All 4 tests pass with race detector:
- `TestStdioCallReturnsOnContextCancel` — cancel unblocks a hung call
- `TestStdioCallTimesOutWithoutDeadline` — 60s timeout fires when ctx has no deadline
- `TestStdioCallRespectsExistingDeadline` — caller's deadline takes precedence
- `TestStdioCallCancelOverridesTimeout` — cancel during timeout returns `Canceled`, not `DeadlineExceeded`

Fixes #4299

### DIFF
--- a/internal/plugin/stdio_cancel_test.go
+++ b/internal/plugin/stdio_cancel_test.go
@@ -2,6 +2,7 @@ package plugin
 
 import (
 	"context"
+	"strings"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


SivanCola:
Thanks @ashishexee. I extracted the still-valid stdio MCP per-call timeout behavior from this PR onto the latest main-v2 in #4628, together with the related reconnect fixes. Closing this PR to keep the integration path focused on #4628.

PR #4616, opened 2026-06-16
Fix stale planner model startup / 修复陈旧规划模型导致启动失败

### LINKED ISSUE #4615: [Bug]: 强制关机之后，重启应用就无法识别模型了
### Version line

### Exact version

v1.8.1

### What happened?

Image

### Steps to reproduce

强制关机之后，重启应用就无法识别模型了

### OS / platform

win11

### Relevant logs or output

```shell

```

GTC2080:
## Problem

#4615 reports that after a forced shutdown, the desktop app can restart into a startup error like:

`planner_model "OpenRouter/moonshotai/kimi-k2.6:free" is not a configured provider`

That leaves the primary chat surface unusable even when the executor model is still configured.

## Root Cause

Two startup paths were too strict for recovered desktop configuration state:

- Model references matched the provider segment with exact casing only, so a recovered `OpenRouter/...` ref could miss a configured `openrouter` provider even though the model id itself was valid.
- `agent.planner_model` is optional, but boot treated a stale planner reference as fatal after the executor model had already resolved successfully.

## Fix

- Resolve the provider segment of `provider/model` refs with exact-match priority and a unique case-insensitive fallback, while keeping model ids case-sensitive.
- Treat an unresolved optional `planner_model` as a warning and continue with the executor model instead of blocking startup.
- Add regression coverage for mixed-case OpenRouter refs and stale optional planner refs during boot.

## Verification

- `go test ./internal/config -run TestResolveModelAcceptsProviderNameCaseInsensitiveRef -count=1`
- `go test ./internal/boot -run TestBuildContinuesWhenOptionalPlannerModelIsStale -count=1`
- `go test ./internal/config ./internal/boot`
- `go test ./...`
- `go vet ./...`
- `git diff --check`
- `gofmt -l .`

Fixes #4615

### DIFF
--- a/internal/boot/boot.go
+++ b/internal/boot/boot.go
@@ -751,9 +751,8 @@ func Build(ctx context.Context, opts Options) (*control.Controller, error) {
 	if pm := cfg.Agent.PlannerModel; pm != "" && !tokenEconomy {
 		pe, ok := cfg.ResolveModel(pm)
 		if !ok {
-			return nil, fmt.Errorf("planner_model %q is not a configured provider", pm)
-		}
-		if pe.Model != entry.Model {
+			sink.Emit(event.Event{Kind: event.Notice, Level: event.LevelWarn, Text: fmt.Sprintf("planner_model %q is not a configured provider; continuing with %s/%s only", pm, entry.Name, entry.Model)})
+		} else if pe.Model != entry.Model {
 			plannerProv, err := NewProviderWithProxy(pe, proxySpec)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


SivanCola:
Thanks @GTC2080. I reviewed this planner_model startup/config resilience PR while consolidating the PE handoff/routing work into #4938.

The integration PR does not include this change because stale planner model startup recovery is a separate issue from PE executor handoff and planner routing. Your work is still credited in #4938 because it helped define what is outside this integration scope.

Closing this PR as part of the consolidation cleanup; if #4615 still needs a fix, it should continue as a separate focused PR.

PR #4618, opened 2026-06-16
fix(agent): persist model in session metadata so resume restores the correct model

### LINKED ISSUE #4519: [Bug]:  Resuming existing session resets model
### Version line

### Exact version

v1.8.1 (reasonix desktop)

### What happened?

Start a conversation with a custom model - eg a model from openrouter. Close reasonix, reopen it and try to resume the conversation. It falls back to the default model. Expected behavior would be that the same model used in the last message should be used to resume it.

### Steps to reproduce

Start a conversation with a custom model - eg a model from openrouter. Close reasonix, reopen it and try to resume the conversation. It falls back to the default model. Expected behavior would be that the same model used in the last message should be used to resume it.

### OS / platform

Linux Mint 22.2 Cinnamon (noble base)

### Relevant logs or output

```shell

```

Juan-LukeKlopper:
Fixes #4519.

### What happened

When a conversation is started with a custom model (e.g. from OpenRouter) and the user closes and reopens Reasonix, resuming that conversation falls back to the default model instead of using the original custom model.

### Root cause

The model ref was never stored inside the session file. It was held only in:

- Desktop tab state (`tab.model` → `desktop-tabs.json`): lost on restart if model resolution silently falls back and overwrites it
- Controller memory (`label`): gone on close
- Session filename: cosmetic, never parsed back

### Fix

Store the full model ref (e.g. `"openrouter/anthropic/claude-sonnet"`) in the `.jsonl.meta` sidecar (BranchMeta) and read it back on every resume path:

- **BranchMeta**: new `Model` field + `LoadSessionModel()` / `SetBranchModel()` helpers
- **Controller**: new `modelRef` field, persisted to BranchMeta on every `snapshot()`
- **CLI** (`--continue`, `--resume`, `chat`, `run`, `serve`): read model from session meta before building controller, unless `--model` is explicitly set
- **Desktop**: `buildTabController` prefers the session meta model over `tab.model`
- **TUI `/resume`**: warns when the target session was created with a different model

### Backward compatibility

Old sessions without a model in their `.jsonl.meta` sidecar fall through to the existing behaviour (`tab.model` → `cfg.DefaultModel`). The model is written on the next snapshot after this change lands.

### DIFF
--- a/desktop/tabs.go
+++ b/desktop/tabs.go
@@ -1442,6 +1442,14 @@ func (a *App) buildTabController(tab *WorkspaceTab) {
 	promoteProviderKeysToCredentials(cfg)
 
 	model := strings.TrimSpace(tab.model)
+	// Prefer the session's original model from BranchMeta — the session file
+	// always knows what model created it, so it survives tab-model drift
+	// across restarts.
+	if tab.SessionPath != "" {
+		if sm, ok := agent.LoadSessionModel(tab.SessionPath); ok {
+			model = sm
+		}
+	}
 	if model == "" {
 		model = cfg.DefaultModel
 	}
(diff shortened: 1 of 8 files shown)


SivanCola:
Closing in favor of #4631, which carries forward this session-model persistence fix on the latest main-v2 and credits @Juan-LukeKlopper as a collaborator/co-author for the original #4618 implementation and problem analysis.

PR #4710, opened 2026-06-17
修复后台任务 teardown 阻塞与延迟清理

lifu963:
## 背景

后台任务的取消依赖 `context.Context` 协作返回。正常的内置 bash、provider stream、MCP 调用大多已经会响应取消，但自定义 provider/tool 或第三方集成仍可能忽略取消，导致 session close、clear、delete、trash 等 teardown 流程被无限拖住。

这会带来两个问题：用户操作可能长时间不返回；如果为了让操作先返回而延后物理清理，又必须保证旧 transcript 和 job artifacts 在 job 真正退出前不会被误删、误移动，且不会被当作普通 session 恢复。

## 做了什么

- 在 `internal/jobs` 增加有界 teardown 等待、超时结果和诊断信息，`Close()` / session destroy 不再被非协作后台 job 无限阻塞。
- 在 control、desktop、HTTP serve、ACP delete/clear/trash 路径中接入 delayed cleanup：teardown 超时后用户操作返回，物理删除或 trash 延后到 job goroutine 真正退出后执行。
- 新增 durable cleanup-pending marker，用于隐藏逻辑删除但尚未完成物理清理的 session；物理清理成功后自动清除 marker。
- 过滤所有普通 list/resume surface：agent list、HTTP `/sessions`、ACP `session/list`、HTTP/ACP/Desktop direct resume、Desktop topic lookup/pinned startup、CLI `run --resume` / `serve --resume` 都不会恢复 cleanup-pending session。
- 修复 Desktop 清理路径对 `.jobs` sidecar 的处理，避免清空或删除 session 后遗留 job artifacts。

## 结果与影响

- 非协作后台 job 不再无限拖住 close/delete/clear/trash；用户可见操作会在 teardown grace 后继续返回。
- 仍可能被运行中 job 持有或写入的 transcript、job logs、metadata 不会在 job 退出前被删除或移动。
- delete/clear 已经返回但物理清理延后的 session，会被 durable marker 隐藏；即使重启或保留旧 path/session id，也不会作为普通可恢复 session 出现。
- 协作型后台任务、活跃 session 的 `wait` / `bash_output` / `kill_shell` 语义保持不变。

## 验证

- `go test ./...`
- `cd desktop && go test ./...`
- `git merge-tree --write-tree HEAD origin/main-v2` 无冲突。

PR #4721, opened 2026-06-17
Improve desktop session restore performance / 优化桌面会话恢复性能

GTC2080:
## Problem

During the desktop performance audit, the session restore/open path had an avoidable hot spot: switching a tab to a different saved session could reload the same JSONL transcript more than once before rendering the restored history.

For small sessions this is hard to notice, but large transcripts with many tool records amplify the repeated disk IO and JSON decoding on an interaction path that users expect to feel immediate.

## Root Cause

`ResumeSessionForTab()` and `OpenChannelSessionForTab()` already loaded the target session to validate it. They then called the generic rebind path, which loaded the same session again, and the controller build path could load the pinned tab session once more before calling `ctrl.Resume()`.

## Fix

- Reuse the already-loaded `*agent.Session` when rebinding a tab to a restored or channel session.
- Thread an optional preloaded session through the desktop controller build path while keeping the existing `buildTabController()` entrypoint unchanged for normal startup.
- Fall back to the old on-demand load behavior when no preloaded session is available.
- Add a regression test proving a preloaded transcript can drive the rebind path without requiring another JSONL read.

## Performance Data

The optimized restore/open path now does one target-session JSONL load in the public entrypoint and reuses that loaded transcript through rebind/controller setup. The previous path could load the same target transcript again in `rebindTabToSessionPath()` and again while resolving the pinned tab session during controller build.

The regression test `TestRebindTabToLoadedSessionReusesPreloadedTranscript` preloads the target transcript, removes the JSONL file, and verifies the rebind still renders the restored history. That locks the no-second-read behavior directly.

Post-change benchmark snapshots:

- Frontend long-history restore benchmark: `10000-turns-archived-1-tool` produced `messages=30000`, `items=30000`, `convertMs=3.80`, `reducerMs=2.78`, `transcriptComputeMs=8.50`.
- Frontend transcript grouping contract: `groups 10k turns in 2.63ms` in the full frontend test run.
- Go history conversion benchmark: `BenchmarkHistoryMessagesSynthetic/1000x10KB-bash-success` ran at `1452255 ns/op`, `2607833 B/op`, `37047 allocs/op`.
- Go history marshal benchmark: `BenchmarkHistoryMessagesMarshalSynthetic/1000x10KB-bash-success` ran at `2024026 ns/op`, `3635856 B/op`, `37066 allocs/op`.
- Session-dir helper benchmark: `BenchmarkDesktopSessionDir` ran at `509.7 ns/op`, `848 B/op`, `6 allocs/op`.

## Verification

- `go test ./...`
- `go vet ./...`
- `cd desktop/frontend && corepack pnpm run test:all`
- `cd desktop/frontend && corepack pnpm run build`
- `cd desktop && go test . -run 'Test(ResumeSessionForTab|RebindTabToLoadedSession|OpenChannelSessionForTab)'`
- `cd desktop/frontend && corepack pnpm exec tsx src/__tests__/render-optimization.test.ts`
- `cd desktop/frontend && corepack pnpm exec tsx src/__tests__/transcript-grouping.test.ts`
- `cd desktop/frontend && corepack pnpm exec tsx src/__tests__/history-performance-benchmark.tsx`
- `cd desktop && go test . -run '^$' -bench 'Benchmark(HistoryMessages|DesktopListSessionsScoped|DesktopSessionDir)' -benchmem -count=1`
- `git diff --check`

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -1926,7 +1926,8 @@ func (a *App) ResumeSessionForTab(tabID, path string) ([]HistoryMessage, error) 
 	if err != nil {
 		return nil, err
 	}
-	if _, err := agent.LoadSession(sessionPath); err != nil {
+	loaded, err := agent.LoadSession(sessionPath)
+	if err != nil {
 		return nil, err
 	}
 	if sessionRuntimeKey(tab.currentSessionPath()) == sessionRuntimeKey(sessionPath) {
@@ -1934,7 +1935,7 @@ func (a *App) ResumeSessionForTab(tabID, path string) ([]HistoryMessage, error) 
 		return a.HistoryForTab(tabID), nil
 	}
 
-	if err := a.rebindTabToSessionPath(tab, sessionPath); err != nil {
+	if err := a.rebindTabToLoadedSessionPath(tab, sessionPath, loaded); err != nil {
 		return nil, err
 	}
 	a.setTabReadOnly(tab.ID, false)
@@ -1951,11 +1952,12 @@ func (a *App) OpenChannelSessionForTab(tabID, path string) ([]HistoryMessage, er
 	if err != nil {
 		return nil, err
 	}
-	if _, err := agent.LoadSession(sessionPath); err != nil {
+	loaded, err := agent.LoadSession(sessionPath)
+	if err != nil {
 		return nil, err
 	}
 	if sessionRuntimeKey(tab.currentSessionPath()) != sessionRuntimeKey(sessionPath) {
-		if err := a.rebindTabToSessionPath(tab, sessionPath); err != nil {
+		if err := a.rebindTabToLoadedSessionPath(tab, sessionPath, loaded); err != nil {
 			return nil, err
 		}
 	}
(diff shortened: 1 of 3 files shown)


SivanCola (approved):
Approved. This is a well-scoped and worthwhile performance fix. The PR addresses redundant desktop session transcript reloads by carrying the already-loaded `*agent.Session` from the restore/open entrypoints through the tab rebind and controller build path, while preserving the existing `buildTabController()` entrypoint and lazy-load fallback for normal startup.

- The optimization remains scoped to the desktop tab/session binding path; it does not change provider request construction, prompt content, tool schemas, or cache-sensitive prefixes.
- The preloaded session is matched by the existing canonical session runtime key before reuse, so unrelated startup and pinned-session paths keep their prior behavior.

Related PR scan: this PR merges cleanly into current `main-v2`. PR #4710 touches the same session cleanup surfaces and will need normal rebase/conflict care if it lands first, but I do not see a blocker for this PR.

Verified:
- `git merge-tree --write-tree origin/main-v2 refs/remotes/origin/pr/4721` passed cleanly
- `git diff --check` passed
- `cd desktop && go test . -run 'Test(ResumeSessionForTab|RebindTabToLoadedSession|OpenChannelSessionForTab)' -count=1` passed
- `cd desktop && go test . -run 'TestRebindTabToLoadedSessionReusesPreloadedTranscript' -count=1 -v` passed

SivanCola:
@codex review

PR #4737, opened 2026-06-17
[codex] Attach workspace image refs to vision turns

Surfire:
## Summary

- Attach workspace image references (`@docs/image.png` and absolute paths under the workspace) to user turns as model image input, matching the existing attachment-image path.
- Keep image refs visible in referenced context, but stop prompting the model to reach for OCR/vision tools when direct image input is available.
- Add regression coverage for workspace-relative and absolute workspace image refs.

## Root Cause

Reasonix only populated `provider.Message.Images` for `.reasonix/attachments/...` image refs. Regular workspace image refs were resolved as file metadata with text saying the bytes were not inlined and suggesting OCR/vision tooling, so vision-capable local providers could be steered away from their native multimodal path.

## Validation

- `GOPROXY=https://goproxy.cn,direct GOSUMDB=sum.golang.google.cn go test ./internal/control ./internal/provider/openai ./internal/provider/anthropic ./internal/config`
- `GOPROXY=https://goproxy.cn,direct GOSUMDB=sum.golang.google.cn go test ./cmd/reasonix ./internal/boot`
- Local smoke test with an oMLX Qwen3.6 vision-capable model using a workspace PNG ref; the model answered from direct image input without requesting OCR.

### DIFF
--- a/internal/control/inputimages_test.go
+++ b/internal/control/inputimages_test.go
@@ -1,6 +1,8 @@
 package control
 
 import (
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 )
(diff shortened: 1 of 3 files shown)


SivanCola:
Follow-up: I opened #5023 to tighten the file-path boundary for the workspace image-ref path added here.

The follow-up addresses CodeQL `go/path-injection` alerts 184, 185, and 186 by keeping file-image reads behind workspace-root containment and documenting the security boundary in `SECURITY.md`.

PR #4754, opened 2026-06-17
fix(desktop): support macOS workbench sidebar titlebar zoom

weikengchen:
## Problem

On macOS with `TitleBarHiddenInset`, the WKWebView consumes double-clicks in the empty 46px titlebar strip above the workbench sidebar, so the native window zoom action is not triggered there.

## Fix

- Keep the existing 46px macOS workbench sidebar padding, preserving the current sidebar geometry.
- Extend the centralized `handleChromeTitlebarDoubleClick` path to recognize only the first 46px of `.sidebar--workbench` on macOS.
- Reuse `app.ToggleMaximiseMainWindow()` and the existing interactive-control exclusions.
- Leave classic and creation styles, and Windows/Linux behavior, unchanged.

The existing centralized handler continues to cover `.app-chrome`, `.topicbar`, and `.workbench-dock__tools`; this PR only adds the missing macOS workbench-sidebar titlebar strip.

## Verification

- [x] `pnpm exec tsx src/__tests__/app-chrome-tabs.test.ts` — 128 passed
- [x] `pnpm build`
- [x] `CI=true wails build -platform darwin/arm64`
- [x] Browser geometry: sidebar content remains 46px from the top, with no synthetic spacer
- [x] Native macOS: top-strip double-click zooms `1240x720` to `1300x768` and restores it; a double-click below the strip leaves the window size unchanged

## Documentation impact

Documentation-impact: none - desktop-only interactive behavior; no product documentation, CLI help, or API surface changes.

### DIFF
--- a/desktop/frontend/src/App.tsx
+++ b/desktop/frontend/src/App.tsx
@@ -2393,6 +2393,7 @@ export default function App() {
         <aside className={sidebarClassName} aria-label={t("sidebar.navigation")}>
           {sidebarWorkbench ? (
             <>
+              {desktopPlatform === "darwin" && <div style={{ height: 46, flexShrink: 0 }} onDoubleClick={() => window.runtime?.WindowToggleMaximise?.()} />}
               <div className="sidebar__head" aria-hidden={sidebarCollapsed}>
                 <div className="sidebar__brand sidebar__brand--workbench">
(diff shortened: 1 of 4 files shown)


weikengchen:
Updated: all three regions now gate on workbench mode
(`sidebarWorkbench`) in addition to macOS, so double-click zoom
only applies when "桌面风格" is set to workbench.

PR #4762, opened 2026-06-17
fix(desktop): prevent stale history after new session

### LINKED ISSUE #4758: [Bug]: 新建会话，会话里内容不为空
### Version line

### Exact version

1.9.1桌面版

### What happened?

新建会话，会话里内容不情清空，而是会引入其他对话的内容。后来解决办法：删除项目目录重新添加

### Steps to reproduce

偶然发现无法复现

### OS / platform

win11

### Relevant logs or output

```shell

```

GTC2080:
## Problem

Opening a new desktop session could still show messages from the previous conversation when an in-flight tab history load completed after the new-session action. This matches #4758's report: the backend session had rotated, but the frontend could repaint stale history into the cleared transcript.

Fixes #4758.

## Root Cause

`loadSessionDataForTab()` already uses per-tab sequence numbers to drop stale loads, but `newSession()` and `clearSession()` did not invalidate that sequence. A slow `HistoryForTab()` from startup, tab restore, or project switching could therefore resolve after the reset and dispatch `history` into the blank session.

## Fix

- Invalidate pending session loads before and after `NewSession` / `ClearSession`.
- Reload current tab data if the backend refuses the action, so startup/failure paths keep the visible transcript consistent.
- Add a jsdom hook-level regression test that resolves an old history promise after `NewSession()` and verifies it cannot repopulate the transcript.

## Verification

- `cd desktop/frontend && corepack pnpm exec tsx src/__tests__/new-session-load-race.test.tsx`
- `cd desktop/frontend && corepack pnpm run test:all`
- `cd desktop/frontend && corepack pnpm run build`
- `go test ./...`
- `git diff --check`
- `gofmt -l $(rg --files -g '*.go')`


SivanCola (approved):
Approved after resolving the main-v2 package.json test-script conflict in commit e2acf5af.

The resolution keeps the latest main-v2 frontend test entries and adds this PRs new-session stale-history regression test to both test scripts.

Verified locally:
- git diff --check
- pnpm --dir desktop/frontend exec tsx src/__tests__/new-session-load-race.test.tsx

The new PR CI run is still pending after the push.

SivanCola:
Follow-up: #5114 builds on this stale-history guard and fixes the remaining desktop session-switch UX issue by making tab switches render optimistically while history, checkpoints, jobs, context usage, and controller readiness hydrate in the background.

It keeps the per-tab sequence guard from this PR, removes the full-screen `ready=false` loading state during normal tab switches, and adds a regression test for slow cross-tab history responses.

Verified:
- `corepack pnpm run test:all` in `desktop/frontend`
- `go test ./...` in `desktop`
- `git diff --check`

PR #4764, opened 2026-06-17
Improve desktop startup bundle splitting / 优化桌面启动包拆分

GTC2080:
## Problem

The desktop startup bundle still carried several non-startup code paths in the initial application graph: secondary panels, PDF/image session export helpers, and the Markdown/KaTeX renderer stack.

That made the first desktop render pay extra parse and preload cost for features users only need after opening settings/history, exporting, or rendering rich Markdown/math content.

## Root Cause

`App.tsx` statically imported settings/history panels and session export helpers, while `Markdown.tsx` statically imported `react-markdown`, remark/rehype plugins, KaTeX CSS, and the full renderer path. Those imports made heavy secondary functionality reachable from the startup graph.

## Fix

- Load `HistoryPanel` and `SettingsPanel` with `React.lazy` only when those drawers open.
- Import `sessionExport` only when exporting PDF or image output.
- Split `Markdown.tsx` into a light wrapper and a lazily loaded `MarkdownRenderer` that owns Markdown, math, KaTeX, code block, and external-link rendering.
- Add a bundle contract test so these heavy modules do not get accidentally reintroduced into the initial chunk.

## Performance Data

Post-change production build output shows the initial `index` chunk at `820.07 kB` minified / `245.96 kB` gzip. Before the split, the same entry chunk was about `1,067.48 kB` minified / `304.16 kB` gzip.

The generated `index.html` now preloads only the app entry, Rolldown runtime, and React vendor chunk. The following paths are emitted as on-demand chunks instead of startup work:

- `SettingsPanel` (`214.14 kB` / `47.77 kB` gzip)
- `HistoryPanel` (`13.79 kB` / `3.96 kB` gzip)
- `vendor-markdown` (`502.48 kB` / `141.27 kB` gzip)
- `sessionExport` (`10.42 kB` / `3.81 kB` gzip)

## Verification

- `corepack pnpm --dir desktop/frontend test:all`
- `corepack pnpm --dir desktop/frontend build`
- `go test ./...` in the desktop module
- `go test ./...` in the repository root
- `git diff --check`


SivanCola:
Closing this as superseded by #4816.

I carried over the lazy-loaded Settings and History panels, lazy Markdown renderer, on-demand session export import, and bundle contract test. The integration PR resolves the App/package script overlap with the lightweight startup settings work so both startup improvements land together.

Authorship is preserved in #4816 via the cherry-picked commit, with SivanCola added as co-author for the integration and conflict-resolution work.

PR #4799, opened 2026-06-18
fix: prevent activeTab fallback from showing wrong conversation on new tab

ttmouse:
## Bug

When creating a new tab (clicking "+" on a project or the "new session" button), the **activeTab** computation in `App.tsx` has a fallback that briefly shows the **old** conversation instead of the new one.

### Root cause

The tab-creation flow is:

```
openProjectTab / openGlobalTab / ensureBlankTab
  → app.OpenProjectTab (backend creates tab, sets activeTabID)
  → setActiveTabId(meta.id)
  → await loadSessionDataForTab(meta.id, ...)
  ...
  → refreshTabMetas()
```

Between `setActiveTabId` and `refreshTabMetas`, there is a timing window where `activeTabId` points to the new tab but `tabMetas` does not yet contain it. During this window the `activeTab` computation:

```ts
tabMetas.find((tab) => tab.id === activeTabId)
  ?? tabMetas.find((tab) => tab.active)    // fallback
```

falls back to `tabMetas.find((tab) => tab.active)`, which returns the **old** tab whose `active` field is still `true` in the stale `tabMetas`. The UI renders the old conversation while the user thinks they are starting a new one.

If the user submits input during this window, the message may go to the wrong tab.

## Fix

Only use the `tab.active` fallback when `activeTabId` is null or undefined (initial load). When `activeTabId` is explicitly set to a tab not yet in `tabMetas`, return `undefined` instead of falling back to a stale tab.

All consumers of `activeTab` use optional chaining (`?.`) and handle `undefined` gracefully.

### DIFF
--- a/desktop/tabs.go
+++ b/desktop/tabs.go
@@ -1115,7 +1115,10 @@ func (a *App) EnsureBlankTab(scope, workspaceRoot string) (TabMeta, error) {
 		tab := a.tabs[id]
 		if a.blankTabMatchesTargetLocked(tab, scope, workspaceRoot) {
 			a.activeTabID = tab.ID
-			tab.TopicTitle = defaultTopicTitle
+			if err := resetReusableBlankTabTitle(tab, scope, workspaceRoot); err != nil {
+				a.mu.Unlock()
+				return TabMeta{}, err
+			}
 			meta := a.tabMeta(tab, true)
 			a.saveTabsLocked()
 			a.mu.Unlock()
@@ -1239,6 +1242,25 @@ func (a *App) blankTabMatchesTargetLocked(tab *WorkspaceTab, scope, workspaceRoo
 		return false
 	}
 	return !messagesHaveConversationContent(tab.Ctrl.History())
+}
+
+func resetReusableBlankTabTitle(tab *WorkspaceTab, scope, workspaceRoot string) error {
+	if tab == nil {
+		return nil
+	}
+	topicID := strings.TrimSpace(tab.TopicID)
+	if topicID == "" {
+		return nil
+	}
+	titleRoot := topicTitleRoot(scope, workspaceRoot)
+	if source := loadTopicTitleSource(titleRoot, topicID); source != topicTitleSourceAuto {
+		return nil
+	}
+	if err := setTopicTitleWithSource(titleRoot, topicID, defaultTopicTitle, topicTitleSourceAuto); err != nil {
+		return err
+	}
+	tab.TopicTitle = defaultTopicTitle
+	return nil
 }
 
 // indexedBlankTopicIDLocked finds a blank topic ID that is indexed on disk
--- a/desktop/tabs.go
+++ b/desktop/tabs.go
@@ -1115,6 +1115,7 @@ func (a *App) EnsureBlankTab(scope, workspaceRoot string) (TabMeta, error) {
 		tab := a.tabs[id]
 		if a.blankTabMatchesTargetLocked(tab, scope, workspaceRoot) {
 			a.activeTabID = tab.ID
+			tab.TopicTitle = defaultTopicTitle
 			meta := a.tabMeta(tab, true)
 			a.saveTabsLocked()
 			a.mu.Unlock()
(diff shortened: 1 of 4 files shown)


ttmouse:
@codex review

ttmouse:
## 场景说明

创建新对话时（点 "+" 或 "新会话"），系统会尝试复用一个已有的空白 tab 而不是新建一个。如果这个空白 tab 之前从某个旧对话清理过来，它的 `TopicTitle` 还残留着旧对话的标题。

当 UI 用这个残留标题渲染时，会让人感觉「旧对话的内容又回来了」—— 实际上这是标题没重置，不是聊天记录被覆盖了。

**修复：** 在 `EnsureBlankTab` 复用时把 `TopicTitle` 重置为默认标题（`"新的会话"`），这样 UI 始终展示正确的标题。聊天记录本身不受影响，只是标题显示的问题。

desktop/tabs.go
```diff
@@ -1115,6 +1115,7 @@ func (a *App) EnsureBlankTab(scope, workspaceRoot string) (TabMeta, error) {
 		tab := a.tabs[id]
 		if a.blankTabMatchesTargetLocked(tab, scope, workspaceRoot) {
 			a.activeTabID = tab.ID
+			tab.TopicTitle = defaultTopicTitle
```
SivanCola: `blankTabMatchesTargetLocked` treats an open blank tab as reusable based on scope, root, runtime state, and history only; it does not check whether the topic title was manually renamed. With this unconditional assignment, a blank topic that the user renamed before sending any messages is immediately rendered as `新的会话` when `EnsureBlankTab` reuses it. A later session-meta write can then persist that default while the topic title source remains manual, so auto-title will not repair it. Please preserve the existing title unless the topic is known to be auto/default, and add a regression for a manually renamed blank topic.

SivanCola (changes requested):
Thanks, this is directionally focused on the stale blank-tab title, but I cannot approve it as-is because the current one-line change regresses manually renamed blank topics.

I also checked the nearby open work in #4762. That PR covers the separate stale-history load race; it does not make this title overwrite safe. This branch applies cleanly to the latest `main-v2` and CI is green, but please preserve the manual-title contract and add a focused regression before merging.

Once this is addressed and verification stays green, I will consider merging it into `main-v2`.

desktop/tabs.go
SivanCola: Fixed in d92103fd. `EnsureBlankTab` now resets a reusable blank tab title only when the topic title source is `auto`, and it persists the default title with the auto source so the returned tab metadata and topic index stay consistent. Manually renamed blank topics are preserved.

Verified:
- `go test ./... -run 'TestEnsureBlankTab|TestRenameTopicLocksTitleManual|TestAutoTitleDoesNotOverrideManualTopicTitle|TestClearSessionCancelsRunningRuntimeAndKeepsTopic' -count=1` from `desktop`
- `go test ./...` from `desktop`
- `git diff --check`

SivanCola (approved):
Approved after the follow-up fix in d92103fd.

This now preserves manually renamed blank topics while still resetting stale auto-sourced blank titles to the default title. The added regression coverage verifies both sides of that contract.

Verified:
- `go test ./... -run 'TestEnsureBlankTab|TestRenameTopicLocksTitleManual|TestAutoTitleDoesNotOverrideManualTopicTitle|TestClearSessionCancelsRunningRuntimeAndKeepsTopic' -count=1` from `desktop`
- `go test ./...` from `desktop`
- `git diff --check`

SivanCola:
@codex review

SivanCola:
Closing this PR in favor of #4819.

Reason: this PR started with an `App.tsx` active-tab fallback approach, but the final verified implementation is a backend `EnsureBlankTab` title-source fix. #4819 carries the final code on top of the latest `main-v2` with a title and body that match the implementation.

Bot feedback audit:
- The three `App.tsx` metadata-pending comments were valid for the intermediate frontend change, but that frontend change was reverted, so they are no longer applicable to the final fix.
- The manually named blank topic feedback was valid and is fixed in #4819.

Authorship note: @ttmouse remains the primary contributor for the fix. @SivanCola contributed as a collaborator by reviewing, fixing the follow-up manual-title regression, and verifying the final state.

PR #4717, opened 2026-06-17
Fix settings credential lookup lag / 优化设置凭据解析延迟

GTC2080:
## Problem

Fixes #4713.

After the v1.9.1 desktop credential-priority changes, Settings page loads and Settings save/reload interactions could feel slower when multiple provider entries shared the same `api_key_env`.

The visible report included broader page-operation lag. The reproducible backend hotspot is the Settings provider/key-state path: opening Settings and saving lightweight Settings controls both rebuild the provider view, and that path was doing repeated credential-store work on the UI interaction path.

## Root Cause

`Settings()` builds both the official provider templates and the configured provider list. Each provider row called `ResolveCredentialForRootGlobalFirst()` independently.

That one-shot resolver checks the global credential store before project/home env files and also computes shadowed sources. Default desktop providers intentionally share keys such as `DEEPSEEK_API_KEY` and `MIMO_API_KEY`, so one Settings refresh could hit the same macOS Keychain / Reasonix credentials lookup multiple times for identical keys.

## Fix

- Add a per-view `CredentialResolver` that caches global-first credential resolution by key while preserving the existing source and shadowed-source metadata.
- Reuse one resolver across `Settings()` official provider rows and configured provider rows.
- Keep the existing one-shot `ResolveCredentialForRootGlobalFirst()` API intact for other callers.
- Add regression coverage proving repeated global-first resolution for the same key performs one stored credential lookup.

## Cache impact

None. This only changes desktop Settings credential-status lookup plumbing. It does not change provider request bodies, model selection semantics, prompt content, tool schemas, or cache-sensitive prefixes.

## Verification

- `go test ./internal/config -run 'TestCredentialResolverCachesGlobalFirstLookups|TestResolveCredentialGlobalFirstDoesNotMutateProjectEnv|TestResolveCredentialSourceShowsCredentials' -count=1`
- `cd desktop && go test . -run 'TestSettings(LoadsActiveWorkspaceCredentialsWithUserConfig|ShowsGlobalCredentialWithoutMutatingWorkspaceEnv|SeedsMissingUserConfigFromLegacyProjectConfig)' -count=1`
- `go test ./internal/config -count=1`
- `cd desktop && go test . -count=1`
- `go test ./...`
- `cd desktop && go test ./...`
- `cd desktop && go run github.com/wailsapp/wails/v2/cmd/wails@v2.12.0 generate module`
- `cd desktop/frontend && corepack pnpm run test:all`
- `cd desktop/frontend && corepack pnpm run build`
[...]

PR #4803, opened 2026-06-18
Reduce desktop startup settings load / 降低桌面启动设置加载开销

GTC2080:
## Problem

Follow-up to #4713, #4717, and #4721.

The desktop frontend still called the full `Settings()` binding during startup and sidebar bot-status refreshes, even though those paths only need chrome preferences, update-check state, display mode, status bar settings, and the bot sidebar snapshot.

On slower Intel Mac WebViews, doing this at startup can amplify the broader 1.9.x responsiveness regression because `Settings()` builds provider rows, official provider templates, provider kinds, credential/key status, sandbox/network settings, and the rest of the full Settings Center payload before the user has opened Settings.

## Root Cause

`App.tsx` reused `app.Settings()` as a convenient all-purpose settings snapshot for desktop startup preferences and bot sidebar state. That tied the first-render/startup path to Settings Center provider construction and credential metadata work.

The full payload is still needed when the Settings Center is open, but it is unnecessary for the startup chrome path.

## Fix

- Add `DesktopStartupSettings()` as a lightweight backend binding for startup chrome preferences and bot sidebar state.
- Keep provider/key status and the rest of the full Settings payload inside `Settings()` for the Settings Center.
- Update frontend startup preference sync and sidebar IM refresh to call the lightweight binding.
- Add Go coverage for the lightweight payload shape, non-nil arrays, and preserved user desktop preferences.
- Add a frontend contract test that prevents startup paths from regressing to full `Settings()` calls.
- Regenerate Wails module bindings locally so the new method is represented in the generated app bridge.

## Performance Data

Synthetic benchmark with 40 configured providers sharing one key:

- `BenchmarkDesktopSettingsPayloads/Settings-18`: `1,437,239 ns/op`, `901,457 B/op`, `7,945 allocs/op`
- `BenchmarkDesktopSettingsPayloads/DesktopStartupSettings-18`: `395,168 ns/op`, `462,747 B/op`, `4,322 allocs/op`

That removes full provider/settings payload construction from the startup chrome path and cuts this measured snapshot by roughly 3.6x in time and about 49% in allocated bytes on the local benchmark machine.

## Verification

- `go test . -run 'Test(DesktopStartupSettingsUsesUserDesktopPreferencesWithoutFullSettingsPayload|BoundArrayPayloadsAreNonNilBeforeStartup)' -count=1` in the desktop module
- `go test . -run '^$' -bench BenchmarkDesktopSettingsPayloads -benchmem -count=1` in the desktop module
- `go run github.com/wailsapp/wails/v2/cmd/wails@v2.12.0 generate module` in the desktop module
- `corepack pnpm --dir desktop/frontend exec tsx src/__tests__/startup-settings-contract.test.ts`
- `corepack pnpm --dir desktop/frontend exec tsc --noEmit -p tsconfig.test.json`
- `corepack pnpm --dir desktop/frontend test:all`
- `corepack pnpm --dir desktop/frontend build`
- `go test ./...` in the desktop module
- `go test ./...` in the repository root
- `go vet ./...` in the desktop module
- `go vet ./...` in the repository root
- `git diff --check`

Refs #4713


SivanCola:
Closing this as superseded by #4816.

I carried over `DesktopStartupSettings()`, the startup/sidebar frontend call-site changes, the non-nil payload guard, and the startup settings contract coverage. The integration PR keeps this lightweight startup path while also absorbing the related Settings credential cache, startup bundle split, and Settings panel render optimizations.

Authorship is preserved in #4816 via the cherry-picked commit, with SivanCola added as co-author for the integration and conflict-resolution work.

Issue #4727, opened 2026-06-17
[Bug]: MacOS端旧会话无法对话，状态按钮弹出立马消失，新建会话可以恢复正常使用


CVEngineer66:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.9.1

### What happened?

<img width="1620" height="908" alt="Image" src="https://github.com/user-attachments/assets/eb3ce1a6-6bcc-4856-87d1-85c2bd6f03e1" />

<img width="1594" height="1104" alt="Image" src="https://github.com/user-attachments/assets/73427836-69f1-4de4-a0e6-cd6a1c8b700f" />

### Steps to reproduce

打开旧会话对话，进行对话，没有任何输出和报错提示

### OS / platform

macOS 26.5

### Relevant logs or output

```shell

```

PR #4811, opened 2026-06-18
fix(agent,provider): unify history normalization into single NormalizeMessages entry point

wangwangcodecode:
## Issue
The back-and-forth cycle of #4727 -> #4775 -> revert came from the load path and send path carrying overlapping historical-message remediation logic. Logic drift made the same replay bug easy to reintroduce.

## Changes
- Kept `NormalizeMessages` as the provider wire-safe normalization entry point for tool-call name backfill, truncated-argument repair, orphan-result cleanup, and unanswered-call placeholders.
- Added `NormalizeSessionMessages` for `LoadSession`, sharing assistant-turn repairs while preserving existing stored tool messages so `Save`/`LoadSession` remains a round-trip contract.
- Refactored `SanitizeToolPairing` to delegate to `NormalizeMessages`.
- Tightened the fast path so healthy tool-call/tool-result histories return the original slice without allocation.
- Added coverage for standalone stored tool results, extra stored tool results, and healthy-history fast-path stability.
- Added the `normalizedDirty` field to `Session` to track sessions that have undergone persisted-safe remediation.

Cache-impact: low - provider-visible bytes for healthy histories stay unchanged, and the fast path now explicitly preserves the original slice for well-formed tool-call histories.
Cache-guard: `go test ./internal/provider -run 'TestSanitizeToolPairing|TestNormalizeSessionMessages|TestBackfill' -count=1`; `go test ./internal/agent -run 'TestSaveLoadRoundTrip|TestSaveLoadLargeMessage' -count=1`; `go test ./...`

## Related Files
- `internal/provider/provider.go` - `NormalizeMessages`, `NormalizeSessionMessages`, fast-path checks
- `internal/agent/normalize.go` - session-safe normalization wrapper
- `internal/agent/save.go` - `LoadSession` call site
- `internal/agent/session.go` - `normalizedDirty` field
- `internal/provider/provider_test.go` - normalization regression coverage

## 问题
#4727 -> #4775 -> revert 的反复折腾，根因是 load 路径和 send 路径各自维护历史消息修复逻辑，逻辑漂移后同类 replay bug 容易反复出现。

## 改动
- 保留 `NormalizeMessages` 作为 provider 发送前的 wire-safe 归一化入口，负责 tool-call name 回填、截断参数修复、孤立结果清理和未应答调用占位。
- 新增 `NormalizeSessionMessages` 给 `LoadSession` 使用，共享 assistant turn 修复，但保留磁盘里已有的 tool message，避免破坏 `Save`/`LoadSession` round-trip 合约。
- `SanitizeToolPairing` 委托 `NormalizeMessages`。
- 强化 fast path：健康的 tool-call/tool-result 历史会原 slice 返回，不分配。
- 补充 standalone/extra stored tool result 以及健康历史 fast-path 稳定性的测试。
- `Session` 新增 `normalizedDirty` 字段标记经过 persisted-safe 修复的会话。

### DIFF
--- a/internal/agent/normalize.go
+++ b/internal/agent/normalize.go
@@ -2,14 +2,12 @@ package agent
 
 import "reasonix/internal/provider"
 
-// NormalizeSession runs the canonical history repairs on a loaded conversation
-// and is the agent-side entry point for making old, partially-saved, or
-// interrupted sessions replayable. It is a thin wrapper over
-// provider.NormalizeMessages — the single source of truth for every repair
-// (tool-call name backfill, truncated-arg closing, orphan-result dropping,
-// placeholder backfill for unanswered calls) — so the load path and the
-// provider send path can never drift apart (the root cause of the
-// #4727 → #4775 → revert churn).
+// NormalizeSession runs the persisted-history-safe repairs on a loaded
+// conversation and is the agent-side entry point for making old, partially
+// saved, or interrupted sessions replayable. It is a thin wrapper over
+// provider.NormalizeSessionMessages, which shares assistant-turn repairs with
+// the provider send path without applying wire-only cleanup such as dropping
+// standalone tool messages.
 //
 // LoadSession calls this right after decoding so a session that was written by
 // an older code version, or that was cut short mid-turn, is corrected in memory
@@ -21,8 +19,8 @@ import "reasonix/internal/provider"
 // load — cheap, because the fast path returns the input slice unchanged.
 //
 // Well-formed histories are returned without allocating (see
-// provider.NormalizeMessages), so this is a no-op in both time and memory for
-// the common case and cannot perturb a provider's prefix-cache key.
+// provider.NormalizeSessionMessages), so this is a no-op in both time and memory
+// for the common case and cannot perturb a provider's prefix-cache key.
 func NormalizeSession(msgs []provider.Message) []provider.Message {
-	return provider.NormalizeMessages(msgs)
+	return provider.NormalizeSessionMessages(msgs)
 }
--- a/internal/agent/normalize.go
+++ b/internal/agent/normalize.go
@@ -1,0 +1,28 @@
+package agent
+
+import "reasonix/internal/provider"
+
+// NormalizeSession runs the canonical history repairs on a loaded conversation
+// and is the agent-side entry point for making old, partially-saved, or
+// interrupted sessions replayable. It is a thin wrapper over
+// provider.NormalizeMessages — the single source of truth for every repair
+// (tool-call name backfill, truncated-arg closing, orphan-result dropping,
+// placeholder backfill for unanswered calls) — so the load path and the
+// provider send path can never drift apart (the root cause of the
+// #4727 → #4775 → revert churn).
+//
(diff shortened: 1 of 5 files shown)


SivanCola (changes requested):
Thanks for tackling the drift between the load path and provider send path. The direction is reasonable, but this cannot merge as-is because the new load-time normalization changes the persisted session round-trip contract.

Blocking issue:
- `internal/agent/save.go`: `LoadSession` now runs `NormalizeSession` on every decoded session. Because `NormalizeMessages` drops any `RoleTool` message that is not immediately preceded by an assistant `tool_calls` turn, previously saved standalone tool-result messages are removed during load. This breaks the existing `reasonix --resume` round-trip contract and is already caught by CI: `TestSaveLoadLargeMessage` loads 2 messages instead of the 3 that were saved.

There is also a cache/perf correctness mismatch to tighten before approval: `tryNormalizeFastPath` only returns the original slice for histories with no tool calls or tool results at all. A healthy history that contains a normal assistant tool call and matching tool result still allocates a new slice, so the `LoadSession` slice-header dirty check marks healthy tool histories as normalized. That does not appear to change provider-visible bytes today, but it contradicts the PR's stated zero-allocation/cache-stability claim and makes `normalizedDirty` unreliable.

Please preserve `LoadSession` round-trip semantics, or move destructive wire-only repairs such as orphan-tool dropping back out of the persisted load path. After that, add focused coverage for both contracts:
- loading a saved standalone large tool message preserves all messages;
- a well-formed tool-call/tool-result history either remains pointer-identical when claimed, or the dirty/cache-stability claim is adjusted accordingly.

Verified:
- `go test ./internal/agent -run TestSaveLoadLargeMessage -count=1` fails with `message count = 2, want 3`
- `go test ./internal/provider -run 'TestSanitizeToolPairing|TestBackfill' -count=1` passes
- `git diff --check origin/main-v2...origin/pr/4811` passes

Also note that the PR is behind the latest `main-v2`, and #4841 overlaps the sanitizer/session-migration area. Please rebase and reconcile the normalization boundary with that work before another review pass.

SivanCola (approved):
Approved after the follow-up fixes in `607cb9b9`.

What changed:
- `LoadSession` now uses session-safe normalization, preserving stored tool messages instead of applying wire-only orphan cleanup during load.
- Provider send-time normalization still drops/reorders orphan results as needed for the API contract.
- The fast path now recognizes healthy tool-call/tool-result histories and returns the original slice.
- The PR was rebased onto the latest `main-v2`, and the PR body now includes the required cache-impact metadata.

Verified:
- `go test ./internal/provider -run 'TestSanitizeToolPairing|TestNormalizeSessionMessages|TestBackfill' -count=1`
- `go test ./internal/agent -run 'TestSaveLoadRoundTrip|TestSaveLoadLargeMessage' -count=1`
- `go test ./internal/agent ./internal/provider -count=1`
- `go test ./...`
- `git diff --check`
- `scripts/check-cache-impact.sh` with the PR body metadata

Latest GitHub checks for the pushed head show `cache-impact` and `label` passing.

PR #4843, opened 2026-06-19
fix(desktop): expand workspace change records

### LINKED ISSUE #4770: [Bug]: 改动记录
### Version line

### Exact version

1.9.1

### What happened?

不是很清楚这个改动记录的原理，这次会话已经进行13轮了，中途已经有多次改动了，但目前就显示一次，而且标记还是两个问号??,
Image

### Steps to reproduce

无

### OS / platform

windows11

### Relevant logs or output

```shell

```

GTC2080:
## Summary

- Fixes #4770 by exposing one desktop workspace-change record per session checkpoint/file instead of only the aggregated changed-file row.
- Keeps the existing aggregated `files` payload for compatibility, while adding `records` for the changes panel to show repeated edits to the same file as separate records.
- Maps raw git porcelain statuses such as `??` to localized user-facing labels, so the panel shows `未跟踪` / `Untracked` instead of leaking implementation codes.
- Adds backend and frontend regression coverage for repeated same-file session edits and readable git status badges.

## Verification

- [x] `go test -run TestWorkspaceChangesIncludesSessionRecordsPerTurn` (from `desktop/`)
- [x] `corepack pnpm tsx src/__tests__/workspace-changes.test.ts` (from `desktop/frontend/`)
- [x] `corepack pnpm run test:typecheck` (from `desktop/frontend/`)
- [x] `go test -run 'TestWorkspaceChanges(UsesRequestedTabCheckpoints|IncludesSessionRecordsPerTurn|GitStatus|GitStatusFromRepoSubdirectory|UntrackedDirectoryListsFiles)'` (from `desktop/`)
- [x] `go run github.com/wailsapp/wails/v2/cmd/wails@v2.12.0 generate module` (from `desktop/`)
- [x] `go test ./...`
- [x] `go test ./...` (from `desktop/`)
- [x] `corepack pnpm run test:all` (from `desktop/frontend/`)
- [x] `corepack pnpm run build` (from `desktop/frontend/`)
- [x] `git diff --check`
- [x] `gofmt -l $(git ls-files '*.go')`
- [x] `go vet ./...`
- [x] `go vet ./...` (from `desktop/`)

## Cache impact

Cache-impact: none - desktop workspace-change payload/UI rendering only; no provider request shape, system prompt, memory prefix, output style, skill index, or tool schema bytes change.
Cache-guard: `go test -run TestWorkspaceChangesIncludesSessionRecordsPerTurn`; `corepack pnpm tsx src/__tests__/workspace-changes.test.ts`; `go test ./...`; `corepack pnpm run test:all`.
System-prompt-review: N/A


esengine:
Thanks for this — your diagnosis was right on both counts, and I'm sorry it sat long enough to go stale.

I've taken the git-status label half onto current `main-v2` in #7469 with a `Co-authored-by:` line for you. One correction I made while porting: matching `s.includes("U")` for unmerged misses `AA` and `DD`, which are also merge conflicts and were being reported as a plain add/delete — the unmerged set is matched as whole codes now.

The per-turn records half doesn't port. `WorkspacePanel.tsx` has been rewritten since June (split view, separate session/git scopes, per-file revert), so the rendering in this PR targets a component that no longer exists. I've left #4770 open with the shape it would take against the current panel, including where the per-file revert button has to land once one file has several rows. If you want to take another run at it I'll review it promptly this time.

PR #4862, opened 2026-06-19
feat: add list_sessions and read_session tools

eghrhegpe:
## Summary

Add two read-only tools that let the AI discover and read past conversation sessions, enabling cross-session AI context sharing. When an AI calls `list_sessions`, it sees available sessions; `read_session(session="...")` returns the full conversation.

### list_sessions
- Scans the session directory and returns all saved `.jsonl` session files
- Parses timestamp and model from filenames, counts user turns per session
- Returns a formatted table sorted newest-first

### read_session
- Loads a session JSONL file by name/path
- Renders the full conversation as readable text organized by speaker turn
- Shows system prompts, user messages, assistant responses (including reasoning), tool calls, and tool results
- Supports `max_turns` parameter (default 50) to limit output length

## Files changed
- `internal/tool/builtin/sessions.go` — new file with both tool implementations (295 lines)
- `internal/boot/boot.go` — register the two tools with session directory during boot
- `internal/boot/boot_test.go` — update expected tool list in token economy test

Cache-impact: low — adds two read-only tools (`ReadOnly() = true`) registered during boot alongside existing tools. No changes to system prompt, provider messages, session compaction, or any prefix-cache-sensitive path. The tools only read existing session files from disk using `os.Open` + `json.NewDecoder`; they don't alter session state or cache behavior.

Cache-guard: existing tests pass (`go test ./internal/tool/builtin/` and `go test ./internal/boot/`). The tool implementations are straightforward wrappers around standard library IO with no shared mutable state.

System-prompt-review: self-reviewed — tool registration only; no changes to system prompts, prefix cache surface, memory loading, skills, or config prompt rendering.


eghrhegpe:
在修bug了，全红可还行

SivanCola (changes requested):
Thanks for the PR. I cannot approve this version yet because it widens the model-visible history surface beyond the safeguards in the existing `history` tool.

Required changes:

- `read_session` should not expose full raw session transcripts as an autonomous read-only tool. The current renderer includes system prompts, `reasoning_content`, tool calls, and normal tool results, while the existing `history` tool intentionally uses search + small `around` windows and excludes normal tool output by default. Session logs can contain secrets from tool results, private user text, and provider reasoning that should not be reintroduced into a later model context without a narrower gate. Please either reuse the existing `history.Searcher` constraints or add an explicit user-gated/export-style flow with redaction, no reasoning output, no system prompt by default, cleanup-pending filtering, and bounded output.

- Please fix and test the `max_turns` contract. The `break` after the truncation message exits only the `switch`, so later assistant/tool messages can still be rendered after the limit. Also the schema says `max_turns: 0` means no limit, but the implementation currently keeps the default 50.

Cache/security notes: adding stable tool schemas is manageable, but this tool can inject large, dynamic past-session content into the current turn and can bypass the existing history privacy boundaries. That needs to be tightened before merge.

Verified while reviewing:

- `go test ./internal/tool/builtin ./internal/boot ./internal/history` passed
- `git diff --check origin/main-v2...HEAD` passed
- GitHub CI is currently green

Once the read surface is narrowed and the truncation behavior is covered by regression tests, I can re-review.

eghrhegpe:
哇塞，你这么一说，问题还挺大的，我以为暴露接口后，调用的原始工具会进行一次初筛，马上改。

eghrhegpe:
暂时修复的问题
#	Reviewer 指出的问题	修复方式
1	read_session 暴露完整原文（含 secrets、私有文本、reasoning）	每条消息truncateRunes(content, 2000)限制，工具结果 2000、参数 1200，复用 history 工具的截断策略
2	reasoning_content 不应输出	完全移除 m.ReasoningContent 输出
3	System prompt 不应默认输出	case RoleSystem: 分支为空，完全跳过
4	cleanup-pending 过滤	isCleanupPending(path) 在 list_sessions 跳过、read_session 拒绝此类会话
5	max_turns 的 break 只退了 switch	用 loop: 标签 + break loop 正确退出外层 for
6	max_turns: 0 应该表示无限制	修复逻辑：*params.MaxTurns == 0 时 maxTurns = 0（无限制）
7	需要回归测试	go test ./internal/tool/builtin/ ./internal/boot/ 通过

SivanCola (changes requested):
Thanks for tightening the first round. The updated version fixes the `max_turns` break bug and stops rendering system prompts / `reasoning_content`, but I still cannot approve this head yet.

Blocking issues:

- `internal/tool/builtin/sessions.go` implements its own cleanup-pending marker as `sessionPath + ".cleanup-pending.json"`. That does not match the real session cleanup contract in `agent.CleanupPendingPath`, which trims `.jsonl` first and writes `session.cleanup-pending.json`. The new tests create the same wrong marker shape as the implementation, so they pass while real cleanup-pending sessions created through `agent.MarkCleanupPending` would still be listed/read. Please reuse the `agent` helper if possible, or at least match the exact existing marker contract and add a regression test using the real helper.

- `read_session` still renders normal tool results by default. The existing `history` tool intentionally excludes normal tool output unless requested because tool results often contain secrets, command output, environment data, or private file contents. Truncating each result to 2000 runes reduces size, but it does not fix the privacy boundary. Please either omit normal tool results by default, expose only tool errors / tool names, or add an explicit opt-in/gated mode with tests.

Verified locally while re-reviewing:

- `go test ./internal/tool/builtin ./internal/boot ./internal/history` passed
- `go vet ./internal/tool/builtin ./internal/boot ./internal/history` passed
- `gofmt -l` on the changed Go files produced no output
- `git diff --check origin/main-v2...HEAD` passed

Once the cleanup-pending guard uses the real contract and the tool-result exposure is narrowed, I can re-review again.

eghrhegpe:
这轮修了两个 reviewer blocking issue：

清理标记路径错位
之前：session.jsonl.cleanup-pending.json（追加 .jsonl 后面）

正确：session.cleanup-pending.json（替换 .jsonl 后缀）

// 修复前
os.Stat(sessionPath + cleanupPendingExt)

// 修复后 — 匹配 agent.CleanupPendingPath
marker := strings.TrimSuffix(sessionPath, ".jsonl") + cleanupPendingExt
工具结果默认不暴露
新增 show_tool_results 参数（默认 false），对齐 history 工具的默认行为：

数据	list_sessions	read_session（默认）	read_session（opt-in）
工具调用名/参数	✅ 始终显示	✅ 始终显示	✅
工具结果内容	—	❌ 仅标 Tool Result: name	✅ 2000 runes 截断

迭代了一次：
PR 从 757 行降到 656 行，减少了 101 行重复代码。

消除的重复：

之前手写，证实可行	现在复用，精简代码
loadSessionJSONL (20行)	agent.LoadSession	整个函数删除
countUserTurns 手读 JSONL (18行)	agent.LoadSession + 遍历	不重复解析器
isCleanupPending 错误标记路径 (6行)	agent.IsCleanupPending	完全匹配合同
os.ReadDir + 过滤 + 排序 (35行)	agent.ListSessionOrder	含 sidecar 元数据
测试中手写标记文件	agent.MarkCleanupPending	真实合同验证
测试中手写 JSONL	agent.NewSession + ses.Save()	标准化写入
cleanupPendingExt 常量	不再需要	无硬编码
加了一个 TestCleanupPendingContract 测试，直接用 agent.MarkCleanupPending 创建标记，验证 read_session 正确拒绝——不再是"自己写标记自己读"的假测试。

SivanCola (approved):
Approved. The two blockers from my previous review are addressed on this head.

What changed:
- `read_session` no longer exposes normal tool result content by default; tool output now requires the explicit `show_tool_results` opt-in, with coverage for the default redaction behavior.
- Cleanup-pending filtering now uses the real agent session contract (`agent.ListSessionOrder`, `agent.IsCleanupPending`, and tests built with `agent.MarkCleanupPending`) instead of duplicating the marker shape.
- I fixed the `read_session` JSON schema quoting issue and added a schema-validity regression so the provider-visible tool schema cannot silently become invalid again.

Verified locally:
- `go test ./internal/tool/sessiontool ./internal/boot ./internal/history` passed
- `go vet ./internal/tool/sessiontool ./internal/boot ./internal/history` passed
- `git diff --check origin/main-v2...HEAD` passed

GitHub CI has restarted for the latest head and should still be allowed to finish before merge.

PR #4889, opened 2026-06-20
perf(desktop): sidebar resilience - timeout, debounce, result cache

ttmouse:
## 问题

PR #4858 解决了重复扫描问题，但在以下场景仍有性能瓶颈：

- **云盘路径阻塞**：Synology Drive / iCloud 路径的 `os.ReadFile` 可能无限阻塞，导致整个 UI 卡死
- **重复构建**：前端连续触发 `emitProjectTreeChanged`，每次都完整重建 `ListProjectTree`
- **goroutine 卡死**：某个会话目录读取异常时 `wg.Wait()` 永不返回
- **冷启动延迟**：`loadTopicTitles` / `loadTopicCreatedAts` 对 15+ 个项目顺序读取，最慢可达 ~5s
- **话题遍历时反复重读文件**：`topicTitleForTab` 在遍历 611 个话题时为每个无标题话题重新读整个 JSON 文件

## 改动

共 7 个 commit，`desktop/tabs.go` + 少量辅助文件：

### 加载性能优化

| 优化 | 效果 |
|------|------|
| 项目话题文件**并发加载** | 15 个项目从串行变成并行，最慢 ~200ms（云盘超时）决定总时间 |
| 禁止话题遍历时**重新读文件** | 直接用已加载的 `titleMap`，`topicTitleForTab` 不再反复调 `loadTopicTitles` |
| **磁盘缓存**（stale-while-revalidate） | 冷启动时先返回缓存树 → 瞬间展示侧边栏 → 后台 goroutine 重建 |
| **移除 200ms 防抖** | `emitProjectTreeChanged` 立即通知前端，不再额外等待 |

### 防御性保护

| 保护 | 解决的问题 |
|------|-----------|
| `readFileWithTimeout` (200ms) | 防止云盘路径 `os.ReadFile` 无限阻塞 |
| `projectTreeResultCache` (2s TTL, 双检锁) | 短时间内多次调用直接返回缓存 |
| `listProjectTreeMu` | 串行化并发调用者 |
| `wg.Wait` 5s 超时 + `timedOut` flag | 防止 goroutine 卡死导致函数永不返回 |
| goroutine panic recovery | 隔离单个目录的 panic |

### 其他改进

- **单实例锁路径哈希**：`com.reasonix.desktop.<binary-path-sha256[:8]>`，dev 和 release 自动可多开
- **配置目录合并**：`desktopConfigDir` 指向 `~/.reasonix/`，与 sessions/projects 统一目录
- **详细 slog 日志**：每个阶段的耗时、缓存命中、迁移状态均可追踪

## 效果（15 个项目的用户实测）

| 场景 | 优化前 | 优化后 |
|------|--------|--------|
| 冷启动首次加载 | ~5s | ~300ms（首次）+ 后续瞬间（磁盘缓存） |
| 重复请求（2s 内） | 重复构建 | 内存缓存命中 |
| 云盘路径阻塞 | 无限挂起 | 200ms 超时自动降级 |

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Build artifacts
 /bin/
+/build/
 /dist/
 /stage/
 /npm/.stage/
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/rand"
+	"crypto/sha256"
 	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
@@ -52,10 +53,26 @@ import (
 // `data:` frames.
 const eventChannel = "agent:event"
 
-// singleInstanceID is used by Wails to route a second desktop launch back to the
-// running instance. Keep it stable across releases so launcher/Dock/taskbar
-// reopen behavior remains predictable on every platform.
-const singleInstanceID = "com.reasonix.desktop"
+// singleInstanceID derives a lock ID from the binary's absolute path so that
+// binaries at different paths (e.g. a dev build vs. a packaged release .app)
+// get distinct single-instance locks and can run simultaneously. The driver is
+// `com.reasonix.desktop.<SHA256-of-path[:8]>` — the static prefix keeps the
+// names recognisable in system logs.
+//
+// When REASONIX_DEV is set, the lock is skipped entirely (see single_instance.go)
+// so contributors can run any number of dev instances without copying.
+func singleInstanceID() string {
+	abs, err := os.Executable()
+	if err != nil {
+		return "com.reasonix.desktop"
+	}
+	abs, err = filepath.EvalSymlinks(abs)
+	if err != nil {
+		abs, _ = filepath.Abs(abs)
+	}
+	h := sha256.Sum256([]byte(abs))
+	return "com.reasonix.desktop." + hex.EncodeToString(h[:8])
+}
 
 // PromptHistoryEntry is one user prompt extracted from a session JSONL file.
 // The frontend uses these for ↑/↓ prompt-history navigation.
--- a/desktop/frontend/src/styles.css
+++ b/desktop/frontend/src/styles.css
@@ -18824,6 +18824,10 @@ a[href] {
   border-bottom-color: var(--border-soft);
 }
 
+:root[data-theme="light"] .topicbar {
+  background: #ffffff;
+}
+
 .app--darwin .app-chrome--tabs {
   padding-right: calc(var(--chrome-toggle-size) + var(--chrome-toggle-size) + 16px);
   background: color-mix(in srgb, var(--bg-soft) 90%, var(--bg-elev));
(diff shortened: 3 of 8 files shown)


SivanCola (changes requested):
Thanks for the performance work here. I think the direction is valuable, but I do not think this is safe to merge yet.

Blocking issues:

- `ListProjectTree` now caches the fully rendered `[]ProjectNode` globally for 2 seconds. That payload includes runtime fields (`Open`, `Running`, `Status`) and is not keyed by the active config/session roots. Calls inside the TTL can return stale project trees, stale runtime status, or even another isolated app/test environment's tree. This is already visible in CI: the `desktop` job fails in `TestOpenProjectTabResolvesProjectSessionFromLegacyDir`, and locally `go test ./...` under `desktop/` also fails with the same cache-hit pattern. `TestProjectTreeShowsBackgroundJobStatus` can also fail because the cached tree does not reflect updated runtime state. Please either cache only the expensive disk-derived inputs and recompute runtime fields each call, or key/invalidate the cache on every state root/runtime-status change.

- `desktopConfigDir()` now moves tabs/projects from the platform config dir to `~/.reasonix`, but `loadTabsFile` and `loadProjectsFile` only read the new location. Existing users with `desktop-tabs.json` / `desktop-projects.json` under the old config dir will start with an empty tab/sidebar state and may overwrite the new location before the old data is migrated. This needs a compatibility read/migration path before changing the write location.

- The new `readFileWithTimeout` returns after 200ms, but the underlying `os.ReadFile` goroutine is still left blocked forever when the filesystem call hangs. Repeated sidebar rebuilds against the same cloud/unmounted path can accumulate stuck goroutines. The caller no longer hangs, which is good, but the implementation still needs bounded worker lifetime or a different cancellable/stat-based strategy.

Also, the PR removes the `REASONIX_DEV` single-instance bypass while the new comment still says it exists. If that bypass is intentionally removed, please update the docs/tests and make sure contributor workflows that rely on same-binary multiple dev instances still have an equivalent path.

Verified:

- GitHub `desktop` check is failing on this PR.
- `go test ./...` from `desktop/` fails locally on the PR head.
- `go test -run TestOpenProjectTabResolvesProjectSessionFromLegacyDir -count=1 -v` passes in isolation, which supports the process-global cache/stale-state diagnosis rather than a pure test setup failure.
- `git diff --check origin/main-v2...origin/pr/4889` passes.

Security/cache note: I did not see new dependencies or network/file-exfiltration behavior. This does not appear to affect provider-visible prompt/cache prefixes, and the repository cache-impact check passed. The blocker is desktop state correctness and stale local data handling.

SivanCola:
Integrated into #4942, which combines the related desktop sidebar/session runtime fixes on top of the latest `main-v2`.

I kept the relevant contribution from this PR credited in the #4942 body and commit co-author trailers. Please use #4942 as the integration path for this set of changes.

SivanCola:
Closing this in favor of the merged integration PR #4942.

What was kept/adapted from this PR:
- bounded sidebar sidecar reads for read-only listing paths;
- serialized `ListProjectTree` rebuilds;
- concurrent project topic sidecar loading;
- fewer repeated topic-title sidecar reads;
- path-scoped single-instance IDs.

What was not adopted:
- the full `ProjectNode` result cache, because runtime fields must not be served from stale tree state;
- the config-dir move, because it needs an explicit migration plan.

@ttmouse is credited in #4942's contribution table and commit co-author trailers. Thank you for the investigation and performance work.

PR #4898, opened 2026-06-20
fix(agent): allow guidance-only executor handoffs

### LINKED ISSUE #4867: [Bug] Plan Execute dual-model: executor handoff nudge forces task invention on guidance-only turns
### Version line

### Exact version

1.9.1

### What happened?

When using the two-model Plan+Execute (PE) collaboration (`planner_model` set), guidance-only tasks trigger a forced nudge loop. The Planner outputs instructions for the user to follow manually (e.g. "download this file", "toggle the switch and listen"), the Executor relays them with zero tool calls — and the harness punishes this as "executor answered without taking any action".

The nudge injects a retry message as a user role turn, forcing the Executor to reply again. Since no plan steps remain, the Executor invents new tasks to satisfy the tool-call requirement, potentially reading/writing files the Planner never asked for.

This is not a theoretical concern — concrete sessions show it happening
(not 100% triggers)

### Steps to reproduce

1. Configure PE dual-model (`planner_model = "deepseek-pro"` in `[agent]`)
2. Ask a guidance-only question, e.g. "I just installed EqualizerAPO, now what?"
3. Planner outputs a step-by-step guide (download URL, click buttons, toggle switch)
4. Executor relays the guide — zero tool calls needed
5. ⚠ Harness emits: "executor answered without taking any action; nudging it to use its tools"
6. Executor receives retry message, invents a file-read or registry-check task to satisfy the tool requirement
7. User sees irrelevant file operations that contradict the conversation state (e.g. checking import status after user already confirmed success)

### OS / platform

Windows 11

### Relevant logs or output

```shell
## Source Code Evidence

**File:** `internal/agent/agent.go` (`main-v2` branch)

| Line          | Code                                                         | Role                          |
| ------------- | ------------------------------------------------------------ | ----------------------------- |
| **85**        | `const maxExecutorHandoffNudges = 1`                         | Nudge limit (intentionally 1) |
| **213-215**   | `executorHandoffGuard bool`                                  | Flag enabled by Coordinator   |
| **689**       | `handoffNudges := 0`                                         | Counter                       |
| **692**       | `executorHandoff := a.executorHandoffGuard && strings.Contains(input, executorHandoffMarker)` | Detection                     |
| **784**       | `if executorHandoff && !usedAnyTool && handoffNudges < maxExecutorHandoffNudges` | Nudge trigger                 |
| **785**       | `handoffNudges++`                                            | Increment                     |
| **786-789**   | Emit warning → push retry message → continue                 | Nudge injection               |
| **1104-1109** | `executorHandoffRetryMessage()` — the retry text             | Retry prompt                  |

The critical check at line 784:


if executorHandoff && !usedAnyTool && handoffNudges < maxExecutorHandoffNudges {
    handoffNudges++
    a.sink.Emit(event.Event{Kind: event.Notice, Level: event.LevelWarn,
        Text: "executor answered without taking any action; nudging it to use its tools"})
    a.session.Add(provider.Message{Role: provider.RoleUser,
        Content: a.withReasoningLanguage(executorHandoffRetryMessage())})
    a.maybeCompact(ctx, usage)
    continue
}


`usedAnyTool` is set to `true` **only inside the tool-execution branch** (line 804), after the model has already made tool calls. If the model produces text with zero tools, `usedAnyTool` stays `false` throughout the entire response, and the nudge fires.

---

## Session Log Evidence (real user session, Windows)

From a real session (HD650 EqualizerAPO setup conversation):

**Trigger 1:** User says "装了" (installed it). Planner asks "Did you check the Peace GUI checkbox?"

- Nudge fires → Executor ignores Planner's instruction and runs registry queries + path checks (3 bash calls)
- None of these were in the plan

**Trigger 2:** User says "开了" (it's on / toggled). Planner says "Go listen to a song and compare the difference."

- Nudge fires → Executor invents "Let me verify your import" → 4 bash calls reading config files and listing directories
- All tools return nothing useful; user had already confirmed the switch was working

**Trigger 3 (different session):** User asks for a TL;DR of update notes

- Executor relays summary correctly, then voluntarily offers "Should I run the upgrade command for you?" — already stepping outside the plan
- Nudge fires → Executor reads executable version info from disk
- Completely invented task, not in any plan

---

## Impact

| Severity   | Scenario                      | Why                                                          |
| ---------- | ----------------------------- | ------------------------------------------------------------ |
| **Medium** | Guidance-only tasks (PE mode) | Repeated nudge waste; Executor invents tasks outside the plan |
| **Low**    | Code tasks (PE mode)          | Never triggers — first tool call happens naturally           |
| **None**   | Single-model mode             | No handoff protocol, no nudge                                |

**Worst-case risk:** When nudge fires after all plan steps are complete, the Executor (a Flash model chosen for execution, not judgment) is forced to invent tasks. This could lead to:

- Unnecessary file writes/edits
- Contradictory or confusing user interactions
- Expensive wrong-direction work before the user can intervene

---

## Suggested Fix

The root cause is that the handoff protocol assumes all Executor tasks involve file manipulation, but the PE executor may also relay guidance-only instructions.

### Option A: Exempt handoff from the tool-use requirement

Treat the executor handoff as an allowed text-only turn type. Add `executorHandoff` to the "final answer" exit path:


// At the final-answer gate:
if executorHandoff {
    // Allow text-only responses — the executor's job may be pure relay
    break // exit normally
}


### Option B: Configurable nudge behavior (least invasive)

Add an `[agent]` config option like `handoff_nudge = false` to let users opt out when they know they're doing guidance-only work.

### Option C: Heuristic check in the nudge gate

Skip nudge when the handoff input contains no file paths, no code blocks, and no tool-specific commands — i.e. it's clearly guidance-only.

---

## Environment

- Reasonix v1.9.1 (agent code confirmed on `main-v2`)
- PE mode: Planner = `deepseek-v4-pro`, Executor = `deepseek-v4-flash`
- Desktop Windows 11
```

GTC2080:
## Summary

- Fixes #4867 by allowing PE executor handoffs to finish with a text-only answer when the original turn or planner output is clearly user-facing guidance, a summary, or a follow-up question.
- Keeps the existing handoff nudge for real work tasks and deferral-style answers, so code/file changes still require executor tool calls.
- Updates the executor handoff instructions to avoid invented local tool calls for pure guidance-only turns.

## Verification

- `go test ./internal/agent -run 'TestCoordinator(AllowsGuidanceOnlyExecutorHandoff|NudgesExecutorThatAnswersWithoutActing|NudgesWorkTaskEvenIfPlannerMentionsUserGuidance|DoesNotNudgeExecutorThatActs)' -count=1`
- `go test ./internal/agent -count=1`
- `go test ./...`
- `(cd desktop && go test ./...)` (passes; local linker prints the existing duplicate `-lobjc` warning)
- `go vet ./...`
- `git diff --check`
- `gofmt -l .`

## Cache impact

Cache-impact: low - touches the executor handoff gate and one handoff user-turn instruction; system prompt and tool schemas are unchanged.
Cache-guard: Added focused coordinator regressions for guidance-only handoffs and for preserving the work-task nudge; also ran `go test ./internal/agent -count=1` and `go test ./...`.
System-prompt-review: N/A

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -781,7 +781,7 @@ func (a *Agent) Run(ctx context.Context, input string) error {
 				a.maybeCompact(ctx, usage)
 				continue
 			}
-			if executorHandoff && !usedAnyTool && handoffNudges < maxExecutorHandoffNudges {
+			if executorHandoff && !usedAnyTool && handoffNudges < maxExecutorHandoffNudges && shouldNudgeExecutorHandoff(input, text) {
 				handoffNudges++
 				a.sink.Emit(event.Event{Kind: event.Notice, Level: event.LevelWarn, Text: "executor answered without taking any action; nudging it to use its tools"})
 				a.session.Add(provider.Message{Role: provider.RoleUser, Content: a.withReasoningLanguage(executorHandoffRetryMessage())})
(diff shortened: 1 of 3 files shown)


SivanCola:
Thanks @GTC2080. I consolidated this guidance-only executor handoff fix into #4938.

What was kept:
- guidance-only, summary, question, and manual-instruction PE handoffs can now finish with a text-only executor answer;
- real work tasks still keep the missing-action nudge so the executor cannot just restate the plan;
- regression coverage preserves both sides of that behavior.

Your investigation and implementation are credited in #4938 under Repository Contributor Credits. Closing this PR in favor of the integration PR.

PR #4923, opened 2026-06-20
fix(tui): forceGotoBottom scrolls regardless of transcript length change (#4797)

### LINKED ISSUE #4797: [Bug]: 多个会话的时候切换回去并不会到最新回答，而是中间某个位置
### Version line

### Exact version

v1.9.1

### What happened?

进行多个会话的时候切换回去一个对话的时候并不会到最新回答的位置，而是中间某个位置，需要人为下滑

### Steps to reproduce

进行多个会话的时候切换回去一个对话的时候并不会到最新回答的位置，而是中间某个位置，需要人为下滑

### OS / platform

window11

### Relevant logs or output

```shell

```

liaoyl830:
## Summary

Fixes #4797

When switching sessions, `replayActiveBranch()` sets `forceGotoBottom=true` to pin the viewport to the latest message. However, `GotoBottom()` was guarded inside a conditional that only fires when the transcript length changes. If the new session had the same number of lines, the viewport stayed at the intermediate scroll position.

## Root Cause

In `internal/cli/chat_tui.go`'s `Update()` wrapper, the scroll-to-bottom logic (`if wasAtBottom || cm.forceGotoBottom`) was inside the `if len(cm.transcript) != prevLines || ...` block. When session switch produced a transcript of equal length, the block was skipped entirely and `GotoBottom()` never ran.

## Fix

Move the `forceGotoBottom` check outside the transcript-change conditional so it always triggers scroll-to-bottom, regardless of content length.

## Tests

- **Added** `TestForceGotoBottomScrollsToBottom` — verifies viewport scrolls to bottom when `forceGotoBottom` is set, even with unchanged transcript length

### DIFF
--- a/internal/cli/chat_tui.go
+++ b/internal/cli/chat_tui.go
@@ -689,10 +689,13 @@ func (m chatTUI) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		wrapped := wrapTranscript(strings.Join(cm.transcript, "\n"), contentW)
 		cm.viewport.SetContent(wrapped)
 		cm.wrappedLines = strings.Split(wrapped, "\n")
-		if wasAtBottom || cm.forceGotoBottom {
+		if wasAtBottom {
 			cm.viewport.GotoBottom() // tail-follow: stay pinned to newest output
-			cm.forceGotoBottom = false
 		}
+	}
+	if cm.forceGotoBottom {
+		cm.viewport.GotoBottom()
+		cm.forceGotoBottom = false
 	}
 	cm.transcriptDirty = false
 	// Any viewport scroll (wheel, PgUp/PgDn, edge auto-scroll, or tail-follow to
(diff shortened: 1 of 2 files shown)


SivanCola:
Thanks @liaoyl830 for isolating the TUI `forceGotoBottom` edge case.

Your PR contributed the TUI fix that has now been consolidated into #5029:
- The root cause that `forceGotoBottom` was incorrectly gated behind transcript length / width / dirty-state changes.
- The fix direction of moving the force-bottom check outside that content-change branch.
- A regression-test shape for same-length transcript/session switches.

#5029 includes this TUI behavior, adds an explicit regression test for force-bottom without transcript changes, and combines it with the desktop transcript bottom-pinning work. I listed your contribution in the #5029 PR body.

Closing this PR as superseded by #5029 so the desktop and TUI scroll fixes can land together.

PR #4924, opened 2026-06-20
fix(agent): check ctx cancellation between tool executions (#4860)

### LINKED ISSUE #4860: [Bug]: 常常出现卡死不能停止的情况
### Version line

### Exact version

1.9 cli

### What happened?

运行复杂任务较久以后，就出现不能esc取消和ctrl+c退出，但shift+tab和ctrl+y有反应

### Steps to reproduce

n/a

### OS / platform

ubuntu 26.04

### Relevant logs or output

```shell

```

liaoyl830:
## Summary

Fixes #4860

When the user presses Escape/Ctrl+C during a running turn, `Cancel()` cancels the context. However, the agent's `executeBatch()` continued executing remaining tools in the batch even after one tool detected `ctx.Done()`. The TUI appeared frozen until all tools finished.

## Root Cause

In `internal/agent/agent.go`, the `executeBatch()` function ran all tools in a serial batch without checking `ctx.Err()` between executions. After `Cancel()` fired, the first tool would detect `ctx.Done()` and return, but the loop continued to start the next tool.

## Fix

- Add `ctx.Err()` checks **before** and **after** each tool execution in serial batches
- When context is cancelled, stop executing remaining tools immediately and fill their results with cancellation markers
- Add `ctx.Err()` check in the main `Run()` loop after `executeBatch()` to exit promptly

## Tests

- **Added** `TestCancelDuringToolExecutionBreaksOutPromptly` — verifies agent exits ~300ms after cancel instead of waiting for all tools (~4s)
- **Added** `TestCancelDuringBatchStopsRemainingTools` — verifies cancelled tools don't start subsequent tools in the batch

## Cache Metadata

Cache-impact: low — changes are confined to the agent's tool execution loop. No system prompt, tool schema, or provider surface modifications. The prompt prefix is unaffected.
Cache-guard: `go test ./internal/agent/ -run TestCancel -v` covers the cancel-during-execution path. Existing `TestRunCancelledMidStreamLeavesResumableSession` in loop_e2e_test.go covers the broader cancel flow.
System-prompt-review: No system prompt paths touched. Changes are internal to the agent execution loop only.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -1297,14 +1297,32 @@ func (a *Agent) executeBatch(ctx context.Context, calls []provider.ToolCall) []s
 		durations[i] = time.Since(start).Milliseconds()
 		results[i] = outcomes[i].output
 	}
+	cancelled := false
+	markCancelled := func(start int) {
+		errMsg := context.Canceled.Error()
+		if err := ctx.Err(); err != nil {
+			errMsg = err.Error()
+		}
+		output := "cancelled: context cancelled before execution"
+		for j := start; j < len(calls); j++ {
+			results[j] = output
+			outcomes[j] = toolOutcome{output: output, errMsg: errMsg}
+		}
+		cancelled = true
+	}
 
 	for _, batch := range partitionToolCalls(a.tools, calls) {
+		if ctx.Err() != nil {
+			markCancelled(batch.start)
+			break
+		}
 		if batch.parallel && batch.end-batch.start > 1 {
 			runParallel(batch.start, batch.end, run)
 			// After parallel execution completes, check if context was cancelled.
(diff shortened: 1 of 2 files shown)


SivanCola:
Superseded by the merged integration PR #4975.

What was kept/strengthened from this PR:
- The agent-level `executeBatch()` cancellation checks.
- Prompt exit after cancellation is observed during tool execution.
- Regression coverage for cancelling during tool execution.

The final integration also keeps the follow-up fix that prevents later batches, including read-only parallel batches, from starting after cancellation, and marks unstarted tool results as cancelled so the UI does not show empty success-looking cards. @liaoyl830 and @SivanCola are credited in #4975 as repository contributors for this work.

Closing this PR because the integrated fix has landed in #4975.

Issue #4926, opened 2026-06-20
[Bug]: Plan-Execute dual-model: per-project handoff buffer not isolated by topicId, causes cross-session plan contamination


halrixx:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.10.0

### What happened?

When using the two-model Plan+Execute (PE) collaboration (`planner_model` set), multiple dual-model tabs under the **same project** can experience cross-session handoff contamination. The Planner output from Tab A is routed to the Executor in Tab B, causing Tab B to receive a task plan that is completely unrelated to its conversation.

This happens because the handoff buffer (or equivalent state object) is maintained at the **project level**, not the **topic/session level**. Since all tabs under the same project share the same workspace root and session storage directory, the Planner→Executor handoff path lacks topicId isolation.

**Cross-project contamination does not occur** because each project has independent workspace paths and session directories.

### Steps to reproduce

1. Configure PE dual-model (`planner_model = "deepseek-pro"` in `[agent]`)
2. Open **two or more tabs** in the Reasonix Desktop under the **same project**
3. In Tab A, start a dual-model conversation about Topic X (e.g. "fix an npm bug")
4. In Tab B, start a dual-model conversation about Topic Y (e.g. "write a GitHub issue")
5. Tab A's Planner completes → writes plan to per-project handoff path
6. Switch focus or let both run:
   - ❌ Tab B's Executor may read Tab A's plan instead of its own
   - ❌ User sees an unrelated plan text injected into the wrong conversation
7. Switch back to Tab A:
   - ❌ Tab A's Executor may have consumed the wrong plan (or Tab B's)
   - ❌ Tab A's correct plan is lost or delivered to Tab B

This is **probabilistic** — it depends on the timing of Planner completions and tab switches.

### OS / platform

Windows 11

### Relevant logs or output

```shell
## Source Code Evidence

**File:** `internal/agent/coordinator.go` (`main-v2` branch)

### The handoff path (lines 166–171)


func (c *Coordinator) Run(ctx context.Context, input string) error {
    // ...
    plan, err := c.plan(ctx, input)
    // ...
    return c.executor.Run(ctx, formatHandoff(input, plan))
}


`formatHandoff()` (lines 117–149) wraps the task + plan into a string. The Coordinator's `Run()` passes it directly to the executor. There is **no topicId, sessionId, or tabId scoping** — this is a pure function call on a local Coordinator instance.

[...]

PR #4930, opened 2026-06-20
test(desktop): cover PE tab handoff isolation

GTC2080:
## Summary

- Adds a desktop regression test for #4926 that builds two Plan+Execute tabs under the same project from the real `boot.Build` path.
- The test blocks both planners until they are concurrently active, then verifies each executor handoff contains only its own task and planner output.
- Current `main-v2` already keeps the runtime path isolated; no production change is included because I could not reproduce a project-level handoff buffer leak on the latest base.

Refs #4926

## Verification

- `cd desktop && go test ./... -run TestPlanExecuteTabsUnderSameProjectKeepPlannerHandoffsIsolated -count=2`
- `go test ./internal/control -run 'Test(NewSession|Resume)ResetsTwoModelPlannerContext' -count=1`
- `go test ./internal/agent ./internal/control -count=1`
- `cd desktop && go test ./... -count=1`
- `go test ./... -count=1`
- `go vet ./...`

## Cache impact

Cache-impact: none — test-only change; no system prompt, tool schema, provider request serialization, or runtime prompt construction changes.
Cache-guard: `cd desktop && go test ./... -run TestPlanExecuteTabsUnderSameProjectKeepPlannerHandoffsIsolated -count=2`
System-prompt-review: N/A

### DIFF
--- a/desktop/tabs_pe_isolation_test.go
+++ b/desktop/tabs_pe_isolation_test.go
@@ -1,0 +1,208 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	"reasonix/internal/provider"
+)
+
+type peIsolationHarness struct {
+	mu               sync.Mutex
+	plannerStarted   int
+	bothPlannersSeen chan struct{}
+	releasePlanners  chan struct{}
+	executorInputs   []string
+}
+
+type peIsolationProvider struct {
+	name    string
+	harness *peIsolationHarness
+}
+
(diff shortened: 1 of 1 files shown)


SivanCola:
Thanks @GTC2080. I reviewed this PE tab handoff isolation test PR while consolidating the related PE work into #4938.

This remains a useful testing direction for #4926, but the current integration PR is focused on executor handoff nudging, no-op planner conclusions, MCP/tool handoff context, and planner routing. Since #4930 is not part of that final code path, I credited it in #4938 and am closing this PR as part of the consolidation cleanup.

If #4926 remains reproducible on current main-v2, a fresh isolated follow-up test/fix can reopen that thread directly.

PR #4934, opened 2026-06-20
fix(agent): reset planner session on tab switch to prevent cross-session contamination

### LINKED ISSUE #4926: [Bug]: Plan-Execute dual-model: per-project handoff buffer not isolated by topicId, causes cross-session plan contamination

liaoyl830:
## Summary

修复 #4926: Plan-Execute dual-model 跨会话计划污染

**问题：** 当使用双模型 Plan+Execute 模式时，同一项目下的多个标签页会发生跨会话污染。Tab A 的 Planner 输出可能被路由到 Tab B 的 Executor。

**根因：** 切换标签页时没有重置 planner session，导致 Coordinator 的 plannerSess 中的旧状态污染新的会话。

**修复：** 将 `Controller.ResetPlannerSession()` 导出为公开方法，并在桌面层的每个标签页激活路径中调用：
- `SetActiveTab`: 用户在标签页间切换
- `openTopicTabWithActivation`: 复用现有标签页
- `CloseTab`: 关闭标签页后切换到下一个

## Changes

- `internal/control/controller.go` - 导出 `ResetPlannerSession()` 方法
- `desktop/tabs.go` - 在4个标签页激活点调用 `ResetPlannerSession()`
- `internal/control/controller_test.go` - 添加测试验证

## Verification

- `go test ./internal/agent/...` passes
- `go test ./internal/control/...` passes (包括新测试)
- `go vet ./...` clean
- `gofmt -w .` clean

Fixes #4926

## Cache impact

Cache-impact: none — 仅添加了方法调用，不改变系统提示或缓存前缀。
Cache-guard: N/A

### DIFF
--- a/desktop/tabs.go
+++ b/desktop/tabs.go
@@ -1002,7 +1002,13 @@ func (a *App) openTopicTabWithActivation(scope, workspaceRoot, topicID, sessionP
 			}
 			if sessionRuntimeKey(tab.currentSessionPath()) == targetKey {
 				if activate {
+					wasActive := a.activeTabID == tab.ID
 					a.activeTabID = tab.ID
+					// Reset planner session when switching to a different tab to
+					// prevent cross-session contamination in dual-model mode (#4926).
+					if !wasActive && tab.Ctrl != nil {
+						tab.Ctrl.ResetPlannerSession()
+					}
 				}
 				meta := a.tabMeta(tab, tab.ID == a.activeTabID)
 				a.saveTabsLocked()
@@ -1014,6 +1020,7 @@ func (a *App) openTopicTabWithActivation(scope, workspaceRoot, topicID, sessionP
 
 	for _, tab := range a.tabs {
 		if tabMatchesTopicTarget(tab, scope, workspaceRoot, topicID) {
+			wasActive := a.activeTabID == tab.ID
 			if activate {
 				a.activeTabID = tab.ID
 			}
@@ -1022,6 +1029,11 @@ func (a *App) openTopicTabWithActivation(scope, workspaceRoot, topicID, sessionP
 			a.saveTabsLocked()
 			a.mu.Unlock()
 			if sameSession {
+				// Reset planner session when switching to a different tab to
+				// prevent cross-session contamination in dual-model mode (#4926).
+				if activate && !wasActive && tab.Ctrl != nil {
+					tab.Ctrl.ResetPlannerSession()
+				}
 				return enrichTabMeta(meta), nil
 			}
 			if err := a.rebindTabToSessionPath(tab, sessionPath); err != nil {
@@ -1362,8 +1374,18 @@ func (a *App) SetActiveTab(tabID string) error {
 		return nil
 	}
 	a.activeTabID = tabID
+	tab := a.tabs[tabID]
 	dir, entries, activeID, version := a.saveTabsCollectLocked()
 	a.mu.Unlock()
+
+	// Reset the planner session when switching tabs to prevent cross-session
+	// contamination in dual-model (Plan+Execute) mode. Each tab's planner
+	// history should be scoped to that tab's conversation; without this reset,
+	// stale planner output from a previously active tab could leak into the
+	// newly active tab's executor handoff (#4926).
+	if tab != nil && tab.Ctrl != nil {
+		tab.Ctrl.ResetPlannerSession()
+	}
 
 	// I/O outside the lock — disk writes can block for hundreds of ms on
 	// Windows when antivirus or the search indexer briefly locks the file.
@@ -1432,6 +1454,11 @@ func (a *App) CloseTab(tabID string) error {
 				nextIndex = len(a.tabOrder) - 1
 			}
 			a.activeTabID = a.tabOrder[nextIndex]
+			// Reset planner session on the newly active tab to prevent
+			// cross-session contamination in dual-model mode (#4926).
+			if nextTab := a.tabs[a.activeTabID]; nextTab != nil && nextTab.Ctrl != nil {
+				nextTab.Ctrl.ResetPlannerSession()
+			}
 		}
 	}
 	a.saveTabsLocked()
--- a/internal/control/controller.go
+++ b/internal/control/controller.go
@@ -1990,7 +1990,7 @@ func (c *Controller) NewSession() error {
 	}
 	c.setActiveJobSession(c.SessionPath())
 	c.executor.SetSession(agent.NewSession(c.systemPrompt))
-	c.resetPlannerSession()
+	c.ResetPlannerSession()
 	c.rebindCheckpoints(c.SessionPath())
 	c.mu.Lock()
 	c.startedOnce = true // NewSession fires SessionStart itself; don't re-fire on the next turn
@@ -2034,7 +2034,7 @@ func (c *Controller) ClearSession() error {
 	}
 	c.setActiveJobSession(c.SessionPath())
 	c.executor.SetSession(agent.NewSession(c.systemPrompt))
-	c.resetPlannerSession()
+	c.ResetPlannerSession()
 	c.rebindCheckpoints(c.SessionPath())
 	c.mu.Lock()
 	c.startedOnce = true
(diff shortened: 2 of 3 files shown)


SivanCola:
Thanks @liaoyl830. I reviewed this tab-switch reset direction while consolidating the PE handoff work into #4938.

I did not adopt the production reset behavior because latest main-v2 passes the planner output directly through the active Coordinator rather than through a project-level handoff buffer, and resetting planner state during tab activation could race with in-flight planning. The #4926 isolation concern should be handled separately if it is still reproducible on the current base.

Your investigation and proposed fix are credited in #4938 under Repository Contributor Credits. Closing this PR in favor of the current integration path.

SivanCola:
Integrated into #4942, which combines the related desktop sidebar/session runtime fixes on top of the latest `main-v2`.

I kept the relevant contribution from this PR credited in the #4942 body and commit co-author trailers. Please use #4942 as the integration path for this set of changes.

PR #4951, opened 2026-06-20
Support codebase-memory MCP auto-indexing / 支持 codebase-memory MCP 自动索引

SivanCola:
## Summary
- Pin codebase-memory-mcp stdio servers to the active workspace root so its session detection sees the opened project.
- Let known session indexers opt into shared-host background startup, allowing codebase-memory-mcp initialize-time auto-indexing to run when a desktop tab opens.
- Keep the behavior scoped to codebase-memory-mcp so existing CodeGraph and generic MCP startup behavior stays unchanged.

## Cache Review
Cache-impact: low - touches MCP/boot runtime wiring but does not change prompt text, tool schemas, or schema ordering; the new startup opt-in is scoped to codebase-memory-mcp.
Cache-guard: `scripts/check-cache-impact.sh` plus `go test ./internal/plugin ./internal/boot`.
System-prompt-review: Codex reviewed the boot/plugin diff under the Reasonix cache gate; no system prompt bytes, memory prefix, or provider-visible tool schema content changed.

## Verification
- `go test ./internal/plugin ./internal/boot`
- `scripts/check-cache-impact.sh`
- `git diff --check`

### DIFF
--- a/internal/boot/boot.go
+++ b/internal/boot/boot.go
@@ -415,9 +415,11 @@ func Build(ctx context.Context, opts Options) (*control.Controller, error) {
 				// subprocess only starts on the first actual tool call.
 				// This avoids spawning MCP processes (e.g. CodeGraph) for
 				// every workspace root on boot — a tab that sits unused for
-				// days never pays the startup cost.
+				// days never pays the startup cost. Known session indexers can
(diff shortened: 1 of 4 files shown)


SivanCola:
@codex review

PR #4963, opened 2026-06-21
fix(desktop): set provider kind default to "openai" for new custom providers

xinyiming:
## Summary
When adding a new custom provider in the desktop Settings panel, the protocol dropdown shows "OpenAI-compatible" but the saved `kind` value is `"anthropic"` instead of `"openai"`.
## Root cause
`provider.Kinds()` returns registered provider kinds sorted alphabetically: `["anthropic", "openai"]`. The `ProviderEditor` component used `kinds[0]` as the default `kind` value for new providers, which resolved to `"anthropic"`. The UI displayed a static "OpenAI-compatible" label that was not bound to the actual `kind` state, so the mismatch was invisible to the user.
## Changes
Three changes in `desktop/frontend/src/components/SettingsPanel.tsx`:
1. **kind initial value** — changed from `initial?.kind ?? kinds[0] ?? "openai"` to `initial?.kind ?? "openai"`
2. **save fallback** — changed from `kind.trim() || kinds[0] || "openai"` to `kind.trim() || "openai"`
3. **new provider UI** — replaced the static `<div>` label with a `<select disabled value="openai">` that explicitly binds the `kind` value, making it visible in the UI
## Verification
- Vite dev server started successfully, UI renders correctly
- No TypeScript errors related to the change
Base 分支： main-v2  
Head 分支： fix/provider-kind-default-openai

### DIFF
--- a/desktop/frontend/src/components/SettingsPanel.tsx
+++ b/desktop/frontend/src/components/SettingsPanel.tsx
@@ -4252,7 +4252,7 @@ function ProviderEditor({
 }) {
   const t = useT();
   const [name, setName] = useState(initial?.name ?? "");
-  const [kind, setKind] = useState(initial?.kind ?? kinds[0] ?? "openai");
+  const [kind, setKind] = useState(initial?.kind ?? "openai");
   const [baseUrl, setBaseUrl] = useState(initial?.baseUrl ?? "");
   const [models, setModels] = useState((initial?.models ?? []).join(", "));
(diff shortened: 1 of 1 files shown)


SivanCola:
Thanks @xinyiming for identifying the same custom-provider default-kind issue and proposing the OpenAI default fix.

We are moving forward with #5641, which carries the same fix direction on the current `main-v2` base and adds regression coverage. Your contribution is credited in #5641's PR body, so I am closing this PR as superseded by #5641.

PR #4999, opened 2026-06-21
fix(desktop): add copy+remove fallback for session trash move on Windows

### LINKED ISSUE #4956: [Bug]: 关于会话回收的新问题
### Version line

### Exact version

reasonix --1.10.0

### What happened?

对于生成的新会话且没有进行对话时，此会话可以删除。但是对于已有对话或者进行过对话的会话无法进行回收，会进行报错

Image

### Steps to reproduce

1.随便选择一个空文件夹创建工作区
2.在工作区创建一个会话并进行对话
3.点击移至回收站，然后点击确认移至回收站
4.报错

1.对已有且进行过对话的会话点击进行移至回收站，然后点击确认移至回收站
2.报错

### OS / platform

win11

### Relevant logs or output

```shell

```

liaoyl830:
## Summary

修复 #4956: 会话回收报错

**问题：** 用户尝试删除有对话历史的会话时，在 Windows 上可能会失败，报错信息为文件被占用。新会话（无对话）可以正常删除。

**根因：** `movePathIfExists` 函数使用 `os.Rename` 移动会话目录到回收站。在 Windows 上，如果文件句柄被短暂占用（如防病毒软件扫描、索引服务、或刚关闭的文件），`os.Rename` 会失败。

**修复：** 添加回退机制：
1. 首先尝试 `os.Rename`（快速、原子操作）
2. 如果失败且错误是跨设备（EXDEV）或文件被占用（Windows ERROR_SHARING_VIOLATION），则回退到复制+删除
3. 复制+删除：递归复制目录树，然后删除源
4. 在 Windows 上，关闭文件句柄后等待 10ms 确保文件完全释放

## Changes

- `desktop/sessions.go` — 添加 `copyAndRemove`, `copyDir`, `copyFile`, `isRenameCrossDeviceOrBusy` 函数
- `desktop/sessions_test.go` — 添加 `TestMovePathIfExistsCopyFallback`, `TestCopyAndRemoveDirectory` 测试

## Verification

- `go test ./desktop/...` passes
- `go vet ./...` clean
- `gofmt -w` applied

Fixes #4956

## Cache impact

Cache-impact: none — 仅修改桌面端会话回收逻辑，不影响系统提示或缓存前缀。
Cache-guard: N/A

### DIFF
--- a/desktop/sessions.go
+++ b/desktop/sessions.go
@@ -344,7 +344,7 @@ func purgeTrashedSessionFile(dir, path string) error {
 }
 
 func movePathIfExists(src, dst string) error {
-	if _, err := os.Stat(src); os.IsNotExist(err) {
+	if _, err := os.Lstat(src); os.IsNotExist(err) {
 		return nil
 	} else if err != nil {
 		return err
@@ -386,18 +386,34 @@ func isRenameCrossDeviceOrBusy(err error) bool {
 // copyAndRemove recursively copies src to dst, then removes src. Used as a
 // fallback when os.Rename fails (cross-device or Windows file-lock races).
 func copyAndRemove(src, dst string) error {
-	info, err := os.Stat(src)
+	if err := copyPath(src, dst); err != nil {
+		return err
+	}
+	// On Windows, wait briefly for any file handle release.
+	time.Sleep(10 * time.Millisecond)
+	return os.RemoveAll(src)
+}
+
+func copyPath(src, dst string) error {
+	info, err := os.Lstat(src)
 	if err != nil {
 		return err
 	}
-	if info.IsDir() {
-		return copyDir(src, dst)
+	mode := info.Mode()
+	switch {
+	case mode&os.ModeSymlink != 0:
+		return copySymlink(src, dst)
+	case mode.IsDir():
+		return copyDir(src, dst, mode.Perm())
+	case mode.IsRegular():
+		return copyFile(src, dst, mode.Perm())
+	default:
+		return fmt.Errorf("unsupported file type in rename fallback: %s", src)
 	}
-	return copyFile(src, dst, info.Mode())
 }
 
-func copyDir(src, dst string) error {
-	if err := os.MkdirAll(dst, 0o755); err != nil {
+func copyDir(src, dst string, mode os.FileMode) error {
+	if err := os.MkdirAll(dst, mode); err != nil {
 		return err
 	}
 	entries, err := os.ReadDir(src)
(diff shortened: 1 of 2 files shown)


desktop/sessions.go
```diff
@@ -350,7 +352,104 @@ func movePathIfExists(src, dst string) error {
...
+	}
+	for _, e := range entries {
+		srcPath := filepath.Join(src, e.Name())
+		dstPath := filepath.Join(dst, e.Name())
+		if e.IsDir() {
+			if err := copyDir(srcPath, dstPath); err != nil {
+				return err
+			}
+		} else {
+			info, err := e.Info()
```
SivanCola: This fallback currently changes `os.Rename` semantics for symlinks. `DirEntry.Info()` follows the link and `copyFile` then opens the target, so a session sidecar directory containing a symlink to an outside file is copied into `.trash` as a regular file when the fallback runs. A rename would have moved the symlink itself without reading the target.

Please switch the recursive copy to use `Lstat`/`Readlink` and recreate symlinks, or explicitly reject symlinks before copying, and add a regression test for a symlink inside a copied artifact directory.

SivanCola (changes requested):
Thanks, this is directionally useful for #4956, but one fallback semantic needs to be fixed before merge.

The new copy+remove path should preserve the behavior of `os.Rename`: it must move directory entries without dereferencing symlinks. The current implementation follows symlinks during fallback and copies the target contents into the trash destination, which can leak content from outside the session artifact tree and changes restore/purge semantics.

Verified:
- Existing focused session/trash tests passed with `go test . -run 'Test(MovePathIfExistsCopyFallback|CopyAndRemoveDirectory|DeleteSessionFile|ReconcileDesktopCleanupPending)'` from the desktop module.
- A temporary symlink regression test fails on this branch because the fallback creates a regular file instead of preserving the symlink.
- GitHub CI is currently green.

Once the symlink handling is fixed and CI stays green, I will consider merging it into `main-v2`.

desktop/sessions.go
SivanCola: Fixed in d3c5f67.

What changed:
- The copy+remove fallback now uses `Lstat`-based dispatch and recreates symlinks with `Readlink`/`Symlink`, so it preserves rename semantics instead of dereferencing symlink targets.
- Added `TestCopyAndRemoveDirectoryPreservesSymlinks` to cover a symlink inside a copied artifact directory.

Root cause:
- `DirEntry.Info()` followed symlinks, and `copyFile` then opened the target path.

Verified:
- `go test . -run 'Test(MovePathIfExistsCopyFallback|CopyAndRemoveDirectory|DeleteSessionFile|ReconcileDesktopCleanupPending)'` passed from the desktop module.
- `git diff --check` passed.
- Pushed to the PR head branch; GitHub CI has started.

SivanCola (approved):
Approved after the follow-up fix in d3c5f67.

The copy+remove fallback now preserves symlink directory entries instead of dereferencing targets, and the new regression test covers symlinks inside copied artifact directories.

Verified:
- `go test . -run 'Test(MovePathIfExistsCopyFallback|CopyAndRemoveDirectory|DeleteSessionFile|ReconcileDesktopCleanupPending)'` passed from the desktop module.
- `git diff --check` passed.
- GitHub CI is green, including desktop, race, Windows test, and cache-impact.

PR #5046, opened 2026-06-22
fix(security): constant-time webhook token, remove QQ_SECRET bypass, add HTTP timeouts

SEPURI-SAI-KRISHNA:
## Summary

  - **fix(feishu):** replace `==` with `crypto/subtle.ConstantTimeCompare` in
    `verificationTokenValid`  plain string equality leaks timing information that
    allows an attacker to brute-force the webhook verification token byte-by-byte.
    The empty-token branch is also changed from fail-open (`return true`) to
    fail-closed (`return false`); webhook mode already refuses to start without a
    token, so the open branch was unreachable dead code.
  - **fix(qq):** remove the undocumented `QQ_SECRET` env-var fallback in
    `appSecret()`. The function previously fell back to a hardcoded variable name
    that bypassed the user-configured `app_secret_env` key entirely. Only the
    configured key is now consulted.
  - **fix(qq, weixin):** replace `http.DefaultClient` (no timeout) with
    package-level `http.Client` values (`qqHTTPClient`, `weixinHTTPClient`) each
    carrying a 30 s `Timeout`. A stalled QQ or WeChat API previously blocked the
    calling goroutine indefinitely, causing goroutine leaks in long-running bot
    deployments.

  ## Verification

  - `go test ./internal/bot/...` all 65 tests pass across `bot`, `bot/feishu`,
    `bot/qq`, and `bot/weixin`.
  - `TestVerificationTokenValidRequiresConfiguredToken` updated: the last assertion
    now checks that an unconfigured token **denies** all callers (fail-closed),
    replacing the old assertion that it allowed any caller (fail-open).

  ## Cache impact

  Cache-impact: none no prompt prefix, tool schema, or provider wire format touched; changes are confined to bot HTTP plumbing and a security primitive swap.
  Cache-guard: N/A no cache-sensitive files changed (confirmed by scripts/check-cache-impact.sh).
  System-prompt-review: N/A

### DIFF
--- a/internal/bot/feishu/feishu.go
+++ b/internal/bot/feishu/feishu.go
@@ -10,6 +10,7 @@ package feishu
 import (
 	"context"
 	"crypto/sha256"
+	"crypto/subtle"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
@@ -417,7 +418,10 @@ func cardActionToast(toastType, content string) *callback.CardActionTriggerRespo
 }
 
 func (a *adapter) verificationTokenValid(token string) bool {
-	return a.cfg.VerificationToken == "" || token == a.cfg.VerificationToken
+	if a.cfg.VerificationToken == "" {
+		return false
+	}
+	return subtle.ConstantTimeCompare([]byte(token), []byte(a.cfg.VerificationToken)) == 1
 }
 
 func firstNonEmpty(vals ...string) string {
--- a/internal/bot/feishu/feishu_test.go
+++ b/internal/bot/feishu/feishu_test.go
@@ -43,9 +43,12 @@ func TestVerificationTokenValidRequiresConfiguredToken(t *testing.T) {
 		t.Fatal("matching token should be accepted")
 	}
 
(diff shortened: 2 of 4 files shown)


SEPURI-SAI-KRISHNA:
@SivanCola can you please review this PR. 
Thanks.

esengine:
This landed as part of `95cf355b2` (`feat(bot): consolidate related gateway improvements`), which names #5105, #5093, #5046 and #5888 as the work it adapts.

Four PRs arrived against the same gateway code inside about a week, each fixing a real problem, and they overlapped heavily. They were consolidated into one change rather than merged in sequence — taking them one at a time would have meant resolving the same conflicts three times over, with each author rebasing after every merge.

Leaving this PR open afterwards was our mistake. From your side it reads as a contribution that was ignored; it had already shipped. Closing it now, with that said rather than letting it lapse quietly.

Thanks in particular for the constant-time comparison and for removing the QQ_SECRET bypass. A token check that returns on the first wrong byte reads as correct and is not, and a bypass that exists only for convenience is precisely the one that outlives its reason.

PR #5048, opened 2026-06-22
fix(provider): force temperature=1 for DeepSeek thinking, preserve explicit temperature=0

SEPURI-SAI-KRISHNA:
## Summary

  - **fix(openai) Bug 2 — DeepSeek temperature:** when `deepseek == true`, always
    set `out.Temperature = TemperaturePtr(1.0)` on the wire, overriding whatever
    the caller passed. DeepSeek reasoning models (`deepseek-reasoner`, R1, V3
    thinking mode) require exactly `temperature=1`; any other value produces
    degraded chain-of-thought or a 400 error from the API. Previously the agent's
    configured temperature (defaulting to `0.0`) was forwarded unchanged and silently
    dropped by `omitempty`, resulting in unpredictable model behaviour.

  - **fix(provider) Bug 3 — temperature=0 silently dropped:** `provider.Request.Temperature`
    was `float64`; the downstream `chatRequest.Temperature` had `json:"temperature,omitempty"`,
    so Go's zero value (`0.0`) was never serialised. Two callers that explicitly
    needed deterministic output (`serve` title generation and `auto_plan_classifier`)
    set `Temperature: 0` but their intent was never sent to the API.

    `Request.Temperature` is now `*float64`. Two helpers are added to the `provider`
    package:
    - `TemperaturePtr(v float64) *float64` — for callers that explicitly want a
      value, including `0` (deterministic output).
    - `OptionalTemperature(v float64) *float64` — returns `nil` for `0` ("not
      configured, use the API default") and a pointer otherwise; used by agent,
      compact, and coordinator where `0` means the user has not set a temperature.

## Verification

  - `go test ./internal/provider/... ./internal/agent/... ./internal/control/... ./internal/serve/...`
    — all packages pass (provider, provider/anthropic, provider/openai, agent,
    agent/testutil, control, serve).
  - `TestProviderAutoPlanClassifierRequestShape` updated to assert
    `Temperature == ptr(0)` instead of `Temperature != 0`.
  - `realcache_test.go` updated to omit `Temperature` (nil) rather than pass the
    untyped constant `0`.

  ## Cache impact

  Cache-impact: low — `provider/openai` wire format gains an explicit `"temperature": 1` field on every DeepSeek request where it was previously absent; the system-prompt and
  tool-schema prefix that governs cache keys is unchanged.
  Cache-guard: `go test ./internal/provider/openai/...` exercises the full SSE stream path including `realcache_test.go`; the temperature field appears in the JSON body after the
  cached prefix, so no cache-key regression is possible.
  System-prompt-review: N/A — no files under `internal/config/system_prompt*`, `internal/outputstyle/`, `internal/skill/`, or `internal/memory/` were modified.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -1319,7 +1319,7 @@ func (a *Agent) stream(ctx context.Context, turn int) (string, string, string, [
 	ch, err := a.prov.Stream(ctx, provider.Request{
 		Messages:    a.session.Messages,
 		Tools:       a.tools.Schemas(),
-		Temperature: a.temperature,
+		Temperature: provider.OptionalTemperature(a.temperature),
 	})
 	if err != nil {
 		return "", "", "", nil, nil, false, false, err
--- a/internal/agent/compact.go
+++ b/internal/agent/compact.go
@@ -610,7 +610,7 @@ func (a *Agent) summarize(ctx context.Context, region []provider.Message, instru
 			{Role: provider.RoleSystem, Content: sys},
 			{Role: provider.RoleUser, Content: renderTranscript(region)},
 		},
-		Temperature: a.temperature,
+		Temperature: provider.OptionalTemperature(a.temperature),
 	})
 	if err != nil {
 		return "", err
--- a/internal/agent/coordinator.go
+++ b/internal/agent/coordinator.go
@@ -254,7 +254,7 @@ func (c *Coordinator) plan(ctx context.Context, input string) (string, error) {
 
 	ch, err := c.planner.Stream(ctx, provider.Request{
 		Messages:    c.plannerSess.Messages,
-		Temperature: c.temperature,
+		Temperature: provider.OptionalTemperature(c.temperature),
 	})
 	if err != nil {
 		return "", err
--- a/internal/control/auto_plan_classifier.go
+++ b/internal/control/auto_plan_classifier.go
@@ -48,7 +48,7 @@ func (c *ProviderAutoPlanClassifier) NeedsPlan(ctx context.Context, input string
 			{Role: provider.RoleSystem, Content: autoPlanClassifierPrompt},
 			{Role: provider.RoleUser, Content: fmt.Sprintf("heuristic_score=%d\n\nUSER_REQUEST:\n%s", score, input)},
 		},
-		Temperature: 0,
+		Temperature: provider.TemperaturePtr(0),
 		MaxTokens:   80,
 	})
 	if err != nil {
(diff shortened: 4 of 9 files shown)


SEPURI-SAI-KRISHNA:
@SivanCola @esengine 

 #5106 and this PR both touch openai.go and agent.go, but they fix entirely different bugs at different lines — there is no overlap in logic or intent.

 What #5106 fixed — stream recovery:
 - agent.go: bumped maxStreamRecoveries from 1 → 3 so the agent retries more on dropped connections
 - openai.go: added a ctx.Err() cancellation check inside readStream so a user-cancelled turn is not misclassified as a dropped stream; added omitempty to Index int in the tool-call
 struct

 What this PR fixes — temperature correctness:
 - openai.go: when c.deepseek == true, the wire request now always sends "temperature": 1. DeepSeek reasoning models (deepseek-reasoner, R1, V3 thinking mode) require exactly 
 temperature=1 — any other value degrades chain-of-thought quality or triggers a 400 from the API. Previously the field was forwarded from req.Temperature which defaulted to 0.0 and
 was then silently dropped by json:"omitempty", so DeepSeek received no temperature at all.
 - provider.go + all call sites: Request.Temperature is changed from float64 to *float64. With a float64 field and omitempty on the downstream struct, callers that explicitly set 
 Temperature: 0 (the title-generation call in serve.go and auto_plan_classifier) had their intent silently discarded — 0.0 is Go's zero value so omitempty dropped it, and the API 
 fell back to its own default (typically 1.0), producing non-deterministic output where deterministic output was required.

 The two PRs are orthogonal. #5106 is already merged and this branch has been rebased on top of it — go test -race ./... passes clean against the merged state.

SivanCola:
Thanks for the investigation here. The explicit-zero temperature part has now landed on `main-v2` via #5883 (merge commit `0b579f2`).

What was adopted:
- `provider.Request.Temperature` now uses `*float64`, so `temperature: 0` can be intentionally serialized.
- Deterministic callers such as auto-plan classification and title generation now pass explicit zero with `provider.TemperaturePtr(0)`.
- Regular agent/configured-zero paths continue to omit temperature when zero means provider default.

What was not adopted:
- The DeepSeek-specific forced `temperature=1` override. #5883 keeps DeepSeek temperature caller-controlled/default-omitted to avoid changing generic OpenAI-compatible behavior.

Verification from #5883:
- `go test ./internal/provider/openai -count=1`
- `go test ./internal/provider/openai ./internal/control ./internal/agent ./internal/serve -count=1`
- `go test ./... -p 1 -count=1`

The adapted work includes your co-author credit in the commit and PR body.

SEPURI-SAI-KRISHNA:
Hey @SivanCola thanks for update and great work. Appreciate your effort, looking forward to contribute further.

Issue #5045, opened 2026-06-22
[Bug]: 为啥更新记忆也要同意啊，让我更新记忆，记录这个修复。需要权限


eghrhegpe:
### Exact version

1.9.10

### What happened?

为啥更新记忆也要同意啊，而且更新记忆的频率、时机没法控制。

### Steps to reproduce

现在狼人被处决时可以发遗言了。让我更新记忆，记录这个修复。
自动同意普通工具：true
需要权限：用户选择同意。
✓
remember
vote-last-words-fix
1951 ms
{
  "body": "**问题**：`cmd_vote`函数缺少遗言处理代码，导致狼人被处决时无法发遗言\n\n**修复**：在`cmd_vote`函数第580行后添加遗言处理代码\n\n**验证**：狼人被处决时可以发遗言，输出\"我是狼人，队友是XX\"",
  "description": "修复cmd_vote缺少遗言代码的bug",
  "name": "vote-last-words-fix",
  "title": "投票遗言修复",
  "type": "project"
}

### OS / platform

window 11

### Relevant logs or output

```shell

```

Zhuocheng-Lang:
bro真的很喜欢他的狼人杀💀

eghrhegpe:
> bro真的很喜欢他的狼人杀💀

看AI狼人杀，睡不着觉了，deep seek能干推理，

并且AI能交流对方的上下文，subagent、跨会话AI居然真的有实际价值。

esengine:
Fixed: memory writes triggered by the remember flow now use a host-issued create-only grant (the auto-memory write queue), so saving a memory the host initiated no longer round-trips through a tool-approval prompt.

Please retest on the latest release; closing as completed.

PR #5049, opened 2026-06-22
fix: remember/forget tools should honor auto-approval mode (#5045)

myipanta:
Remove hardcoded always-ask rules for memory tools so they follow the
normal policy flow consistent with every other writer tool.

Two changes:
- newInteractiveGate(): no longer appends explicit Ask rules for
  remember/forget. They flow through Policy.DecideSubject normally —
  prompted in ask mode (Mode fallback), auto-allowed in auto/yolo.
- requiresFreshApprovalTool(): only planApprovalTool remains.
  Session grants, "always allow" persistence, and YOLO drain now
  work for memory tools — previously all blocked.

After fix:

| Mode | remember/forget |
|---|---|
| ask (default) | Prompt |
| auto | Auto-allow |
| yolo | Auto-allow |
| User adds Ask:remember | Prompt even in auto |
| User adds Deny:remember | Blocked in all modes |

Memory tools are safer than file-mutation tools (write to
~/.reasonix/memory/, never touch workspace code). They now receive
the same approval treatment as bash and edit_file — no special
hardcoding.

Tests updated: three renamed to match new behavior, approval
notifications now include subject text. time.Sleep replaced with
event-channel synchronization per review.

Cache-impact: low — only changes approval decision flow, no prompt
prefix or tool schema modifications

Cache-guard: go test ./internal/control/ ./internal/permission/

System-prompt-review: no system prompt, memory prefix, or skill index
changes — default config Ask rules were removed before merge

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1161,6 +1161,9 @@ func Default() *Config {
 		// Mode "ask" with no rules keeps `reasonix run` autonomous (no TTY → ask
 		// resolves to allow) while `reasonix` prompts before writers. Users add
 		// deny/allow rules to harden or quiet specific tools.
+		// Memory tools (remember/forget) follow the normal policy flow — prompted
+		// in ask mode, auto-allowed in auto/yolo. Add to ask to always require
+		// manual review.
 		Permissions: PermissionsConfig{Mode: "ask"},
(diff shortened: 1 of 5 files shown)


myipanta:
```
 Reasonix coding agent AI 辅助、人审批
```

PR #5076, opened 2026-06-22
Fix short choice replies in dual-model context / 修复双模型短选项回复上下文

### LINKED ISSUE #5027: [Bug]: 大模型无法感知上下文
### Version line

### Exact version

 1.11.0

### What happened?

Image

刚刚问了我需要通过哪种方式进行处理,回答了 1,马上就不知道 1 表示的是什么了,之前的版本不会出现

### Steps to reproduce

1. 具体执行的时候,忘了当前是什么意思

### OS / platform

MacOS 26 Arm64

### Relevant logs or output

```shell

```

GTC2080:
## Summary

- Treat short context-dependent replies like `1`, `A`, `好的`, `继续`, and `选 1` as executor-only turns in two-model mode, so they keep the previous assistant answer from the executor session instead of starting a fresh planner pass without context.
- Keep auto-plan scoring aligned with the same short-reply guard so these replies do not enter read-only planning.
- Add regression coverage for the reported numbered-choice flow: after the assistant offers numbered options, a user reply of `1` goes directly to the executor, preserves the previous assistant options in the request, and is not wrapped as a planner handoff.

Fixes #5027

## Verification

- `go test ./internal/control -run 'Test(TaskWarrantsPlanner|TwoModelShortChoiceReplySkipsPlanner)' -count=1`
- `go test ./internal/control -count=1`
- `go test ./internal/agent -run TestCoordinator -count=1`
- `go test ./...`
- `go vet ./...`
- `go vet ./...` in `desktop/`
- `gofmt -l .`
- `git diff --check`

## Cache impact

Cache-impact: low — this is a routing/classification change for short context-dependent replies; it does not change system prompts, tool schemas, tool ordering, request serialization, memory prefixes, compaction, or reasoning upload behavior.
Cache-guard: `go test ./internal/control -run 'Test(TaskWarrantsPlanner|TwoModelShortChoiceReplySkipsPlanner)' -count=1`
System-prompt-review: N/A

### DIFF
--- a/internal/control/auto_plan.go
+++ b/internal/control/auto_plan.go
@@ -145,6 +145,9 @@ func isContextDependentShortReply(text string) bool {
 		return true
 	}
 	lower := strings.ToLower(text)
+	if containsAny(lower, complexIntentTerms) || containsAny(lower, lowRiskWorkRequestTerms) {
+		return false
+	}
 	if shortContextReplies[lower] {
 		return true
 	}
--- a/internal/control/auto_plan.go
+++ b/internal/control/auto_plan.go
@@ -17,6 +17,8 @@ const (
 )
 
 var numberedListRE = regexp.MustCompile(`(?m)^\s*(?:[-*]|\d+[.)])\s+\S`)
+var directOptionReplyRE = regexp.MustCompile(`(?i)^\s*(?:\d+|[a-z])\s*[.)、。]?\s*$`)
+var prefixedOptionReplyRE = regexp.MustCompile(`(?i)^\s*(?:选|选择|就|用|按|走|执行|choose|pick|use|option|choice|方案)\s*(?:第\s*)?(?:方案|选项|option|choice)?\s*(?:\d+|[一二三四五六七八九十]|[a-z])\s*(?:个|号|项|种|条|方案|option|choice)?\s*[.)、。!！?？]?\s*$`)
 
 type AutoPlanClassifier interface {
 	NeedsPlan(ctx context.Context, input string, score int) (bool, string, error)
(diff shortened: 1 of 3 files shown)


SivanCola (changes requested):
Thanks, this fixes the reported numbered-choice path and the added regression is useful. Before merging, please narrow the short-reply guard so it does not classify terse work requests as context-dependent replies.

The new `isContextDependentShortReply` check runs before the work-request/complex-intent checks in both `TaskWarrantsPlanner` and `autoPlanScore`. Because the coordinator gate only receives the current input, prefixes such as `执行`, `开始`, `继续`, and `continue` now skip the planner even for standalone short tasks like `执行修复`, `开始迁移`, or `继续重构`. Those contain existing complex/work terms and should still warrant planning in two-model mode, so this regresses the existing contract that terse work requests still get planned.

Suggested direction: keep bare confirmations/options executor-only, but let inputs that also contain work-request/complex-intent terms fall through to the existing planner heuristics, and add regression cases for those short work requests.

Verified:
- `go test ./internal/control -run "Test(TaskWarrantsPlanner|TwoModelShortChoiceReplySkipsPlanner)" -count=1`
- `go test ./internal/control -count=1`
- PR CI checks are currently green

SivanCola (approved):
Approved. The follow-up commit `6ad1df0` addresses the short-reply guard regression by preserving executor-only routing for bare confirmations/options while allowing short inputs with existing work-request or complex-intent terms to fall back to the planner heuristics.

Verified:
- `go test ./internal/control -run "Test(TaskWarrantsPlanner|TwoModelShortChoiceReplySkipsPlanner)" -count=1`
- `go test ./internal/control -count=1`
- `git diff --check`
- PR CI checks are green

PR #5077, opened 2026-06-22
fix(desktop): dedupe skill roots in settings

### LINKED ISSUE #4779: [Bug]: skills 目录显示重复
### Version line

### Exact version

1.9.1

### What happened?

Image

skill 目录显示重复

### Steps to reproduce

点击进入skill设置

### OS / platform

macos27

### Relevant logs or output

```shell

```

Ninot1Quyi:
## Summary

- Collapse duplicate skill source rows by canonical path in desktop Skills settings.
- Preserve configured/removable metadata and discovered skill items when a convention root is also listed in `[skills].paths`.
- Add regression coverage for `~/.reasonix/skills` being configured as a custom skill path.

Fixes #4779

## Verification

- `cd desktop && go test . -run 'TestSkillRootsView(DedupesConfiguredConventionRoot|CountsProjectSkills|MarksEnvConfiguredCustomRoot|OmitsExcludedConventionRoot)|Test(AddSkillPathRestoresConventionRootWithoutCustomPath|RemoveSkillPathPseudoDeletesConventionRoot)' -count=1`
- `go test ./...`
- `cd desktop && gofmt -l .`
- `git diff --check`

## Cache impact

Cache-impact: none — desktop skills settings rendering only; no provider-visible prompt, tool schema, memory prefix, output style, skill index, or request serialization changes.
Cache-guard: not applicable — no cache-sensitive surface changed.
System-prompt-review: N/A

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -4295,7 +4295,7 @@ func cloneSkillRootViews(in []SkillRootView) []SkillRootView {
 func rootActive(roots []SkillRootView, path string) bool {
 	want := config.CanonicalSkillPath(path)
 	for _, r := range roots {
-		if r.Scope == string(skill.ScopeCustom) && config.CanonicalSkillPath(r.Dir) == want {
+		if config.CanonicalSkillPath(r.Dir) == want {
 			return true
 		}
 	}
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -4196,6 +4196,7 @@ func skillRootsViewFrom(cwd string, cfg, userCfg *config.Config) []SkillRootView
 		}
 	}
 	out := []SkillRootView{}
+	seenRoots := map[string]int{}
 	for _, r := range roots {
 		dir := config.CanonicalSkillPath(r.Dir)
 		view := SkillRootView{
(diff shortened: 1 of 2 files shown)


SivanCola (changes requested):
Thanks, this is directionally right, but one duplicate-root case still slips through. When a user has a project convention root (for example, <project>/.reasonix/skills) in [skills].paths, the store returns the project convention root before the duplicate custom root. The new merge keeps the first row's project scope, and the later inactive-user-path pass calls rootActive(), which only treats matching custom-scope rows as active. That appends a second inactive custom row for the same directory, so the Skills settings page can still show duplicate rows.

Please either make rootActive() consider an already-present matching active root regardless of scope, or otherwise preserve enough custom-active metadata after merging duplicate roots. Please also add a regression test for a configured project convention root.

Verified locally:
- The PR's focused Skills settings tests pass.
- git diff --check passes.
- An additional probe for configured <project>/.reasonix/skills still fails with two rows: active project plus inactive custom.

Once this is addressed and CI stays green, I will consider merging it into main-v2.

SivanCola (approved):
Approved. I pushed a small follow-up commit that addresses the duplicate project convention root case I requested changes for.

What changed:
- Treat an already-rendered canonical skill root as active when deciding whether to add an inactive user-configured placeholder.
- Add a regression test for a configured project convention root so <project>/.reasonix/skills does not render as active project plus inactive custom.

Verified locally:
- cd desktop && go test . -run TestSkillRootsView... focused skill roots suite -count=1 passed
- git diff --check passed

Cache impact remains none: this is still limited to desktop Skills settings rendering and tests, with no provider-visible prompt, skill index, tool schema, or request serialization changes.

SivanCola:
Thanks, this has been merged into main-v2.

The final version fixes duplicate skill source rows for both configured home convention roots and configured project convention roots, with the added project convention root regression test. All GitHub checks are passing, including cache-impact, desktop, lint, race, coverage, multi-platform tests, and CodeQL.

This also closes #4779.

PR #5106, opened 2026-06-23
Fix interrupted stream recovery / 修复长任务断流恢复

SivanCola:
## Summary
- Treat user cancellation as cancellation instead of a dropped model stream, preventing reconnects after the user stops a long-running turn.
- Allow the agent loop to recover from up to three interrupted partial streams before surfacing an actionable error.
- Omit response-only `tool_calls[].index` from replayed assistant messages while preserving streaming delta parsing.

## Cache and provider compatibility
Cache-impact: low - Touches provider request serialization and agent recovery limits, but does not change the stable prompt prefix, tool schema bytes or order, model list, or provider thinking controls. It only omits a response-only zero-valued tool-call index from replayed assistant messages.
Cache-guard: `TestBuildRequestOmitsResponseOnlyToolCallIndex`; `go test ./...`

- Aligns replayed `tool_calls` with the OpenAI-compatible message shape (`id`, `type`, `function`) while keeping streaming `delta.tool_calls[].index` parsing intact.
- Related nearby PRs checked: #5063, #5059, #5048. They touch adjacent provider/thinking behavior but not this stream-recovery path.

## Verification
- `git diff --check`
- `go test ./...`

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -81,7 +81,7 @@ var planModeGoWriteOrExecArgs = map[string]bool{
 
 const maxFinalReadinessBlocks = 3
 const maxEmptyFinalBlocks = 3
-const maxStreamRecoveries = 1
+const maxStreamRecoveries = 3
 const maxExecutorHandoffNudges = 1
 
 // Renderer redraws the assistant's final-answer text as styled output. It is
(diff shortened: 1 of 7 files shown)


SivanCola:
@codex review

SivanCola:
Follow-up: #5067 has been linked and closed as fixed after this PR was merged into `main-v2`.

Verification before/after merge: local `git diff --check`, local `go test ./...`, and PR CI were green, including cache-impact, lint, desktop, race, platform tests, and CodeQL.

PR #5223, opened 2026-06-24
Default workspace panel to collapsed / 默认收起工作区面板

SivanCola:
## Summary
- Default the desktop workspace/right dock panel to collapsed on startup.
- Keep the existing manual expand affordance and persisted width preferences unchanged.
- Add a source-level regression check so the default does not drift back to expanded.

## Verification
- `pnpm exec tsx src/__tests__/app-chrome-tabs.test.ts`
- `pnpm test:typecheck`
- Browser reload verified the dock is not rendered by default and the toggle offers "Expand workspace".


SivanCola:
@codex review

PR #5230, opened 2026-06-24
fix: 工作台/创作模式下新建对话显示旧对话的问题

ttmouse:
工作台和创作模式使用 `ensureBlankSurface`（单 surface 布局）创建新对话时，
与 mount/onReady 时 fire-and-forget 调用的 `syncActiveTabFromBackend`
存在竞态条件：后者异步返回后会用旧的 backend tab 覆盖新创建的空白 tab。

修复：给 `syncActiveTabFromBackend` 增加 `guard` 参数，mount/onReady 的
fire-and-forget 调用传入 `guard=true`，在异步获取 backend 结果后检查
`activeTabIdRef.current` 是否已被用户操作改变，若是则跳过覆盖。

### DIFF
--- a/desktop/frontend/src/__tests__/new-session-load-race.test.tsx
+++ b/desktop/frontend/src/__tests__/new-session-load-race.test.tsx
@@ -181,6 +181,52 @@ eq(controller?.state.items.length, 0, "stale history load cannot repopulate a ne
 await act(async () => {
   root.unmount();
 });
+
+const guardedStartupTabs = deferred<TabMeta[]>();
+let ensureBlankSurfaceCalls = 0;
+window.go.main.App = {
+  ListTabs: async () => guardedStartupTabs.promise,
+  MetaForTab: async () => meta(),
+  ContextUsageForTab: async () => context,
+  EffortForTab: async () => effort,
+  BalanceForTab: async () => balance,
+  JobsForTab: async () => jobs,
+  CheckpointsForTab: async () => checkpoints,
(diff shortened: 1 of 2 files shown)


SivanCola (approved):
Approved. This is a well-scoped fix for a real startup race in the single-surface desktop layouts. The guarded mount/onReady sync now avoids overwriting a foreground tab change that happened while the backend active-tab lookup was still in flight, while the direct sync paths used by close/switch flows keep their existing behavior.

- The frontend remains the owner of immediate visible active-tab state after user actions such as creating a blank surface.
- The backend remains the source of truth for startup hydration, but fire-and-forget startup sync is now guarded against stale async results.

I added `Add startup active-tab race regression` to cover the exact race: startup `ListTabs()` returns an older active tab after `EnsureBlankSurface()` has already activated the new blank surface, and the active tab must remain the new one.

Security/cache review:
- No new dependencies, network/file/command execution, persistence, or privilege surface.
- No provider-visible prompt prefix, tool schema/order, request serialization, compaction, memory, or reasoning upload changes. `cache-impact` passed.

Verified:
- `pnpm --dir desktop/frontend exec tsx src/__tests__/new-session-load-race.test.tsx` passed
- `wails generate module` followed by `pnpm --dir desktop/frontend typecheck` passed
- `git diff --check` passed
- GitHub checks passed on the updated PR head: CI, desktop, race, coverage, lint, govulncheck, CodeQL, and cache-impact

PR #5262, opened 2026-06-24
Fix subagent cancellation contract

lifu963:
## Summary

- 背景：取消 subAgent 时，阻塞的 provider stream 或并行子任务聚合可能让外层调用无法及时退出，造成“取消了但还在等”的体验。
- 做了什么：让 Agent stream、OpenAI/Anthropic provider chunk 发送、foreground subagent 路径和 `parallel_tasks` 聚合都显式响应 `context` 取消。
- 为什么这样做：取消语义应由上层 `context` 统一驱动，不能依赖每个 provider 或 tool 都一定会主动、及时返回。

## Expected result

- 前台 subagent、`runAs=subagent` skill 和 `parallel_tasks` 在取消后能快速返回，不会被卡死的 provider channel 拖住。
- provider producer 在 reader 放弃消费后能退出，降低 goroutine 泄漏风险。
- `parallel_tasks` 取消时保留已完成子任务结果，并把未完成任务标记为 cancelled/skipped，避免把取消误看成全部成功。

## Verification

- `git ls-files '*.go' ':!desktop/*' | xargs gofmt -l`：通过，无输出。
- `go vet ./... && go test ./...`：通过。
- `git merge-tree --write-tree HEAD origin/main-v2`：通过，无冲突。
- `git diff --check origin/main-v2...HEAD`：通过，无 whitespace 问题。

## Cache impact

Cache-impact: none - 不改变系统提示、工具 schema、provider request 序列化或稳定 prompt 前缀；仅修复运行时取消路径。
Cache-guard: `go vet ./... && go test ./...`，包含新增的取消路径回归测试。
System-prompt-review: 未改动 system prompt、memory prefix、output style 或 skill index 文本；`internal/skill/tools_test.go` 仅新增 `runAs=subagent` cancellation wiring 测试，请维护者按 runtime cancellation 变更 review，无需单独 system-prompt review。


SivanCola (changes requested):
Thanks for tackling this; the provider-side unblock and the partial `parallel_tasks` aggregate are directionally good. I found one cancellation race that should be fixed before merge.

`internal/agent/agent.go` now selects between `ctx.Done()` and the provider chunk channel. In the `!ok` branch it finalizes the stream as a normal completion without checking `ctx.Err()`. The provider changes in this PR make that reachable: after `sendChunk` observes cancellation, OpenAI/Anthropic can return and close `out` without delivering a `ChunkError`. When the context is canceled and the channel is also closed, Go can choose the closed-channel case, so `Run` may treat the canceled stream as a successful empty completion and eventually report an empty-final error instead of `context.Canceled`.

A minimal regression shape is: provider returns an already-closed chunk channel while the run context is already canceled; repeating that loop exposes `Run` returning `model finished without a visible final answer 3 times` instead of `context.Canceled`.

Please check `ctx.Err()` before normal finalization in the closed-channel branch (or otherwise guarantee cancellation is surfaced as an error) and add that regression coverage.

Verified:
- Focused PR cancellation tests pass for the existing agent/provider/skill cases.
- Added a local regression for canceled context + closed provider channel; it fails on this PR head as described.

SivanCola (approved):
Approved. I pushed `105ff371` to address the cancellation race from my previous review.

What changed:
- `Agent.stream` now checks `ctx.Err()` before treating a closed provider chunk channel as normal completion.
- Added a regression test for the canceled-context + closed-provider-channel race, so cancellation cannot fall through into the empty-final retry path.

Verified:
- `go test ./internal/agent -run 'TestCanceledContextClosedProviderStreamReturnsCancel|TestCancelDuringStuckProviderStreamReturnsPromptly|TestParallelTasksCancelReturnsPartialAggregate|TestTaskToolCancelDuringStuckProviderReturnsPromptly' -count=1` passed.
- `go test ./internal/provider/openai ./internal/provider/anthropic -run 'TestReadStreamSendUnblocksOnContextCancel|TestStreamStallTimesOut' -count=1` passed.
- `go test ./internal/skill -run 'TestRunSkillSubagentCancellationReachesRunner' -count=1` passed.
- `go test ./internal/agent ./internal/provider/openai ./internal/provider/anthropic ./internal/skill -count=1` passed.
- `git diff --check` passed.

The GitHub checks have restarted on the pushed head.

PR #5282, opened 2026-06-25
fix: Stop hook 使用 context.Background() 替代已取消的 ctx

### LINKED ISSUE #5281: [Bug]: 用户取消回复时 Stop hook 命令无法执行（ctx 已取消）
## Bug 描述

用户点击 Stop（或按 Esc）取消正在运行的 AI 回复时，Stop hook 的 defer 虽然触发，但 shell 命令无法执行——因为取消时 turn context 已经被 Controller.Cancel() 取消了，exec.CommandContext(cancelledCtx, ...) 立即失败。

## 根因

internal/control/turn_orchestrator.go:47，Stop hook 直接接收 turn context：

```
defer func() { c.hooks.Stop(ctx, ...) }()
```

正常完成时 ctx 有效，hook 正常执行。但用户点击 Stop → Cancel() → ctx 取消 → DefaultSpawner 内 context.WithTimeout(cancelledCtx, 30s) 子 ctx 也取消 → 命令立即失败，返回 DecisionError。

Stop 是非阻塞清理 hook（IsBlocking(Stop) = false），不应受 ctx 影响。

## 修复

一行代码：
```
- defer func() { c.hooks.Stop(ctx, ...) }()
+ defer func() { c.hooks.Stop(context.Background(), ...) }()
```

## 复现步骤

1. 在 .reasonix/settings.json 中配置 Stop hook
2. 发送消息给 Reasonix
3. 在回复完成前点击 Stop/Esc
4. Hook 命令未被执行

## 环境

- Reasonix Desktop v1.11.1
- macOS arm64

manza7014:
## 概要

用户点击 Stop（或 Esc）取消正在运行的回复时，Stop hook 虽然通过 defer 触发，但 shell 命令无法执行——turn context 已被 `Controller.Cancel()` 取消，`exec.CommandContext(cancelledCtx, ...)` 立即失败。

本 PR 将 `ctx` 替换为 `context.Background()`，使清理 hook 不受取消影响。

## 关联 Issue

Closes #5281

## 变更

`internal/control/turn_orchestrator.go` —— 一行：

```
- defer func() { c.hooks.Stop(ctx, ...) }()
+ defer func() { c.hooks.Stop(context.Background(), ...) }()
```

## 为什么用 context.Background()

- Stop 是非阻塞清理 hook，不是门控事件
- 不需要 turn context，只执行用户配置的命令
- `context.Background()` 给 hook 干净的 30s 超时窗口

## 验证

- `go build ./cmd/reasonix` 通过
- 本地编译桌面版测试：取消后 Stop hook 正常执行

Cache-impact: none
Cache-guard: existing hook tests in internal/hook/

### DIFF
--- a/internal/control/controller.go
+++ b/internal/control/controller.go
@@ -1115,7 +1115,7 @@ func (c *Controller) Run(ctx context.Context, input string) error {
 		if block, _ := c.hooks.PromptSubmit(ctx, input, c.turn); block {
 			return nil
 		}
-		defer func() { c.hooks.Stop(ctx, lastAssistantText(c.History()), c.turn) }()
+		defer func() { c.hooks.Stop(context.Background(), lastAssistantText(c.History()), c.turn) }()
 	}
 	return c.runner.Run(ctx, input)
 }
--- a/internal/control/controller_test.go
+++ b/internal/control/controller_test.go
@@ -70,6 +70,15 @@ func (r *sessionContextRunner) Run(ctx context.Context, input string) error {
 	return nil
 }
 
+type cancelingRunner struct {
+	cancel context.CancelFunc
+}
+
+func (r cancelingRunner) Run(_ context.Context, _ string) error {
+	r.cancel()
+	return nil
+}
+
 type fakeControlTool struct{ name string }
 
 func (t fakeControlTool) Name() string { return t.name }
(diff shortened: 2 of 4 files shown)


SivanCola (changes requested):
Thanks, this is directionally right: Stop is a non-blocking cleanup hook, and the desktop/guarded turn path should not pass an already-canceled turn context into the hook spawner.

Before merging, please address one remaining contract gap:

- Apply the same cancellation-independent Stop-hook context to the synchronous/headless `Controller.Run(ctx, ...)` path in `internal/control/controller.go`. This PR fixes `turn_orchestrator.go`, but `Controller.Run` still defers `c.hooks.Stop(ctx, ...)`; if that caller context is canceled before `Run` returns, `DefaultSpawner` derives its timeout from an already-canceled context and the Stop command can still fail with `context.Canceled`.
- Please add a focused regression test that proves a canceled turn/caller context still lets the Stop hook receive a usable context. Ideally cover both the guarded turn path and the `Controller.Run` path, or put the contract at the shared owner level if you factor the Stop-hook launch into one helper.

Security/cache review: no new dependency, network, prompt-prefix, tool-schema, provider-serialization, or cache-shape risk in the current diff. The broader hook execution remains bounded by the hook timeout.

Verified locally:
- `go test ./internal/control ./internal/hook` passed on the PR head.

SivanCola (approved):
Approved. This is a well-scoped and worthwhile fix. The PR correctly keeps Stop hooks alive after turn cancellation by launching them with a cancellation-independent context in both the guarded and synchronous Controller.Run paths.

- Stop is a cleanup hook with its own per-hook timeout, so it should not inherit turn or caller cancellation state.
- The controller/orchestrator boundary owns prompt and Stop hook lifecycle, so this is the right owner to enforce the contract.

The main thing I wanted to tighten was regression coverage for that cancellation contract. I pushed e507555 to add focused tests for both entry points so a canceled turn/caller context does not cancel the Stop hook launch.

Verified:
- `go test ./internal/control ./internal/hook` passed
- `git diff --check` passed

SivanCola (approved):
Approved. The latest head only reformats the added Stop-hook regression test, with no semantic change to the fix.

Re-verified on `c4a11bf`:
- `go test ./internal/control ./internal/hook` passed
- `git diff --check` passed

PR #5284, opened 2026-06-25
feat(i18n): add MB unit to download progress display

qiukeren:
Show 'MB' after the numeric values in the updater's downloading status text across all three locales (en, zh, zh-TW).

image

## Summary

```
  "updater.downloading": "下载中… {done} / {total}（{pct}%）",
```

to

```
  "updater.downloading": "下载中… {done} MB / {total} MB（{pct}%）",
```

## Cache impact

NO


SivanCola (approved):
Approved. I verified that the updater progress values are byte counts formatted with the existing 1024 * 1024 frontend helper before interpolation, so adding the MB label is consistent with the current UI convention and does not change download logic. Checks are green.

PR #5285, opened 2026-06-25
fix(serve/auth): handle crypto/rand.Read error in signSession

YoungDan-hero:
signSession() silently discarded the error from rand.Read, leaving the
session nonce all-zero on failure. While this does not directly bypass
authentication (forging a cookie still requires the PBKDF2-derived
sessKey), it weakens the unpredictability of password-mode session
cookies and is an explicit anti-pattern inconsistent with the same
file's generateToken() and sessionKeyForPasswordHash(), both of which
panic on rand.Read failure.

Now panics with a descriptive message, matching the existing convention.

### DIFF
--- a/internal/serve/auth.go
+++ b/internal/serve/auth.go
@@ -458,8 +458,10 @@ func (ag *authGate) signSession() string {
 	nonce := make([]byte, 16)
 	if _, err := rand.Read(nonce); err != nil {
 		// crypto/rand.Read cannot fail on modern systems; panic rather than
-		// fall back to an all-zero nonce that would make the session cookie
-		// trivially forgeable and bypass authentication.
+		// fall back to an all-zero nonce. Forging a cookie still requires the
+		// PBKDF2-derived sessKey, so this is not an auth bypass, but a constant
+		// nonce weakens session token uniqueness/unpredictability and is the
+		// same anti-pattern generateToken/sessionKeyForPasswordHash panic on.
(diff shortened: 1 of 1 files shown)


SivanCola (approved):
Approved. Well-scoped, worthwhile hardening fix. `signSession()` was discarding the `rand.Read` error and falling back to an all-zero nonce; it now panics with a descriptive message, matching the existing `generateToken()` and `sessionKeyForPasswordHash()` convention in the same file.

Why this fits:
- The error path is consistent with the file's established "crypto/rand.Read cannot fail on modern systems; panic" convention, rather than introducing a new silent-degradation behavior.
- Scope is a single function in `internal/serve/auth.go`; no auth/session logic changed, no cache-sensitive prompt/tool surfaces touched.

The one thing I tightened was the new inline comment. The original wording said an all-zero nonce would make the cookie "trivially forgeable and bypass authentication," but `verifySession` still requires the PBKDF2-derived `sessKey` to produce a valid HMAC, so a constant nonce is not an auth bypass — it weakens session token uniqueness/unpredictability. Your PR description already had this exactly right; I pushed a one-line follow-up commit to make the code comment match it, so it doesn't mislead future security review. Original authorship on your commit is preserved.

Verified:
- `go test ./internal/serve` passed
- `go vet ./internal/serve` passed
- `gofmt -l internal/serve/auth.go` clean
- `git diff --check` clean

Thanks for catching and fixing this. Once CI stays green I'll merge it into main-v2.

PR #5301, opened 2026-06-25
fix: use OSC 52 directly for clipboard instead of platform-tool-then-…

meteor614:
Remove the direct `atotto/clipboard` write path from `copyToClipboard` and the two-tier strategy that tried platform clipboard tools (`xclip`/`wl-copy`/`pbcopy`) before falling back to OSC 52.

Over SSH, a platform clipboard tool can succeed on the remote host, which prevents the OSC 52 fallback from reaching the user's terminal-side clipboard. Using Bubble Tea's `tea.SetClipboard` keeps transcript selection copy on the terminal clipboard path directly.

## Summary

- Simplify transcript selection copy to use `tea.SetClipboard` / OSC 52 directly.
- Remove the test-only `clipboardWriteAll` injection and the platform-success/fallback split.
- Keep Ctrl+C selection tests focused on TUI behavior instead of platform clipboard stubs.
- Clarify that OSC 52 depends on terminal support and avoids remote-host clipboard writes in common SSH/tmux flows.

## Verification

- `go test ./internal/cli`

## Cache impact

Cache-impact: none — CLI/TUI clipboard-copy behavior only; no provider request serialization, system prompt, memory prefix, tool schema, or compaction behavior changes.
Cache-guard: existing cache-impact guard passed on the PR; no cache-sensitive surface changed.
System-prompt-review: N/A — no provider-visible prompt or prefix changes.

### DIFF
--- a/internal/cli/chat_tui_test.go
+++ b/internal/cli/chat_tui_test.go
@@ -10,7 +10,6 @@ import (
 	"time"
 
 	tea "charm.land/bubbletea/v2"
-	"github.com/atotto/clipboard"
 	"github.com/charmbracelet/x/ansi"
 
 	"reasonix/internal/agent"
@@ -1860,10 +1859,6 @@ func TestCtrlCClearsThenDoublePressQuits(t *testing.T) {
 // with an active text selection copies the selected text to clipboard instead
 // of arming the double-press quit gesture.
 func TestCtrlCCopySelection(t *testing.T) {
-	var copied string
-	clipboardWriteAll = func(text string) error { copied = text; return nil }
-	defer func() { clipboardWriteAll = clipboard.WriteAll }()
-
 	m := newTestChatTUI()
 	ctrlC := tea.KeyPressMsg{Code: 'c', Mod: 4}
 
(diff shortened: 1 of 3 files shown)


SivanCola (approved):
Approved. This is a well-scoped and worthwhile CLI/TUI fix. The PR correctly addresses the SSH/tmux clipboard issue by removing the platform-tool-first path for transcript selection copy and sending selection copies through Bubble Tea's OSC 52 clipboard command directly.

Why this fits the current implementation:
- `copyToClipboard` is only used for explicit transcript-selection copy actions, so routing it through `tea.SetClipboard` keeps the behavior scoped to the terminal clipboard path.
- Paste handling still uses `atotto/clipboard` where the dependency is needed for reading clipboard text, so the PR does not accidentally remove that remaining contract.
- Bubble Tea handles the returned set-clipboard message by executing `ansi.SetSystemClipboard`, matching the intended OSC 52 mechanism.

Security and cache review:
- No new dependencies, network calls, command execution, file access, or hidden persistence were introduced.
- No provider-visible prompt, tool schema, request serialization, memory prefix, reasoning upload, or compaction behavior changes.

Verified:
- `git diff --check` passed
- `go test -count=1 ./internal/cli` passed
- GitHub `label` and `cache-impact` checks passed

ddrpa:
I suggest refine or temporarily revert this change to avoid breaking copy functionality for a large number of users.

Standard VTE-based terminals (including Ptyxis, GNOME Terminal) do not support OSC 52 by default due to security policies. Forcing OSC 52 will leave these users unable to copy text.

A safer approach would be to implement a dynamic runtime check:

- Use OSC 52 if the terminal supports it.
- Fall back to `atotto/clipboard` if it doesn't.

This ensures not breaking compatibility for users running default desktop terminal environments.

Issue #5286, opened 2026-06-25
[Bug]: 中断的任务会在后续追加执行


fixed-link:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

reasonix npm-v1.11.1-rc.1

### What happened?

示例：
新增一个配置文件abc
（中间ctrl+c中断）
新增一个配置文件123
（正常执行）

结果：
新增配置文件abc
新增配置文件123

期待：
只新增配置文件123

### Steps to reproduce

见上

### OS / platform

Ubuntu

### Relevant logs or output

```shell

```

SivanCola:
Closing this as fixed by #5303 (merged as 7056884afc6c0ff9b6d97166db0f9e7f0bbe0e3c).

The fix strips incomplete assistant/tool messages when a turn is explicitly cancelled by the user, rebuilds todo state after the strip, and flushes the cleaned transcript so the next prompt does not inherit or continue the interrupted work.

Verification in the PR: focused cancel/flush regression tests with `go test -race`, plus `go test ./internal/agent/ ./internal/control/` passed.

Release status: the fixing commit is included in v1.13.0 / Desktop v1.13.0, so users should pick up the fix by upgrading to v1.13.0 or newer.

PR #5303, opened 2026-06-25
fix(control): strip incomplete turn messages after user cancel (#5286)

myipanta:
When the user presses Ctrl+C to cancel a turn, the agent's incomplete assistant messages and tool results are already saved to the session. On the next turn, the model sees leftover in-progress todo items and partial tool calls and may re-execute the interrupted work.

Fix: in turn_orchestrator.runTurnWithRawDisplay, when Run() returns context.Canceled AND the cancel was user-initiated (CancelRequested() == true), strip the session back to the pre-turn message count so the next prompt starts clean.

The guard (CancelRequested) ensures internal context cancellations (stream recovery, timeouts) are unaffected — only explicit user cancels trigger the strip.

## Summary

- Strip incomplete turn messages from session after user cancel so the next turn starts clean (#5286)
- Reset Agent.todoState after strip so Controller.Todos(), goal readiness, and the task panel no longer see cancelled in_progress items
- Flush cleaned transcript to disk after strip to overwrite any stale mid-turn autosave

## Verification

- `go test -race -run 'TestTurnOrchestratorCancelStripsIncompleteTurn|TestTurnOrchestratorCancelFlushesCleanTranscriptToDisk' ./internal/control/` passes
- `go test ./internal/agent/ ./internal/control/` passes (both packages, no regressions)

## Cache impact

Cache-impact: none - The change to internal/agent/agent.go only adds a public wrapper method RebuildTodoState() that calls the existing private rebuildTodoState(). It does not touch system prompt, tool list, tool schemas, provider serialization, compaction, or reasoning upload. No provider-visible prefix changes.
Cache-guard: go test ./internal/agent/ ./internal/control/ — existing schema stability tests pass; no cache-sensitive surface is touched.
System-prompt-review: N/A

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -1086,6 +1086,13 @@ func (a *Agent) emitTodoState(todos []evidence.TodoItem, itemIndex int) {
 	a.sink.Emit(event.Event{Kind: event.ToolResult, Tool: t})
 }
 
+// RebuildTodoState re-derives canonical task state from the current session
+// transcript. Call after externally truncating the session (e.g. after a
+// user-cancel strip) so Agent.todoState stays consistent with the messages.
+func (a *Agent) RebuildTodoState() {
+	a.rebuildTodoState(a.Session().Snapshot())
+}
+
 // rebuildTodoState reconstructs the canonical task list from a transcript: the
 // latest successful todo_write is the base, then every complete_step after it
 // advances an item. Deterministic from persisted messages, so it survives a
--- a/internal/control/controller.go
+++ b/internal/control/controller.go
@@ -2116,6 +2116,23 @@ func (c *Controller) messageCount() int {
 	return len(c.executor.Session().Snapshot())
 }
 
+// stripTurnMessagesAfter truncates the executor's session to keep only messages
+// before the given index, discarding an incomplete turn (the user prompt plus
+// every assistant / tool message that followed).  It is called when the user
+// explicitly cancels a turn so the next prompt starts clean — the model won't
+// see leftover in-progress todo items or partial tool calls and re-execute
+// interrupted work.
+func (c *Controller) stripTurnMessagesAfter(idx int) {
+	if c.executor == nil {
+		return
+	}
+	msgs := c.executor.Session().Snapshot()
+	if len(msgs) <= idx {
+		return
+	}
+	c.executor.Session().Replace(msgs[:idx])
+}
+
 func (c *Controller) snapshotActivityIfChanged(startMessages int) {
 	if c.messageCount() <= startMessages {
 		return
--- a/internal/control/controller.go
+++ b/internal/control/controller.go
@@ -2131,6 +2131,24 @@ func (c *Controller) stripTurnMessagesAfter(idx int) {
 		return
 	}
 	c.executor.Session().Replace(msgs[:idx])
+	// Rebuild canonical todo state from the truncated transcript so
+	// Controller.Todos(), goal readiness, and the task panel no longer see
+	// the in_progress items written by the cancelled turn.
+	c.executor.RebuildTodoState()
+	// The mid-turn autosave may have already written a partial transcript to
+	// disk.  snapshotActivityIfChanged skips the write when messageCount()
+	// returns to startMessages, so force a flush here to overwrite the stale
+	// file.  We call Session.Save directly to cover the edge case where the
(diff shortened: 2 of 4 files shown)


SivanCola (changes requested):
Thanks, this is directionally right and the mechanism is sound. The `CancelRequested()` guard is the correct discriminator, and adding the strip in both the first-run and plan-approved-execution branches is the right scope. Before merging, please address two incomplete contracts:

**[P1] `todoState` is not reset after the strip**

`stripTurnMessagesAfter` truncates the transcript via `Session.Replace`, but `Agent.todoState` is independent in-memory state. `Session.Replace` does not call `rebuildTodoState` — that only fires from `SetSession` (session pointer swap) and compaction. So after a cancelled turn that had a successful `todo_write`, the next call to `Controller.Todos()`, `goalTodos()`, goal readiness checks, and the task panel still see the stale `in_progress` item from the stripped turn.

The fix: after `stripTurnMessagesAfter`, call `rebuildTodoState` (or an equivalent public wrapper) on the truncated message slice so `todoState` is consistent with the transcript. For the plan-approved-execution branch, restore to the post-seed state (after `seedPlanTodos`) rather than the pre-plan state — the seed's todos are legitimate, only the execution leftovers should go.

**[P2] A mid-turn autosave can leave stale bytes on disk**

`autosaveWhileRunning` snapshots to disk periodically during a turn. `snapshotActivityIfChanged(startMessages)` is deferred, but its guard is `messageCount() > startMessages`; after the strip, the count is back to `startMessages`, so the defer exits early without writing the cleaned transcript. A restart or session resume will reload the stale file and the model will see the partial turn again.

The fix: after stripping, call `SnapshotActivity()` (or `snapshot(true)`) unconditionally to flush the cleaned transcript to disk. Make sure the "new session with only a system message" edge case (where `HasContent()` would return false) is also covered, since a cancelled first-turn leaves a session file with only the partial turn stripped away.

**What already works well:**
- The `CancelRequested()` guard correctly limits the strip to explicit user cancels, leaving internal timeouts and stream-recovery cancels unaffected.
- Adding the strip in both the first-run branch and the plan-approved-execution branch is the right scope.
- `stripTurnMessagesAfter` itself is correct and safe.
- The test accurately captures the transcript strip. Once P1 is addressed, add a parallel assertion that `c.Todos()` (or the agent's `CanonicalTodoState()`) returns the pre-turn list, and that a session reload yields the clean transcript (P2).

Once these are addressed and CI stays green, I will consider merging this into `main-v2`.

SivanCola (approved):
Both issues are addressed in commit 65f18ea. Here is what changed:

**P1 — todoState now rebuilt after strip**

Added `Agent.RebuildTodoState()` (a thin public wrapper over the existing `rebuildTodoState` internal method) and call it in `stripTurnMessagesAfter` immediately after `Session.Replace`. The rebuild derives canonical state from the truncated transcript, so `Controller.Todos()`, goal readiness, and the task panel see the correct (empty) state after a user cancel.

The test `TestTurnOrchestratorCancelStripsIncompleteTurn` now pre-seeds the executor's todoState via `ReplaceTodoState` (simulating a successful `todo_write` mid-turn) and asserts `c.Todos()` returns empty after the cancel strip.

**P2 — disk flushed after strip**

Added `Session.Save(path)` directly after the strip in `stripTurnMessagesAfter`. This bypasses the `HasContent()` guard in `snapshot()` so the clean transcript is written even when the strip leaves only a system message. The `snapshotActivityIfChanged` guard (`messageCount() > startMessages`) would have skipped the write because the count is back to startMessages after stripping.

Added `TestTurnOrchestratorCancelFlushesCleanTranscriptToDisk`: it spins up a controller with a real session path, runs a cancel, then reloads the file with `LoadSession` and asserts the on-disk non-system message count matches the pre-cancel count.

**Verified:**
- `go test -race -run 'TestTurnOrchestratorCancelStripsIncompleteTurn|TestTurnOrchestratorCancelFlushesCleanTranscriptToDisk|TestTurnOrchestratorStopHookCancelledContext|TestTurnOrchestratorApprovedPlanSharesOneStopHook' ./internal/control/` passed
- `go test ./internal/agent/ ./internal/control/` passed (both packages, no -race regressions)
- Failure-mode verified: P1 test fails without `RebuildTodoState`, P2 test fails without `Session.Save`
- `gofmt -l` clean, `git diff --check` clean

PR #5308, opened 2026-06-25
fix(bot): run turn off dispatch goroutine so approval replies arent deadlocked

### LINKED ISSUE #4402: [Bug]: 微信机器人功能异常：会话跳转按钮失效 & 权限审批时卡死
### Version line

### Exact version

1.7.0

### What happened?

当前使用 Reasonix v1.7.0 微信机器人，存在两个功能问题：
1.「打开对应会话」按钮无法使用
机器人已正常连接微信，且已与用户产生对话，发送者 ID 也在白名单内，但后台的「打开对应会话」按钮始终处于不可点击状态，无法直接跳转到对应的聊天窗口查看 / 管理对话。

Image

2.权限审批流程中机器人卡死
工具审批为跟随默认，当对话触发权限审批（如用户请求敏感操作、需要管理员审批的指令）时，机器人会直接停止响应，既无法继续推进审批流程，也不能处理后续的新消息 / 新对话，只能删除机器人并重新添加才能恢复正常。

Image

Image

### Steps to reproduce

1.配置并启动 Reasonix v1.7.0 微信机器人，确认连接状态为「已连接」，且用户 ID 已加入白名单。
2.用微信向机器人发送消息，生成有效对话记录。
3.进入后台的「IM 连接详情」页面，此时「打开对应会话」按钮仍为灰色，无法点击。
4.向机器人发送需要权限审批的请求（如敏感指令、需管理员确认的操作）。
5.回复1批准。
6.观察机器人状态：连接保持正常，但不再处理任何消息，审批流程停滞，也无法响应新对话。

### OS / platform

Windows 11

### Relevant logs or output

```shell

```

leslie3747:
**Problem**
A bot turn that triggers interactive tool approval blocks inside RunTurn waiting for ctrl.Approve. The only code path that calls those is handleSlashCommand, which runs on the same per-adapter dispatchLoop goroutine that drives runTurn inline.

So once a turn blocks on approval, the dispatch loop can never read the /approve reply that would unblock it.

**Fix**
Run runTurn on its own goroutine: go gw.runTurn(ctx, binding.Adapter, key, msg)

Per-session serialization is unchanged (session lock). Also initialize render sink fully before publishing.

Fixes #4402 #4701 #4863

### DIFF
--- a/internal/bot/gateway.go
+++ b/internal/bot/gateway.go
@@ -336,7 +336,15 @@ func (gw *BotGateway) handleMessage(ctx context.Context, binding AdapterBinding,
 		return
 	}
 
-	gw.runTurn(ctx, binding.Adapter, key, msg)
+	// Run the turn on its own goroutine so the dispatch loop stays free to read
+	// the next inbound message. A turn that hits interactive approval/ask blocks
+	// inside RunTurn waiting for ctrl.Approve/AnswerQuestion — and the ONLY path
+	// that calls those is handleSlashCommand on this same dispatch goroutine. Run
+	// it inline and the loop can never deliver the /approve (or card) reply that
+	// would unblock it: the session wedges until restart (#4701, #4863, #4402).
+	// Per-session serialization is still held by the session lock (active[key]),
+	// which the deferred Release inside runTurn clears.
+	go gw.runTurn(ctx, binding.Adapter, key, msg)
 }
 
 func (gw *BotGateway) addPendingReaction(ctx context.Context, plat Platform, adapter Adapter, msg InboundMessage) {
@@ -821,6 +829,9 @@ func (gw *BotGateway) runTurn(ctx context.Context, adapter Adapter, key string, 
 			gw.mu.Unlock()
 		},
 	)
+	// Finish initializing the sink before publishing it as the live target: once
+	// setTarget runs, other goroutines can reach this sink via state.sink.Emit.
+	sink.ctrl = state.ctrl
 	state.sink.setTarget(sink)
 	defer state.sink.setTarget(nil)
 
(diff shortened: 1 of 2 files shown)


SivanCola (approved):
Approved. Well-scoped fix for a real deadlock, and I added a regression test to guard the contract.

**The bug is real.** `ctrl.Approve` / `ctrl.AnswerQuestion` are only ever called from `handleSlashCommand`, which runs on the same per-adapter `dispatchLoop` goroutine that drives `runTurn` inline. So once a turn parks inside `RunTurn` waiting on interactive approval, the dispatch loop is blocked in that same call and can never read the `/approve` (or card) reply that would unblock it — the session wedges until restart (#4402, #4701, #4863). Moving `runTurn` onto its own goroutine keeps the loop free to deliver the reply.

**Concurrency is still safe.** Per-session serialization is held by `SessionManager.active[key]`, not by the dispatch goroutine, and the deferred `Release(key)` inside `runTurn` clears it (and drains the pending queue). Non-bypass messages for a busy session still merge/queue via `TryAcquire`, so one goroutine per active session per key is preserved.

**On the second hunk** (moving `sink.ctrl = state.ctrl` before `setTarget`): `renderSink.ctrl` is currently only ever written, never read, so this is a harmless no-op today. It's a reasonable defensive ordering for the moment a reader appears, but the commit message's "race" framing slightly overstates the current code. No objection to keeping it.

**Cache/security:** no impact. The diff only touches bot turn scheduling — no system prompt, tool list/schema, provider serialization, compaction, or reasoning-upload surface, and no new deps, network, file, or command-execution paths.

**What I tightened:** there was no test exercising the deadlock path. I added `TestGatewayApprovalReplyUnblocksTurnOffDispatchGoroutine`, which drives the real `dispatchLoop` with a fake controller whose `RunTurn` parks until `Approve` is called, then sends a normal turn followed by `/approve`. I verified it fails against the inline implementation ("approval reply was never delivered: dispatch loop is wedged") and passes with the fix.

Verified:
- `go test -race ./internal/bot/` passed
- New test fails on the pre-fix inline `runTurn`, passes on `go gw.runTurn(...)`
- `gofmt -l` clean, `git diff --check` clean

Issue #5329, opened 2026-06-26
[Bug]: 无限发送对话30轮


eghrhegpe:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.12.0

### What happened?

太恐怖了，桌面端，多开任务的时候，突然发现有一个窗口频繁发送对话框。

### Steps to reproduce

🟡 潜在问题：多 .pmx 并发加载 main.tsL179-L181 for (const path of paths) { handleDropFi...
5 个工具

[已粘贴文本 #45 · 71 行]
9 个工具

记录文档

记录工作到docs/status.md

记录工作到docs/status.md
6 个工具

[已粘贴文本 #1 · 25 行]
56 个工具

[已粘贴文本 #1 · 108 行]
32 个工具

<memory-compiler-execution> {"type":"memory_v5_execution_contract","instructi...

<memory-compiler-execution> {"type":"memory_v5_execution_contract","instructi...

<memory-compiler-execution> {"type":"memory_v5_execution_contract","instructi...

<memory-compiler-execution> {"type":"memory_v5_execution_contract","instructi...

<memory-compiler-execution> {"type":"memory_v5_execution_contract","instructi...

<memory-compiler-execution> {"type":"memory_v5_execution_contract","instructi...

<memory-compiler-execution> {"type":"memory_v5_execution_contract","instructi...
30+重复发送
<memory-compiler-execution>
[...]

PR #5331, opened 2026-06-26
fix(agent): skip Memory v5 compiler on goal continuation turns (fixes infinite loop #5325)

eghrhegpe:
## Summary

Memory v5 execution compiler was re-compiling synthetic goal continuation turns on every loop iteration, producing the same `memory_v5_execution_contract` JSON repeatedly. Some models echoed this JSON back as text, causing an infinite loop where the same compiled contract was generated, fed to the model, echoed back, and re-compiled 30+ times.

## Root Cause

In `internal/agent/agent.go`, the `Run()` method calls `memCompiler.StartTurn()` for every user turn, including goal continuation turns whose input is always `"Continue pursuing the active goal..."`. The Memory v5 compiler sees the same input every time, produces the same execution contract, and the model sometimes echoes the JSON format back as assistant text, creating an infinite loop.

## Fix

Add `isGoalLoopContinue()` check in `agent.go` to skip the Memory v5 compiler on turns whose leading text matches the synthetic goal continuation instruction. The compiled contract is now only injected on user-authored tasks, not on goal loop iterations.

## Changes

- `internal/agent/agent.go`: Add `isGoalLoopContinue()` helper, wrap `StartTurn()` call with check

## Testing

- `go test ./internal/agent/` — all tests pass
- `go vet ./internal/agent/` — clean
- `gofmt` — clean

Cache-impact: low — only affects agent.go's Run() loop path for goal continuation turns
Cache-guard: `go test ./internal/agent/` passes

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -358,6 +358,15 @@ func (a *Agent) SetResponseLanguage(lang string) {
 // It is safe for desktop settings to call while other tabs are idle or running;
 // an already-started turn keeps its own Turn handle and future turns observe the
 // new runtime.
+// isGoalLoopContinue detects synthetic goal continuation turns so the
+// Memory v5 compiler can skip them — re-compiling the same synthetic
+// "Continue pursuing the active goal..." instruction on every loop
+// produces the same execution contract JSON, which some models echo back
+// as text, creating an infinite loop (#5325).
+func isGoalLoopContinue(input string) bool {
+	return strings.HasPrefix(strings.TrimSpace(input), "Continue pursuing the active goal")
+}
+
 func (a *Agent) SetMemoryCompiler(rt *memorycompiler.Runtime) {
 	if a == nil {
 		return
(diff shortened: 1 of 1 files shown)


SivanCola:
fix in PR #5387: fix(memory-v5): stop synthetic turns from spinning the goal loop (#5342, #5329)

SivanCola:
Follow-up after the integration pass: this PR was reviewed while consolidating the Memory v5 / approved-plan fixes in #5391.

The core diagnosis here was correct: synthetic goal-continuation turns should not be compiled into a fresh `memory_v5_execution_contract` on every loop iteration. We did not merge this exact diff because the final fix landed in #5387 at the producer/orchestrator layer instead: synthetic turns are marked with `agent.WithMemoryCompilerSkip(ctx)` by the turn orchestrator, using the existing synthetic-message registry, and `Agent.Run` honors that skip. That covers goal continuation, plan-approved execution, and future synthetic turns without hard-coding one English prefix in `agent.go`.

#5391 then built on that merged base and fixed the remaining desktop restore/noise paths. This PR is credited in #5391 as a reviewed-but-not-adopted/superseded contribution. Keeping this PR closed is the right final state; thanks for the early root-cause work.

PR #5347, opened 2026-06-26
fix(cli): folded paste content was invisible to the model

liugh-dev:
## Summary
A folded multi-line paste ([Pasted text #N · M lines]) was never seen by the model — only the placeholder label was. The CLI passed the folded label (the un-expanded `line`) as the controller's `raw` value. The Memory v5 compiler (enabled by default) takes `raw` as the source_event of its execution contract, and that contract replaces the whole user turn — so the model's only view of the paste became the label, with the actual content lost.

## Verification

Copy multi-line text and paste into the input erea, then ask the agent how many lines it can see. If it can't see any pasted text, the bug exists.

## Cache impact

Cache-impact: TODO
Cache-guard: TODO
System-prompt-review: N/A

### DIFF
--- a/internal/cli/chat_tui.go
+++ b/internal/cli/chat_tui.go
@@ -1188,7 +1188,12 @@ func (m chatTUI) update(msg tea.Msg) (tea.Model, tea.Cmd) {
 				return m, finalize(m, cmds)
 			}
 
-			cmds = append(cmds, m.startTurnWithRaw(sentLine, sentLine, line, line))
+			// `raw` is the un-resolved user prompt used for auto-plan scoring AND the
+			// memory compiler's source_event. It must be the EXPANDED paste content
+			// (sentLine), not the folded label (line) — otherwise the memory compiler's
+			// execution contract replaces the user turn with one whose source_event is
+			// just the placeholder label, and the model never sees the pasted content.
+			cmds = append(cmds, m.startTurnWithRaw(sentLine, sentLine, line, sentLine))
 			return m, finalize(m, cmds)
 		}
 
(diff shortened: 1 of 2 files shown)


SivanCola (approved):
Approved. This is a well-scoped fix for the folded-paste regression.

The important part is that the controller `raw` value now carries the expanded paste text instead of the visible folded placeholder, so Memory v5's `source_event` no longer loses the actual pasted content. The added regression captures both the runner input and the memory-compiler source input, which is the right contract to pin down here.

Cache impact: low. This changes the dynamic user-turn source text for pasted content only; it does not alter the stable provider-visible prefix, tool schemas, system prompt, or request serialization.

Verified:
- `go test ./internal/cli -count=1` passed with this PR stacked alongside the current non-draft TUI changes that touch the same file.
- The PR cherry-picks cleanly onto the latest `main-v2` together with #5349 and #5378.

SivanCola:
Pushed f6b7b42 to fix the failing `race` check.

Root cause: `TestPasteFoldExpandOnSubmit` waited with a fixed sleep and then read the recording runner output while the controller goroutine could still be writing it under `go test -race`.

Fix: the test now wires a test event sink and waits for `event.TurnDone` before reading the recorded runner input. Runtime behavior is unchanged; this only gives the async test path a real happens-before edge.

Verified:
- `go test -race ./internal/cli -run TestPasteFoldExpandOnSubmit -count=20`
- `go test -race ./internal/cli -count=1`
- `go test ./internal/cli -count=1`
- `go test -race ./...`
- Clean merge check against latest `main-v2`, plus `go test -race ./internal/cli -count=1`

PR #5349, opened 2026-06-26
feat(tui): add /copy and /export commands

liugh-dev:
## Summary

- `/copy` opens a modal picker (up/down, Enter, Esc) to select an assistant response from the current turn and copy it to the clipboard.
- `/copy N` copies the Nth-newest assistant response from the current turn directly.
- `/export` saves the full session as a Markdown file, merging consecutive same-role messages and stripping thinking/tool calls.
- `/export` strips compose prefixes and expanded referenced-context payloads before writing user messages to Markdown.
- `/export` reports an empty export instead of writing a header-only Markdown file when no visible user or assistant content remains after filtering.
- Both commands read visible `Content` only; `ReasoningContent`, `ToolCalls`, and system/tool messages are excluded.
- Add i18n strings for en, zh-Simplified, zh-Traditional.

## Verification

- `go test ./internal/cli -run 'TestSlashCopy|TestSlashExport'`
- `go test ./internal/cli`

## Cache impact

Cache-impact: low - CLI TUI copy/export/display behavior only; no provider request serialization, stable prompt prefix, or tool schema bytes changed.
Cache-guard: `go test ./internal/cli -run 'TestSlashCopy|TestSlashExport'` and `go test ./internal/cli`
System-prompt-review: N/A


SivanCola (changes requested):
Thanks, the `/copy` and `/export` direction is useful, but I found one blocker before this can merge.

`/export` currently renders directly from `Controller.History()` and only strips compose prefixes from user messages. For turns that used `@` references, the stored provider history can contain the expanded `Referenced context:` preamble plus referenced file/resource contents. Exporting that directly means the Markdown file can include reference payloads instead of only the user's visible prompt. Please route exported user text through the same display-safe stripping used for referenced context, for example by applying `control.StripReferencedContextPrefix` in addition to `StripComposePrefixes`, and add a regression test proving exported Markdown does not include the expanded reference block or referenced file contents.

Please also run `gofmt`; the new i18n field alignment currently leaves these files unformatted:
- `internal/i18n/i18n.go`
- `internal/i18n/messages_en.go`
- `internal/i18n/messages_zh.go`
- `internal/i18n/messages_zh_tw.go`

The current coverage only updates slash completion. Please add focused coverage for the new behavior, at least for `/copy` selection/direct index and `/export` filtering of system/tool/internal referenced-context content.

SivanCola:
Addressed in `63575c9e0` and follow-up `dde60a424`.

What changed:
- `/export` now strips compose prefixes and expanded referenced-context preambles before writing user messages to Markdown.
- `/export` now reports an empty export instead of writing a header-only Markdown file when no visible user or assistant content remains after filtering.
- Added focused coverage for `/copy` picker selection, `/copy N` direct index, `/export` filtering of system, tool, reasoning, tool-call, and referenced-context payloads, plus system-only and filtered-only empty export cases.
- Ran `gofmt` on the i18n files called out in review.

Verified:
- `go test ./internal/cli -run 'TestSlashCopy|TestSlashExport'`
- `go test ./internal/cli`

Cache impact remains low: this is CLI export/display behavior only; no provider request serialization, stable prompt prefix, or tool schema bytes changed.

SivanCola (approved):
Approved. The earlier export blocker is addressed: user messages are display-safe before Markdown export, referenced-context payloads are filtered out, and empty/fully filtered sessions no longer create header-only export files.

Coverage now includes `/copy` picker selection, `/copy N`, `/export` filtering of system/tool/reasoning/tool-call/referenced-context payloads, and empty export cases. Verified with `go test ./internal/cli -run 'TestSlashCopy|TestSlashExport'` and `go test ./internal/cli`.

Cache impact remains low because the changes are limited to CLI TUI copy/export/display behavior and do not affect provider request serialization, stable prompt prefix, or tool schemas.

PR #5355, opened 2026-06-26
fix(desktop): paginate restored transcript warm turns

GTC2080:
## Summary

- Fix restored transcript warm/cold pagination so long histories initially render only the latest warm page instead of every older warm turn.
- Reuse the transcript grouping helper layer for the pagination math and cover the boundary with a regression test.
- Reduce Desktop startup/topic-open render work for large restored transcripts.

## Verification

- PASS `corepack pnpm --dir desktop/frontend exec tsx src/__tests__/transcript-grouping.test.ts`
- PASS `corepack pnpm --dir desktop/frontend exec tsx src/__tests__/app-chrome-tabs.test.ts`
- FAIL (existing) `corepack pnpm --dir desktop/frontend test:typecheck` — `src/__tests__/capabilities-panel-actions.test.ts(127,5): error TS2322: Type '"auto"' is not assignable to type 'Mode'.`

## Cache impact

Cache-impact: none, UI-only transcript rendering change; no provider prompt or cache-prefix behavior changes.
Cache-guard: `corepack pnpm --dir desktop/frontend exec tsx src/__tests__/transcript-grouping.test.ts`
System-prompt-review: N/A

Refs #5306

### DIFF
--- a/desktop/frontend/src/__tests__/transcript-grouping.test.ts
+++ b/desktop/frontend/src/__tests__/transcript-grouping.test.ts
@@ -3,7 +3,7 @@
 import { readFileSync } from "node:fs";
 import { fileURLToPath } from "node:url";
 import { dirname, resolve } from "node:path";
-import { buildTurnGroups } from "../lib/transcriptGrouping";
+import { buildTurnGroups, warmPagination } from "../lib/transcriptGrouping";
 import type { Item } from "../lib/useController";
 
 let passed = 0;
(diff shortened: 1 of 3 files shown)


SivanCola (changes requested):
Thanks, this is directionally the right fix for the restored-transcript render spike, but one state-reset issue still blocks merging.

Before merging, please reset the warm/cold pagination state when the rendered session changes. `Transcript` is explicitly reused across tab switches rather than keyed by `tabId`, but `coldPage` and `expandedWarmTurns` are component-local state. After a user clicks "show earlier history" a few times in one long transcript, switching/opening another long restored transcript reuses that non-zero `coldPage`. With the new `warmPagination` semantics, `coldPage = 3` on a 100-turn transcript makes `warmStartTurn = 0`, so the next session initially renders all 70 non-hot warm turns again and loses the performance guarantee this PR is trying to add. Leaked `expandedWarmTurns` can also expand unrelated turn indexes in the next session.

A small reset keyed by the session identity/tab reveal boundary, plus a regression test that simulates loading earlier history and then changing sessions, should cover this.

Verified during review:
- `corepack pnpm -C desktop/frontend --ignore-workspace exec tsx src/__tests__/transcript-grouping.test.ts` passed
- `corepack pnpm -C desktop/frontend --ignore-workspace exec tsx src/__tests__/app-chrome-tabs.test.ts` passed
- `git diff --check origin/main-v2...origin/pr/5355` passed
- Latest `origin/main-v2` has no overlapping changes to the three touched files, and the merge-tree applies cleanly

Once this state reset is addressed and CI stays green, I will consider merging it into `main-v2`.

SivanCola (approved):
Approved. This PR is now a well-scoped fix for the restored-transcript render spike.

The original pagination change correctly limits initial warm-turn rendering for long restored transcripts. I added two follow-up fixes during review:
- `667451b` resets warm/cold pagination state across session/tab boundaries, so an expanded old session cannot make the next long transcript render too much history on first paint.
- `5ccacbc` reveals the required warm page before QuestionJumpBar scrolls to an older cold-zone question, so jumps to earlier questions still work after the render window is tightened.

- The pagination state is now scoped by the rendered transcript identity instead of being leaked through the reused `Transcript` component instance.
- Jump navigation asks the warm/cold pagination owner to include the target turn before touching the DOM, so it preserves the performance optimization without breaking navigation.

Verified:
- `corepack pnpm -C desktop/frontend --ignore-workspace exec tsx src/__tests__/transcript-grouping.test.ts` passed
- `corepack pnpm -C desktop/frontend --ignore-workspace exec tsx src/__tests__/app-chrome-tabs.test.ts` passed
- `corepack pnpm -C desktop/frontend --ignore-workspace exec tsc --noEmit` passed
- `git diff --check` passed
- GitHub Actions are green for the latest head SHA

PR #5360, opened 2026-06-26
fix(agent): avoid Memory v5 goal loop injection

### LINKED ISSUE #5342: [Bug]: Memory v5 功能缺陷
### Version line

### Exact version

1.12.0

### What happened?

这个功能是否有点恐怖，开启后我在对话里面安排一项任务，它不停的出来捣乱，会话我只说了2句，他一口气让Reasonix的回答干到了53轮，Token不消耗吗？而且它参与进来所提的都与项目没多大关联，直接导致工作无法进行。它可能有两个严重问题：

它选的策略完全不相关——比如一直推荐 frontend-visual-verify（前端 UI 审查），跟我的项目这个后端项目没有关系
它有死循环 Bug——每次你发一条消息，它就注入一次，收集到的反馈又触发下一次注入

### Steps to reproduce

1、打开设置，选择通用；2、找到Memory v5，点击“开启”；3、回到会话进行提问。4、然后会话就开始无限制的自言自语，提示了很多与项目不相关的内容，不停止会话，他就一直自我对话下去。

### OS / platform

Windows Server 2025

### Relevant logs or output

```shell

```

GTC2080:
## Summary

- Fixes Memory v5 compiling synthetic goal-continuation turns into repeated `memory_v5_execution_contract` user turns.
- Fixes short strategy preconditions like `ui` so they match tokens only, preventing `pursuing` from selecting `frontend-visual-verify`.
- Closes #5342.

## Verification

- `go test ./internal/agent ./internal/memorycompiler -count=1`
- `go test ./internal/control -run 'TestIsSyntheticUserMessage|TestGoal|TestTurnOrchestrator|TestAutoPlan' -count=1`
- `git diff --check`

## Cache impact

Cache-impact: low — only changes provider-visible Memory v5 behavior for synthetic goal-continuation turns and short-token strategy matching; system prompt, tool schemas, and provider prefix are unchanged.
Cache-guard: focused agent, memorycompiler, and control tests listed above.
System-prompt-review: N/A

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -681,7 +681,9 @@ func (a *Agent) Run(ctx context.Context, input string) (runErr error) {
 		memoryCompilerInput = sourceInput
 	}
 	input = a.withTurnPreferences(rawInput)
-	if memCompiler := a.memoryCompilerRuntime(); memCompiler != nil {
+	memoryCompilerDisplayInput := strings.TrimSpace(StripTransientUserBlocks(rawInput))
+	skipMemoryCompiler := strings.HasPrefix(memoryCompilerDisplayInput, "Continue pursuing the active goal.")
+	if memCompiler := a.memoryCompilerRuntime(); memCompiler != nil && !skipMemoryCompiler {
 		if compiledInput, turn := memCompiler.StartTurn(ctx, memoryCompilerInput, a.session.Snapshot()); turn != nil {
 			a.compilerTurn = turn
 			a.emitMemoryCompilerStats(turn)
--- a/internal/agent/loop_e2e_test.go
+++ b/internal/agent/loop_e2e_test.go
@@ -91,6 +91,32 @@ func TestRunUsesMemoryCompilerContractAsUserTurn(t *testing.T) {
 	}
 }
 
+func TestRunSkipsMemoryCompilerForGoalContinuationTurn(t *testing.T) {
+	rt := memorycompiler.New(t.TempDir())
+	_, seed := rt.StartTurn(context.Background(), "fix a bug", nil)
+	seed.RecordToolResults([]memorycompiler.ToolRecord{
+		{Name: "bash", Error: "exit status 1"},
+		{Name: "bash", Error: "exit status 1"},
+	})
+	seed.Finish(nil)
+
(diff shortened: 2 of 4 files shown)


esengine:
Thanks for diagnosing this and pinpointing both root causes — the goal-loop contract re-injection and the bare-substring `ui` mismatch.

A more general fix for the same two issues has just been merged in #5387 (closes #5342 and #5329):

- **Synthetic-turn skip.** Instead of special-casing the `"Continue pursuing the active goal."` prefix inside `agent.go`, #5387 marks the turn at the layer that already knows it is synthetic — the turn orchestrator — via a new `agent.WithMemoryCompilerSkip(ctx)`, keyed off the existing canonical registry `control.IsSyntheticUserMessage` / `syntheticPrefixes`. That registry already lists the goal-loop messages (and plan-approved execution, stream-recovery retries, etc.), so this skips compilation for **all** synthetic turns uniformly and avoids hard-coding the message text / coupling the agent layer to `control`’s wording.
- **Token matching.** Your `strategyPreconditionMatches` idea was adopted: short preconditions (≤2 runes, e.g. `ui`) now match only on whole-token boundaries, so `pursuing`/`build`/`quiet` no longer route to `frontend-visual-verify`.

Regression tests cover both the agent gate and the orchestrator marking, plus the `ui` token boundary.

Since #5387 supersedes the changes here, this PR can probably be closed — leaving that to @esengine. Thanks again for the clear root-cause writeup, which made the generalized fix straightforward.

SivanCola:
Follow-up after the integration pass: this PR was reviewed again while consolidating the related Memory v5 / approved-plan issues in #5391.

The investigation here materially shaped the final fix: it identified both the synthetic goal-continuation compilation loop and the short-token strategy matching issue (`ui` matching inside words like `pursuing`). The final merged implementation landed in #5387, with two adjustments:

- The Memory v5 skip is applied by the turn orchestrator for all registered synthetic turns via `agent.WithMemoryCompilerSkip(ctx)`, rather than only checking the goal-continuation prefix inside `agent.go`.
- The short-precondition token-boundary behavior from this PR's direction was adopted so `ui` no longer matches inside unrelated words.

#5391 then built on #5387 and fixed the remaining desktop restore/noise paths. This PR is credited in #5391 as a reviewed-but-not-adopted/superseded contribution. Keeping this PR closed is the right final state; thank you for the clear diagnosis and test direction.

PR #5362, opened 2026-06-26
fix(boot): hint migration when api_key_env holds a literal key value

### LINKED ISSUE #5325: [Bug]: 自建API接入异常
### Version line

### Exact version

1.11.1

### What happened?

公司内，自建服务器，从1.11.0正常，1.11.1、1.12开始，API接入异常，设置 刷新模型 无法识别key，
接入显示：填写密钥后才可选择这些模型。

model "**" is selected but its API key ***** is not set — requests will fail until you set it。
实际已经填写key，安装恢复至1.11正常

### Steps to reproduce

公司内，自建服务器，从1.11.0正常，1.11.1、1.12开始，API接入异常，设置 刷新模型 无法识别key，
接入显示：填写密钥后才可选择这些模型。

model "**" is selected but its API key ***** is not set — requests will fail until you set it。
实际已经填写key，安装恢复至1.11正常

### OS / platform

Win11

### Relevant logs or output

```shell

```

aznikline:
Resolves #5325.

Since 1.11, api_key_env resolves strictly as an env-var or credentials name — the pre-1.11 fallback that treated the field's value as the key itself was removed to keep secrets out of the config file. Configs carried over from 1.10 (or written assuming the field took the key) now resolve to an empty APIKey and surface only the generic 'key not set' notice, which does not point at the real cause.

This adds a conservative LooksLikeLiteralAPIKeyValue() heuristic (long + lowercase/punctuated, or a known provider prefix like sk-/AIza/sk-ant) and, when it fires, emits a targeted migration hint naming the 1.11 change and a suggested credential name — instead of the generic message. It does NOT restore the literal-value fallback, which was an intentional security tightening.

Changes:
- config: ProviderEntry.LooksLikeLiteralAPIKeyValue() + nil-safe
- boot: branch the missing-key notice; suggestAPIKeyEnvName() helper
- tests: 11 heuristic cases + a Build-level notice test

Verification: go build ./... ok; go test ./internal/config/... ./internal/boot/... ok (incl. new TestBuildNoticesLiteralAPIKeyValue). gofmt clean.

---

Cache-impact: none - the new branch runs only inside the existing `if entry.RequiresAPIKey() && entry.APIKey() == ""` startup notice path; it changes no tool schema, no prompt assembly, no message structure, so the model-facing cache shape is unaffected.
Cache-guard: TestBuildNoticesMissingAPIKey (existing, generic unset path) + TestBuildNoticesLiteralAPIKeyValue (new, literal-value hint path) together guard the notice branch — the generic message still fires for real env-var names, the migration hint only for literal values.
System-prompt-review: not applicable — this PR changes only a human-facing startup notice emitted via sink.Emit; it touches no system prompt, no tool schema, no model-facing content. boot.go is flagged as system-prompt-sensitive by the guard because adjacent code assembles prompts, but the changed lines are notice-only.

### DIFF
--- a/internal/boot/boot.go
+++ b/internal/boot/boot.go
@@ -236,7 +236,16 @@ func Build(ctx context.Context, opts Options) (*control.Controller, error) {
 	// (RequireKey is false so the UI stays reachable) and then fail silently on the
 	// first request, showing as an empty/dead model. Surface the cause up front.
 	if !opts.RequireKey && entry.RequiresAPIKey() && entry.APIKey() == "" {
-		sink.Emit(event.Event{Kind: event.Notice, Text: "Selected model is missing its API key.", Detail: fmt.Sprintf("model %q is selected but its API key %s is not set — requests will fail until you set it", modelName, entry.APIKeyEnv)})
+		if entry.LooksLikeLiteralAPIKeyValue() {
+			// Before 1.11, api_key_env tolerated a literal key value placed
+			// directly in the field; that fallback was removed for security.
+			// A config carried over from 1.10 (or written assuming the field
+			// took the key itself) now resolves empty — point the user at the
+			// migration rather than the generic "not set" message.
+			sink.Emit(event.Event{Kind: event.Notice, Text: fmt.Sprintf("model %q: api_key_env looks like a raw key value, not an environment-variable or credential name. Since 1.11, api_key_env must name an env var or a credentials entry (the literal key is no longer accepted in config). Move the value into your credentials file under a name like %s and set api_key_env to that name.", modelName, suggestAPIKeyEnvName(entry.APIKeyEnv))})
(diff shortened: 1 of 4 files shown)


aznikline:
Rebased onto `main-v2` to clear the merge conflict. The only content conflict was in `internal/boot/boot.go`, where both this branch and `main-v2` had touched the "API key not set" notice — resolved by keeping the literal-value detection branch (the point of this PR) and adopting `main-v2`'s `Text`/`Detail` split for the generic path.

Verified locally: `go build ./...` ok, `gofmt` clean, `go test ./internal/config/ ./internal/boot/` green (incl. `TestBuildNoticesLiteralAPIKeyValue` and `TestBuildNoticesMissingAPIKey`). Closes #5325.

PR #5376, opened 2026-06-26
Feature/20260626/fix memory v5 pipeline pollution

myipanta:
## Summary

-

## Verification

-

## Cache impact

Cache-impact: low — adds new methods to Agent struct, no system-prompt or tool-schema changes
Cache-guard: go test ./internal/agent/ -run "TestShouldInject|TestMarkCompiler|TestRunUsesMemoryCompiler" -count=1
System-prompt-review: N/A

### DIFF
--- a/desktop/frontend/src/__tests__/memory-compiler-display.test.ts
+++ b/desktop/frontend/src/__tests__/memory-compiler-display.test.ts
@@ -1,44 +1,45 @@
-// Run: tsx src/__tests__/memory-compiler-display.test.ts
-//
-// Display-boundary safety net for #5361: a corrupted/accreted Memory v5 contract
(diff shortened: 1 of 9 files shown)


SivanCola:
Thanks @myipanta for investigating the Memory v5 pipeline pollution/noise path.

I'm closing this PR as superseded by #5493. The selected fix keeps the same product goal but uses the newer Memory v5 path: task/chat classification gates Memory v5 startup, simple greetings and acknowledgements skip execution-contract injection, real task/failure descriptions still use Memory v5, and compiler citations are kept out of the visible transcript. #5493 also has focused backend/frontend coverage and is clean against the current `main-v2` after #5487.

We are not adopting this branch because it is based on an older version of the flow and leaves more of the raw Memory v5 contract/UI behavior exposed than the selected approach. Thanks again for the exploration here.

PR #5448, opened 2026-06-27
feat(plugin): add disabled_tools filter for MCP servers

### LINKED ISSUE #5444: [Feature]: feat: per-plugin `disabled_tools` filter for MCP servers
### Version line

MCP servers often expose more tools than a workspace wants the model to see. Today there is no way to suppress individual tools: every tool the server advertises in `tools/list` is registered and surfaced to the model, counted toward the token budget, and shown in `/mcp`. Users work around this by editing server config or running patched servers, which is brittle and leaks unwanted tool definitions into the prompt prefix.

### Proposed solution

Add a per-plugin `disabled_tools` field to `[[plugins]]` config:

```toml
[[plugins]]
name = "my-server"
command = "my-server"
disabled_tools = ["write_file", "delete_file"]
```

Names use the raw form the server reports in `tools/list` (e.g. `"write_file"`), not the namespaced `"mcp__<server>__<tool>"` form. Disabled tools are filtered out before they reach the registry, so they never enter the model context, appear in `/mcp`, or count toward tokens. Empty/nil preserves current behavior (no filtering) for backward compatibility.

### Follow-up: config persistence gap

The initial implementation missed both TOML rendering paths (`RenderTOMLForScope` and `RenderTOMLProjectDelta`), so any config rewrite (e.g. via `/model` or `/theme`) would silently drop `disabled_tools` from disk. This was fixed in the PR #5448

ADX15xs:
Closed #5444 

## Summary

- Add per-plugin `disabled_tools` config field: a list of raw MCP tool names (as reported by the server in `tools/list`) to exclude from registration. Disabled tools never enter the model context, never appear in `/mcp`, and never count toward token budgets.
- Filter is applied at both tool-loading paths — `listTools` (eager, post-handshake) and `LazyToolset` (cache-hit, pre-handshake) — so every downstream consumer (registry, on-disk schema cache, `Host.ToolsFor`) sees a consistent view.
- `disabled_tools` is folded into `SpecFingerprint` so changing the list invalidates the on-disk schema cache (next launch re-handshakes instead of serving stale tools).
- Backward compatible: `nil`/empty `disabled_tools` produces the same fingerprint and behavior as a pre-field Spec, so existing caches keep hitting.

## Verification

- `go vet ./internal/plugin/... ./internal/config/... ./internal/boot/...` — clean.
- `go build ./...` — clean.
- `gofmt -l` on changed Go files — empty (clean).
- `go test ./internal/plugin/ ./internal/config/` — `ok reasonix/internal/plugin`, `ok reasonix/internal/config`.
- New tests: `TestPluginEntryTOMLDisabledTools`, `TestPluginEntryTOMLDisabledToolsEmpty`, `TestListToolsFiltersDisabledTools` (incl. short-circuit cache path + `c.tools`), `TestIsToolDisabledNilAndEmpty`, `TestSpecFingerprintNilVsEmptyDisabledTools`, `TestSpecFingerprintWithoutDisabledTools`, `TestLazyToolsetCacheHitFiltersDisabledTools`, `TestCacheInvalidatesWhenDisabledToolsChanges`.

## Cache impact

Cache-impact: low — `disabled_tools` is added to `SpecFingerprint`. Toggling the list changes the fingerprint and correctly invalidates the on-disk schema cache (forcing a re-handshake). Existing configs without `disabled_tools` keep the same fingerprint as before the field existed, so warm caches are preserved.

Cache-guard: `TestSpecFingerprintNilVsEmptyDisabledTools` (nil vs empty parity) + `TestCacheInvalidatesWhenDisabledToolsChanges` (fingerprint changes → cache miss) + `TestSpecFingerprintWithoutDisabledTools` (omitted field = old-struct parity).

System-prompt-review: Pending review from @esengine  — this PR modifies boot/config for tool filtering; registry-level change only, no prompt-prefix content alterations. Awaiting explicit sign-off.
The changes for sysPrompt assembly in boot.Build — `disabled_tools` acts on the tool registry (`listTools`/`LazyToolset` → `reg.Add`), downstream of and separate from `ResolveSystemPromptForRoot` + `memory.Compose` + `skill.ApplyIndex` + `outputstyle.Apply`; no provider-visible prompt-prefix content changes.

### DIFF
--- a/internal/boot/boot.go
+++ b/internal/boot/boot.go
@@ -1486,6 +1486,15 @@ func PluginSpecsForRootWithOptions(entries []config.PluginEntry, workspaceRoot s
 
 func pluginSpecFromEntryWithOptions(e config.PluginEntry, workspaceRoot string, opts PluginSpecOptions) plugin.Spec {
 	e = e.ExpandedPlugin() // resolve ${VAR} / ${VAR:-default} from the environment
+	var disabled map[string]bool
+	if len(e.DisabledTools) > 0 {
+		disabled = make(map[string]bool, len(e.DisabledTools))
+		for _, name := range e.DisabledTools {
+			if name = strings.TrimSpace(name); name != "" {
+				disabled[name] = true
+			}
+		}
+	}
 	return plugin.ApplyKnownOverrides(plugin.Spec{
 		Name:               e.Name,
 		Type:               e.Type,
@@ -1498,6 +1507,7 @@ func pluginSpecFromEntryWithOptions(e config.PluginEntry, workspaceRoot string, 
 		CallTimeout:        secondsDuration(e.CallTimeoutSeconds),
 		ToolTimeouts:       toolTimeoutDurations(e.ToolTimeoutSeconds),
 		ReadOnlyToolNames:  trustedRawReadOnlyToolNames(e.TrustedReadOnlyTools),
+		DisabledTools:      disabled,
 	}, workspaceRoot)
 }
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1316,6 +1316,12 @@ type PluginEntry struct {
 	// AutoStart controls whether the server connects during session startup.
 	// Nil preserves historical behavior: configured servers start automatically.
 	AutoStart *bool `toml:"auto_start"`
+	// DisabledTools lists raw MCP tool names (as reported by the server in
+	// tools/list) to exclude from registration. Disabled tools never enter the
+	// model context, never count toward token budgets, and never appear in
+	// /mcp output. Names use the raw form ("read_file"), not the namespaced
+	// "mcp__<server>__<tool>" form. Empty/nil = no filtering (backward compat).
+	DisabledTools []string `toml:"disabled_tools"`
 	// Tier is a legacy compatibility field. New config rendering omits it; enabled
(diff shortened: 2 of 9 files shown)


ADX15xs:
@esengine @SivanCola Could you please review the system-prompt changes in boot.go when you have a moment? This PR is ready for your explicit sign-off. Thanks

ADX15xs:
## Fix: TOML rendering (60c56c9d)

The initial implementation (54048f68) added the struct field and filtering logic but missed both TOML rendering paths (`RenderTOMLForScope` and `RenderTOMLProjectDelta`). Any config rewrite triggered by `/model`, `/theme`, or `saveProjectIncremental` would silently drop `disabled_tools` from the on-disk config.

> The field is readable in Go code but disappears after `/model` — the struct was updated but `RenderTOMLForScope` was not, so a full config rewrite drops it.

**Root cause**: The `[[plugins]]` loops in both render functions output each plugin field with explicit `fmt.Fprintf` calls. Since `disabled_tools` was not added to these loops, neither path persisted the field.

**Fix**: Added `disabled_tools` rendering blocks after `trusted_read_only_tools` in both `[[plugins]]` loops within `render.go` (`RenderTOMLForScope` and `RenderTOMLProjectDelta`).

PR #5450, opened 2026-06-27
fix(memory,security): fix 13 bugs (5 CRITICAL, 8 HIGH) in memory, sandbox, permission, agent

bfxh:
﻿## Summary

Deep code review of the memory system, sandbox, permission layer, and agent core. Found and fixed 13 bugs (5 CRITICAL, 8 HIGH) via static analysis, logical review, and dynamic testing. All existing tests pass.

### 1. memory: compounding confidence decay zeroes all memories within weeks
`applyDriftControl` wrote the decayed confidence back to `node.Confidence` while never advancing `node.Timestamp`. Each subsequent turn recomputed decay from the original timestamp against the now-lower base, producing exponential decay (`0.95^(n(n+1)/2)` instead of `0.95^n`). After ~20 turns, all non-TruthLocked memories fell below `staleConfidenceThreshold` and were marked `QualityNoise`.

**Fix**: Compute decayed confidence for immediate use only; do not write it back.

### 2. memory: flushIndexIn silently drops hand-edited MEMORY.md content
`flushIndexIn` rewrote MEMORY.md from the `lines` map alone, discarding headings, paragraphs, sub-lists, and any line not matching `indexLineRe`. A single `Store.Save` call would wipe user notes.

**Fix**: Read existing content, preserve non-managed lines, update managed lines in place, append new entries.

### 3. memory: indexLineRe over-matches any markdown link
The regex `\]\(([^)]+)\.md\)` matched any `](xxx.md)` link, not just managed index lines. `Delete("design")` would clobber a user-written `see [design](design.md)` reference.

**Fix**: Tighten to `(?m)^\s*-\s\[.+?\]\(([^)]+)\.md\)\s*—\s.*$`.

### 4. security: env command auto-approved as read-only despite executing arbitrary commands
`env` was in `ReadOnlyCommands`, but `env rm -rf /` executes `rm -rf /`. `hasUnsafeReadOnlyArgs` had no `env` case, so `env <malicious>` was auto-approved without user confirmation.

**Fix**: Remove `env` from `ReadOnlyCommands`.

### 5. agent: parallel_tasks goroutine leak on ctx cancel
When `ctx.Done()` fired, the drain loop used `default: break drain` and returned immediately without calling `wg.Wait()`. Spawned sub-agent goroutines became orphans, holding session/provider/tool resources.

**Fix**: Add `wg.Wait()` before returning.

### 6. permission: find missing unsafe flags
`hasUnsafeReadOnlyArgs` for `find` only checked `-exec`, `-execdir`, `-delete`. Missing: `-ok`, `-okdir` (execute commands), `-fls`, `-fprint`, `-fprintf` (write files).

### 7. sandbox: Linux bwrap missing toolchain/cache dirs
macOS seatbelt binds `/tmp`, `~/.cache`, `~/.cargo`, `~/.npm`, `~/go`, etc. Linux bwrap only had `--tmpfs /tmp`, breaking `go build`, `cargo build`, `npm install` inside the sandbox.

### 8. memorycompiler: corrupt state.json silently overwritten
`loadStateLocked` returned empty state on JSON parse failure; next turn overwrote the original file, losing all learning history.

**Fix**: Back up corrupt file to `state.json.corrupt-<ts>` before returning empty state.

### 9. memory: slug returns empty for non-ASCII names
`slug("用户偏好")` returned `""`, blocking CJK users from saving memories.

**Fix**: Regex `[^\p{L}\p{N}]+` preserves Unicode letters/digits; empty fallback to `memory-<ts>`.

### 10. memory: resolvePath allows @~/.ssh/id_rsa import
`@~` and `@/absolute` imports were allowed, leaking private keys into LLM prompt context via malicious AGENTS.md.

**Fix**: Reject `~` and absolute paths; enforce baseDir subtree via `filepath.Rel`.

### 11. memorycompiler: retainMemoryNodes evicts TruthLocked nodes
When node count exceeded `maxMemoryGraphNodes` (300), TruthLocked nodes had no special protection, contradicting `upsertNode`'s "TruthLocked cannot be overwritten" semantics.

**Fix**: Unconditionally retain all TruthLocked nodes; allocate remaining quota by priority.

### 12. memorycompiler: traceID collisions on Windows
`time.Now()` on Windows has ~15ms resolution. Fast consecutive turns produced identical traceIDs, causing learnings to be deduplicated and lost.

**Fix**: Append `rand.Int63()` suffix.

### 13. agent: SummarizeFrom/SummarizeUpTo missing IncrementRewrite
These methods called `session.Replace(next)` without `IncrementRewrite()`, breaking `cache_shape` diagnostics (cache miss wrongly attributed instead of `log_rewrite`).

## Test plan

- [x] `go build ./...` passes
- [x] `go test -count=1 ./internal/memory/... ./internal/memorycompiler/... ./internal/agent/... ./internal/sandbox/... ./internal/permission/...` all pass
- [x] `compression_test.go` updated: `TestTruthLockedImportanceDecaysForCompressionPriority` now asserts TruthLocked preservation (was asserting eviction)
- [x] `store_extra_test.go` updated: empty-name validation moved to raw input layer to coexist with slug fallback

### DIFF
--- a/internal/agent/compact.go
+++ b/internal/agent/compact.go
@@ -303,6 +303,7 @@ func (a *Agent) SummarizeFrom(ctx context.Context, fromIdx int) error {
 		Content: "Summary of the later conversation (compacted from here on):\n" + summary,
 	})
 	a.session.Replace(next)
+	a.session.IncrementRewrite()
 	a.sink.Emit(event.Event{Kind: event.Notice, Level: event.LevelInfo,
 		Text: fmt.Sprintf("summarized %d later messages → summary", len(region))})
 	return nil
@@ -336,6 +337,7 @@ func (a *Agent) SummarizeUpTo(ctx context.Context, toIdx int) error {
 	})
 	next = append(next, msgs[toIdx:]...)
 	a.session.Replace(next)
+	a.session.IncrementRewrite()
 	a.sink.Emit(event.Event{Kind: event.Notice, Level: event.LevelInfo,
 		Text: fmt.Sprintf("summarized %d earlier messages → summary", len(region))})
 	return nil
--- a/internal/agent/parallel_tasks.go
+++ b/internal/agent/parallel_tasks.go
@@ -316,6 +316,7 @@ func (p *ParallelTasksTool) Execute(ctx context.Context, args json.RawMessage) (
 				}
 			}
 			markCancelled(err)
+			wg.Wait() // ensure spawned goroutines exit before returning (was leak on ctx cancel)
 			return formatParallelTasksAggregate(outputs, taskErrs, statuses, true), err
 		}
 	}
--- a/internal/memory/doc.go
+++ b/internal/memory/doc.go
@@ -245,19 +245,30 @@ func importTarget(line string) (string, bool) {
 	return p, true
 }
 
-// resolvePath turns an import token into a filesystem path: ~ expands to home,
-// absolute paths pass through, everything else is relative to baseDir.
+// resolvePath turns an import token into a filesystem path. Home-relative ("~")
+// and absolute imports are refused so a memory can't read sensitive files
+// outside the project tree (e.g. ~/.ssh/id_rsa). Only paths relative to
+// baseDir are allowed, and the resolved target must stay inside baseDir's
+// subtree; an import that escapes returns "" so the caller skips it.
 func resolvePath(p, baseDir string) string {
-	if strings.HasPrefix(p, "~") {
-		if home, err := os.UserHomeDir(); err == nil {
-			rest := strings.TrimLeft(p[1:], "/\\")
-			return filepath.Join(home, rest)
-		}
+	cleaned := strings.TrimSpace(p)
+	if cleaned == "" || strings.HasPrefix(cleaned, "~") || filepath.IsAbs(cleaned) {
+		return ""
 	}
-	if filepath.IsAbs(p) {
-		return p
+	joined := filepath.Join(baseDir, cleaned)
+	abs, err := filepath.Abs(joined)
+	if err != nil {
+		return joined
 	}
-	return filepath.Join(baseDir, p)
+	baseAbs, err := filepath.Abs(baseDir)
+	if err != nil {
+		return abs
+	}
+	rel, err := filepath.Rel(baseAbs, abs)
+	if err != nil || rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator)) {
+		return ""
+	}
+	return abs
 }
 
 // absOf returns the absolute form of p, falling back to a cleaned p on error so
--- a/internal/memory/store.go
+++ b/internal/memory/store.go
@@ -184,10 +184,10 @@ func (s Store) Save(m Memory) (string, error) {
 	if dir == "" {
 		return "", fmt.Errorf("memory store unavailable (no user config dir)")
 	}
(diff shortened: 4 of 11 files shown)


SivanCola:
Thanks @bfxh for the deep investigation in this PR. I reviewed the patch against the current `main-v2` implementation and opened #5452 to carry forward the valid fixes in a smaller integration PR with additional regression coverage and a few safety adjustments.

What was kept or adapted in #5452:

- Memory drift control now uses decayed confidence for stale detection without writing the decayed value back every turn.
- `MEMORY.md` reindexing preserves hand-written content and only updates managed memory-index lines.
- Managed memory-index matching is narrowed so ordinary Markdown links are not treated as saved-memory entries.
- Unicode memory names are supported.
- Unsafe memory imports are blocked, including absolute paths, home-relative paths, parent escapes, and symlink escapes.
- `env` is removed from read-only shell auto-approval.
- `find` write/execute predicates are blocked, with coverage extended to `-fprint0` as well.
- Corrupt memory-compiler `state.json` files are backed up before falling back to empty state.
- Trace IDs include a collision guard for coarse platform clocks.
- Manual summarize operations increment the rewrite counter for cache diagnostics.
- Foreground `parallel_tasks` waits for spawned workers to exit after cancellation.
- Linux bubblewrap cache-directory binding is kept in intent, but adapted to bind only directories that actually exist and to preserve the newer forbid-read behavior.

What was not adopted as-is:

- The unconditional `TruthLocked` retention change is not safe. Existing compression tests intentionally allow stale truth-locked nodes to lose retention priority under tight limits, and retaining every truth-locked node can violate `maxMemoryGraphNodes`.
- The random slug fallback for empty or punctuation-only names is not safe because `slug()` is also used by lookup, recall, archive, and delete paths. #5452 keeps Unicode slugs deterministic and rejects empty/punctuation-only names instead.
- The lexical-only import containment check was incomplete because symlinks could still escape the importing directory. #5452 adds symlink-escape coverage.
- The original Linux bwrap cache-dir patch could make sandbox startup fail on machines without every cache directory, because `bwrap --bind` requires an existing source path. #5452 skips missing directories.

Closing this PR in favor of #5452, with @bfxh credited there as a repository contributor for the original investigation and valid fix directions.

SivanCola:
Follow-up: #5452 has now been merged into `main-v2` in 20e99f6.

That integration PR carries forward the valid fixes from this investigation, including memory drift non-persistent decay, `MEMORY.md` preservation, managed-index matching, unsafe import containment, `env` read-only removal, `find` predicate hardening, corrupt state backup, trace ID collision guarding, summarize rewrite diagnostics, `parallel_tasks` cancellation waiting, and Linux bwrap cache-dir handling.

Verification completed on #5452:

- GitHub checks passed: `test (ubuntu-latest)`, `test (macos-latest)`, `test (windows-latest)`, `race`, `lint`, `coverage`, `desktop`, `CodeQL`, `cache-impact`, and `govulncheck`.
- Local verification before merge included `go test -count=1 ./...` and `GOOS=windows GOARCH=amd64 go test -c ./internal/sandbox`.

Release availability: the fix is merged on `main-v2` but is not included in the current published `v1.12.0` / `desktop-v1.12.0` releases. It will become user-visible after the next release that includes 20e99f6.

Thanks again @bfxh for the original investigation and fix directions; #5452 keeps contributor credit for this work.

Issue #5456, opened 2026-06-27
[Bug]: Version 1.12.0 broke the TUI, model output no longer visible


drm-nz:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.12.0

### What happened?

Version 1.12.0 broke the TUI. When you ask a question from Reasonix, the thinking tokens, model response, choices are displayed outside of the user's view. You can only tell Reasonix is doing something by the "thinking XXs..." counter at the bottom, but the actual model output is invisible. You also can't scroll up and down as scroll moves the terminal window itself (pretty much what you would see when you scroll up and down in nano). This leads me to believe that the scroll action isn't captured by Reasonix and it is applied to the terminal itself.

If you quit reasonix with CTRL + C and then restart it, then resume the previous session by running "/resume 1", you can see what happened, but by that time the choices are no longer available as you cancelled them out when you quit.

Rolling back to the previous version 1.11.1 fixes this.

----

  ## Rollback
  
  Since reasonix is a cask (not a formula), brew switch doesn't apply. The standard approach uses the tap's git history:

  ### 1. Find the tap directory and the commit that had 1.11.1
  `cd $(brew --repository esengine/reasonix)`
  `git log --oneline Casks/reasonix.rb`

  ### 2. Check out the old cask definition (use the commit hash for 1.11.1)
  `git checkout <commit-hash> Casks/reasonix.rb`

  ### 3. Uninstall current version and install from the old definition
  `brew uninstall --cask reasonix`
  `brew install --cask reasonix`

  ### 4. Pin it so brew doesn't auto-upgrade it again
  `brew pin reasonix`

  ### 5. Restore the tap to HEAD (so other tap operations aren't broken)
  `git checkout HEAD Casks/reasonix.rb`

### Steps to reproduce

1, Start Reasonix
2, Ask a question that requires a long answer that doesn't fit the initial screen
3, Note screen no longer scrolls to follow the model response
4, Note you can't scroll manually either
5, Quit Reasonix with CTRL + C
6, Restart Reasonix
7, Resume previous session (/resume 1)
8, Notice the output is there it just wasn't visible

### OS / platform

macOS 26.5.1

### Relevant logs or output

```shell

```

SivanCola:
Thanks for the report. This regression has been addressed in v1.13.0.

[...]

PR #5378, opened 2026-06-26
fix(tui): restore mouse wheel scrolling in transcript viewport

yekern:
# PR: Restore mouse wheel scrolling in CLI TUI transcript viewport

Closes #5334, closes #5311, closes #5324

### Summary

Commit `edacf1a8` ("fix(tui): preserve native terminal selection") removed
`v.MouseMode = tea.MouseModeCellMotion` from the `View()` method, which disabled
mouse reporting in the terminal. Without mouse mode, terminals do not emit
`MouseWheelMsg` events. Many terminal emulators translate mouse wheel input into
**Up / Down arrow key** sequences instead. These arrow keys are handled by
`recallSubmittedInput()`, causing the **input history** to scroll instead of the
transcript viewport.

This PR restores `MouseModeCellMotion` on the **alt-screen path only** so that
mouse wheel events reach the viewport and scroll the transcript as expected.

### Affected issues

| Issue | Description | Platform |
|-------|-------------|----------|
| #5334 | CLI 输出的内容无法上下滚动查看，滚动时成了滚动输入框中的内容了 | Windows 11 |
| #5311 | 鼠标滚轮之前是滚动对话记录, 现在变成了滚动历史命令 | Windows Terminal → macOS |
| #5324 | Scrolling up to view past messages does not work anymore (CLI) | macOS 26.5.1 |

All three issues report the same regression caused by `edacf1a8`.

### Root cause

| Before `edacf1a8` | After `edacf1a8` | This PR |
|---|---|---|
| `MouseModeCellMotion` enabled | Mouse mode removed | `MouseModeCellMotion` restored |
| Wheel → `MouseWheelMsg` → `viewport.ScrollUp/Down` ✅ | Wheel → terminal sends ↑/↓ keys → `recallSubmittedInput()` ❌ | Wheel → `MouseWheelMsg` → `viewport.ScrollUp/Down` ✅ |

### What changed

- **`internal/cli/chat_tui.go`** (+1 line): Added back `v.MouseMode = tea.MouseModeCellMotion` in the `View()` method, only on the alt-screen (non-Termux) path.
- **`internal/cli/chat_tui_test.go`** (+5 / -4 lines): Updated `TestViewAltScreenFillsHeight` to expect `MouseModeCellMotion` instead of `MouseModeNone`. The Termux test (`TestViewTermuxUsesNativeScrollback`) is unchanged because the Termux path returns before the MouseMode line.

### What is NOT affected

[...]

PR #5476, opened 2026-06-28
fix(tui): add regression test verifying viewport auto-follows streaming answer (#5456)

ttmouse:
## 修复内容

为 Issue #5456 添加回归测试，验证视口在流式输出时自动跟随。

### 变更说明
- 新增  — 
- 测试序列：填充溢出内容 → 触发流式 event.Text → 验证视口始终保持在底部
- 补充了现有  的逆命题

### 验证
- go test ./internal/cli/ 全部通过
- 三种代码走查全部通过（内置 review / ponytail-review / pr-review）
- 最小改动：仅新增一个测试文件（46 行）

### DIFF
--- a/internal/cli/auto_scroll_test.go
+++ b/internal/cli/auto_scroll_test.go
@@ -1,0 +1,46 @@
+package cli
+
+import (
+	"testing"
+
+	tea "charm.land/bubbletea/v2"
+	"reasonix/internal/control"
+	"reasonix/internal/event"
+)
+
+func TestAutoFollowStreamingAnswerAtBottom(t *testing.T) {
+	ctrl := control.New(control.Options{})
+	ch := make(chan event.Event, 1)
+	adv := func(m chatTUI, msg tea.Msg) chatTUI {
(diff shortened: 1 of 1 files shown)


SivanCola:
Thanks for adding coverage here.

I do not think this PR should be treated as the fix or regression guard for #5456 in its current form. The issue's root cause was the removal of `v.MouseMode = tea.MouseModeCellMotion` from the alt-screen TUI path, which prevented terminals from delivering wheel input as `MouseWheelMsg` events to the transcript viewport. That was restored in #5378 and then hardened further in #5400, both of which are included in v1.13.0.

I also checked the proposed test against the v1.12.0 code plus this test file: it still passes there, so it would not have caught the reported #5456 regression.

If we keep this PR, I think it should be scoped as supplemental coverage for answer-stream auto-follow while already pinned to the bottom, and the title/body should remove the #5456 fix implication. Otherwise, I think it is reasonable to close this as superseded by #5378/#5400.

PR #5506, opened 2026-06-28
fix(desktop): show fallback UI in ErrorBoundary instead of blank screen

HUQIANTAO:
Currently ErrorBoundary renders null when a child crashes, leaving users with a completely blank area and no recovery path. This PR adds a fallback card with a "Try again" button.

**Before:** `return this.state.crashed ? null : this.props.children;`
**After:** Renders a fallback `<div role="alert">` with error message and retry button that resets error state.


SivanCola:
Thanks for the improvement. This is frontend error-boundary UI work, and the current consolidation is intentionally non-frontend, so I did not include it in #5620.

Closing this PR for this batch; a frontend follow-up should include the UI copy/localization and visual validation together.

PR #5507, opened 2026-06-28
fix(desktop): readPref() now reads from localStorage for locale persistence

HUQIANTAO:
readPref() always returned "" (auto-detect from OS), causing user-set language preferences to be lost on restart. Now reads from localStorage using the same key as readLegacyLangPref().

**Fix:** `readPref()` now actually reads from localStorage instead of returning a stub.


SivanCola:
Thanks for looking at language persistence. I am not adopting this change because the current desktop language preference is persisted through the user-level desktop config; `localStorage` is only read for legacy migration. Reading it again in `readPref()` would reintroduce the old preference source and conflict with the current model.

PR #5508, opened 2026-06-28
fix(desktop): add aria-label to Composer send button for screen readers

HUQIANTAO:
The send button in Composer had no aria-label, appearing as an unlabeled button to screen readers. Adds `aria-label={t("composer.sendAriaLabel")}` with locale keys in en/zh/zh-TW.


SivanCola:
Thanks for the accessibility cleanup. This touches the frontend composer, and the current consolidation is intentionally non-frontend, so I did not include it in #5620.

Closing this PR for this batch to keep the maintenance branch narrowly scoped.

PR #5509, opened 2026-06-28
fix(desktop): handle rand.Read error in newTopicID with timestamp fallback

HUQIANTAO:
newTopicID() ignored the error return from rand.Read, producing all-zero random bytes on system entropy failure. Now falls back to nanosecond timestamp, matching the pattern used by newTabID().

### DIFF
--- a/desktop/tabs.go
+++ b/desktop/tabs.go
@@ -5143,7 +5143,10 @@ func newTabID() string {
 
 func newTopicID() string {
 	var b [8]byte
-	rand.Read(b[:])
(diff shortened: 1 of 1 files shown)


SivanCola:
Thanks for the contribution. I consolidated this into #5620, keeping the `rand.Read` failure fallback for `newTopicID` and applying it on the latest `main-v2` together with the related desktop tray lifecycle fix.

Closing this source PR so the maintenance work proceeds through one focused branch.

PR #5510, opened 2026-06-28
fix(serve): add capacity limit and eviction to prevent titleCache unbounded growth

HUQIANTAO:
titleCache had no eviction policy — entries only added, never removed. Over long server uptimes the in-memory map and on-disk JSON grow without bound. Adds maxTitleCacheEntries=1000 with oldest-entry eviction in put().


SivanCola:
Thanks for the cache-cap idea. I am not adopting this implementation as written: it describes oldest-entry eviction, but deleting the first key from a Go map uses random iteration order, not insertion or age order. A title-cache cap may still be useful, but it needs explicit ordering and coverage.

PR #5511, opened 2026-06-28
fix(desktop): add context cancellation check to asyncRuntimeEmitter

HUQIANTAO:
asyncRuntimeEmitter.run() looped indefinitely until queue drained with no way to exit on app shutdown. Adds a ctx field checked via ctx.Done() in the run loop so the goroutine exits cleanly when the owning context is cancelled.


SivanCola:
Thanks for looking at shutdown behavior. I am not adopting this implementation as written: adding a `ctx` field to `asyncRuntimeEmitter` does not wire that field into the emitter instance used by the sink, so the new cancellation check would normally not observe app shutdown.

This needs a different fix with a focused regression test. No code from this PR was included in #5620, so I am closing it.

PR #5513, opened 2026-06-28
perf(desktop): wrap Tooltip and CopyButton with React.memo

HUQIANTAO:
Tooltip and CopyButton receive stable props but re-render on every parent update, causing unnecessary reconciliation overhead — especially during streaming where the Composer re-renders frequently. Wrapping with React.memo skips re-renders when props are shallow-equal.


SivanCola:
Thanks for the performance cleanup. I am not adopting this one: the benefit is unmeasured, and `React.memo` around these leaf components may not help much when JSX children/props are recreated by parents.

PR #5514, opened 2026-06-28
fix(plugin): memoize empty shell-PATH probe results to prevent repeated probes

HUQIANTAO:
The retry-after-empty behavior in cachedShellPATH is intentional — PATH may become available later (e.g. after network home directories mount). Memoizing empty results broke TestCachedShellPATHRetriesAfterEmpty which validates this design. Reverted the behavior change and documented the rationale in the function comment.

Cache-impact: none — comment-only change, no cache prefix shift
Cache-guard: comment-only


SivanCola:
Thanks for checking the shell PATH cache behavior. I am not adopting this PR: the final diff is comment-only after reverting the behavior change, and the PR currently has a failing cache-impact check. The retry-after-empty behavior should remain as-is.

PR #5515, opened 2026-06-28
fix(desktop): broaden clipboard image MIME detection to all image/* types

HUQIANTAO:
Previous clipboardHasImageHint used substring matching for png/jpeg/jpg/tiff, missing webp, gif, bmp, svg+xml, avif and other standard image types. Also had overly broad matching that could false-positive on non-image types. Now uses startsWith("image/") covering all IANA image MIME types.


SivanCola:
Thanks for reviewing clipboard image detection. I am not adopting this change: the current check already accepts standard `image/*` MIME types, while this PR mainly removes the extra substring fallback. That is not necessary for this maintenance pass.

PR #5516, opened 2026-06-28
fix(desktop): use aria-live=assertive for warn/error toast notifications

HUQIANTAO:
All toasts previously shared one container with aria-live=polite, meaning warn/error notifications (failed paste, failed send) were announced with the same low priority as info toasts. Split into two containers: polite for info, assertive for warn+error so critical notifications interrupt screen readers immediately.


SivanCola:
Thanks for the accessibility idea. I am not adopting this implementation as written: splitting toasts into two separate fixed containers can overlap when polite and urgent toasts coexist. This should be solved with a single layout/queue model if we revisit it.

PR #5517, opened 2026-06-28
fix(desktop): show error state in past:chats when ListSessions fails

HUQIANTAO:
When app.ListSessions() fails, pastChats was set to empty array with no error indication — users saw "No recent sessions" even on network/backend errors. Now shows localized error message with retry button.


SivanCola:
Thanks for the error-state improvement. This is frontend composer/menu behavior, and the current consolidation is intentionally non-frontend, so I did not include it in #5620.

Closing this PR for this batch; it can be reconsidered separately as part of a frontend-focused pass.

PR #5518, opened 2026-06-28
fix(desktop): cancel pending clipboard image paste timer on any key press

HUQIANTAO:
The native clipboard image paste timer (160ms after Ctrl+V) only had clearNativeClipboardPasteTimer() called on paste shortcuts. If the user started typing within 160ms of pasting, the async image attachment would still fire and could insert an image path after typed text. Now clears the timer on every keydown.


SivanCola:
Thanks for the paste-race fix. This touches frontend composer input behavior, and the current consolidation is intentionally non-frontend, so I did not include it in #5620.

Closing this PR for this batch to keep the maintenance branch narrowly scoped.

PR #5519, opened 2026-06-28
perf(desktop): preload MarkdownRenderer code-split chunk to avoid Suspense flash

HUQIANTAO:
MarkdownRenderer is code-split via React.lazy. On first render, users see a raw-text fallback flash before the chunk loads. Now preloads eagerly at module import time so the chunk is ready before the first message.


SivanCola:
Thanks for the rendering polish idea. I am not adopting this change: eager preloading offsets the existing code-splitting behavior, and there is no measurement showing the Suspense flash is worth that tradeoff.

PR #5520, opened 2026-06-28
feat(serve): add atomic dropped-frame counter to Broadcaster

HUQIANTAO:
Broadcaster silently drops frames when subscriber buffers are full, with no visibility into how many frames are lost — making debugging slow-client SSE issues impossible. Adds an atomic dropped counter incremented per drop, exposed via Dropped() for monitoring.


SivanCola:
Thanks for the diagnostics idea. I am not adopting this implementation as written: adding a dropped-frame counter without logging, metrics, an API, or another consumer does not yet provide operational visibility.

PR #5521, opened 2026-06-28
perf(desktop): avoid unnecessary uniqueStrings allocations in helpers

HUQIANTAO:
containsDesktopString called uniqueStrings on every invocation turning O(n) into O(n) with allocations. removeString deduplicated twice. Now both iterate the original slice directly.


SivanCola:
Thanks for the small allocation cleanup. I am not adopting this one: it is a micro-optimization without measured user impact, and it is outside the focused non-frontend maintenance fix in #5620.

PR #5523, opened 2026-06-28
fix(desktop): guard setSubmitting against unmounted component in submit()

HUQIANTAO:
The submit() finally block called setSubmitting(false) after onSend completed. When the user switches tabs mid-flight, the component unmounts while the async send is still pending, causing React to warn about state updates on an unmounted component. Added a mountedRef guard.


SivanCola:
Thanks for the defensive React cleanup. I am not adopting this one in the current pass: it is frontend-scoped and primarily guards a low-impact unmounted-state warning rather than a user-visible failure.

PR #5524, opened 2026-06-28
fix(desktop): use private DBus connection in traySupported()

HUQIANTAO:
SessionBus() returns a shared singleton — closing it can interfere with other DBus users in the process. Switch to SessionBusPrivate() which creates an independent connection that can be safely closed with Auth() and Hello().


SivanCola:
Thanks for the contribution. I consolidated this into #5620, keeping the final shared-`SessionBus` lifecycle fix: tray probing should not close the shared DBus session connection.

Closing this source PR so the maintenance work proceeds through one focused branch.

PR #5525, opened 2026-06-28
fix(serve): document titleCache.load() must be called with lock held

HUQIANTAO:
load() accesses loaded and entries without internal locking — safe under current callers (get/put both hold mu) but future callers could race. Documents the precondition explicitly.


SivanCola:
Thanks for clarifying the lock precondition. I am not adopting this PR because it is comment-only and does not change behavior or add coverage for this maintenance pass.

PR #5526, opened 2026-06-28
fix(desktop): log warning when SHA-256 hashing fails in attachDedup

HUQIANTAO:
When Web Crypto API is unavailable or digest fails, sha256() silently returned empty string — disabling hash-based dedup with zero diagnostic signal, making debugging impossible. Adds a one-shot console.warn so the failure is visible in dev tools.


SivanCola:
Thanks for the diagnostics cleanup. I am not adopting this one: it is frontend-scoped and only adds a one-shot console warning when hash-based attachment dedup is unavailable. That does not address a user-visible failure in this maintenance pass.

PR #5527, opened 2026-06-28
docs(memorycompiler): document runtimeLocks non-eviction as intentional design

HUQIANTAO:
runtimeLocks sync.Map entries are never deleted — each is a lightweight *sync.Mutex and the number of unique project directories is bounded in practice. Documents this as an intentional trade-off to prevent future readers from misidentifying it as a leak.

### DIFF
--- a/internal/memorycompiler/runtime.go
+++ b/internal/memorycompiler/runtime.go
@@ -42,6 +42,13 @@ const (
 	staleConfidenceThreshold  = 0.2
 )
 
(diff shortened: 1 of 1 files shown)


SivanCola:
Thanks for documenting the `runtimeLocks` lifecycle. I am not adopting this PR because it is comment-only, and the cleanup guidance would need a more concrete owner/lifecycle model before being useful.

PR #5538, opened 2026-06-29
fix(desktop): empty placeholder session overwrites real session, dropping system prompt

drafish:
## Bug

When a new desktop tab is created, the saved session file has no system
message — the agent identity contract (Reasonix persona, AGENTS.md,
MEMORY.md) is lost.  When the session is reloaded the model has no
identity context and falls back to its training-data defaults, giving
wrong answers to identity questions.

## Root cause

In `desktop/tabs.go`:

1. `createEmptySessionFile(dir, "")` pre-creates an empty placeholder `.jsonl`
   file (0 bytes) for a new tab.
2. `buildTabControllerWithContext` builds a real session with the full
   system prompt via `boot.Build()`.
3. `loadPinnedTabSessionWithPreload` loads the placeholder via
   `LoadSession`, which returns a **non-nil** `&Session{}` with zero
   messages for an empty file (EOF is not an error).
4. Because it's non-nil, `ctrl.Resume(emptySession, path)` is called,
   which calls `executor.SetSession(emptySession)` — **replacing the
   real session with an empty one**.
5. The first turn adds a user message and `snapshot()` saves the
   session — without any system message.

## Fix

One logical line in `loadPinnedTabSessionWithPreload`: treat a
0-message session the same as a missing file (return nil), so the
caller keeps the session built by `boot.Build`.

Defensive guard in `controller.snapshot()`: refuse to persist a
session that has user/assistant/tool content but no system message,
preventing future regressions from silently producing broken transcripts.

## Verification

- `go test ./desktop -count=1` — all pass
- `go test ./internal/agent/... ./internal/control/... -count=1` — all pass
- Wails build: `wails build` — successful
- Includes `HasSystemMessage()` unit tests covering the normal,
  missing, post-Replace, and post-compaction cases

### DIFF
--- a/desktop/tabs.go
+++ b/desktop/tabs.go
@@ -5203,6 +5203,13 @@ func loadPinnedTabSessionWithPreload(dir, sessionPath string, preloaded loadedTa
 		}
 		return nil, "", false
 	}
+	// An empty file (0 messages) is a pre-created placeholder, not a real
+	// session to resume.  Treating it as valid would make ctrl.Resume replace
+	// the executor's live session (with system prompt) with the empty one,
+	// causing the saved transcript to lack the agent identity contract.
+	if len(loaded.Snapshot()) == 0 {
+		return nil, path, true
+	}
 	return loaded, path, true
 }
 
--- a/internal/agent/session.go
+++ b/internal/agent/session.go
@@ -77,3 +77,13 @@ func (s *Session) HasContent() bool {
 	}
 	return false
 }
+
+// HasSystemMessage reports whether the session starts with a system message,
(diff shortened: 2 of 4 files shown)


SivanCola:
Thanks @drafish. I absorbed this fix into #5630 and kept your contribution explicit there: the empty-placeholder session guard, the `HasSystemMessage()` persistence guard, and the focused agent/control coverage are preserved and credited as co-created work.\n\n#5630 extends it with the desktop rebuild/resume system-prompt refresh needed for #5618, so I'm closing this PR as superseded by #5630 rather than merging two overlapping branches.

PR #4180, opened 2026-06-12
feat(desktop): add embedded terminal panel with multi-tab support

myipanta:
- Add embedded xterm.js terminal panel as a new dock tab (Terminal) alongside Context / Files / Changed in the right workspace dock
- Multi-tab terminal: support opening multiple shell sessions, switch between tabs, close individual tabs
- Right-click file/folder in workspace tree → 'Open in terminal' creates a new terminal tab cd'd to that directory
- System terminal launchers for macOS (Terminal.app), Linux (auto-detect 12 terminals), Windows (wt.exe / cmd.exe)
- Go PTY backend with per-session management and Wails event streaming for real-time terminal I/O

Files:
- desktop/terminal.go: multi-session PTY manager
- desktop/open_terminal_*.go: platform-specific system terminal launcher
- desktop/app.go: OpenTerminal, OpenTerminalAt, StartTerminal, StartTerminalAt, TerminalInput, StopTerminal, TerminalResize
- frontend/src/components/TerminalPanel.tsx: xterm.js terminal panel
- frontend/src/components/WorkspacePanel.tsx: right-click menu items
- frontend/src/App.tsx: Terminal dock tab integration
- frontend/src/lib/bridge.ts: Wails bindings + browser dev mock
- frontend/src/locales/{en,zh}.ts: i18n keys

PR #5576, opened 2026-06-29
feat(desktop): add embedded terminal panel

myipanta:
- Add terminal panel with session rail, view, and PTY backend
- Terminal Go bindings: OpenTerminal, StartTerminal, StopTerminal, TerminalInput, TerminalResize
- Frontend: TerminalPanel, TerminalSessionRail, TerminalView components
- Keyboard shortcuts for terminal panel toggle
- i18n support for terminal UI
- Fix z-index hardcoded values to use --z-* tokens

## Summary

-

## Verification

-

## Cache impact

Cache-impact: TODO
Cache-guard: TODO
System-prompt-review: N/A


myipanta:
image
 在桌面端嵌入完整终端面板，支持 PTY、多会话管理、xterm.js 渲染、快捷键、深色/浅色主题

esengine:
The integrated terminal shipped, and you are credited on it — `170b23fbd` (#6994, merged 2026-07-29) carries `Co-authored-by: ipanta` on three of its commits, and `desktop/frontend/src/components/TerminalPanel.tsx` is in the tree today.

You proposed this twice, in #4180 and again in #5576, and both are still sitting open. That is the part we got wrong. The feature you asked for is running; the PRs that asked for it were never closed, so from your side it looks like two rejected attempts followed by someone else building it. The credit lines say otherwise, but a `Co-authored-by` in a squashed commit is invisible from a PR page, which makes it worth very little as an acknowledgement.

What shipped differs from both proposals in scope — sessions are bound to the active chat tab rather than global, with ConPTY on Windows, bounded session counts, and an explicit user action to put terminal output into the composer instead of any automatic prompt injection. The direction was yours.

Closing this now, with the account written where you can see it. Thanks for pushing on it twice — the second attempt is why it stayed on the list.

PR #5603, opened 2026-06-29
fix(memory): read listed markdown names

### LINKED ISSUE #5597: bug: memory tool — .md suffix in list links fails on read (slug mismatch)
## Bug Description

`memory(operation="list")` displays link text with `.md` extension (e.g., `[坤哥的铁规矩](kunge-rules.md)`), but `memory(operation="read", name="kunge-rules.md")` fails with "not found" because the tool slug-normalizes the name by replacing dots with dashes.

## Steps to Reproduce

1. Create a memory: `memory(operation="remember", name="my-rules", body="test")`
2. List memories: `memory(operation="list")` → shows `[my-rules](my-rules.md)` 
3. Try to read with the displayed name: `memory(operation="read", name="my-rules.md")`
4. ❌ Error: `memory "my-rules-md" not found`
5. Read without extension: `memory(operation="read", name="my-rules")`
6. ✅ Works

## Impact

Users lose 1-2 tool calls debugging the slug format. The list output is misleading — the displayed reference is not a valid input for the read operation.

## Expected Behavior

Either:
- `memory(read)` accepts the `.md` suffix and strips it internally, OR
- `memory(list)` displays links without the `.md` extension

## Environment

- Reasonix: Go 1.0 (main-v2)

cyq1017:
## Problem
`memory(operation="list")` renders saved memories as links like `name.md`, but `memory(operation="read", name="name.md")` slugifies the suffix and fails to find the saved memory.

## Change
Strip a trailing `.md` from memory read names before resolving the saved memory slug.

Cache-impact: low - touches the memory read resolver, but does not change the memory tool description, schema, list output, or composed system-prompt text.
Cache-guard: `go test ./internal/memory -run 'TestRecallToolSchemaIsCacheStable|TestRecallToolReadsMemoryByListedMarkdownName' -count=1`
System-prompt-review: self-reviewed; provider-visible memory tool description/schema are unchanged and covered by TestRecallToolSchemaIsCacheStable.

## Verification
- `go test ./internal/memory -run TestRecallToolReadsMemoryByListedMarkdownName -count=1`
- `go test ./internal/memory -count=1`
- `go test ./internal/memory ./internal/boot -count=1`
- `go test ./internal/memory -run 'TestRecallToolSchemaIsCacheStable|TestRecallToolReadsMemoryByListedMarkdownName' -count=1`
- `git diff --check`

Closes #5597

### DIFF
--- a/internal/memory/recall.go
+++ b/internal/memory/recall.go
@@ -196,6 +196,7 @@ func filterMemories(memories []Memory, typ Type) []Memory {
 }
 
 func readMemoryByName(store Store, name string) (Memory, bool) {
+	name = strings.TrimSuffix(strings.TrimSpace(name), ".md")
 	name = slug(name)
 	if name == "" {
 		return Memory{}, false
--- a/internal/memory/recall_test.go
+++ b/internal/memory/recall_test.go
@@ -157,6 +157,27 @@ func TestRecallToolReadsMemoryByName(t *testing.T) {
 	}
 }
 
+func TestRecallToolReadsMemoryByListedMarkdownName(t *testing.T) {
+	store := Store{Dir: t.TempDir()}
+	saveMemory(t, store, Memory{
+		Name:        "user-prefers-tabs",
+		Title:       "Prefers tabs",
+		Description: "User prefers tabs for indentation",
+		Type:        TypeUser,
+		Body:        "Use tabs unless the repository style clearly says otherwise.",
+	})
+
+	out, err := NewRecallTool(store).Execute(context.Background(), []byte(`{"operation":"read","name":"user-prefers-tabs.md"}`))
+	if err != nil {
+		t.Fatalf("Execute read with listed markdown name: %v", err)
+	}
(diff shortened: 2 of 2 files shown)


esengine:
This is fixed. `memory` accepts a name with or without the `.md` suffix today — `store_v2.go` and `forget.go` both normalise it at the parse boundary, and `recall_test.go` pins the exact case with `{"operation":"read","name":"listed-memory.md"}`.

It was fixed as part of the Context Engine v2 rewrite (#7001) rather than by merging this diff, because that change replaced the addressing scheme wholesale: memories are referenced by a stable id instead of a mutable slug, so suffix handling moved to one place rather than each call site.

Worth noting two people filed this independently within a day — #5603 and #5604 — which is usually a sign the papercut was worse than its size suggested.

Leaving this open afterwards was our mistake: the bug is gone, but your PR looks unread. Closing it now and saying where it went.

Thanks for it. A name that renders one way in a list and is rejected in another form is exactly the kind of thing that makes a tool feel unreliable.

PR #5604, opened 2026-06-29
fix(memory): strip .md suffix in read to match list display format

### LINKED ISSUE #5597: bug: memory tool — .md suffix in list links fails on read (slug mismatch)

salarkhannn:
## Summary

`memory(list)` displays links with `.md` suffix (e.g. `[my-rules](my-rules.md)`), but `memory(read, name="my-rules.md")` fails because `slug()` converts the dot to a dash, producing `"my-rules-md"` instead of `"my-rules"`.

Fix: strip `.md` suffix in `readMemoryByName` before slug normalization, so the displayed link format works as a direct input to `memory(read)`.

## Verification

- `go test ./internal/memory/ -run TestRecallTool` — 9 tests pass (8 existing + 1 new)
- `go test ./internal/memory/... ./internal/control/...` — all pass
- `go build ./...` — clean
- `gofmt -l .` — no changes

Closes #5597

Cache-impact: low — tool schema and description unchanged; only adds input normalization in the read path.
Cache-guard: TestRecallToolReadsMemoryByName + TestRecallToolReadsMemoryWithDotMdSuffix
System-prompt-review: No schema, description, or prompt changes — only read-path input normalization. TestRecallToolSchemaIsCacheStable confirms schema stability.

### DIFF
--- a/internal/memory/recall.go
+++ b/internal/memory/recall.go
@@ -196,6 +196,7 @@ func filterMemories(memories []Memory, typ Type) []Memory {
 }
 
 func readMemoryByName(store Store, name string) (Memory, bool) {
+	name = strings.TrimSuffix(name, ".md")
 	name = slug(name)
 	if name == "" {
 		return Memory{}, false
--- a/internal/memory/recall_test.go
+++ b/internal/memory/recall_test.go
@@ -136,6 +136,32 @@ func TestRecallToolExcludesArchivedMemories(t *testing.T) {
 	}
 }
 
+func TestRecallToolReadsMemoryWithDotMdSuffix(t *testing.T) {
+	store := Store{Dir: t.TempDir()}
+	saveMemory(t, store, Memory{
+		Name:        "test-rules",
+		Title:       "Test rules",
+		Description: "Test description",
+		Type:        TypeProject,
+		Body:        "Test body.",
+	})
+
+	for _, tc := range []struct{ name, label string }{
+		{"test-rules.md", ".md suffix from list display"},
+		{"test-rules", "plain slug (backwards compat)"},
+	} {
+		out, err := NewRecallTool(store).Execute(context.Background(), []byte(`{"operation":"read","name":"`+tc.name+`"}`))
+		if err != nil {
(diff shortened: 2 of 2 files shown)


esengine:
This is fixed. `memory` accepts a name with or without the `.md` suffix today — `store_v2.go` and `forget.go` both normalise it at the parse boundary, and `recall_test.go` pins the exact case with `{"operation":"read","name":"listed-memory.md"}`.

It was fixed as part of the Context Engine v2 rewrite (#7001) rather than by merging this diff, because that change replaced the addressing scheme wholesale: memories are referenced by a stable id instead of a mutable slug, so suffix handling moved to one place rather than each call site.

Worth noting two people filed this independently within a day — #5603 and #5604 — which is usually a sign the papercut was worse than its size suggested.

Leaving this open afterwards was our mistake: the bug is gone, but your PR looks unread. Closing it now and saying where it went.

Thanks for it. A name that renders one way in a list and is rejected in another form is exactly the kind of thing that makes a tool feel unreliable.

PR #5649, opened 2026-06-30
fix(math): display-math blocks render empty or swallow prose between two blocks

drafish:
## Problem

Display-math formulas (`$$...$$`) fail to render in two scenarios:

1. **Single-line **: The old repair regex moved the closing $$ to a new line but left the opening $$ with content after it. remark-math requires opening $$ on its own line, so it parsed the block as empty — formulas disappeared.

2. **Two display blocks separated by prose**: The  replacement regex used `[\s\S]*?` which matched across both blocks, consuming the prose paragraph between them as math content.

## Fix

Replace all regex-based $$ handling with a line-by-line parser (`normaliseDisplayBlocks`) that:

- Walks lines sequentially, opening a display block on finding $$ and collecting content until the closing $$
- Emits each block as $$\nformula\n$$ (the format remark-math requires)
- Correctly separates adjacent display blocks with prose between them
- Skips digits before $$ (e.g. `c^2$$`) to avoid false triggers
- Handles orphan opening $$ (no closing) by moving $$ to its own line

## Tests

All 189 golden-case tests pass. Updated test expectations for the new multi-line output format (e.g. `normalizeMath("\\[E=mc^2\\]")` now returns `"$$\nE=mc^2\n$$"` instead of `"$$E=mc^2$$"`).

### DIFF
--- a/desktop/frontend/src/__tests__/math-golden.test.ts
+++ b/desktop/frontend/src/__tests__/math-golden.test.ts
@@ -282,6 +282,12 @@ check("digit before $$ is NOT a prose boundary (preserves c^2$$)", () => {
   const out = normalizeMath("c^2$$ x $$");
   return out === "c^2$$ x $$";
 });
+eq(normalizeMath("intro$$x+1"), "intro\n$$\nx+1", "orphan opening $$ is not duplicated");
+eq(
+  normalizeMath("first$$a$$ middle $$b$$ end"),
+  "first\n$$\na\n$$\n middle \n$$\nb\n$$\n end",
+  "multiple display blocks on one line are all normalised",
+);
 
 console.log("\nnormalizeMath — non-math dollar filtering");
 eq(normalizeMath("costs $1$ today"), "costs &#36;1&#36; today", "$1$ not math");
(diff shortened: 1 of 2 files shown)


SivanCola (approved):
Approved. I pushed a follow-up commit a3ddb99 to cover the two display-block parser edge cases found in review.

- The fix stays inside the existing normalizeMath display-block pre-pass and preserves the code-protection, Young-diagram expansion, inline classifier, and KaTeX-normalization flow.
- It does not change dependencies, network/file behavior, or provider/cache-visible surfaces.

Review fixes:
- Orphan opening `$$` now consumes the current line after emitting the repaired block, so the original line is not duplicated.
- Trailing text after a same-line `$$...$$` pair is normalised again, so later display blocks on the same physical line are not left as raw `$$...$$` text.
- Added golden cases for both contracts.

Verified:
- `./node_modules/.bin/tsx src/__tests__/math-golden.test.ts` passed (191/191)
- `./node_modules/.bin/tsc --noEmit` passed
- `git diff --check origin/main-v2...HEAD` passed

SivanCola:
Maintainer update: I opened #5662 as the integration PR for this display-math parser work and the related classifier fixes from #4543.

#5662 keeps/adapts this PR's line-by-line display parser direction, including the orphan opening `$$`, adjacent display blocks, and blockquote display-math regression coverage. I am leaving this PR open for traceability until the integration PR lands.

PR #5658, opened 2026-06-30
fix(desktop): ensure tab.mode is normalized on load, add reverse reconciliation (#5480)

erphenheimer:
## Summary

Two minimal defense-in-depth fixes for tab mode persistence (#5480, #5481).

### Changes

**`desktop/app.go`:**
1. `restoreOrBuildTabs` now uses `normalizeTabMode(entry.Mode)` instead of `persistedTabMode(entry.Mode)`. This ensures `tab.mode` is always a valid normalized value ("normal", "plan", "yolo", "plan-yolo") rather than `""` for default mode. The old JSON format (where `omitempty` drops `""`) is correctly handled by `normalizeTabMode("")` → `"normal"`.
2. Added reverse reconciliation for the edge case where `entry.Mode=""` (old format) + `entry.ToolApprovalMode="yolo"`. This pairs with the existing reconciliation (mode implies yolo → fix toolApproval) to also handle the reverse (toolApproval says yolo → fix mode).

### Verification
- `TestTabModeLoadedAsNormalNotBlank` — verifies that loading a tab with `mode=""` from JSON produces a normalized `"normal"` value
- `TestTabModeReconciliationEdgeCases` — verifies all edge cases of old-format mode/toolApproval combinations

### What this does NOT change
- `persistedTabMode`/`persistedToolApprovalMode` are left unchanged (they return `""` for default values as before)
- `Meta.PendingPrompt` field is NOT added (already exists in `TabMeta` from `RuntimeStatus()`)
- No changes to `tabs.go` save logic or frontend

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -41,20 +41,11 @@ import (
 	"reasonix/internal/fileref"
 	fileenc "reasonix/internal/fileutil/encoding"
 	"reasonix/internal/i18n"
-	"reasonix/internal/jobs"
-	"reasonix/internal/mcpcatalog"
 	"reasonix/internal/mcpdiag"
-	"reasonix/internal/mcptrust"
 	"reasonix/internal/memory"
-	"reasonix/internal/notify"
 	"reasonix/internal/plugin"
-	"reasonix/internal/pluginpkg"
 	"reasonix/internal/provider"
-	"reasonix/internal/repair"
-	"reasonix/internal/sandbox"
 	"reasonix/internal/skill"
-	"reasonix/internal/store"
-	"reasonix/internal/tool"
 )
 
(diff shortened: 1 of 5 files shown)


esengine:
Thanks for the defense-in-depth work here. Closing as superseded: the mode-flip-on-switch class this targeted was root-caused as a compound navigation race and fixed structurally — tab switches are now coalesced through a single navigation scheduler (#5404) and event-sink teardown/setup was unified (#5406) — and #5481 is already closed. The branch has also drifted into conflict with the reworked tab code. The remaining normalize-on-load hardening is marginal since every consumer of `tab.mode` normalizes at the use site; if a concrete mode-restore bug survives on a current release, a fresh small PR against the new code would be the way to go. Appreciate the contribution!

PR #5703, opened 2026-07-01
feat(permission): match compound bash commands segment-by-segment

morningtzh:
## Summary

Extends the bash permission gate to evaluate compound commands segment-by-segment against the existing rule table, so prefix rules (`Bash(git push:*)`, `Bash(go test:*)`, …) created by the current auto-save path can cover common compound flows without any change to how rules are stored.

- Adds `DecomposeBashCommand` in `internal/permission/bash_decompose.go` — a hand-rolled tokenizer that splits on `;`, `|`, `||`, `&`, `&&`, and newlines while treating single/double quotes, `$(...)`, backticks, and backslash escapes as opaque. `2>&1` / `>&2` are recognized as file-descriptor duplication (not splitters). Unbalanced quotes/parentheses return `nil` so callers fall back to whole-string matching.
- Routes `Policy.DecideSubject`'s bash branch through a new `decideBashSegments` helper when decomposition yields ≥ 2 segments. Each segment is evaluated independently against the rule table with the existing `deny > ask > allow > read-only fallback` precedence. Segments recognized as read-only by `shellsafe` allow without a rule, matching the behavior of an atomic read-only bash call.
- **Storage semantics are unchanged.** `BashCommandPrefix` still returns nil for compound subjects, and "always allow / save to config" continues to store the exact literal string for compound commands. The maintainer's red line in #3882 — no prefix synthesized from a compound — is preserved. What changes is only the matching algorithm.

Effect on user experience: a user who has approved `Bash(git add:*)` + `Bash(git commit:*)` + `Bash(git push:*)` — via the existing atomic prefix save path or manual config — now passes `git add . && git commit -m … && git push` silently, without a re-prompt on wording changes.

Motivation and design comparison against Claude Code / Codex are laid out in the linked discussion issue.

## Out of scope for this PR

- **Redirect stripping inside individual segments** (`2>/dev/null`, `> file`). `bashPrefixMatches` still rejects subjects with any shell syntax, so a segment carrying a redirect misses even a matching prefix rule. This is a natural follow-up PR — same design, additional tokenizer work — and keeping it separate keeps this diff reviewable.
- **Per-segment approval UI in the TUI.** This change is match-side only; the TUI still shows the full compound in the approval card. A future UI change could let users approve missing segments individually.

## Verification

- `go test ./internal/permission/... -count=1` — all green (15 tokenizer cases + 6 policy cases added)
- `go build ./cmd/reasonix` — clean
- `go vet ./...` — clean

Key test coverage:

- Compound splitting on `&&` / `||` / `|` / `;` / newline
- Operators inside single quotes, double quotes, `$(…)`, backticks are not splitters
- `2>&1` / `>&2` are not splitters
- Unbalanced quotes / parentheses return `nil` (safe fallback to exact match)
- Policy layer: three segments all allowed → Allow; any segment denied → Deny; any segment uncovered → Ask; deny wins over allow within the same compound

## Cache impact

Cache-impact: none — permission is evaluated before any LLM request, so this touches no system prompt bytes, memory prefix, tool schema, provider request serialization, or compaction behavior.

Cache-guard: N/A.
System-prompt-review: N/A.

## Refs

- Discussion of the design: #5702
- Prior art on why prefix synthesis from a compound is off-limits: #3882 (closed by-design). This PR takes a different route: match-side decomposition against existing atomic rules, no synthesis on save.
- Adjacent, already merged: #3815 makes the atomic prefix save path live in the first place — this PR lets compound commands make use of those atomic rules on the match side.

### DIFF
--- a/internal/permission/bash_decompose.go
+++ b/internal/permission/bash_decompose.go
@@ -15,27 +15,45 @@ import (
 //
 // It splits on the shell control operators `;`, `&`, `&&`, `|`, `||`, and
 // newlines. Quoting (single, double, backslash-escapes inside double quotes)
-// and $(...) / `...` command substitutions are treated as opaque — operators
-// inside them do NOT split the outer command. File-descriptor duplication
-// like `2>&1` is recognized (the `&` following an unquoted `>` does not
-// split).
+// and $(...) / <(...) / >(...) / `...` command / process substitutions are
+// treated as opaque — operators inside them do NOT split the outer command.
+// File-descriptor duplication like `2>&1` is recognized (the `&` following an
+// unquoted `>` does not split).
+//
+// Known out-of-scope shapes — the parser refuses to decompose these to keep
+// downstream matching safe, so callers fall back to whole-string matching:
+//   - heredocs (`cat <<EOF … EOF`): the delimiter body isn't shell syntax,
+//     but tokenizing it as one is wrong; we bail on any unquoted `<<`.
+//   - leading operator (`&& ls`, `; ls`): malformed shell.
+//   - unbalanced quotes, `$(...)`, `<(...)`, `>(...)`, or backticks.
 //
 // Returns nil when the input has no control operator to split on, or when the
-// parser encounters unbalanced quotes/parentheses/backticks — callers are
-// expected to fall back to whole-string matching in that case so no compound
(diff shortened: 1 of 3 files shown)


morningtzh:
谢谢补上,这个漏了。测试你也覆盖到了,接受合入。

PR #5709, opened 2026-07-01
fix(agent): make parallel tasks read-only

CVEngineer66:
## Summary

This PR narrows `parallel_tasks` to read-only sub-agent dispatch so concurrent subtasks cannot mutate the workspace or invoke writer tools.

## Why

`parallel_tasks` previously ran multiple sub-agents concurrently while exposing the normal writable sub-agent tool registry. That made it possible for parallel children to edit files or run write-capable tools at the same time without write-set isolation, locking, conflict detection, or merge semantics.

The tool schema also intentionally hid `depends_on`, but the runtime still parsed and honored it. That created an undocumented behavior path where callers could trigger dependency scheduling that was not part of the provider-visible contract.

## Changes

- Mark `parallel_tasks` as read-only and plan-mode safe.
- Run each child with `ReadOnlySubagentToolRegistry` instead of the writable sub-agent registry.
- Reject unknown arguments with `DisallowUnknownFields`, including hidden `depends_on`.
- Remove the hidden dependency/wisdom-file scheduling path.
- Remove the background-job fallback that could route back through writable `task` execution.
- Add regression tests for the read-only contract, hidden dependency rejection, and writer-tool isolation.

## Cache Impact

Cache-impact: medium - changes the provider-visible `parallel_tasks` tool description, so sessions started after this version may get a different cached tool-prefix entry. It does not introduce per-turn dynamic schema changes, and the JSON schema remains stable.
Cache-guard: `go test ./internal/agent -run 'TestParallelTasks'` covers schema stability, read-only classification, hidden `depends_on` rejection, and writer-tool isolation.

## Validation

- `go test ./internal/agent -run 'TestParallelTasks'`
- `go test ./internal/agent`
- `go test ./internal/control`
- `go test ./internal/...`

### DIFF
--- a/internal/agent/parallel_tasks.go
+++ b/internal/agent/parallel_tasks.go
@@ -1,23 +1,21 @@
 package agent
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
-	"os"
-	"path/filepath"
 	"strings"
 	"sync"
 
 	"reasonix/internal/event"
-	"reasonix/internal/jobs"
 	"reasonix/internal/provider"
 	"reasonix/internal/tool"
 )
 
-// ParallelTasksTool dispatches multiple sub-agent tasks concurrently and
-// collects all results. Each sub-task runs as a foreground sub-agent in its
+// ParallelTasksTool dispatches multiple read-only sub-agent tasks concurrently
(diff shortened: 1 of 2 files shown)


CVEngineer66:
@codex review

CVEngineer66:
@codex review

PR #5710, opened 2026-07-01
fix(memory): inject feedback memories into prompt

### LINKED ISSUE #5595: bug: global memories (type: feedback) not auto-injected at session start — user preferences silently ignored
## Bug Description

Global memories stored in `memory/global/` (particularly `type: feedback` rules like user preferences, communication style, behavioral constraints) are **not auto-injected** into the model's context at session start. The model must manually call `memory(operation="list")` and `memory(operation="read")` to discover them — but nothing in the system prompt signals this is necessary.

## Impact

User preferences and rules are silently ignored until the user notices and complains. In observed usage, this caused 3+ turns of incorrect behavior (model using polite/professional tone instead of the user's preferred aggressive style) before the user had to manually intervene.

## Steps to Reproduce

1. Create a global memory with `type: feedback` in `memory/global/`:
   ```
   memory(operation="remember", type="feedback", name="my-rules", 
          body="Always start replies with 'Hey boss'. Use casual tone.")
   ```
2. Start a new session
3. Observe: the memory-compiler injects `<memory-compiler-execution>` blocks with stale `learning:*` constraints, but does **not** inject any constraints from the global feedback memory
4. The model replies with default tone, completely unaware of the rule

## Expected Behavior

- Global memories with `type: feedback` should be auto-injected into the execution contract (`planner_ir.constraints`) at session start
- Or at minimum, surfaced in the system prompt similarly to how "Pinned Skills" are displayed
- The model should not need to manually probe for memories to follow user rules

## Actual Behavior

- Only compiler-derived `learning:*` traces and `noise_filter`/`strategy_selector` constraints appear in planner_ir
- Global memories are indexed but inactive — the model gets zero signal to read them
- The user must notice the problem and manually ask the model to check memories

## Environment

- Reasonix version: Go 1.0 (main-v2)
- OS: Windows 11
- Memory types tested: `feedback`, `user`

cyq1017:
Fixes #5595.

Problem:
- Saved memories of type `feedback` are only exposed through the MEMORY.md index at session start, so standing user guidance is not applied unless the model reads the linked file manually.

Change:
- Capture feedback memories during `memory.Load`.
- Inject feedback bodies as standing guidance in the Memory prompt block.
- Keep non-feedback saved memories linked-only as background context.

Cache-impact: medium - feedback memory bodies now become part of the cached Memory prompt block when saved feedback exists.
Cache-guard: `go test ./internal/memory -run TestBlockInjectsFeedbackMemoryBodies -count=1` verifies feedback bodies are injected while non-feedback bodies stay linked-only.
System-prompt-review: self-reviewed for cache prefix impact; change is limited to saved feedback guidance loaded at session start.

Verification:
- `go test ./internal/memory -run TestBlockInjectsFeedbackMemoryBodies -count=1`
- `go test ./internal/memory -count=1`
- `go test ./internal/control -run 'TestMemory|TestCompose' -count=1`\n- `git diff --check`

### DIFF
--- a/internal/boot/boot_test.go
+++ b/internal/boot/boot_test.go
@@ -166,7 +166,7 @@ model = "x"
 	if _, err := store.Save(memory.Memory{
 		Name:        "synthesis-cache-policy",
 		Description: "Stable conclusions should be reused from memory",
-		Type:        memory.TypeFeedback,
+		Type:        memory.TypeProject,
 		Body:        "Use a synthesis cache document when expensive retrieval produced a stable conclusion.",
 	}); err != nil {
 		t.Fatalf("save memory: %v", err)
--- a/internal/memory/memory.go
+++ b/internal/memory/memory.go
@@ -13,11 +13,12 @@ import (
 // UserDir are retained so the controller can resolve quick-add targets without
 // re-deriving discovery context.
 type Set struct {
-	Docs    []Source // REASONIX.md / AGENTS.md, ascending precedence
-	Store   Store    // auto-memory store (may be a zero/disabled Store)
-	Index   string   // MEMORY.md contents at load time
-	CWD     string   // project working dir used for discovery
-	UserDir string   // user config root (may be "")
+	Docs     []Source // REASONIX.md / AGENTS.md, ascending precedence
+	Store    Store    // auto-memory store (may be a zero/disabled Store)
+	Index    string   // MEMORY.md contents at load time
+	Feedback []Memory // saved feedback bodies captured at load time
+	CWD      string   // project working dir used for discovery
+	UserDir  string   // user config root (may be "")
 }
 
 // Options configures discovery. CWD defaults to "." and UserDir is the user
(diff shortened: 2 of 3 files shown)


esengine:
This is implemented. The prompt now carries a "Global preferences and feedback" section that inlines the **bodies** of `user` and `feedback` memories rather than linking to them — `BackgroundBlock()` in `internal/memory/memory.go`, fed by a filter that admits exactly those two types.

Your diagnosis was right, and it is the part that mattered: guidance sitting behind a link in MEMORY.md does not get applied, because nothing makes the model open it.

What changed relative to your patch is where the resulting text sits in the authority order. It is presented as preferences to apply when relevant, explicitly ranked below the current request and below standing instruction files, rather than as standing guidance itself. That separation exists so a stale saved fact cannot acquire instruction authority — a real risk precisely once bodies are inlined, which is why the section carries its own caveat sentence.

So the behaviour you asked for shipped, with a narrower claim about how binding it is.

Leaving the PR open afterwards was our mistake. Closing it now with that stated rather than letting it lapse.

Thanks for it — and for calling out the cache impact in the description, which is the part most people leave out.

PR #5745, opened 2026-07-01
fix: crash/performance overlay copy buttons now use 3-tier clipboard fallback

ttmouse:
## Problem

When the performance report panel or crash overlay appears (e.g. after Reasonix detects response lag), the **copy button does nothing** — clicking it has no effect.

## Root cause

Both panels' copy buttons (`paintPerformancePrompt` and `paint` in `crash.ts`) only call `navigator.clipboard?.writeText()` with **no fallback**. In Wails desktop webview, `navigator.clipboard.writeText` often fails silently because:

- The webview loads from `file://` which is **not a secure context**
- The overlay may lose transient activation
- `navigator.clipboard` itself may be `undefined`

Meanwhile, the regular React `CopyButton` component already had a full 3-tier fallback chain that the crash panels were not sharing.

## Fix

1. **Extracted** `writeClipboardText` + `fallbackCopyText` into a shared `src/lib/clipboard.ts` utility
2. **Updated `CopyButton.tsx`** to import from the shared utility instead of defining its own local copy
3. **Fixed `crash.ts`** — both copy buttons now use the shared `writeClipboardText` with 3-tier fallback:

   ```
   navigator.clipboard.writeText() → window.runtime.ClipboardSetText() → textarea + execCommand('copy')
   ```

### DIFF
--- a/desktop/frontend/src/components/CopyButton.tsx
+++ b/desktop/frontend/src/components/CopyButton.tsx
@@ -1,55 +1,7 @@
 import { useEffect, useRef, useState } from "react";
 import { Check, Copy } from "lucide-react";
 import { useT } from "../lib/i18n";
-
-function fallbackCopyText(value: string): boolean {
-  const activeElement = document.activeElement;
-  const selection = document.getSelection();
-  const ranges: Range[] = [];
-  if (selection) {
-    for (let index = 0; index < selection.rangeCount; index += 1) {
-      ranges.push(selection.getRangeAt(index));
-    }
-  }
-  const textarea = document.createElement("textarea");
-  textarea.value = value;
-  textarea.setAttribute("readonly", "");
-  textarea.style.position = "fixed";
-  textarea.style.inset = "0 auto auto 0";
-  textarea.style.width = "1px";
-  textarea.style.height = "1px";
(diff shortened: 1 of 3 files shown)


SivanCola:
Thanks @ttmouse for the three-tier clipboard fallback. That approach — async Clipboard API → Wails `ClipboardSetText` bridge → hidden-textarea `execCommand` — is now the shared `writeClipboardText` helper that the crash and performance overlay copy buttons route through, integrated via #6424 (with a native-bridge fallback regression test). Consolidating there to review and merge in one place; your contribution is preserved via that commit's `Co-authored-by` trailer. Closing as superseded — thank you!

感谢 @ttmouse 的三层剪贴板兜底方案。那套思路(async Clipboard API → Wails `ClipboardSetText` 桥 → 隐藏 textarea `execCommand`)现已作为共享的 `writeClipboardText` helper,供崩溃与性能弹窗的复制按钮统一调用,并通过 #6424 整合(含原生桥兜底的回归测试)。为集中评审、合并,这里合并进 #6424;你的贡献已通过该 commit 的 `Co-authored-by` 保留归属。作为已被取代关闭,谢谢!

PR #5820, opened 2026-07-02
 fix(serve): deduplicate tool card on partial dispatch + align card width

ZhUyU1997:
## Summary

Two fixes in `internal/serve/index.html`:

1. **Tool card width alignment** — Added `max-width: 760px` to `.card` so tool cards match `.msg` width instead of stretching full container width.

2. **Duplicate tool card on partial→full dispatch** — The backend emits `tool_dispatch` twice for the same tool call: first with `{partial: true}` (name only, no args), then with full args. The original code created a new card each time and overwrote `toolCards[tool.id]`, leaving the first card stuck in loading state forever. Now `renderToolDispatch` reuses the existing card element (clears `innerHTML` and repopulates) when a dispatch for the same `tool.id` arrives, preventing duplicates.

## Verification
| Before | After |
|--------|-------|
|  |  |
- Manually tested: running `ls` or `read_file` produces exactly one card per tool call, no leftover spinning card
- Card width now matches chat message width

### DIFF
--- a/internal/serve/index.html
+++ b/internal/serve/index.html
@@ -825,7 +825,9 @@ function finalizeMsg() { if(currentMsg){const c=currentMsg.querySelector('.curso
 
 // ── tool cards (v1 pattern) ──
 function renderToolDispatch(tool) {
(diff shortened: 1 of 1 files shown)


ZhUyU1997:
@SivanCola  please help check this PR for serve

PR #5841, opened 2026-07-03
fix: prevent session data loss from non-atomic writes and silent corr…

### LINKED ISSUE #5794: [Bug]: 对话框会吞记录
### Version line

### Exact version

1.15

### What happened?

今天早上使用中生成了代码并保存，下午来看代码一直在那里，但工作中和ai的对话记录都不见了

### Steps to reproduce

none

### OS / platform

win11

### Relevant logs or output

```shell
记录丢失
```

myipanta:
…uption (#5794)

Three data-loss vectors fixed:

1. copyOnto in ReplaceFile now writes to a sibling temp file, fsyncs, then renames atomically — never truncates the destination directly with os.WriteFile, which could leave a corrupted file on crash or power loss. This path is hit frequently on Windows when antivirus or search indexer briefly locks the destination.

2. Session.Save and saveSessionTitles/saveSessionDisplays now call tmp.Sync() before Close(), matching the AtomicWriteFile pattern. Without fsync, the OS may not flush the temp file to disk before the rename, leaving a truncated file on crash.

3. ListSessions now shows corrupted sessions (non-empty file that fails LoadSession) with a "⚠️ Session may be corrupted" warning in the sidebar, instead of silently hiding them.

Cache-impact: none — session persistence layer; no tool/agent/prompt changes. The write path changes are runtime-only and invisible to the model and prefix cache.

Cache-guard: go test ./internal/agent/ covers ListSessions behavior; the new TestListSessionsShowsCorruptedSession and
TestListSessionsSkipsEmptyFile validate the corruption-detection path.

Fixes #5794

## Summary

-

## Verification

-

## Cache impact

Cache-impact: TODO
Cache-guard: TODO
System-prompt-review: N/A

### DIFF
--- a/desktop/sessions.go
+++ b/desktop/sessions.go
@@ -84,6 +84,11 @@ func saveSessionTitles(dir string, m map[string]string) error {
 		os.Remove(tmpPath)
 		return err
 	}
+	if err := tmp.Sync(); err != nil {
+		tmp.Close()
+		os.Remove(tmpPath)
+		return err
+	}
 	if err := tmp.Close(); err != nil {
 		os.Remove(tmpPath)
 		return err
@@ -842,6 +847,11 @@ func saveSessionDisplays(dir string, m sessionDisplayMap) error {
 	}
 	tmpPath := tmp.Name()
 	if _, err := tmp.Write(b); err != nil {
+		tmp.Close()
+		os.Remove(tmpPath)
+		return err
+	}
+	if err := tmp.Sync(); err != nil {
 		tmp.Close()
 		os.Remove(tmpPath)
 		return err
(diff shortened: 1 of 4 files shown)


SivanCola:
Thank you @myipanta. #8053 supersedes this source PR after reviewing it against current main-v2. The integration keeps and adapts the durable sidecar-write and corrupted-session visibility objectives, including the current planner-display sidecar, and credits you with commit-level Co-authored-by attribution. It intentionally does not copy the proposed second-temp fallback because it would not remove the underlying rename/filter-driver failure mode. Focused session, Desktop, race, full-root, and cache-guard verification passed.

PR #5105, opened 2026-06-23
fix(bot): run turn off dispatch goroutine so approval replies aren't deadlocked

cecil-su:
### Problem

A bot turn that triggers interactive tool approval (or an `ask`) blocks inside `RunTurn` waiting for `ctrl.Approve` / `AnswerQuestion`. The **only** code path that calls those is `handleSlashCommand`, which runs on the **same** per-adapter `dispatchLoop` goroutine that drives `runTurn` inline (`gateway.go`).

So once a turn blocks on approval, the dispatch loop can never read the `/approve`, `/answer`, or card-action reply that would unblock it — the session deadlocks until the app is restarted. The Feishu/Lark card click returns its "操作已提交" toast (that callback runs on the lark-ws goroutine), but the queued `InboundMessage` is never consumed. The 30‑minute `ApprovalTimeout` only auto‑*denies*; it never lets a user actually approve over the bot.

Reproduced by:
- #4701 — 飞书/Lark 机器人发送审批/问答卡片后，机器人彻底不响应
- #4863 — 飞书审批请求无法响应问题
- #4402 — 微信机器人：权限审批时卡死

This is platform-independent (the deadlock is in `internal/bot`, shared by feishu/qq/weixin).

### Fix

Run `runTurn` on its own goroutine so the dispatch loop stays free to deliver the reply:

```go
go gw.runTurn(ctx, binding.Adapter, key, msg)
```

- Per-session serialization is unchanged — it's still held by the session lock (`active[key]`); the deferred `Release` continues to drain the pending queue, so one session never runs two turns concurrently.
- The render sink is now fully initialized (`sink.ctrl = state.ctrl`) **before** it is published via `state.sink.setTarget`, now that it can be reached concurrently.

### Test

`TestGatewayApprovalReplyUnblocksWedgedTurn` drives a real turn through the dispatch loop with a controller blocked on approval, then delivers `/approve` and asserts the turn completes. It **fails in 2s (deadlock) on the old inline call** and **passes with the fix**.

### Verification

- `go test ./internal/bot/...` — pass (bot, feishu, qq, weixin)
- `go test -race -count=5 ./internal/bot/...` — pass (race-clean)
- `go vet ./internal/bot/...` — clean
- `go test -run 'Approval|Approve|Ask' ./internal/control/` — pass

Fixes #4701
Fixes #4863
Fixes #4402

PR #5093, opened 2026-06-22
Resolve Windows Execution Issues, Statusline Test   Failures, and Bot Gateway Shortcut Normalization

Juko-div:
## Description
    This pull request resolves a series of platform-specific
  execution issues on Windows, corrects test failures in the custom   
  status line runner, improves the reliability of process tree        
  cancellation tests, and enhances user interaction handling in the   
  bot gateway.

    ---

    ## Key Changes

    ### 1. ACP Session Prompt Replay
    - **File**: `internal/acp/service.go`
    - **Details**: When reconnecting to an existing session, re-      
  emits `ApprovalRequest` and `AskRequest` events so that the
  frontend UI matches the active backend loop state.

    ### 2. Bot Gateway Ask Shortcut Normalization (Closes #5043)      
    - **Files**: `internal/bot/gateway.go`,
  `internal/bot/gateway_test.go`
    - **Details**: Extends input normalization to non-command user    
  inputs when a pending ask is active. Instead of only accepting      
  exact numeric choices, the bot gateway now wraps freeform replies   
  or multiple/semicolon-separated options (e.g. `"1;2"`) as
  `/answer <askID> <reply>`, improving usability. Fixes issue where   
  the bot hangs when the agent calls the ask tool in IM mode.

    ### 3. PowerShell Script Execution Policy Bypass (Closes #5035)   
    - **Files**: `internal/sandbox/shell.go`,
  `internal/sandbox/sandbox_test.go`
    - **Details**: Spawning PowerShell commands inside the sandbox    
  now specifies `-ExecutionPolicy Bypass`. This avoids permission     
  denied errors on Windows hosts running strict script execution      
  policies, preventing command execution failures.

    ### 4. Cross-Platform Statusline Test Stability
    - **File**: `internal/cli/statusline_test.go`
    - **Details**: Adapts the helper commands in
  `TestModelSwitchRefreshesCustomStatusline` to use `more` instead    
  of `cat` on Windows platforms, preventing test failures due to      
  missing POSIX utilities.

    ### 5. Windows Process Cancellation Test Optimization
    - **File**: `internal/tool/builtin/bash_cancel_windows_test.go`   
    - **Details**: Replaces the slow, nested PowerShell subprocess    
  spawn with `cmd.exe` (boosting boot performance), replaces static   
  sleeps with dynamic file/process polling, and raises the timeout    
  threshold to 60s for bulletproof parallel runs.

    ### 6. Git Ignore Lists
    - **File**: `.gitignore`
[...]

PR #5888, opened 2026-07-03
fix(feishu): switch to Interactive Card JSON 2.0 for proper markdown rendering

zlwu:
## Summary

Fix Feishu/Lark bot reply markdown rendering by switching to Interactive Card JSON 2.0, while preserving ReplyMessage API for proper reply/thread semantics.

## Problem

1. **Markdown lost in rendering** — The previous markdown-to-post conversion (`SimpleMarkdownToPost` from the Lark SDK) only supported `[links](url)` and `<at>` tags. Bold, italic, inline code, code blocks, and headings were all displayed as raw markdown syntax.

2. **ReplyMessage API was removed** — The first iteration dropped `ReplyMessage` in favor of `CreateMessage` to avoid the platform-level quote prefix (`Reply @user: original message…`), but this broke reply/thread semantics in group and topic conversations.

## Fix

- **Preserve ReplyMessage API** — `sendSDKContent` now uses `ReplyMessage` API when `ReplyToMsgID` is set, and `CreateMessage` when it is empty. This keeps reply/thread semantics intact while fixing markdown rendering. The `ReplyInThread` field is left at its default (not set) to match current thread behavior.

- **Switch to Interactive Card JSON 2.0** — `sendMessage` builds a JSON 2.0 card with a `markdown` element via `buildMarkdownCard`, sending with `MsgTypeInteractive`. This gives native CommonMark rendering (bold, italic, inline code, code blocks, headings, etc.) with no custom markdown parser needed.

- **Card size limit fallback** — if the card exceeds Feishu 30KB body limit (error codes 11310/11325), the adapter automatically retries as plain text.

- **Updated test** — `TestBuildMarkdownCard` validates the new card format. Added `TestSendWithReplyToMsgID` and `TestSendWithoutReplyToMsgID` to prove `ReplyToMsgID` is not dropped.

## Future consideration

In direct messages (DM), the platform-level reply quote prefix (`Reply @user: …`) is unnecessary since both parties already know the context. A future improvement could differentiate: suppress the reply prefix in DMs (use `CreateMessage`) while preserving it in group chats (use `ReplyMessage`). This would give cleaner DM output without sacrificing group-thread semantics.

## Verification

- `go test ./internal/bot/...` — all pass
- `go vet ./internal/bot/...` — clean
- `gofmt -l internal/bot/feishu/` — clean

## Cache impact

Cache-impact: none — `internal/bot/` is not in cache-sensitive paths.
Cache-guard: existing tests cover the changed code paths.
System-prompt-review: N/A

### DIFF
--- a/internal/bot/feishu/feishu.go
+++ b/internal/bot/feishu/feishu.go
@@ -557,6 +557,29 @@ func (a *adapter) sendSDKContent(ctx context.Context, msg bot.OutboundMessage, m
 	if err != nil {
 		return bot.SendResult{}, err
 	}
+
+	replyTo := strings.TrimSpace(msg.ReplyToMsgID)
+	if replyTo != "" {
+		req := larkim.NewReplyMessageReqBuilder().
+			MessageId(replyTo).
+			Body(larkim.NewReplyMessageReqBodyBuilder().Content(content).MsgType(msgType).Build()).
+			Build()
+		resp, err := client.Im.Message.Reply(ctx, req)
+		if err != nil {
+			return bot.SendResult{}, err
+		}
+		if resp == nil {
+			return bot.SendResult{}, fmt.Errorf("feishu reply error: empty response")
+		}
+		if !resp.Success() {
+			return bot.SendResult{}, fmt.Errorf("feishu reply error: %s", feishuCodeError(resp.Code, resp.Msg))
+		}
+		if resp.Data == nil {
+			return bot.SendResult{}, nil
+		}
+		return bot.SendResult{MessageID: stringPtrValue(resp.Data.MessageId)}, nil
+	}
+
 	chatID := strings.TrimSpace(msg.ChatID)
 	if chatID == "" {
 		return bot.SendResult{}, fmt.Errorf("feishu chat_id is empty")
--- a/internal/bot/feishu/feishu_test.go
+++ b/internal/bot/feishu/feishu_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"net/http"
 	"strings"
 	"sync"
 	"testing"
@@ -13,6 +14,7 @@ import (
 	"reasonix/internal/bot"
 	"reasonix/internal/config"
 
+	lark "github.com/larksuite/oapi-sdk-go/v3"
 	"github.com/larksuite/oapi-sdk-go/v3/event/dispatcher/callback"
 )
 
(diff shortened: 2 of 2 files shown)


zlwu:
如图：
image

SivanCola (changes requested):
Thanks for the screenshots and the focused Feishu rendering work. I don't think this is ready to merge yet.

The PR drops the Feishu ReplyMessage path whenever OutboundMessage.ReplyToMsgID is set. The gateway still fills ReplyToMsgID for bot responses, so this turns replies into standalone CreateMessage sends. That can regress Feishu reply/thread semantics, especially in group/topic-style conversations. Feishu's ReplyMessage API supports the same message types used here and exposes reply_in_thread, so the adapter should not silently discard the reply target while fixing markdown rendering.

Please preserve the ReplyMessage path when ReplyToMsgID is present, or make the no-reply behavior an explicit product/config decision with regression coverage. If Interactive Card markdown remains the desired rendering path, send it through ReplyMessage for reply sends where possible, and add a test proving ReplyToMsgID is not dropped.

I re-checked this against latest main-v2 after #5890 merged. The branch is mechanically mergeable and CI is green, and local integration checks passed:
- go test ./internal/bot/...
- go vet ./internal/bot/...
- git diff --check
- gofmt -l internal/bot/feishu/feishu.go internal/bot/feishu/feishu_test.go

zlwu:
Hi @SivanCola, thanks for the feedback! I have updated this PR:

- **Preserved ReplyMessage API** — `sendSDKContent` now uses `ReplyMessage` when `ReplyToMsgID` is set, and `CreateMessage` when empty. The Interactive Card (JSON 2.0 markdown) is sent through `ReplyMessage` for reply sends.
- **Added tests** — `TestSendWithReplyToMsgID` and `TestSendWithoutReplyToMsgID` use a mock HTTP client to verify the correct API path is called, proving `ReplyToMsgID` is not dropped.
- **All checks pass** — `go test ./internal/bot/...`, `go vet ./internal/bot/...`, `gofmt -l` all clean.

> **Future considerations**
> 
> In direct messages (DMs), the platform-level "Reply @user: ..." prefix is redundant — both parties already share the conversation context. A future improvement could differentiate behavior by chat type: use CreateMessage (no prefix) for DMs while continuing to use ReplyMessage for group/topic chats where the prefix helps maintain conversation context. This would give cleaner DM output without sacrificing group-thread semantics.
> 

Please take another look when you have time. Thanks!

---

Screenshots of this fix 

image

esengine:
This landed as part of `95cf355b2` (`feat(bot): consolidate related gateway improvements`), which names #5105, #5093, #5046 and #5888 as the work it adapts.

Four PRs arrived against the same gateway code inside about a week, each fixing a real problem, and they overlapped heavily. They were consolidated into one change rather than merged in sequence — taking them one at a time would have meant resolving the same conflicts three times over, with each author rebasing after every merge.

Leaving this PR open afterwards was our mistake. From your side it reads as a contribution that was ignored; it had already shipped. Closing it now, with that said rather than letting it lapse quietly.

Thanks for it. Interactive Card 2.0 is the difference between markdown rendering as markdown and rendering as literal asterisks — which is the difference between a usable bot and a demo.

PR #5890, opened 2026-07-03
fix(feishu): remove pending reaction emoji after turn completes

zlwu:
## Summary

The "OnIt" reaction emoji added when the bot starts processing a message was never removed. It stayed visible after the turn completed, which is confusing.

## Problem

`AddPendingReaction` in `gateway.go` added the reaction on message receipt, but there was no corresponding removal logic. The emoji persisted indefinitely.

## Fix

Changed `AddPendingReaction` to return a `func()` cleanup closure. The gateway defers this cleanup in `runTurn`, so the emoji is automatically removed when the turn finishes (success or error).

Only the Feishu adapter implements this interface. QQ and WeChat are unaffected.

Files: `internal/bot/gateway.go`, `internal/bot/feishu/feishu.go`, `internal/bot/gateway_test.go`.

## Verification

- `go test ./internal/bot/...` — 65/65 pass
- `go vet ./...` — clean
- `gofmt -l .` — no changes

## Cache impact

Cache-impact: none — `internal/bot/` is not in cache-sensitive paths.
Cache-guard: existing tests cover the changed code paths.
System-prompt-review: N/A

### DIFF
--- a/internal/bot/feishu/feishu.go
+++ b/internal/bot/feishu/feishu.go
@@ -570,14 +570,14 @@ func (a *adapter) sendSDKContent(ctx context.Context, msg bot.OutboundMessage, m
 	return bot.SendResult{MessageID: stringPtrValue(resp.Data.MessageId)}, nil
 }
 
-func (a *adapter) AddPendingReaction(ctx context.Context, messageID string) error {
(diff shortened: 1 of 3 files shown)


zlwu:
### **Before Fix**

image

---

### **After Fix**

image

PR #5904, opened 2026-07-04
fix(openai): add vLLM reasoning field compatibility

FightZhanAng:
## Problem

When Reasonix connects to a **vLLM-backed OpenAI-compatible provider**, chain-of-thought / reasoning text is silently dropped and never displayed, even though:

1. The model is configured with `reasoning_protocol = "auto"`
2. The API correctly returns the `reasoning` field
3. The model supports thinking/reasoning mode

### Root Cause

Reasonix's OpenAI provider (`internal/provider/openai/openai.go`) only parses the `reasoning_content` JSON field from API responses. However, **vLLM returns thinking text in the `reasoning` field** instead.

### Reproduction

1. Deploy a thinking-capable model on vLLM with thinking enabled:
   ```bash
   vllm serve Qwen/Qwen3.6-35B-A3B \
     --reasoning-parser qwen3 \
     --default-chat-template-kwargs '{"enable_thinking": true}'
 
Cache-impact: low - only adds a fallback field for vLLM, does not change cache hit logic
Cache-guard: run test ./internal/provider/openai/... to verify reasoning field parsing


esengine:
This landed. The `delta.reasoning` fallback is in the OpenAI provider today — `internal/provider/openai/openai.go` reads the field on the streaming path — and `3de4e1739` records the co-authors.

What happened is that you and the author of the sibling PR proposed the same approach independently, within a day of each other, against a branch that already carried a version of it. That is why neither diff was merged as written: the change was already in flight, and a credit commit was the honest way to record where the approach came from.

Leaving this PR open afterwards was an oversight on our side. From your side it reads as a contribution that was ignored, when it had in fact already shipped. Closing it now, and saying that out loud rather than letting it quietly disappear.

Thanks for it. A provider compatibility gap that only appears against a self-hosted vLLM deployment is not something we would have found without someone running one.

PR #5973, opened 2026-07-05
feat(ui): add mouse_capture config option for terminal-native text selection

nick-maderight:
## Summary

Adds a `[ui].mouse_capture` config option to `~/.reasonix/config.toml` so users can disable in-app mouse capture persistently.

When running inside terminal multiplexers like **tmux** (with `set -g mouse on`) or **screen**, the TUI captures all mouse events via `\x1b[?1002h` / `\x1b[?1006h` / `\x1b[?1003h` (Bubble Tea `MouseModeCellMotion`), which prevents the multiplexer from intercepting mouse drags for native text selection. The `MouseDragEnd1Pane → pbcopy` binding never fires.

Setting `mouse_capture = false` sets `v.MouseMode = tea.MouseModeNone` in View(), releasing mouse ownership back to the terminal so click-drag selection and the right-click context menu work normally. In-app scrollbar drag and wheel-scrolling are sacrificed while it is off.

## Changes

| File | Change |
|------|--------|
| `internal/config/config.go` | Add `MouseCapture *bool` field to `UIConfig` (`toml:"mouse_capture"`) |
| `internal/cli/cli.go` | Check `cfg.UI.MouseCapture` after model init and set `mouseCaptureOff` when explicitly `false` |
| `reasonix.example.toml` | Document the new option in the `[ui]` section |

The existing `REASONIX_DISABLE_MOUSE` env var continues to work unchanged — this is purely additive. The `*bool` pointer type means nil/absent preserves the current env-var-based default.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -651,6 +651,9 @@ func chatREPL(args []string) int {
 		m.statuslineCmd = cfg.Statusline.Command // custom status-line command, "" = built-in row
 		m.showReasoning = cfg.UI.ShowReasoning   // /verbose persistence: start with config default
 		m.cfg = cfg
+		if cfg.UI.MouseCapture != nil && !*cfg.UI.MouseCapture {
+			m.mouseCaptureOff = true
+		}
 	}
 
 	// /model support: a pure builder the TUI calls to rebuild on a different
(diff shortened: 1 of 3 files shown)


nick-maderight:
Closing and reopening with corrected authorship

Issue #5947, opened 2026-07-04
[Bug]: 自动打开对话


liko135:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.16.0

### What happened?

<img width="2067" height="1695" alt="Image" src="https://github.com/user-attachments/assets/bce3ab30-c7b0-4204-88af-c05f544ac7f0" />

### Steps to reproduce

1.让他清理reasonix历史文件，会打开所谓的恢复分支，然后软件就卡死了

### OS / platform

Windows

### Relevant logs or output

```shell

```

liko135:
是否可以去掉所谓的恢复分支，不然太麻烦了

SivanCola:
来看下

SivanCola:
Status update — not closing yet.

What shipped toward this:
- #5976 hid conflict-recovery copies from the normal topic list, added a crash-safe/bounded session event log, and swept stale lock/log sidecars from cleanup — reducing the recovery-cascade behavior discussed around this issue and #5940/#5960.

What's not resolved:
- The PR is explicit it does not claim to fix an unrelated freeze without a concrete crash log — the hang you reported after opening the recovery branch hasn't been specifically confirmed fixed by this change.

Please retest on v1.17.8 (current release): if cleaning up history and opening the recovery branch still hangs, a repro with the Reasonix log/backtrace at the freeze point would help pin down the remaining cause.

PR #5976, opened 2026-07-05
Fix session recovery persistence cleanup / 修复会话恢复持久化清理

SivanCola:
## Summary

- Hide conflict-recovery copies from the normal topic list and stop showing the recovery badge/title as a regular conversation. Recovery keeps the original topic; copies stay reachable from History and are marked as conflict copies in the session tree and History drawer.
- Add an append-only session event log with a compatibility checkpoint: normal saves append, while explicit rewrites remain versioned events.
- Make the event log crash-safe and bounded: torn tails left by a crash or full disk are truncated by the next save instead of failing `LoadSession`; replace events and checkpoint rewrites are fsynced; past a size threshold the log compacts to a single replace event and the `.jsonl` checkpoint is refreshed, so direct readers and older binaries stay bounded-stale instead of frozen at first save.
- Never touch files the native layer does not own: a legacy v0.x `.events.jsonl` transcript sitting at the native log path is read-ignored and preserved byte-for-byte; logs written by a newer schema fail loudly instead of being silently truncated.
- Keep one-shot saves single-file: force saves (subagents, guardian, migrations, forks) never bootstrap an event log and fold an existing one into a single replace event.
- Sweep the event log/index from every deletion surface through `store.SessionSidecarFiles` — `/clear`, serve delete, ACP delete, subagent delete, and desktop trash — so deleted conversations can neither leak on disk nor be resurrected by `LoadSession`.
- Move session discovery and artifact lifecycle onto the centralized session filename helpers; guardian sidecars are excluded from transcript discovery.
- Clean up stale lease metadata, session lock sidecars, and orphan display entries during safe desktop cleanup paths. Desktop trash/delete now refuses to touch a session whose lease is held by another live process, preserving cross-process mutual exclusion.
- Migrate checkpoint-only readers to event-aware APIs: prompt history (↑ recall), the serve session list, automatic topic titles, duplicate-trash comparison, history search recency, and doctor byte totals.
- Count recovery-copy activity in topic summaries (recency, unread, time filters) without double-counting turns, and hide recovery-only topics only when they are unpinned with no live runtime session.

## Context

This addresses the recovery-chain and stale artifact behavior discussed around #5940, #5947, and #5960. It does not claim to fix an unrelated process crash without a concrete crash log; it reduces the recovery cascade surface, makes the authoritative transcript tolerant of the crashes that used to trigger cascades, and prevents leftover lock/display/log sidecars from accumulating through the built-in cleanup and delete flows.

## Verification

- `go test -p 1 ./...`
- `go test .` from `desktop/`
- `git diff --check`
- New regression coverage: torn-tail self-heal, foreign/legacy log read-ignore (boot migration end-to-end), future-schema rejection, broken append chains, concurrent load-during-append, compaction growth bound, force-save single-file semantics, deletion sweeps on all five surfaces asserting no resurrection via `LoadSession`, external-lease trash guard, event-aware prompt history / serve list / auto titles, duplicate-trash content comparison, and the recovery-only topic hiding matrix.


SivanCola:
Review follow-ups addressed in 7bb7290c2 (after merging `main-v2` with #5977):

**Lease split-window on delete/trash paths (P1).** The one-shot busy probe is gone. Desktop trash, trash-reconcile, and permanent delete now acquire an `agent.SessionRemovalGuard` that non-blockingly takes both the session's save lock and lease lock **before** any artifact is moved, holds them across the whole operation, and deletes the lock sidecars atomically with the release — built directly on #5977's delete-while-holding-flock primitive (`tryTakeSessionLockFile` / `RemoveAndUnlock`, Unix unlink-under-flock, Windows delete-disposition). A live holder of either lock (another process, or another runtime in this process) surfaces as a sanitized "session is in use" error instead of a deletion. Regression tests hold real locks: the guard blocks lease acquisition while held, all three desktop entry points refuse while a lease is live and leave every file untouched, and the sidecars are gone after a successful trash.

**Interaction found while merging #5977.** Its overlong-filename test caught that the event-index sidecar (`<stem>.event-index.json`) exceeds NAME_MAX for pre-bounded recovery-cascade names, failing saves with ENAMETOOLONG. Overlong-named sessions now run checkpoint-only (no event log/index) until `reconcileOverlongSessionFilenames` renames them.

**Lint (P2).** QF1003 fixed (tagged switch on the save mode). `golangci-lint run` is clean on both the root and desktop modules locally with the CI version.

**#5968.** No dependency from this PR; it touches the same desktop files and will need a rebase after whichever lands first.

Verification on 7bb7290c2: `go test -p 1 ./...` (all packages), `go test .` in `desktop/`, `golangci-lint run` on both modules, `pnpm --dir desktop/frontend typecheck && check:css`, `git diff --check`.

Issue #5998, opened 2026-07-05
[Bug]: 检测到保存冲突


Suryxin-xx:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

v1.17.0

### What happened?

正常对话过程中，反复提示冲突

### Steps to reproduce

<img width="1000" height="700" alt="Image" src="https://github.com/user-attachments/assets/3a0a297b-27d8-4696-9083-0f0e5b39f3dd" />

### OS / platform

windows

### Relevant logs or output

```shell

```

Issue #5990, opened 2026-07-05
[Bug]: 切换模型失败


KangLong:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.16.2

### What happened?

切换模型的时候报错：

模型切换失败：session lease held by another runtime: c:\users\xxx\appdata\roaming\reasonix\projects\c--users-bwill-appdata-roaming-reasonix-global-workspace\sessions\20260705-071320.193344800-deepseek-deepseek-v4-flash.jsonl is held by DESKTOP-N1EBCJK-22972-546fba434eabbe16

### Steps to reproduce

1.打开Reasonix
2.在已有的会话中切换模型。
3.新建一个会话，切换模型
都会报错

### OS / platform

windows 10, windows11

### Relevant logs or output

```shell

```

Issue #5989, opened 2026-07-05
模型切换失败 [Bug]:


xky5000:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

v1.16.2

### What happened?

切换模型失败;

模型切换失败：this session is already open in another Reasonix window or still running in the background; close the other window or open a copy before changing model

### Steps to reproduce

模型切换失败：this session is already open in another Reasonix window or still running in the background; close the other window or open a copy before changing model

### OS / platform

win10

### Relevant logs or output

```shell

```

PR #6002, opened 2026-07-05
fix(agent): auto-reclaim stale session leases from crashed processes

HUQIANTAO:
## Problem

When a process crashes or is force-killed without releasing its session lease, the OS file lock persists and blocks all subsequent attempts to open that session. Users see: "this session is already open in another Reasonix window or still running in the background".

This is the root cause of multiple reported issues on Windows where the app crashes and sessions become permanently locked.

**Issues:** #5996, #5998, #5990, #5989

## Root Cause

`session_lease.go` — `TryAcquireSessionLease` fails with `ErrSessionLeaseHeld` when the OS lock is held, but does not check whether the holder process is still alive. The existing `TryReclaimCurrentProcessSessionLease` only handles same-process stale leases.

## Changes

| File | Change |
|------|--------|
| `proc_alive_unix.go` (new) | `processAlive(pid)` via `syscall.Kill(pid, 0)` |
| `proc_alive_windows.go` (new) | `processAlive(pid)` via `OpenProcess` + `GetExitCodeProcess` |
| `session_lease.go` | Add `tryReclaimStaleLease()` that probes the holder PID; call it from `TryAcquireSessionLease` on `ErrSessionLeaseHeld` |
| `session_lease_test.go` | Add tests: stale foreign lease reclaim, live process not reclaimed |

## How It Works

1. `TryAcquireSessionLease` fails with `ErrSessionLeaseHeld`
2. Read `.lease.json` to get the holder PID
3. Call `processAlive(pid)` — platform-specific check
4. If PID is dead: remove orphaned lock file + lease info, retry acquisition
5. If PID is alive: return the original error (holder is genuinely active)

## Testing

- All 9 existing session lease tests pass
- New test `TestSessionLeaseReclaimsStaleForeignLease`: verifies stale lease from dead PID is reclaimed
- New test `TestSessionLeaseDoesNotReclaimLiveProcess`: verifies live holder is not reclaimed

### DIFF
--- a/internal/agent/proc_alive_unix.go
+++ b/internal/agent/proc_alive_unix.go
@@ -1,0 +1,11 @@
+//go:build !windows
+
+package agent
+
+import "syscall"
+
+// processAlive reports whether the given PID is still running.
+func processAlive(pid int) bool {
+	// kill(pid, 0) sends no signal but checks whether the process exists.
+	return syscall.Kill(pid, 0) == nil
+}
--- a/internal/agent/proc_alive_windows.go
+++ b/internal/agent/proc_alive_windows.go
@@ -1,0 +1,21 @@
+//go:build windows
+
+package agent
+
+import "golang.org/x/sys/windows"
+
+// processAlive reports whether the given PID is still running.
+func processAlive(pid int) bool {
+	h, err := windows.OpenProcess(windows.PROCESS_QUERY_LIMITED_INFORMATION, false, uint32(pid))
+	if err != nil {
+		return false
+	}
+	var exitCode uint32
+	if err := windows.GetExitCodeProcess(h, &exitCode); err != nil {
+		windows.CloseHandle(h)
+		return false
+	}
+	windows.CloseHandle(h)
+	// STILL_ACTIVE (259) means the process has not exited.
+	return exitCode == windows.STILL_ACTIVE
+}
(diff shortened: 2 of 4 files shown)


SivanCola:
Thanks @HUQIANTAO for digging into stale session leases.

I am closing this as superseded by the newer lease/CAS work in #6023 and #6030 rather than merging it as-is.

Why:
- Current `main-v2` now uses the OS lock as the arbiter and already handles the common orphan states this PR was targeting: foreign/corrupt/missing lease metadata with a free lock is cleaned up or rewritten, and `.lease.lock` sidecars are retired safely.
- This branch is now behind `main-v2` and conflicts in `internal/agent/session_lease_test.go`.
- The new test here creates a `.lease.lock` file but does not actually hold the OS lock, so it does not prove the proposed reclaim path. Current `main-v2` already succeeds in that plain-sidecar case.
- Reclaiming by checking only PID liveness and then removing `.lease.lock` is not safe enough for this code path. PID reuse and permission failures can make process liveness ambiguous, and unlinking a lock file by path can let a new process create a second lock file while another process still owns the original lock inode. We need to keep the existing lock-arbiter pattern: take the lock non-blocking first, then clean up atomically.

If a reproducible case remains where a dead process leaves a lease that current `main-v2` cannot recover, please open a new issue or PR against current `main-v2` with that reproduction. The fix should preserve the lock-first invariant and include a test that actually exercises the held-lock/reclaim behavior.

Thanks again — this helped validate the right failure area, but the safe pieces have landed through the newer stack and this implementation should not be merged directly.

Issue #6043, opened 2026-07-05
[Bug]: 模型设置推理强度设置无效


ranwangPk:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

v1.17.1

### What happened?

<img width="1593" height="814" alt="Image" src="https://github.com/user-attachments/assets/509906e0-22c7-451b-acde-fbc677fc1190" />

模型接入 -> 配置 -> 兼容设置，
设置推理强度为关闭后 ，新建会话，会话窗口依旧是默认的 auto

### Steps to reproduce

模型接入 -> 配置 -> 兼容设置，
设置推理强度为关闭后 ，新建会话，会话窗口依旧是默认的 auto

### OS / platform

Win11

### Relevant logs or output

```shell

```

esengine:
Reasoning-effort handling was fixed since: the effort pipeline now respects the provider's declared effort vocabulary (#7350, fixing #7273's class of silently-ignored settings). Please update to a current release; if a specific provider still ignores your effort setting there, see also #7451 (provider-side effort capability detection) and reopen with the provider details.

PR #6061, opened 2026-07-06
Fix/6043 provider effort persist

TonyWu340826:
## Summary

- Fix provider effort level not being persisted when saving provider settings (#6043)
- Add `e.Effort = p.Effort` in `saveProviderConfig` to persist the effort level
- Add `effort` field to frontend `ProviderView` interface for proper data binding

## Verification

- Manual testing: Set effort level in provider settings → save → restart app → verify effort is preserved
- Code review: Confirm `Effort` field flows from frontend → backend → config storage

## Cache impact

Cache-impact: none
Cache-guard: none
System-prompt-review: N/A

- `Cache-impact`: `none` — This change only affects configuration persistence, not runtime caching or session state.
- `Cache-guard`: none — No cache-related tests needed; existing config tests cover the persistence path.
- `System-prompt-review`: N/A — No changes to provider-visible system prompts, memory prefix, output style, or skill index behavior.


TonyWu340826:
这个问题可以合并吗

PR #6077, opened 2026-07-06
fix(control): refuse NewSession while a turn is running

SivanCola:
## Summary

- `ClearSession` (and `Compact`) refuse to run while a turn is in flight, but `NewSession` had no such guard. Both the `Submit` `/new` path (`go c.NewSession()`) and the bot gateway invoke it asynchronously, so a mid-turn `/new` could swap the executor session out from under the run loop's direct session reads and race the turn-end snapshot onto the fresh path.
- Mirror `ClearSession`'s running guard in `Controller.NewSession` ("cannot start a new session while a turn is running"). Callers already surface errors: TUI notices `SlashNewFailed`, Submit notices "new session failed", serve returns the error, desktop propagates to the frontend.
- The bot gateway cancels the in-flight turn right before rotating, and that cancel is asynchronous — it now waits a bounded window (5s/10ms, same shape as the desktop's `waitControllerStopped`) for the turn to unwind, and on failure replies 新会话创建失败 instead of claiming a new session started while silently keeping the old one.

## Verification

- New test `TestNewSessionRefusesWhileTurnRunning`: refusal while running (path unrotated, executor session untouched), successful rotation once the turn stops.
- `go test ./internal/control ./internal/cli ./internal/serve ./internal/bot` — all green.

## Cache impact

Cache-impact: none — session lifecycle guard only; no provider request, system prompt, or tool schema changes.
Cache-guard: not applicable; existing NewSession/ClearSession controller tests plus the new guard test cover the change.
System-prompt-review: N/A

Follow-up to the /new triage around #5699.

### DIFF
--- a/internal/bot/gateway.go
+++ b/internal/bot/gateway.go
@@ -1126,8 +1126,18 @@ func (gw *BotGateway) handleSlashCommand(ctx context.Context, adapter Adapter, k
 			if cancel != nil {
 				cancel()
 			}
+			// NewSession refuses to rotate while a turn is running; the cancel
+			// above is asynchronous, so give the turn a bounded window to
+			// unwind before rotating.
+			deadline := time.Now().Add(5 * time.Second)
+			for state.ctrl.Running() && time.Now().Before(deadline) {
+				time.Sleep(10 * time.Millisecond)
+			}
 			if err := state.ctrl.NewSession(); err != nil {
 				gw.logger.Warn("new session failed", "err", err)
+				gw.sessions.ForceRelease(key)
+				_ = gw.sendText(ctx, adapter, msg, "新会话创建失败，请稍后重试。")
+				return
 			}
 			gw.rememberSessionReady(msg, state.ctrl)
 		}
--- a/internal/control/controller.go
+++ b/internal/control/controller.go
@@ -2639,9 +2639,17 @@ func (c *Controller) summarizeAt(ctx context.Context, turn int, from bool) error
 	if c.executor == nil {
 		return c.rewindFail(fmt.Errorf("checkpoints unavailable"))
 	}
-	if c.Running() {
-		return c.rewindFail(fmt.Errorf("cannot summarize while a turn is running"))
+	// Summarize rewrites the live session AFTER a provider round-trip, so the
+	// bare Running() check left a seconds-wide window for a turn to start and
+	// then have the log replaced under it. Hold the rotation gate from the
+	// boundary read through the post-rewrite snapshot.
+	if err := c.beginRotation(); err != nil {
+		if errors.Is(err, errTurnRunningRotation) {
+			return c.rewindFail(fmt.Errorf("cannot summarize while a turn is running"))
+		}
+		return c.rewindFail(err)
 	}
+	defer c.endRotation()
 	boundary, hasBound := c.checkpoints.boundary(turn)
 	if !hasBound {
 		return c.rewindFail(fmt.Errorf("summarize unavailable for turn %d (resumed session)", turn))
--- a/internal/control/controller.go
+++ b/internal/control/controller.go
@@ -59,6 +59,14 @@ import (
 // while one is already active in the same Controller.
 var ErrTurnRunning = errors.New("turn already running")
 
+// errTurnRunningRotation and errRotationInProgress are returned by the
+// session-rotation gate (beginRotation) when a rotation cannot proceed: a turn
+// is in flight, or another rotation already holds the gate.
+var (
+	errTurnRunningRotation = errors.New("cannot start a new session while a turn is running")
+	errRotationInProgress  = errors.New("cannot start a new session while another session change is in progress")
+)
+
 // errNoSessionPath is returned by snapshot when a session has content to persist
 // but no resolved session path — a misconfiguration (e.g. an unresolvable data
 // dir in a bot deployment) that previously dropped conversations silently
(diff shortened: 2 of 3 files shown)


SivanCola:
Confirmed the P1 and fixed it in a6f96dc.

The reviewer is right: the entry-time `if c.running` check was released before `Snapshot()`, and a turn starting inside that window (running still false) would then have its session replaced by the later `SetSession` — the guard only closed the "already running at call time" case, not "turn starts mid-rotation".

Fix: a rotation gate instead of a point-in-time read.
- `beginRotation()` atomically refuses when `c.running` OR another rotation is in progress, and sets `c.rotating` under `c.mu`; it stays held from before `Snapshot()` through the swap (`NewSession` and `ClearSession` both use it via `defer endRotation()`).
- `runGuarded` and `RunTurn` now also refuse to start while `c.rotating`. So turn-start and session-rotation are mutually exclusive in both directions — a rotation can't begin under a live turn, and a turn can't begin under an in-flight rotation. The run loop's session reference can no longer change under it.

Regression test `TestNewSessionRefusesTurnStartedDuringSnapshot` forces exactly the reported interleaving deterministically: it parks `NewSession` inside `Snapshot()`'s recovery callback (still holding the gate), then asserts a concurrent `RunTurn` is refused with `ErrTurnRunning`, `running` stays false, and the session is not mutated until the rotation completes. I verified the test **fails without the `c.rotating` checks** (panics at the RunTurn-refused assertion) and passes with them.

Verification:
- `go test ./internal/control ./internal/bot ./internal/cli ./internal/serve ./internal/acp` — green
- `go test -race ./internal/control -run 'TestNewSessionRefuses|TestConcurrentSnapshots|TestRunTurn|TestClearSession'` — green
- `go vet ./internal/control ./internal/bot` — clean

SivanCola:
Extended the fix to close the same TOCTOU window in every other session-mutating entry point (252ff75).

Auditing all live-session mutations (`SetSession`/`Session().Replace`/`CompactNow`) turned up the identical bare-`Running()`-then-mutate-later shape in five more interactive commands, all dispatched from `submit()`'s synchronous slash-command path so they can race a streaming turn:

| method | shape |
|---|---|
| `forkNamed` (`/branch N`, Fork) | `Running()` → `Snapshot()` → `SetSession` |
| `Branch` (`/branch`) | read running → `Snapshot()` → `SetSession` |
| `SwitchBranch` (`/switch`) | read running → `Branches()`/`LoadSession` → `SetSession` |
| `Rewind` (`/rewind`) | `Running()` → `Session().Replace` + `SnapshotRewrite` |
| `Compact` (`/compact`) | `Running()` → `CompactNow` rewrites live session |

All five now go through the same `beginRotation()`/`endRotation()` gate, held from before the snapshot/load through the swap; `runGuarded`/`RunTurn` already refuse while `c.rotating`. Each keeps its original "cannot X while a turn is running" message on the running path and reports `errRotationInProgress` for the window this closes.

Safety checks I verified:
- **No re-entrancy / deadlock**: none of these is called from inside a turn body — all callers are the slash-command dispatch and Fork wrappers. Their internal `Snapshot`/`Branches`/`SnapshotRewrite` calls touch only `snapshotMu`, never `rotating`.
- **Lock order unchanged**: `beginRotation` holds `c.mu` only briefly to set the flag, then releases; the gate is a flag, not a held mutex, so no interaction with `snapshotMu`/`c.mu` ordering.
- **Error shapes preserved**: `rewindFail` wrapping, `agent.BranchInfo{}` returns, and every user-facing message are byte-identical on the running path.

Regression test `TestSessionMutationsRefuseWhileRotating` asserts all five (plus `RunTurn`) refuse while a rotation holds the gate with the live session untouched, and that the gate can't be claimed while a turn runs. Verified it **fails when the rotating check is removed**.

Verification: `go test ./internal/control ./internal/bot ./internal/cli ./internal/serve ./internal/acp` green; `go test -race ./internal/control` green; desktop build+tests green; `go vet` clean.

SivanCola:
Confirmed and fixed in 203ab34. The finding is correct — and it was the worst instance of the class: `summarizeAt` rewrites the live session **after a provider round-trip**, so the unguarded window was seconds wide, not microseconds.

Why my previous audit missed it: I enumerated direct mutation sites in controller.go (`SetSession` / `Session().Replace` / `CompactNow`), but summarize delegates the `Session.Replace` to the agent layer (`Agent.SummarizeFrom/UpTo` in `internal/agent/compact.go`). I have now completed the audit from the opposite direction — every executor method the controller calls, classified by whether it mutates the session:

- `SetSession` ×7, `CompactNow`, `SummarizeFrom`, `SummarizeUpTo` — all now behind the rotation gate.
- `PruneStaleToolResults` — mutates the session, but its only caller is `maybeColdResumePrune` on the `Resume` path (session load, no concurrent turn on this controller), same class as `Resume` itself.
- `ReplaceTodoState`/`SeedTodoState`/`RebuildTodoState` — todo sidecar state, not the message log.

So the gated set is now closed over all user-triggerable live-session mutations: NewSession, ClearSession, forkNamed, Branch, SwitchBranch, Rewind, Compact, SummarizeFrom/UpTo.

Fix mirrors the others: `beginRotation()`/`defer endRotation()` from the boundary read through the post-rewrite `SnapshotRewrite`, original "cannot summarize while a turn is running" message preserved on the running path.

Test hardening: your finding also exposed that my `TestSessionMutationsRefuseWhileRotating` assertions were too weak — several checked only `err != nil`, which a bare-Running() revert can satisfy by failing later for an unrelated reason (e.g. missing boundary). Every gated method now asserts the `errRotationInProgress` sentinel specifically, plus the summarize entries and the "cannot summarize while a turn is running" running-path message. Reverse-verified: reverting summarizeAt to the bare check fails the test at exactly the SummarizeFrom assertion.

Verification: `go test ./internal/control ./internal/bot ./internal/cli ./internal/serve ./internal/acp` green; `go test -race ./internal/control` over the rotation tests green; desktop tests green; `go vet` clean.

PR #6094, opened 2026-07-06
test: guard prompt-prefix byte stability at boot and desktop rebind level

SivanCola:
## Summary

治本工程收尾：给 provider 前缀缓存加最终防线测试，补上 desktop 成本排查时确认的两个 CI 盲区——此前没有任何测试验证 (a) prompt 组装是确定性的、(b) desktop rebind 路径保持请求字节不变。

- **`TestBuildComposesByteStableSystemPrompt`** (internal/boot): two `Build`s over the same workspace/config must compose **byte-identical** system prompts. On failure it prints a window around the first diverging byte, so the drifting prompt section is named directly instead of dumping both prompts. Probe stability across processes is already covered by the persisted-snapshot tests from #6093; this pins the rest of the composition (memory compose, skills index, output style, workspace line, policies) against future nondeterminism.
- **`TestRebindKeepsRequestPrefixBytes`** (desktop): drives the real desktop rebind shape — `agent.LoadSession` + `sessionWithFreshSystemPrompt` + `Resume` on a freshly built controller — with a request-capturing provider, and asserts the next turn's leading request messages are **byte-identical** to the request sent before the rebind. This is the prefix-cache invariant itself: if it holds, a desktop rebuild costs zero cache misses.
- **`TestRebindWithDriftedPromptBreaksRequestPrefix`** pins the failure mode the guard exists for: a drifted fresh prompt rewrites the first message and the prefix diverges. If the swap policy ever changes to keep the persisted prompt for resumed conversations, this test flips and forces a conscious update of both guards.

Tests only — no production code changes.

## Verification

- `go test ./internal/boot` (16s) and full desktop suite (39s) green; `go vet` clean.
- Negative control built in: the drifted-prompt test demonstrates the capture-and-compare method detects prefix divergence.

## Cache impact

Cache-impact: none - test-only change; adds the guards that future cache-sensitive changes will be caught by.
Cache-guard: this PR IS the guard — boot-level composition determinism plus desktop rebind request-prefix byte equality.
System-prompt-review: requested from @SivanCola — test-only change (internal/boot/* glob matches the new test file); no provider-visible content changes, the new test pins existing composition bytes.

Refs #2945, #5614; builds on #6093.

### DIFF
--- a/desktop/session_prompt_bytes_test.go
+++ b/desktop/session_prompt_bytes_test.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"encoding/json"
+	"os"
 	"path/filepath"
 	"sync"
 	"testing"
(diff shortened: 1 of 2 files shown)


SivanCola:
Both findings addressed.

**P1 (cache-impact red)**: fixed before this comment — the PR body's `System-prompt-review: N/A` was rejected by the gate (working as designed: `internal/boot/*` matches the system-prompt-sensitive glob, test files included). The body now names the requested review and the gate re-ran green; all 15 checks pass.

**P2 (guard weaker than its claim)**: correct, and fixed in 1192fad. The old comparison (`after[:len(before)]` vs the first turn's request) only proved `[system, user1]` stability — the resumed prefix's assistant messages were never compared. Rewritten to the comparison you suggested: from ONE transcript saved after turn one, the same follow-up turn runs twice — on the original controller (no rebind, cache-warm baseline) and after the desktop rebind path — and the two requests must be **byte-identical end to end** (system prompt, both prior turns, and the composed follow-up). `TestRebindReproducesRequestBytes` replaces the old test; the drifted-prompt canary got the same baseline structure and now also pins the divergence to the leading system message.

Worth noting: the strengthened comparison immediately caught a bug — in the *test itself*: the baseline turn autosaves onto the shared session path, so the rebind was loading a transcript that already contained the baseline exchange. The rebind now loads an independent copy of the after-turn-one save (`copySessionFiles`). With that isolation the full-request comparison passes, i.e. load + swap + resume genuinely reproduces the live request bytes.

Verification: full desktop suite green (62s) including both rewritten tests; `internal/boot` unchanged since your review.

PR #6140, opened 2026-07-07
Support Claude plugin manifest installs / 支持 Claude 插件清单安装

SivanCola:
## Summary
- support `.claude-plugin/plugin.json` as a compatible plugin package manifest
- preserve existing Reasonix and Codex manifest behavior while labeling Claude manifests as `manifestKind: "claude"`
- update plugin package docs in English and Chinese

Cache-impact: low - touches install_source plugin detection only; it does not change existing provider-visible tool schemas or system prompt content.
Cache-guard: `go test ./internal/pluginpkg ./internal/installsource ./internal/hook`; full `go test ./...`; isolated install verification for `nextlevelbuilder/ui-ux-pro-max-skill`.

## Verification
- `go test ./internal/pluginpkg ./internal/installsource ./internal/hook`
- `go test ./...`
- verified `https://github.com/nextlevelbuilder/ui-ux-pro-max-skill` plans and installs successfully in an isolated Reasonix home

### DIFF
--- a/docs/PLUGIN_PACKAGES.md
+++ b/docs/PLUGIN_PACKAGES.md
@@ -17,8 +17,8 @@ directory.
   `https://github.com/obra/superpowers`.
 - A GitHub branch or subdirectory URL, such as
   `https://github.com/owner/repo/tree/main/path/to/plugin`.
-- A local directory that contains `reasonix-plugin.json` or
-  `.codex-plugin/plugin.json`.
+- A local directory that contains `reasonix-plugin.json`,
+  `.codex-plugin/plugin.json`, or `.claude-plugin/plugin.json`.
 
 Preview the install plan without writing files:
 
@@ -224,8 +224,9 @@ third-party install scripts during plugin installation.
 
 ## Codex Compatibility
 
-Reasonix also reads Codex plugin manifests at `.codex-plugin/plugin.json`.
-For packages such as Superpowers and Claude-style skill packs, Reasonix maps:
+Reasonix also reads Codex plugin manifests at `.codex-plugin/plugin.json` and
+Claude Marketplace manifests at `.claude-plugin/plugin.json`. For packages such
+as Superpowers and Claude-style skill packs, Reasonix maps:
 
 - `skills` to Reasonix skill roots.
 - `hooks/session-start-codex` to the Reasonix `SessionStart` hook when present.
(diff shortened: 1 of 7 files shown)


SivanCola:
Review applied directly in 6a68799 (plus PR-body cache-impact lines for the gate). Findings and fixes:

**[P1] Standard Claude plugins installed as zero capabilities, silently.** The official Claude plugin shape is a metadata-only `plugin.json` with skills auto-discovered from the conventional `skills/` directory (packs in the wild also use `.claude/skills/`) — the explicit `skills` field this PR read is the exception, not the rule (the validation repo `ui-ux-pro-max-skill` happens to declare it, which masked the gap). Claude-kind parsing now falls back to the conventional directories when the manifest declares none, adopting a directory only if it actually contains a `<name>/SKILL.md`. Explicit declarations disable the fallback; codex parsing is untouched (pinned by `TestParseCodexManifestNotAffectedByClaudeFallback`). Reverse-verified: removing the fallback fails `TestParseClaudePluginConventionSkillDirs`.

**[P2] Unmapped Claude capabilities were silently dropped.** `commands/`, `agents/`, `hooks/hooks.json`, `.mcp.json` now surface as install warnings ("Reasonix does not map yet; that capability will not be installed") through the existing warnings channel, so an install that lands half a package says so.

**[P3] Negative test for the new `includeCodexSessionStartHook` flag**: a claude-kind package shipping a `hooks/session-start-codex` file must NOT get it registered as an executable SessionStart hook — that convention belongs to codex manifests, and this is the security boundary of the flag.

**Docs**: compatibility section retitled to "Codex & Claude", fallback + warnings documented, and `marketplace.json` multi-plugin indexes explicitly noted as out of scope.

Also audited and confirming clean: all `ManifestKind` consumers (hook.go was the only kind-switch and this PR already converts it to `ManifestPath`); installed packages re-parse through `ParseDir` at runtime so the claude branch reaches every downstream; no new command-execution surface.

Verification: `go test ./internal/pluginpkg ./internal/installsource ./internal/hook` green, `go build ./...` + `go vet` clean; 6 new tests.

Remaining non-blocking (fine as follow-ups): `ParseDir` masks a parse error in an existing-but-corrupt manifest as "no ... found" (pre-existing pattern, worth a dedicated fix); GitHub probing now fetches up to 3 manifest paths per branch.

PR #6141, opened 2026-07-07
Fix quoted node eval hook autorepair / 修复 node -e hook 引号自愈

SivanCola:
## Summary
- Add conservative command normalization for copied JSON-escaped `node -e` commands that read JSON payloads from stdin.
- Apply the normalization when loading hooks, when saving desktop hook settings, and before running the TUI custom statusline command.
- Cover per-project repeated hook loading, safe non-repair cases, desktop settings persistence, and TUI statusline execution.

## Tests
- `go test ./internal/cli -run 'TestRunStatuslineCmd|TestModelSwitchRefreshesCustomStatusline' -count=1`
- `go test ./internal/hook -count=1`
- `cd desktop && go test . -run 'TestSaveHooksSettings|TestProjectHooksSettings|TestTrustProjectHooks' -count=1`

### DIFF
--- a/desktop/hooks_settings_app.go
+++ b/desktop/hooks_settings_app.go
@@ -69,6 +69,7 @@ func (a *App) SaveHooksSettingsForRoot(scope, projectRoot string, hooks []HookCo
 		if cmd == "" {
 			continue
 		}
+		cmd = hook.NormalizeCommand(cmd)
 		settings.Hooks[event] = append(settings.Hooks[event], hook.HookConfig{
 			Match:       strings.TrimSpace(h.Match),
 			Command:     cmd,
--- a/desktop/settings_app_test.go
+++ b/desktop/settings_app_test.go
@@ -1137,6 +1137,30 @@ func TestSaveHooksSettingsPreservesUnknownSettingsKeys(t *testing.T) {
 	}
 }
 
+func TestSaveHooksSettingsNormalizesQuotedNodeEvalHookCommand(t *testing.T) {
+	isolateDesktopUserDirs(t)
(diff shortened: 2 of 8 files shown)


SivanCola:
Review applied in 9535e41. One important correction to my own review first:

**The P2 I flagged ("well-formed commands lose shell expansion via direct exec") turned out to be a false alarm** — reverse-verification proved it. `shellparse.StaticFields` rejects runtime expansions outright, so a command containing `$VAR`/backticks never reached the direct-exec branch in the original code; only fully-static commands did, and for those `sh -c` vs direct exec are semantically equivalent. The original design was already safe on this axis.

What this commit does instead:

- **Makes the boundary explicit and pins it**: `spawnCommand` now direct-execs only commands it actually repaired, plus (Windows-only) recognized node stdin-hook evals where `cmd /c` mangles quoted JS — with a comment stating the contract, and `TestWellFormedNodeEvalKeepsShellSemantics` asserting `$VAR` expansion survives for a well-formed command. Behaviorally identical today; the test is the tripwire if `StaticFields` ever loosens or the exec-vehicle logic changes.
- **Hardens the quote-trim helpers**: they now require the closing marker instead of prefix-only matching, so an unterminated candidate (`\"abc` with no tail) can no longer be "repaired" into an unintended script.
- **Removes the unexplained `runtime.Gosched()`** from the test mock provider — hook/cli/agent/control suites plus a 5× statusline stress run pass without it. If it was papering over a real flake, it should come back with a comment naming the flake.

Verification: `go test ./internal/hook ./internal/cli ./internal/agent` green, desktop hooks-settings tests green, `go vet` clean. Merging once CI confirms.

PR #6142, opened 2026-07-07
fix(desktop): play attention sound for background prompts

SivanCola:
## Summary
- Play the awaiting-response attention sound from the global runtime event stream, so approval and ask prompts in background tabs notify immediately.
- Remove the modal-mount sound side effect and dedupe by tab and prompt id so replayed pending prompts do not chime again while same-id prompts from other tabs still notify.
- Add a focused frontend regression test for attention-event keying and replay dedupe.

## Validation
- `pnpm --dir desktop/frontend exec tsc --noEmit`
- `pnpm --dir desktop/frontend exec tsx src/__tests__/sound.test.ts`

## Cache-impact
None. Frontend-only notification sound handling.


SivanCola:
Review applied in 65cded0. Findings and fixes:

**[P1] Ask prompts double-chimed.** The PR moved the chime to the global runtime event stream and removed the ApprovalModal mount side effect — but `AskCard` still played `playAttentionChime()` in its `useEffect([ask.id])`, so every ask prompt on the active tab chimed twice (event stream + mount). The mount chime is removed with a comment pointing at the event-stream path as the single source; background-tab asks keep working through the global path.

**[P3] `pnpm test` skipped the new suite** — `sound.test.ts` was only wired into `test:all`. Added to the `test` script at the same position.

**[P4] Unbounded dedupe set** — one entry per prompt id for the app's lifetime. Now self-prunes past 512 entries (drops the oldest half by insertion order); replay dedupe only needs recently seen ids. Pinned by a 2000-prompt bounded-growth test that also asserts recent ids stay deduped after pruning.

Confirmed clean during review: the event stream's `tabId` field is real (Go `json:"tabId"`, passed through `onEvent` verbatim), replay/cross-tab dedupe semantics are correct, restart re-chimes pending prompts once (desired), and plan approvals behave identically to the previous mount-based path.

Verification: `sound.test.ts` 11/11, `tsc --noEmit` clean, `vite build` OK. Merging once CI confirms.

PR #6146, opened 2026-07-07
fix(desktop): re-arm the prompt chime after runtime rebuilds

SivanCola:
## Summary

#6142 合并后的评审发现（成立）：审批/ask id 是 **per-controller 的递增计数器**（`approvalManager.nextID`，从 "1" 开始），而桌面端切模型/effort/设置会**热重建**该 tab 的 controller——重建后的第一个审批往往复用同 tab 早前已消费过的 id，提示音去重把它误判为 replay，**静音**。操作序列很常见：批过一次审批 → 切个模型 → 下一个审批不响。

修复：在 `agent:ready`（(重)建完成信号，恰好是 id 重新编号的时刻）清除去重键——带 tab 的 ready 清该 tab 前缀；设置重建发出的**无 tab** ready 清全部。方向性取舍：多清只会让 replay 的 pending prompt 多响一次（那本来就是用户需要的提醒），少清则静音一个活的提示——宁多勿漏。

## Verification

- `sound.test.ts` 14/14（新增 3 个用例：重建后同 id 再响、tab 级清理不误伤他 tab、无 tab ready 清全部）；
- `tsc --noEmit` clean；`vite build` OK。

## Cache impact

Cache-impact: none - frontend notification sound handling only.
Cache-guard: not applicable.
System-prompt-review: N/A

Follow-up to #6142.

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -1792,6 +1792,7 @@ func (a *App) clearActiveSessionRuntime(tab *WorkspaceTab, oldCtrl control.Sessi
 	tab.resetTelemetry(path)
 	oldCtrl.CloseAfterDestroy()
 	a.emitProjectTreeChanged()
+	a.notifyTabRuntimeRebuilt(tab.ID)
 	return nil
 }
 
@@ -7345,6 +7346,7 @@ func (a *App) SetModelForTab(tabID, name string) error {
 	// The runtime now reflects the on-disk config; drop any deferred refresh.
 	a.clearDeferredRebuild(tab.ID)
 	a.persistTabSessionPath(tab, path)
+	a.notifyTabRuntimeRebuilt(tab.ID)
 	return nil
 }
 
@@ -7496,6 +7498,7 @@ func (a *App) SetEffortForTab(tabID, level string) error {
 	// The rebuilt runtime reflects the on-disk config; drop any deferred refresh.
 	a.clearDeferredRebuild(tab.ID)
 	a.persistTabSessionPath(tab, path)
+	a.notifyTabRuntimeRebuilt(tab.ID)
 	return nil
 }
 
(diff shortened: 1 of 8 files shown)


SivanCola:
Valid blocker — fixed in 7c437b2.

Confirmed every leg: `SetModelForTab`, `SetEffortForTab`, and `SetTokenModeForTab` swap the controller and return with no `agent:ready`, so the onReady clearing never fired for exactly the rebuilds the PR body names as the reproduction. Auditing all `tab.Ctrl = newCtrl` swap sites also turned up a fourth silent one the review didn't list: `clearActiveSessionRuntime` (clear-while-running) — same id-restart, same stale-key mute.

Between the two suggested fixes I took the dedicated event: emitting `agent:ready` from these paths would drag its reload semantics along (useController re-hydrates the session on ready — churn these switches deliberately avoid today). All four swap paths now emit a lightweight `runtime:rebuilt` with the tab id via `notifyTabRuntimeRebuilt`; the frontend subscribes alongside onReady and clears that tab's chime keys.

The requested test, `TestRuntimeRebuildsEmitRuntimeRebuiltForTab`, drives **real** model, effort, and token-mode switches end-to-end against a boot.Build'd controller and asserts each emits `runtime:rebuilt` carrying the tab id — reverse-verified (removing the effort-path emit fails it).

One piece of test-infra fallout worth flagging: the new emissions ride `asyncRuntimeEmitter`, whose nil-hook fallback called Wails `runtime.EventsEmit` directly — which **log.Fatals on the plain Background contexts desktop tests use**, killing the whole test binary from any emitting path (ten existing rebuild-path tests would have started dying one at a time). Rather than whack-a-mole hooks into each test, the fallback now routes through `runtimeEventsEmitFallback`, neutralized once in the package `TestMain`; per-instance `runtimeEvents.emit` hooks keep precedence, so event-asserting tests are unaffected and future tests can't step on this mine.

Verification: full desktop suite green (37.6s), `sound.test.ts` 14/14, `tsc --noEmit` clean, `vite build` OK.

PR #6189, opened 2026-07-08
Fix/api key passive leak 6178

myipanta:
## Summary

-

## Verification

-

## Cache impact

Cache-impact: low — additive: new SensitiveDataPolicy constant appended at end; RedactCredentials is post-processing, does not change provider cache prefix
Cache-guard: go test ./internal/sanitize/ -v (22 tests covering all credential patterns + zero false positives)
System-prompt-review: low — SensitiveDataPolicy is a new additive constant appended at end of all system prompts, no existing content modified

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -2431,7 +2431,7 @@ func (a *Agent) executeOne(ctx context.Context, call provider.ToolCall) toolOutc
 		if !json.Valid([]byte(call.Arguments)) {
 			detail = strings.TrimRight(detail, "\n") + "\nThe arguments were not valid JSON. Re-emit them exactly per this schema:\n" + string(t.Schema())
 		}
-		result = sanitize.RedactCredentials(result)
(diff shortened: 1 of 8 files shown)


SivanCola:
Thanks for taking this on @myipanta! #6178 has now been fixed by #6182 (merged as 7d68fb1a9), which covers the ground this PR targeted — tool-output redaction before model context and session-persistence redaction — plus config-gated subprocess-env filtering and sensitive-path hiding, digest-stable save/load behavior, and a `reasonix doctor redact-sessions` command for scrubbing pre-existing logs. Closing this one as superseded.

One idea here that #6182 deliberately did **not** adopt is the system-prompt `SensitiveDataPolicy` clause: prompt text is part of Reasonix's cache-stable prefix, so a prompt-layer ban needs its own cache-impact review (see the repo's Cache-impact PR checklist). If you'd like to pursue that as a focused follow-up on top of the merged runtime layers, it would be welcome.

PR #6258, opened 2026-07-09
Fix/right dock max width

myipanta:
## Summary

-

## Verification

-

## Cache impact

Cache-impact: TODO
Cache-guard: TODO
System-prompt-review: N/A


myipanta:
image
不应该限制宽度，可以让用户 自行决定

PR #6266, opened 2026-07-09
fix(serve): strip transient blocks from /history and /checkpoints

ZhUyU1997:
## Summary

`GET /history` and `GET /checkpoints` returned raw composed user messages
containing transient XML blocks (`<reasoning-language>`,
`<response-language>`, `<memory-update>`) meant only for the model. After a
page refresh, these tags appeared in the transcript and rewind picker.

Both endpoints now strip transient blocks server-side via
`agent.StripTransientUserBlocks()`, so the inline HTML client and standalone
React client receive clean text. The CLI/TUI and desktop already handled this;
serve was the only frontend missing it.

## Verification

- `go build ./internal/serve/` — compiles clean
- `go vet ./internal/serve/` — no warnings
- `go test ./internal/serve/ -count=1` — all tests pass

## Cache impact

Cache-impact: none — `internal/serve/` is not a cache-sensitive path (only
HTTP response formatting, no system-prompt or tool-schema changes).
Cache-guard: N/A — existing test suite covers the changed code paths.
System-prompt-review: N/A

### DIFF
--- a/internal/serve/serve.go
+++ b/internal/serve/serve.go
@@ -616,6 +616,9 @@ func historyMessages(msgs []provider.Message) []historyMessage {
 			}
 		}
 		hm := historyMessage{Role: string(m.Role), Content: m.Content}
+		if m.Role == provider.RoleUser {
+			hm.Content = agent.StripTransientUserBlocks(m.Content)
+		}
 		if m.Role == provider.RoleAssistant {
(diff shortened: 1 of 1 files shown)


ZhUyU1997:
@SivanCola pleased help review

PR #6269, opened 2026-07-09
fix(serve): make rewind checkpoint click go directly to scope selection

ZhUyU1997:
## Summary

After clicking a checkpoint in the rewind picker, the dialog flashes and
reappears as the checkpoint list instead of showing the scope selection
(both / conversation / code / etc.). The issue is that the click handler
only updated the selection index without advancing to the next stage.

## Verification

- `go build ./internal/serve/` — compiles clean

## Cache impact

Cache-impact: none
Cache-guard: N/A
System-prompt-review: N/A


ZhUyU1997:
@SivanCola please help review

PR #6276, opened 2026-07-09
Fix Windows encoded config reads

SivanCola:
## Summary

Fix Windows/global configuration paths that could fail on non-UTF-8 user-edited files, especially GB18030/GBK Chinese text in global hooks and related settings.

## Root Cause

Several user-editable text paths still handed raw bytes directly to strict JSON/TOML/dotenv/Markdown parsers. On Windows, files saved by common editors can be GB18030/GBK, UTF-8 BOM, or UTF-16, so the same content could work at project scope while failing at global scope depending on which loader was used.

## Changes

- Add shared UTF-8 decoding helpers for text-like user files (`fileencoding.DecodeToUTF8` / `ReadFileUTF8`, a thin wrapper over the existing detection cascade used by `read_file`; valid UTF-8 input is byte-identical).
- Route config, hooks, trust, dotenv, credentials, `.mcp.json`, system prompt files, commands, skills, output styles, plugin packages, memory, desktop state/settings, and related JSON/JSONL sidecars through the shared decoder.
- Keep raw-byte reads for binary, source editing, attachment, update/signature, checkpoint encoding detection, and copy-only paths. Job artifacts (captured subprocess output) also stay raw on both read paths: the incremental `bash_output` reader is byte-offset based, so decoding only the whole-file path would render one artifact in two encodings and risk garbling binary output via UTF-16 misdetection.
- Update Windows-facing path docs/help text from hardcoded `~/.reasonix/...` to `<Reasonix home>/...` where appropriate.
- Add GB18030 regression coverage for hooks, config, dotenv, `.mcp.json`, system prompts, commands, skills, output styles, plugin manifests, memory, heartbeat config, and `.gitignore` patterns.

## Compatibility

| Area | Old-data behavior | Conclusion |
| --- | --- | --- |
| Existing valid UTF-8 (no BOM) files | `Detect` short-circuits to identity; bytes fed to parsers and prompts are unchanged | safe |
| GB18030 / UTF-16 / BOM files | Previously failed strict parsers or injected mojibake/BOM bytes; now decoded to UTF-8 — the fix itself | intended change |
| Read-modify-write paths (memory quick-add, store index) | Whole-file rewrite normalizes such files to UTF-8; content is preserved, no mixed-encoding files | safe |
| Reasonix-owned JSON/JSONL sidecars | Always written UTF-8 by us; decode is identity | safe |

## Validation

- `git diff --check`
- `gofmt -l` clean (including the new helpers)
- `go test ./...`
- `./scripts/cache-guard.sh` — 8/8 pass

Cache-impact: low - prompt-feeding sources (system_prompt_file, skills, memory, output styles) are byte-identical for valid UTF-8 files (identity short-circuit); only previously-broken GB18030/BOM/UTF-16 files change bytes, which is the fix itself and a one-time invalidation for exactly those users. No tool schema, ordering, or serialization changes.
Cache-guard: scripts/cache-guard.sh pass (8/8 cases, tail_avg 92-95, threshold 90).
System-prompt-review: no prompt template bytes change — only the byte source for user-supplied prompt files is decoded; templates, skills index rendering, and tool schemas untouched.


YingshuoZhang:
good one

PR #6278, opened 2026-07-09
fix(desktop): preserve cancelled turn display history

### LINKED ISSUE #6260: [Bug]: 手动取消的对话再次对话会丢失部分历史记录
### Version line

### Exact version

v1.17.9

### What happened?

手动取消的对话再次对话有时候丢失历史记录，终止后再次对话操作记录会丢失，重新切换会话原本的思考过程，一些执行步骤都消失。甚至，原来进入plan mode对话，终止后退出plan mode再次对话，ai还是认为处于plan mode，但实际没有执行限制，已经不是了。

Image

### Steps to reproduce

1

### OS / platform

win11

### Relevant logs or output

```shell

```

GTC2080:
## Summary

- Preserve reasoning, streamed text, and tool cards from manually cancelled Desktop turns through session reloads using the existing display-only history sidecar, while keeping incomplete assistant/tool messages out of the model transcript.
- Reduce the preserved cancelled user message to user-authored text so a stale Plan Mode marker cannot affect the next turn after Plan Mode is disabled.
- Mark cancel-completed turns and add regression coverage for display restoration, canonical transcript cleanup, and on-disk cleanup.
- Fixes #6260.

## Verification

- `go test ./internal/control -run 'TestTurnOrchestratorCancel(PreservesVisibleUserPrompt|FlushesCleanTranscriptToDisk)|TestResumeClearsStale(Visible|Synthetic)InFlightTurn|TestCancelClearsPendingApprovalRuntimeStatus' -count=1`
- `cd desktop && go test . -run 'TestHistoryForTabRestores(Planner|CancelledExecutor)DisplayAfterReload|TestRecordPlannerDisplayEventKeepsNoticeDetail' -count=1`
- `gofmt -l internal/event/event.go internal/control/controller.go internal/control/runtime_status_test.go internal/control/turn_orchestrator_test.go desktop/tabs.go desktop/history_test.go desktop/tabs_planner_notice_test.go`
- `git diff --check HEAD^ HEAD`

## Cache impact

Cache-impact: low — only the provider-visible history after a cancelled turn changes; the system prompt and tool schemas are unchanged.
Cache-guard: targeted controller tests verify the cleaned in-memory and on-disk cancelled transcript.
System-prompt-review: N/A


SauronSkywalker:
If the content saved after the session message interruption is only visible to the user, but the model's modifications to the file have already occurred, could it lead the user to believe that the modifications have already taken place, which is indeed the case, but the model is unaware of this at the beginning of the next session.This further leads to the model not realizing that some code has been modified. If the user requests the model to roll back the recent changes at this point, the model would be unsure of what to do, as there is no record of the changes in the context，Perhaps a reminder should be added to inform users that the model will "forget" any content interrupted during the session

SivanCola:
Thank you for this contribution — it established the right architecture: keep a cancelled turn's already-rendered output as display-only history, separate from the canonical model transcript.

This work has been adopted and merged via #6666 (merge commit 2b43e0f8), which adapts your mechanism to the current main-v2 turn-finalization path and adds stable notice localization, cross-process-safe sidecar persistence, lifecycle cleanup (delete/purge/prune), and expanded end-to-end and race coverage.

Your contribution is credited with a commit-level Co-authored-by trailer on the merge commit, so it appears in your GitHub contribution record. Closing this PR as superseded by #6666.

PR #6304, opened 2026-07-10
fix(mcp): forward MCP tool result images to vision models / 将 MCP 工具结果图片结构化传给视觉模型

### LINKED ISSUE #6169: [Bug]: MCP 工具调用返回的图片被删除
### Version line

### Exact version

1.17.6 desktop

### What happened?

我正在尝试使用 https://github.com/alisaitteke/photoshop-mcp 连接到 Photoshop 让 agent 读取参数和图片信息，但 `photoshop_get_preview` 工具返回的图片被移除了，没有传递给 LLM，也没有在 UI 中显示

这个问题会影响很多测试和前端开发流程

补充信息：vscode 上的 copilot 也有问题，那边是在 UI 中显示图片，并且在给 LLM 的信息中将图片转成了一个 `vscode-chat-response-resource://`，但因为 https://github.com/microsoft/vscode/issues/313710 ，Agent 无法在后续自行读取这个图片，只能让用户自行复制聊天历史中的图片，粘贴到新的回复

没有安装 ps 的话，可以尝试用这些工具来验证：
- https://github.com/sam2332/Mcp-Server-Placeholder-Image-Generator
- https://github.com/ChromeDevTools/chrome-devtools-mcp (注意 [take_screenshot](https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/main/docs/tool-reference.md#take_screenshot) 在不使用 `filePath` 参数时才会将截图附加到回复)

### Steps to reproduce

1. 安装 https://github.com/alisaitteke/photoshop-mcp 并启动 Photoshop
2. 要求一个带视觉功能的 agent 描述当前 Photoshop 中的内容

模型可能会根据图层的文本信息猜测内容，要追问是否真的看到了图片

### OS / platform

Windows 11

SivanCola:
## Description

Fixes #6169

MCP tools that return image content items (e.g. `photoshop_get_preview`, chrome-devtools `take_screenshot`) had those items silently dropped by `parseToolResult`, so vision models never saw the image and the UI had nothing to show. This PR forwards tool result images to vision models through a structural channel instead of inlining base64 into the tool text.

### Why a structural channel

Tool output text passes through `truncateToolOutput` (32KB head+tail splice) before reaching providers. Any in-band base64 markdown larger than that — which is every real screenshot — gets spliced, and the splice notice contains no `)`, so a data-URI regex still matches across the gap and ships corrupted base64 to the API (reproduced during review of #6227: the captured "image" contained the splice notice text, which Anthropic rejects as a 400). Keeping images outside the text makes corruption structurally impossible.

### Changes

1. **`internal/tool/tool.go`** — new optional `ImageTool` capability interface (`ExecuteWithImages`), following the existing `Previewer`/`PlanModeClassifier` pattern.
2. **`internal/plugin/plugin.go`** — `parseToolResult` now returns images as data URLs alongside the text, leaving a short `[image: <mime>]` placeholder at each image's position. Each item is validated instead of trusted from the MCP server: mime whitelist (jpeg/png/gif/webp — the set every supported vision API accepts), 4MB base64 per image, 5 images per result, whitespace-normalized and decode-checked base64; anything else becomes a text placeholder rather than poisoning the request. `remoteTool` implements `ImageTool`.
3. **`internal/agent/agent.go`** — `executeOne` prefers `ExecuteWithImages` and carries the images on `toolOutcome`; the session's `RoleTool` message stores them in the existing `Message.Images` field. Text truncation applies to text only.
4. **`internal/provider/anthropic/anthropic.go`** — with vision enabled, a tool message with images builds `tool_result.content` as an array (text block + image source blocks — the API's supported shape). Text-only results keep plain string content with byte-identical serialization.
5. **`internal/provider/openai/openai.go`** — with vision enabled, tool images are injected as a synthetic user message after the turn's full run of tool results, because the OpenAI API accepts only text content parts under `role:"tool"`. Histories without tool images serialize byte-identically.

Non-vision models (e.g. DeepSeek) see only the short placeholder text — consistent with how user image attachments are already gated on `vision`.

### Backward compatibility

| Touch point | Old-data behavior | Conclusion |
| --- | --- | --- |
| `Message.Images` on tool messages (session JSON) | Field already exists with `omitempty`; old sessions deserialize with nil Images. New sessions read by an older binary: images ignored on tool messages (old providers embed user-role images only), no error. | safe |
| `contentBlock.Content` `string` → `any` (anthropic) | Request-side only, never persisted; a string value marshals byte-identically, guarded by `TestBuildRequestToolResultTextOnlyKeepsStringContent`. | safe |
| openai message loop (indexed → append + flush) | With no tool images the emitted message list is unchanged, guarded by `TestBuildRequestSkipsToolImagesWithoutVision`. | safe |
| `tool.Tool` interface | Unchanged; `ImageTool` is optional, non-implementing tools use the exact previous path. | safe |

### Relationship to #6227

Supersedes #6227 by @ttmouse — thank you for the investigation and the first implementation. Kept/adapted from it: the `parseToolResult` image-item parsing (Data/MimeType fields, png default) and the Anthropic `tool_result` array-content shape; ttmouse is credited with a commit `Co-authored-by` trailer. Reworked relative to it: images no longer round-trip through markdown text (the truncation-corruption blocker), OpenAI no longer puts `image_url` parts in tool messages (the API rejects them), and non-vision models no longer receive raw base64 text.

Follow-up (not in this PR): rendering tool-result images in the desktop transcript UI — the images are now on the session message, so the UI half of #6169 has the data it needs.

### Verification

- `go test ./internal/plugin/ ./internal/provider/... ./internal/agent/` — pass (new: 8 parseToolResult image tests, agent truncation-bypass test, 3 anthropic + 3 openai request-shape tests)
- `go test ./...` — pass
- `./scripts/cache-guard.sh` — pass (`long-tool-loop tail_avg=95 threshold=90`)
- `go vet`, `gofmt -l`, `git diff --check` — clean

---

Cache-impact: low — system prompt and tool schemas untouched; request serialization changes only when a tool message carries images (new data), guarded by byte-stability regressions for text-only histories on both providers

Cache-guard: ./scripts/cache-guard.sh — pass (tail_avg=95, threshold=90)

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -1231,11 +1231,12 @@ func (a *Agent) Run(ctx context.Context, input string) (runErr error) {
 			return &maxStepsPause{steps: a.maxSteps, key: a.maxStepsKey}
 		}
 
-		results := a.executeBatch(ctx, calls)
+		results, images := a.executeBatch(ctx, calls)
 		for i, call := range calls {
 			a.session.Add(provider.Message{
 				Role:       provider.RoleTool,
 				Content:    results[i],
+				Images:     images[i],
 				ToolCallID: call.ID,
 				Name:       call.Name,
 			})
@@ -1964,8 +1965,9 @@ func (a *Agent) systemPrompt() string {
 // goroutines; unknown and writer calls run as single-call serial segments so
 // write/read ordering stays provider-ordered. ToolResult events are emitted
 // after the batch in call order, so emission stays serial even when execution
-// parallelised.
-func (a *Agent) executeBatch(ctx context.Context, calls []provider.ToolCall) []string {
+// parallelised. The second return carries each call's tool-result images (nil
+// for most calls), aligned by index with the first.
+func (a *Agent) executeBatch(ctx context.Context, calls []provider.ToolCall) ([]string, [][]string) {
 	for _, c := range calls {
 		t, ok := a.tools.Get(c.Name)
 		ev := event.Tool{ID: c.ID, Name: c.Name, Args: c.Arguments, ReadOnly: ok && t.ReadOnly()}
@@ -2092,7 +2094,11 @@ func (a *Agent) executeBatch(ctx context.Context, calls []provider.ToolCall) []s
 	if !cancelled {
 		a.applyStormBreaker(calls, outcomes, results, receiptMark)
 	}
-	return results
+	images := make([][]string, len(calls))
+	for i := range outcomes {
+		images[i] = outcomes[i].images
+	}
+	return results, images
 }
 
 func (a *Agent) withPreviewFileDiffs(calls []provider.ToolCall) []provider.ToolCall {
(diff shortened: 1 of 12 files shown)


SivanCola:
Closing: per maintainer decision this fix was delivered directly on #6227 (commit 2e854a5c) instead of a separate PR.

PR #6317, opened 2026-07-11
feat: agent intelligence, provider optimizations, phantom panel UI, and safety systems

bfxh:
## Summary

This PR adds a comprehensive set of features across the agent core, provider layer, desktop UI, and safety systems:

### Provider Layer
- **LogProbs support** (OPT-04): Token-level log probabilities in Request/Usage/Chunk for confidence calibration
- **Circuit breaker** (OPT-06): Prevents cascading failures on overloaded APIs with per-provider tracking
- **Request deduplication** (OPT-09): DeduplicatingProvider wrapper merges concurrent identical requests

### Agent Core
- **Adaptive compaction**: Tail budget scales with context window size (32K→1M)
- **Semantic pruning**: Tool results pruned by importance (errors/tests > paths > normal output)
- **Parallel task safety** (P0-1): Deep copy of task list to prevent TOCTOU race conditions

### Agent Intelligence
- **Scene classifier**: Classifies user input by scene/complexity/tool-needs
- **Scene policy provider**: Per-scene policies (think mode, reasoning effort, pride signal limits)
- **Isolation system**: Four-level isolation (sandbox/zoned/observed/merged) for multi-session safety

### Agent Resilience
- **Incremental cache**: Buffers streaming content for interruption recovery
- **Side-effect tracker**: WAL-style recording of tool execution effects for crash recovery

### Agent Interaction (Multi-Panel System)
- **Phantom panel**: Real-time window projection and session transfer
- **Eye tracker**: Fixation detection and gaze-to-region mapping
- **Review gate**: Tiered code change review (auto/manual/blocked)
- **Tab manager**: Multi-tab session management

### Desktop UI
- **Phantom Panel**: Zero-token session status panel with Go channel-based updates
  - Dashed-border style, name-sorted entries, click-to-jump
  - Isolation level color coding and comm badges
  - 8 unit tests covering all functionality

### Safety
- **Safe delete system**: Path-protected deletion with quarantine support
  - Whitelist for Documents/Desktop/.git/go.mod etc.
  - Wired into Windows sandbox ACL cleanup

### Boot/Prompt Optimization
- **Prompt cache layering**: System prompt reordered into L1 (stable) / L2 (session) / L3 (volatile) layers
- **Tool schema compression** (OPT-05): Strips non-essential JSON Schema fields (14-70% token reduction)
- **Tool description sanitization** (OPT-08): Filters instruction injection in MCP tool descriptions

## Test Plan
- [x] Go backend compiles (`go build .`)
- [x] Frontend builds (`vite build`)
- [x] 8 Phantom Panel unit tests pass
- [x] 6 Agent integration tests pass
- [x] CLI binary runs correctly (`reasonix --version`, `reasonix doctor`)


SivanCola:
Thanks for the ambitious work here. The overall direction is valuable, especially the zero-token multi-session status surface, provider resilience, and cache-aware prompt organization. I reviewed head `a7343804d` against the latest `main-v2`, and I do not think this is ready to merge as one PR yet.

Key blockers:

1. **The provider test suite currently fails.** The new jitter is described as `+/-50%`, but the implementation only adds positive jitter and can exceed the existing capped backoff contract. `go test ./internal/agent ./internal/provider ./internal/plugin` fails at `TestBackoffDelay` with a delay above `maxBackoff + 250ms`.

2. **Request deduplication can merge semantically different requests.** `RequestKey` hashes only the provider name and the first five message roles/contents. It ignores later messages, tools, temperature, max tokens, images, and other request fields. Once enabled, two requests with the same prefix can receive the same replayed response even when their actual current turns differ. The dedup test only checks interface conformance and does not exercise concurrent or collision behavior.

3. **Phantom isolation does not currently enforce an isolation boundary.** `filterByIsolation` shallow-copies `PhantomEntry`, then mutates the shared `Conclusion` pointer when hiding or truncating the summary. Update events also publish the original, unfiltered entry to subscribers. This can both modify stored state and expose content that the selected isolation level is supposed to hide. Immutable/deep-copied filtered snapshots and focused event-stream tests are needed.

4. **Several advertised systems are not wired into the production boot path.** Incremental cache, side-effect tracking, request deduplication, agent Phantom UI, eye tracking, and review gating are only enabled by options that have no production callers. The scene classifier is initialized by default, but its policies do not control reasoning effort, thinking mode, or sampling; they currently produce notices only. Desktop conclusion and communication-badge update helpers also have no production call sites.

5. **The default MCP tool contract is changed without cache/behavior evidence.** Tool descriptions are truncated and schemas are rewritten for every MCP tool. This may reduce tokens, but it may also remove model-visible usage constraints. The cache-impact check is currently red because the PR does not provide the required cache-impact, cache-guard, and system-prompt-review evidence. Please add provider-visible schema stability/behavior tests before enabling this globally.

6. **Safe delete needs path-boundary semantics and tests.** `IsProtected` uses substring matching for entries such as `dev` and `sys`, so unrelated temporary paths can be classified as protected and silently left behind by cleanup callers. The new `internal/safeops` package currently has no tests.

Verification performed:

- `go test ./internal/agent ./internal/provider ./internal/plugin` -> agent and plugin passed; provider failed at `TestBackoffDelay`.
- `go test -race . -run TestPhantom -count=1` in the desktop module -> passed.
- `go test ./internal/safeops ./internal/sandbox` -> passed, but `safeops` has no test files.
- The PR is currently behind the latest base and conflicts in `desktop/frontend/src/styles.css`, `internal/agent/agent.go`, and `internal/boot/boot.go`.

I recommend splitting this into focused PRs: provider resilience, compaction/pruning, MCP schema/cache changes, safe delete, and a minimal end-to-end Phantom Panel. That would make each contract testable and allow the useful parts to land independently.

Issue #6337, opened 2026-07-11
[Bug]: AI编辑不了文件，即便已经读取过，只能执行替换


eghrhegpe:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.17

### What happened?

AI编辑不了文件，即便已经读取过，只能执行替换

### Steps to reproduce

Good, the file is fresh. Now let me make the edit to add the `_updateCheckerTexture` function.

✗
edit_file
frontend/src/scene/env/env-impl.ts
+65 -0
思考过程
I need to re-read the full file to get a fresh view. Let me do that.

已读 1 个文件
607→    if (v < 0) v += 1;
608→    tex.uOffset = u;
609→    tex.vOffset = v;
610→}
611→
612→/**
613→ * 同步地面法线贴图（bumpTexture）。
614→ * 仅 texture 模式且指定了法线路径时挂载；否则移除。
615→ */
616→function _syncGroundNormalTexture(mat: StandardMaterial, state: EnvState): void {

[more lines below; pass offset=616 to continue]
思考过程
Good, now I can see line 610 has `tex.vOffset = v;` and line 611 has `}` followed by line 612 with `/**`. Let me now make the edit to add the `_updateCheckerTexture` function between lines 610 and 612.

Actually, I see the issue - the fresh read only shows lines 598-648. Let me try the edit with the more specific context.

✗
edit_file
frontend/src/scene/env/env-impl.ts
+63 -0
    if (v < 0) v += 1;
    tex.uOffset = u;
    tex.vOffset = v;
}
/**
 * 棋盘格/程序化纹理原地更新：重新绘制 canvas 并替换现有贴图。
 * 不重建整个 material，避免失去引用（如反射纹理、边缘 fade）。
 */
function _updateCheckerTexture(mat: StandardMaterial, state: EnvState): void {
    const scene = getScene();
    const canvas = document.createElement('canvas');
    canvas.width = 128;
    canvas.height = 128;
    const ctx = canvas.getContext('2d');
    if (!ctx) return;
    const tileSize = Math.max(4, Math.round(16 * state.groundGridSize));
    const c0 = `rgb(${Math.round(state.groundColor[0] * 255)},${Math.round(state.groundColor[1] * 255)},${Math.round(state.groundColor[2] * 255)})`;
[...]

PR #6339, opened 2026-07-11
fix(prompt): guide model to re-read file before retrying edit_file/multi_edit

eghrhegpe:
Add instruction to edit_file and multi_edit descriptions and schema documentation: if 'old_string not found', first re-read the file with read_file — the content may have changed since last read — then retry.

**Cache-impact:** medium — changes tool descriptions and schema documentation strings, which are part of the system prompt prefix
**Cache-guard:** existing — tool schema changes are caught by cache-hit regression tests
**System-prompt-review:** nudge only — schema descriptions are parameter documentation, not behavioral logic

**Also addresses:** #6337

### DIFF
--- a/internal/tool/builtin/editfile.go
+++ b/internal/tool/builtin/editfile.go
@@ -24,11 +24,11 @@ type editFile struct {
 func (editFile) Name() string { return "edit_file" }
 
 func (editFile) Description() string {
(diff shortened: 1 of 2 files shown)


eghrhegpe:
Closing — this prompt-level change has marginal value; the oldStringNotFoundHint in edit_file's error message already tells the model to re-read. The agent-layer auto-annotation PR #6338 is the more impactful fix.

PR #6413, opened 2026-07-13
fix(control): check preApproved unconditionally in gateApprover (#3271)

### LINKED ISSUE #3271: [Bug]: tools use permission - keep asking
### Version line

### Exact version

desktop 1.2.1

### What happened?

if you start the prompt in auto mode, it will ask for permission every time, no matter which you clicked, allow for session or always allow.

however, if you start in plan then start exec, it is ok.

### Steps to reproduce

1. start in auto mode
2. ask to fix code
3. it will ask for permissions

### OS / platform

mac

### Relevant logs or output

```shell

```

myipanta:
The preApproved() check (YOLO mode, plan-execution window, session grants) was incorrectly gated behind guardianSess != nil. For users without a guardian session, 'Allow for session' and 'Always allow' were stored but never checked — every tool call re-prompted regardless.

Move the preApproved check before the guardian condition so it always short-circuits the prompt when a session grant, plan-execution bypass, or YOLO posture covers the call.

Fixes: #3271

## Summary

-

## Verification

-

## Cache impact

Cache-impact: TODO
Cache-guard: TODO
System-prompt-review: N/A

### DIFF
--- a/internal/control/auto_plan.go
+++ b/internal/control/auto_plan.go
@@ -38,6 +38,15 @@ func normalizeAutoPlan(mode string) string {
 }
 
 func (c *Controller) maybeAutoPlan(ctx context.Context, input string) {
+	c.mu.Lock()
+	suppressed := c.suppressAutoPlan
+	if suppressed {
+		c.suppressAutoPlan = false
+	}
+	c.mu.Unlock()
+	if suppressed {
+		return
+	}
 	if c.shouldAutoPlan(ctx, input) {
 		c.SetPlanMode(true)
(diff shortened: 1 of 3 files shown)


esengine:
This landed. `8c548f25d` (`chore(mcp): credit shared contributors`) records the work integrated from #6413 alongside #6033.

Checking `preApproved` unconditionally is the right shape — a guard that is only consulted on some paths is not a guard, and the asymmetry is invisible until the one path that skips it is the one that matters.

Leaving the PR open after integrating it was an oversight. Closing it now rather than letting it sit there looking unread.

Thanks for it.

PR #6416, opened 2026-07-13
fix(desktop): add accent colour underline to active tab / 选中tab底部增加主题色条

### LINKED ISSUE #6391: [Feature]: 顶部 tab 的选中效果太弱了
### Version line

Image

### Proposed solution

_No response_

JesonChou:
## Summary

Fix #6391

Add a 2px accent-coloured underline to the active tab using `box-shadow: inset`, making the selected tab immediately distinguishable from inactive ones. Uses `var(--project-accent, var(--accent))` to respect per-project accent colours.

`::after` pseudo-elements were avoided because `.tabbar__tab--drop-after::after` already occupies the `::after` slot for the drag-drop indicator.

- **macOS**: accent bar at bottom edge — the always-on `[data-theme-style]` layer flattens the tab and owns the accent, so it renders like the other platforms
- **Windows / Linux native-tabs**: accent bar at bottom edge (`inset 0 -2px`)
- **Classic mode** (base `.tabbar__tab--active`): accent bar at bottom edge
- **workspace-tabs-bar**: accent bar at bottom edge
- **`[data-theme-style]`** creation mode: accent bar at bottom edge, composed with existing graphite shadow

> Maintainer note: `applyTheme` always sets `data-theme-style`, so the `[data-theme-style]` rule wins on every platform and renders the accent as a uniform bottom-edge marker. The earlier macOS top-edge (`inset 0 2px`) declaration never applied and has been reverted to its base box-shadow in a follow-up commit; a `.tabbar__tab--active:focus-visible` rule was also added so keyboard focus keeps the active marker.

## Verification

- **Active tab indicator**: Open multiple tabs. The active tab shows a 2px accent-coloured underline. Switch between tabs — the underline moves to the newly selected tab.
- **Theme switching**: Cycle through the 6 built-in themes. The underline colour changes to match the current theme's accent.
- **Project colours**: Open tabs from different projects with custom accent colours. Each tab's underline uses its project's colour.
- **Drag-drop**: Drag a tab to reorder. The drop indicator (vertical accent line) and the active underline do not interfere.
- **macOS styling**: macOS active tabs keep their original border and chat-bg background; the accent renders as a subtle bottom line through the themed layer.

## Cache impact

Cache-impact: none — pure CSS, no model/provider/serialization path touched.
Cache-guard: `pnpm --dir desktop/frontend build` passes; 77/77 frontend CSS contract tests pass.
System-prompt-review: N/A — strictly client-side CSS, no prompt or memory prefix changed.

## Result

QQ_1783936613609

### DIFF
--- a/desktop/frontend/src/__tests__/app-chrome-tabs.test.ts
+++ b/desktop/frontend/src/__tests__/app-chrome-tabs.test.ts
@@ -115,6 +115,30 @@ ok(
 );
 
 ok(
+  finalDeclaration(":root[data-theme-style] .tabbar__tab--active", "box-shadow")?.includes(
+    "inset 0 -2px 0 var(--project-accent, var(--accent))",
+  ),
+  "active themed tab carries the project-accent underline",
+);
+
+ok(
+  finalDeclaration(":root[data-theme-style] .tabbar__tab--active:focus-visible", "box-shadow")?.includes(
+    "inset 0 -2px 0 var(--project-accent, var(--accent))",
+  ) &&
+    finalDeclaration(":root[data-theme-style] .tabbar__tab--active:focus-visible", "box-shadow")?.includes(
+      "0 0 0 3px var(--accent-soft)",
+    ),
+  "keyboard focus on the active tab keeps both the focus ring and the accent underline",
+);
+
+ok(
+  matchingBlocks(".app--darwin .app-chrome--tabs .tabbar__tab--active").every(
+    (block) => !block.includes("inset 0 2px"),
+  ),
+  "macOS active tab declares no dead top-edge accent (the themed bottom-edge layer owns it)",
+);
+
+ok(
   /workbenchChrome \? \(\s*<span className="app-chrome__spacer" aria-hidden="true" \/>/s.test(appChromeSource),
   "AppChrome workbench branch skips the tab strip",
 );
--- a/desktop/frontend/src/styles.css
+++ b/desktop/frontend/src/styles.css
@@ -19782,7 +19782,9 @@ body > .mermaid-diagram--fullscreen {
   color: var(--fg);
   background: var(--chat-bg);
   border-color: var(--border-soft);
-  box-shadow: 0 -1px 0 color-mix(in srgb, var(--fg) 5%, transparent);
+  box-shadow:
+    0 -1px 0 color-mix(in srgb, var(--fg) 5%, transparent),
+    inset 0 2px 0 var(--project-accent, var(--accent));
 }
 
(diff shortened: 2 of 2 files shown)


SivanCola:
Coordination note re: overlap with #6433.

#6433 originally also added an active-tab accent underline on the same `:root[data-theme-style] .tabbar__tab--active` box-shadow this PR changes. I've since narrowed #6433 to cover only tab boundaries (spacing, per-tab outlines, and an inter-tab divider) and dropped its underline, so this PR owns the active-tab accent marker. It's the better home for it: it fixes #6391, places the accent per platform (top edge on macOS), and honours per-project accent colours via `var(--project-accent)`.

The only overlap left is a trivial adjacent-line merge in that one rule — this PR's `box-shadow` vs #6433's `font-weight: 600` — which resolves by keeping both. Suggested order: this PR first, then #6433 rebases on top.

Nice, well-scoped fix — thanks!

SivanCola:
Pushed two small follow-up fixes to this branch (maintainer edit, `d9aaecc6`) after tracing the active-tab cascade:

1. **macOS accent direction.** `applyTheme()` always sets `data-theme-style`, so the later `:root[data-theme-style] .tabbar__tab--active` rule (and the grouped `.app--darwin` rule) win over the earlier `.app--darwin .app-chrome--tabs .tabbar__tab--active` declaration. The `inset 0 2px` top-edge accent therefore never rendered — every platform already shows the themed bottom-edge accent. I reverted that dead top-edge declaration to its base `box-shadow`; the accent stays a uniform bottom-edge marker owned by the themed layer.

2. **Keyboard focus dropped the active marker.** `:root[data-theme-style] .tabbar__tab:focus-visible` has higher specificity than `.tabbar__tab--active` and rewrote the whole `box-shadow`, so a keyboard-focused active tab lost its underline. Added `.tabbar__tab--active:focus-visible` composing the focus ring, graphite shadow, and project-accent underline; non-active focused tabs keep just the ring.

Also added `app-chrome-tabs` contract coverage for the active underline, focus preservation, and the absence of the dead top-edge accent (both new assertions fail on the pre-fix CSS and pass after). Verified: `pnpm exec tsx src/__tests__/app-chrome-tabs.test.ts` (77 passed), `pnpm typecheck`, and `pnpm check:css` all pass.

I also updated the PR description's macOS lines to match. Note that #6433 touches the same `.tabbar__tab--active` rule for boundary styling — it's been narrowed to defer the accent to this PR, so the plan is to land this one first and rebase #6433 on top. Thanks for the clean fix!

Issue #5561, opened 2026-06-29
[Bug]: 计划已就绪待确认时候切换到别的会话再回来会导致待确认框消失


SauronSkywalker:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.13.1

### What happened?

https://github.com/user-attachments/assets/5cc212e5-553f-4f76-adbf-f67c97830adb
计划已就绪待确认时候切换到别的会话再回来会导致待确认框消失
同时也没法退出对话，只能退出客户端

### Steps to reproduce

计划已就绪待确认时候切换到别的会话再回来会导致待确认框消失

### OS / platform

win11

### Relevant logs or output

```shell

```

SivanCola:
已在最新 `main-v2` 里覆盖这个根因。

复查下来，这个现象是计划确认框属于 pending approval 事件；切换会话后如果前端错过或清掉了这个事件，后端仍在等待 `exit_plan_mode` 确认，界面就会进入“还在运行，但没有确认框可点，也不好停止”的死锁状态。

相关修复已经在 `main-v2`：
- `bd6b41439`：切换 tab 时重新调用 `ReplayPendingPrompts()`，让 plan approval / ask / tool approval 重新发给前端并恢复确认框。
- `761e96690`：前端重连/重载后也会重放仍在阻塞的 approval/ask。
- 最新 `main-v2` 还包含 PR #5704，对切换历史会话时的权限/模式 profile 做了按 session 恢复，避免 Ask/Auto/YOLO 状态串台。

我复查的最新 `main-v2` head 是 `ebd73457d`，并验证了：
- `go test ./internal/control -run 'TestReplayPendingPrompts|TestRuntimeStatus'`

等桌面构建包含上述提交后，请再验证一次；如果仍能复现，请补充构建号/commit、切换路径，以及当时是否同时有多个会话在生成计划。

esengine:
Fixed by `95390ad18` — live approval and ask prompts are now preserved across stale runtime snapshots.

The cause was that switching away and back replayed a snapshot taken before the prompt existed, so a pending approval or a plan awaiting confirmation was overwritten by a state that predated it. The prompt had not been answered or cancelled; it was simply painted over.

Closing. If a prompt still disappears on a current build, that is a new defect — worth its own issue with the sequence of tabs you switched through.

PR #6432, opened 2026-07-13
fix(desktop): keep pending plan approval visible across session switches / 修复切换会话后待确认审批不显示

### LINKED ISSUE #6429: [Bug]: windows desktop 切换会话导致待确认不显示
### Version line

### Exact version

1.17.12

### What happened?

切换会话导致待确认不显示
切到别的会话，提示音提示我上一个任务的计划系统列出来了，等我审批是同意计划还是修改计划，左边导航提示待确认，点进去却不显示计划确认选项，只显示输入框，来回切几次会话，双击几次会话在导航中的标题，就有概率显示出来
Image

### Steps to reproduce

切换会话导致待确认不显示
切到别的会话，提示音提示我上一个任务的计划系统列出来了，等我审批是同意计划还是修改计划，左边导航提示待确认，点进去却不显示计划确认选项，只显示输入框，来回切几次会话，双击几次会话在导航中的标题，就有概率显示出来

### OS / platform

win11

### Relevant logs or output

```shell

```

SivanCola:
Fixes #6429. Same root-cause family as #5561 and #5481, which remain open for separate verification against their recordings.

## Symptom

Switching to a session whose plan approval or question was pending could briefly render the prompt and then lose it. Clicking the session again often restored it. Adjacent races could also resurrect an already answered prompt or hide a new prompt after a controller rebuild.

## Root causes

The frontend stores pending prompts in `state.approval` / `state.ask`, while Wails calls and runtime events are delivered independently:

1. A runtime snapshot fetched before a replayed prompt could be dispatched after it and clear the live prompt as stale backend state.
2. A delayed replay could re-arm a prompt already resolved optimistically in the frontend.
3. Approval IDs restart from `1` after a controller rebuild, so ID-only async callbacks from the old controller could affect a new same-ID prompt.
4. Auto/YOLO mode changes drain only the approvals the new policy can actually allow. Fresh plan, memory, managed-config, and sandbox-escape decisions remain pending, so blanket frontend clearing strands those turns.

## Fix

- Stamp prompt arrivals and runtime snapshot fetches with monotonic clocks; ignore snapshots that predate the visible prompt and schedule an authoritative follow-up reconcile.
- Tombstone locally resolved prompt IDs so delayed same-ID replays stay suppressed. Failed approve/answer RPCs remove that tombstone and request a backend replay.
- Track a per-tab `promptEpoch`, incremented on controller rebuild and session reset. Submit failures and mode-drain results carry the epoch captured when their RPC started, so late old-controller callbacks cannot affect a new same-ID prompt.
- Make `ApplyMode` / `ApplyToolApprovalMode` return exactly the approval IDs drained by the backend. The frontend dismisses only those IDs; fresh or policy-uncovered approvals remain visible.
- Preserve tab-scoped routing throughout activation, hydration, cancellation, mode changes, and prompt replay.

## Protected-workflow audit

- Send / stop / cancel: explicit cancel still clears the current prompt and the existing cancellation reconcile remains authoritative.
- Tab switch / hydration: activation resets uncertain local lifecycle state, then backend replay restores genuinely pending prompts; stale snapshots cannot erase them.
- Model / effort / token switch: controller rebuild advances the prompt epoch before reused IDs are accepted.
- Mode switch: same-epoch drained approvals disappear and stay tombstoned; retained approvals remain answerable; old-epoch drain results are ignored.
- Failed answer submission: same-epoch failures recover the backend-held prompt, while old-epoch failures cannot disturb the replacement controller.

## Compatibility and cache audit

| Surface | Older/newer behavior | Conclusion |
| --- | --- | --- |
| Persisted sessions/config | No persisted fields or formats changed | Compatible |
| Wails mode setters | New backend returns drained ID arrays; older frontends ignore the return value, while the new frontend normalizes `null`/`void` to `[]` | Compatible |
| Frontend prompt epoch | In-memory, per-tab state only | Compatible |
| Provider-visible prompt/tools | No prompt, tool schema/order, provider serialization, memory, or compaction changes | Cache-neutral |

## Security audit

No new dependency, network call, credential handling, file access, command execution, persistence hook, sandbox expansion, or permission bypass is introduced. Auto/YOLO continues to retain approvals that require a fresh human decision.

## Verification

- `npm test` (full frontend suite) - pass
- `npm run typecheck` - pass
- `npm run test:typecheck` - pass
- `pending-prompt-stale-status.test.tsx` - 78 passed, 0 failed
- `go test ./internal/control` - pass
- `cd desktop && go test ./...` - pass
- `go test ./...` - pass
- Mutation proof for the final race: removing the epoch guard makes four reducer/hook assertions fail; restoring it makes all 78 focused assertions pass.


SivanCola:
Author note — hardening for the reverse race raised in review.

The initial commit (`95390ad1`) fixed the forward direction (#6429): a runtime
snapshot fetched before a live approval/ask event could be dispatched after it
and wrongly clear the prompt. Review then surfaced the reverse direction — a
prompt that was already answered getting resurrected — in two rounds. Both are
now addressed.

**Round 2 — `ad11beb3` (anchor freshness to the prompt id):**
The first guard refreshed the freshness anchor on every approval/ask event.
Because the backend re-emits the pending prompt on `ReplayPendingPrompts` and the
event crosses the Wails bridge asynchronously, a replay snapshotted before the
user answers can be delivered after — carrying the just-answered id. That
advanced the anchor, so an authoritative idle snapshot fetched right after the
answer looked stale and was discarded, and `turn_done`'s `keepPlanApproval`
resurrected the resolved `exit_plan_mode` card. Fix: anchor `promptArrivedAt` to
the prompt id — a same-id replay keeps the original arrival time; only a
genuinely new id (backend ids are monotonic within a controller) re-anchors.
Approve-failure recovery is preserved (a same-id replay still re-arms the prompt;
only the timestamp is left untouched).

**Round 3 — `6e4596cc` (authoritative reconcile backstop):**
Round 2 still had a hole across a navigation boundary: on a rapid A→B→A switch
(or a single-surface state wipe), `backend_activation_start` drops the anchor, so
a delayed same-id replay re-anchors to a late time and the zombie returns. A
frontend anchor alone cannot distinguish genuine recovery of a still-pending
prompt from a delayed replay of a resolved one, because navigation can erase the
anchor the distinction relies on. Fix: whenever `dispatchRuntimeStatusForTab`
rejects a stale idle snapshot, schedule one debounced `reconcileTabRuntime` that
refetches backend state. If the backend resolved the prompt, the fresh snapshot
(taken after any in-flight replay) is newer than the anchor and clears the
zombie; if the prompt is genuinely pending, the fresh snapshot keeps it and
re-replays. The reducer still rejects the stale snapshot in place (it cannot
authoritatively resolve the ambiguity locally); the hook-level reconcile is the
authority — the "reject then reconcile" option from review.

**Verification:**
- `pnpm test` (full frontend suite, incl. the reducer reverse-race cases in both
  orderings and the hook-level A→B→A convergence case) — passed
- `pnpm run typecheck` and `pnpm run test:typecheck` — passed
- Mutation checks: reverting the id anchor fails 5 assertions; disabling the
  reconcile reschedule fails the 2 convergence assertions (zombie survives, input
  stays locked)
- CI green on head `6e4596cc`; no persisted-format, Wails-contract, or
  cache-sensitive surface changed (frontend in-memory state only)

PR #6439, opened 2026-07-13
fix(desktop): resolve context_window=0 for custom providers, move setting to visible area

ashishexee:
## Problem

When adding a custom model provider via **Settings → Models → Add Provider** (desktop UI), the "Context window" field was buried inside the collapsed "Compatibility settings (usually leave unchanged)" section. Most users never saw it, leaving the value at 0.

The CLI/TUI path was unaffected — it already hardcodes `128000` for custom providers.

## What This PR Does

**Moved the context window field** from the collapsed advanced section to the visible area (after manual models input, before the model picker), matching the TUI's existing behavior.

**Pre-fills 128000** for new custom providers — same value the TUI uses — instead of leaving it empty.

## What This PR Does NOT Do

- Does **not** backfill `context_window = 0` to 128000 (respects #6677's semantic: `0 = disable compaction`)
- Does **not** clamp the save value (explicit 0 saves as 0, preserving the compaction-disable escape hatch)
- Does **not** change locale strings (already updated by #6677 to say "0 = disable auto-compaction")

## Changes

**`desktop/frontend/src/components/SettingsPanel.tsx`** (1 file):
- Removed context window input from `<details>` "Compatibility settings" advanced section
- Added it to the visible area before `ProviderEditorModelPicker`
- New providers pre-fill with `128000`; existing providers keep their saved value

## After / Before

**Before:** Context window hidden in collapsed advanced section, defaults to empty/0, context gauge shows `64987/0`, compaction disabled

**After:** Context window visible in main settings area, pre-fills 128K for new providers, context gauge shows `64987/128000`, compaction works. Explicit 0 still means "disable compaction".

## Testing

- Config tests — PASS
- Desktop preset tests — PASS (including `TestAddEveryProviderPresetAccessInstallsTemplate`)
- Provider model override tests — PASS

---
Closes #6439

### DIFF
--- a/desktop/frontend/src/components/SettingsPanel.tsx
+++ b/desktop/frontend/src/components/SettingsPanel.tsx
@@ -5627,8 +5627,9 @@ function ProviderEditor({
   const [keyDraft, setKeyDraft] = useState("");
   const [balanceUrl, setBalanceUrl] = useState(initial?.balanceUrl ?? "");
   // Empty when unset so the placeholder (and its "0 = disabled" hint) reads instead
-  // of a bare "0"; saved back as 0.
-  const [ctx, setCtx] = useState(initial?.contextWindow ? String(initial.contextWindow) : "");
+  // of a bare "0"; saved back as 0. New providers start with 128000 (matching the
+  // TUI's hardcoded default for custom providers).
+  const [ctx, setCtx] = useState(initial?.contextWindow ? String(initial.contextWindow) : (initial ? "" : "128000"));
   const [modelContextWindows, setModelContextWindows] = useState<Record<string, string>>(
     () => providerModelContextWindowDrafts(initial?.modelOverrides),
   );
@@ -5950,17 +5951,6 @@ function ProviderEditor({
           onChange={(e) => setBalanceUrl(e.target.value)}
         />
         <div className="mem-hint">{t("settings.balanceUrlHint")}</div>
-        <label className="set-label">{t("settings.providerContextWindow")}</label>
-        <input
-          className="mem-input"
-          inputMode="numeric"
-          min={0}
-          placeholder={t("settings.contextWindowPlaceholder")}
-          type="number"
-          value={ctx}
-          onChange={(e) => setCtx(e.target.value)}
-        />
-        <div className="mem-hint">{t("settings.contextWindowHint")}</div>
       </div>
     </details>
   );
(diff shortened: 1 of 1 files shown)


ashishexee:
## CI Update: Fixed `desktop` test failure

The `desktop` CI job failed on `TestAddEveryProviderPresetAccessInstallsTemplate` after the initial push.

### What was happening

The `backfillCustomProviderContextWindows()` function we added runs during config loading (`loadForRoot()`). It was designed to fix custom providers with `context_window = 0`, but it was **too broad** — it also mutated curated provider presets (qwen-cn, stepfun, novita, gmi, huggingface, nvidia, kilocode, ollama-cloud, etc.) that ship with `context_window = 0` in their template definitions.

The test installs every preset, then compares the installed provider against the original template. The backfill changed `context_window` from `0 → 128000` on these presets, causing a mismatch:

```
Template:  context_window = 0      (original, untouched)
Installed: context_window = 128000 (mutated by backfill)
```

### The fix

Added a `PresetID` guard to `backfillCustomProviderContextWindows()` so it only applies to **user-added custom providers** (which have no PresetID), not curated presets:

```go
if cfg.Providers[i].ContextWindow <= 0 && cfg.Providers[i].PresetID == "" {
    cfg.Providers[i].ContextWindow = defaultCustomContextWindow
}
```

Curated presets (identified by `PresetID`) are now left untouched — their templates remain authoritative.

### Rebase

Branch has been rebased on the latest `origin/main-v2` (`4828ef14`) and force-pushed.

SivanCola:
Thanks for tackling this — the underlying gap is real: a custom provider saved with `context_window = 0` currently gets no compaction and a dead context gauge, and moving the setting out of the advanced section is the right direction.

One required change before this can merge: please do not resolve the gap by backfilling `context_window = 0` to 128000. In the current code, 0 is a meaningful value — `agent.Options.ContextWindow <= 0` disables compaction (`internal/agent/agent.go`, `internal/agent/compact.go`), `docs/SPEC.md` documents "0 disables", and #6677 now documents and relies on the same semantics. `backfillCustomProviderContextWindows` (plus the `Math.max(1, ...)` save clamp) silently rewrites intentional 0s and removes that escape hatch. We've decided to keep "0 = disable compaction".

What we would like instead:

- Drop `backfillCustomProviderContextWindows` and the `load.go` changes.
- Keep moving the input to the visible area, and keep pre-filling 128000 for *new* providers — but let an explicit 0 save as 0 (no clamp), so disabling compaction stays possible.
- Align the locale wording with the "0 = disable auto-compaction" semantics. Note that #6677 edits the same three locale keys and the same `ProviderEditor` region, so whichever lands second will need a rebase.

Once this is addressed and CI stays green, I will consider merging it into main-v2.

ashishexee:
Thanks @SivanCola for the clear guidance on the rework.

**What changed in this rework:**

1. **Dropped**  and the  changes — no longer overwriting `context_window = 0` to 128000. The `0 = disable compaction` semantic from #6677 is fully preserved.

2. **Kept** the context window field in the visible area (moved from collapsed advanced section to before the model picker), matching the TUI's existing behavior.

3. **Pre-fills 128000** for new providers (matching the TUI's hardcoded default for custom providers). Existing providers keep their saved value.

4. **Explicit 0 saves as 0** — no `Math.max(1, ...)` clamp. Disabling compaction stays possible.

5. **Locale strings** already say "0 = disable auto-compaction" (updated by #6677).

The result: new providers get a sensible 128K default, the field is visible without expanding advanced settings, and `context_window = 0` still means "disable compaction" as documented by #6677.

PR #6494, opened 2026-07-15
Show pinned conversations in classic sidebar / 经典侧边栏独立显示置顶会话

SivanCola:
## Summary

- add a dedicated pinned section to the classic desktop sidebar
- remove pinned conversations from their project rows while they are pinned, avoiding duplicate entries
- return conversations to their original project order after unpinning
- preserve the existing classic row density, activity time, hover actions, active state, and project ordering

## Root cause

The classic layout only sorted pinned conversations before unpinned siblings inside each project. The shared pinned-section splitter and section markup were gated to the workbench layout, so pinning changed the button state without creating a visible top-level group.

## User impact

Classic-layout users now get the same clear pinned hierarchy as workbench users while retaining classic styling. Archive, History, Trash, restore, and persisted pin metadata behavior are unchanged.

## Compatibility

| Surface | Existing data behavior | Conclusion |
| --- | --- | --- |
| Topic pin metadata | Reads the existing `pinned` field from project-tree nodes | Compatible |
| Project and topic persistence | No format or write-path changes | Compatible |
| History and Trash | Existing APIs and session records are untouched | Compatible |
| Workbench and Creation layouts | Workbench keeps pinned projects and topics; Creation remains unsplit | Compatible |

## Verification

- `tsx src/__tests__/project-tree-runtime.test.ts` (44 passed)
- `tsc --noEmit`
- `tsc --noEmit -p tsconfig.test.json`
- CSS syntax and z-index token checks
- browser DOM/click verification for pin, deduplication, unpin restore, project order, classic row geometry, and zero console warnings/errors
- `git diff --check`

### DIFF
--- a/desktop/frontend/src/__tests__/project-tree-runtime.test.ts
+++ b/desktop/frontend/src/__tests__/project-tree-runtime.test.ts
@@ -11,6 +11,7 @@ import {
   projectTreeShouldRenderTopicActions,
   projectTreeTopicMetaLine,
   arrangeClassicProjectTree,
+  splitPinnedProjectTree,
   classicTopicWindow,
   projectTreeTopicHoverCardModel,
   projectTreeTopicMenuOffersPin,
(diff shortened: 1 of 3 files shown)


SivanCola (commented):
Reviewed at `82fa7c98f` against `main-v2` `417335f02`. No blocking findings.

Functional correctness:
- The change reuses the existing pinned-tree split at the presentation boundary, so persisted topic/project data and backend pin APIs remain unchanged.
- Classic now extracts pinned topics into one top-level section, removes them from project children to avoid duplicates, and returns unpinned topics to normal project ordering.
- Workbench still extracts pinned projects and topics; Creation remains unsplit.
- The existing topic row renderer remains responsible for active state, activity time, shortcuts, archive, context menus, and open behavior.

Regression and compatibility:
- Focused coverage verifies cross-project pinned ordering, deduplication, unpin restoration, classic project ordering, and preserved workbench project pinning.
- History, Trash, restore, and persistence contracts are untouched.
- No new dependencies, network access, file access, permissions, secret handling, or provider-visible/cache-sensitive data paths are introduced.

Related PRs:
- #5899 is complementary rather than duplicative: it changes the existing workbench pinned header and unread action visibility, but does not add a classic pinned section. It is based on an older merge-base and currently conflicts in `ProjectTree.tsx`, so it should be rebased after this PR while preserving this classic split.
- #6119 shares `ProjectTree.tsx` for worktree badges but its ProjectTree changes merge cleanly with this PR; its broader stale branch has unrelated conflicts elsewhere.

Verified:
- focused project-tree runtime tests: 44 passed
- frontend app and test TypeScript checks passed
- CSS syntax and z-index token checks passed
- browser pin/unpin, deduplication, project-order, geometry, and console checks passed
- all GitHub CI, race, CodeQL, cache-impact, and cross-platform checks passed
- `git diff --check` passed

PR #6497, opened 2026-07-15
fix(agent): enforce serial todo progress / 强制串行 Todo 推进

SivanCola:
## Summary

- enforce a serial Todo state machine: completed items form a prefix, exactly one unfinished item is current, and `complete_step` cannot sign off pending work
- make the serial contract level-aware while preserving the provider-visible single-in_progress contract: in a two-level list the active level-1 sub-step is the only `in_progress` item and its level-0 phase stays `pending`; sub-steps complete in order, the phase becomes `in_progress` to sign off only after every sub-step has completed, `complete_step` refuses a phase whose sub-steps are unfinished, and a level-1 item with no phase above it is rejected
- keep that contract across user turns and desktop history restoration while normalizing legacy out-of-order Todo snapshots in memory; live host advance, session rebuild, and desktop history replay all share one `evidence.AdvanceSerialTodo` promotion path
- block batched sign-offs, completed-prefix duplication/reordering, numeric-title identity confusion, and title churn that previously looked like progress
- manage execution automatically with an adaptive host-observed progress lease: nudge after 8 no-progress tool rounds, pause with saved work after 16, and renew on a new completion, unique read, command, or mutation
- retire hidden `[agent].max_steps` and `planner_max_steps` behavior: old TOML remains parseable, but values are ignored and removed by a one-time boot migration
- retain explicit one-off CLI `--max-steps`, unattended `[bot].max_steps`, and the provider-visible `task.max_steps` tool parameter
- preserve Plan mode plan replacement and the provider-visible prompt/tool contract

## Root cause

The workflow validated evidence but did not own one serial state-machine contract across all producers and consumers. `complete_step`, canonical replay, and desktop history replay could independently advance any non-completed match. The previous fixed stall counter also treated legitimate long-running work like drift. Keeping old hidden step-limit values active would additionally let stale small limits preempt the new adaptive progress policy and truncate inherited subagent work.

## User impact

Failed sign-off now leaves the current Todo active and the model must repair and retry that item before moving on. Existing malformed sessions restore as a valid serial list instead of displaying a later item as completed. Normal users do not need to choose executor/planner step limits: real work automatically extends the run, exact repeats do not, and a genuinely stalled run first gets a recovery prompt and then pauses safely with resumable work. Existing hidden step-limit keys no longer silently weaken this behavior.

## Compatibility

| --- | --- | --- |
| Session transcripts | Existing tool calls and JSON fields deserialize unchanged; malformed Todo ordering is normalized only in memory during restoration | Backward compatible; no session rewrite or format change |
| Config TOML | Retired `[agent].max_steps` and `planner_max_steps` fields still decode; runtime normalizes them to automatic mode and atomically removes standard assignments once, with a notice | Backward readable; intentional one-time behavior migration prevents stale limits |
| Cross-version config | The migrated file only omits retired keys, so older releases still parse it and fall back to their existing zero defaults | Safe for old/new binaries sharing the config |
| Desktop/Wails JSON | Existing `maxSteps` and `plannerMaxSteps` fields and method parameters remain; old frontend calls are accepted and normalized to zero | Backward compatible; no payload-shape or binding break |
| Explicit budgets | CLI `--max-steps`, `[bot].max_steps`, and `task.max_steps` remain supported and keep distinct runtime labels | Operational cost controls remain available without normal settings complexity |
| Provider tool contract | Tool names, order, schemas, descriptions, and request serialization are unchanged | Stable provider-visible prefix and prompt-cache contract |
| Two-level task lists | Legacy flat or out-of-order phase snapshots replay through the level-aware normalizer; a historical phase sign-off completes on replay only when its sub-steps had completed; legacy orphan level-1 items advance as standalone steps | The `todo_write` description and schema ("exactly one in_progress") stay true byte-for-byte; no provider-visible contract change |
| Plan mode | Prior executor Todo state is not injected while planning, so a new plan can replace old execution state | Existing planning workflow preserved |

## Validation

- `DEEPSEEK_API_KEY= go test ./...`
- focused config/boot regression tests for parse-only loading, atomic migration, one-time notice, second-boot silence, and Safe Mode leaving config byte-for-byte unchanged
- focused level-aware chain tests: validator accept/reject matrices (including phase-in_progress-before-sub-steps and orphan level-1 rejection), legacy chain normalization, `AdvanceSerialTodo` walking a full two-phase chain and advancing orphan sub-steps as standalone steps, `complete_step` phase gating, active sub-step continuity, Plan approval seed validity, and desktop history replay gating
- `./scripts/cache-guard.sh` (long tool-loop tail cache hit rate: 95%)
- `wails generate module` produced no binding diff
- `cd desktop/frontend && pnpm build && pnpm test:all`
- `go test -race ./internal/evidence ./internal/tool/builtin ./internal/agent ./internal/control ./internal/boot`
- real DeepSeek `deepseek-v4-flash` tool-call smoke: one layered Todo advanced child A1 → child A2 → phase A → phase B and returned `TODO_PROTOCOL_OK`
- merged the latest `main-v2`; GitHub reports the PR mergeable
- `git diff --check`

Cache-impact: none — no provider-visible tool schema/order/description, stable prompt prefix, request serialization, or compaction behavior changes. The adaptive reassessment and explicit-budget labels are appended only at the dynamic conversation tail when their guards fire.
Cache-guard: `./scripts/cache-guard.sh` passed; the long tool-loop tail cache hit rate remained 95%.
System-prompt-review: `DefaultSystemPrompt` and stable prompt assembly are unchanged.


SivanCola:
Review note: the serial task-list validator was level-blind at 791ec4d.

`ValidateSerialTodos` treated two-level lists (level-0 phase + level-1 sub-steps) as one flat sequence, and the first level-aware pass (fb37b9b) fixed the completion order but required a phase and its active sub-step to both be `in_progress` — contradicting the provider-visible todo_write contract ("keep exactly one item in_progress at a time" / "Keep at most one in_progress"), and it left orphan level-1 items able to stall serial advancement.

**Update 2: reworked in c8723be96 to preserve the single-in_progress contract.**

- The active level-1 sub-step is now the *only* `in_progress` item; its level-0 phase stays `pending` while sub-steps complete in order. Once every sub-step is completed the phase becomes the single `in_progress` item and signs off last — so the tool description and schema stay true byte-for-byte, with no provider-visible change.
- `complete_step` still refuses a phase whose sub-steps are unfinished, and the pending-rejection hint names the active sub-step.
- A level-1 item with no phase above it is rejected by `todo_write`; legacy orphan sub-steps in existing histories advance as standalone steps, so completing one promotes the next pending unit instead of stalling the list.
- Live host advance, session rebuild, and desktop history replay keep sharing the one `evidence.AdvanceSerialTodo` promotion path.

Verified at c8723be96: root and desktop `go test ./...`, `go test -race` on agent/evidence/builtin, `./scripts/cache-guard.sh`, `gofmt -l`, and `git diff --check` all pass.

Issue #6499, opened 2026-07-15
[Bug]: 配置丢失


keleyoudu:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.17.12

### What happened?

让reasonix自己帮我装一个mcp，装完以后我配置的供应商配置没了。

<img width="823" height="896" alt="Image" src="https://github.com/user-attachments/assets/62cb6aa9-9bab-44bc-bda7-0d929148f2f7" />

<img width="1135" height="402" alt="Image" src="https://github.com/user-attachments/assets/48bf48aa-d346-466a-ad65-79bc3aee1858" />

### Steps to reproduce

1，mcp add codegraph
2，执行完就没了

### OS / platform

windows 11

### Relevant logs or output

```shell

```

SauronSkywalker:
看看他是怎么安mcp的?别是直接创建一个新的配置文件全量覆盖了

keleyoudu:
{
  "title": "Global / mcp add codegraph",
  "exportedAt": "2026-07-15T03:31:00.283Z",
  "items": [
    {
      "kind": "notice",
      "id": "n0",
      "level": "info",
      "text": "当前模型缺少 API key。",
      "detail": "model \"deepseek-flash/deepseek-v4-flash\" is selected but its API key DEEPSEEK_API_KEY is not set — requests will fail until you set it"
    },
    {
      "kind": "user",
      "id": "u1",
      "text": "mcp add codegraph",
      "createdAt": 1784082709989
    },
    {
      "kind": "assistant",
      "id": "a2",
      "text": "",
      "reasoning": "The user wants to add an MCP (Model Context Protocol) server called \"codegraph\". This seems to be a request to configure/install an MCP server. Let me understand the context - this looks like a configuration task for an MCP client (like Claude Desktop or similar).\n\nLet me look at the workspace to understand what we're working with.",
      "streaming": false,
      "reasoningComplete": true,
      "reasoningDurationMs": 1960,
      "workDurationMs": 3118
    },
    {
      "kind": "phase",
      "id": "p3",
      "text": "商汤 · planning"
    },
    {
      "kind": "tool",
      "id": "call_11fba5dbd5934912b581bd7b",
      "name": "ls",
      "args": "{\"path\": \".\"}",
      "readOnly": true,
      "status": "done",
      "summary": "1 项",
      "dataArchived": true
    },
    {
      "kind": "tool",
[...]

PR #6527, opened 2026-07-15
Fix/6499 config overwrite on mcp install

myipanta:
## Summary

-

## Verification

-

## Cache impact

Cache-impact: none — config load path only, no prompt/tool/system-prefix changes
Cache-guard: go test ./internal/installsource/ -count=1 covers MCP install apply path
System-prompt-review: N/A

### DIFF
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -618,6 +618,22 @@ func loadForEditStrict(path string, loadCredentials, persistMigrations bool) (*C
 	return cfg, nil
 }
 
+// LoadForEditSafe is like LoadForEdit but returns an error when the config file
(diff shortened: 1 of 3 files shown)


SivanCola:
Thanks for the fix. #6784 has now landed strict, no-write-on-parse-error handling for permission approval persistence. For the install_source / #6499 path covered here, #6505 is the broader implementation: it applies strict loading across MCP and skill install, remove, and uninstall paths, and includes wider regression coverage.

Closing this PR as superseded by #6505 so the remaining config-edit work can be reviewed in one place. Thank you @myipanta for identifying and implementing the fail-before-write approach.

PR #6528, opened 2026-07-15
Fix/6473 bash reap memory leak

myipanta:
## Summary

-

## Verification

-

## Cache impact

Cache-impact: none — bash tool output buffering only, no prompt or tool definition changes
Cache-guard: go test ./internal/tool/builtin/ -count=1 covers foreground bash execution
System-prompt-review: N/A

### DIFF
--- a/internal/tool/builtin/bash.go
+++ b/internal/tool/builtin/bash.go
@@ -28,6 +28,12 @@ import (
 
 const (
 	bashWaitDelay = 5 * time.Second
+	// maxBashOutput caps the combined stdout+stderr a single command can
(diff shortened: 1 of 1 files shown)


SivanCola:
Thank you @myipanta. #8053 supersedes this source PR with an adaptation at the shared shellrun owner: foreground combined output is capped at 10 MiB for every caller, retains the opening context plus a rolling final 64 KiB, and is regression-tested under concurrent stdout writes. Both integration commits carry your public Co-authored-by identity, so the adopted contribution receives repository contributor credit after merge.

PR #6529, opened 2026-07-15
Fix/6475 session recovery silent failure

myipanta:
## Summary

-

## Verification

-

## Cache impact

Cache-impact: TODO
Cache-guard: TODO
System-prompt-review: N/A

### DIFF
--- a/internal/agent/save.go
+++ b/internal/agent/save.go
@@ -1151,6 +1151,16 @@ func loadSessionUnlocked(path string) (*Session, error) {
 			// branch. Anchor the baseline on digest+version only until a
 			// successful save re-learns the revision.
 			s.markPersistedRevisionUnknown(path, digest, s.version, s.rewriteVersion)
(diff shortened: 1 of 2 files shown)


SivanCola:
Heads-up: #6667 changes `Controller.snapshot` to a three-parameter signature and restructures the conflict-recovery block this PR's `internal/control/controller.go` hunk patches, and it also touches `internal/agent/save.go`. The two changes are semantically complementary — the `conflictDropped` user notice and the stale-writer CAS disarm here remain wanted — but this branch will need a rebase once #6667 lands.

SivanCola:
Thank you @myipanta. #8053 records and credits the stale-writer and silent-conflict investigation from this PR. The old CAS disarm is not copied because current lease ownership, resume hygiene, and topic/session binding fixes address the reported stale-session path, while the old rule could disarm a valid writer after a normal process restart. The full disposition is preserved in the integration ledger.

PR #6560, opened 2026-07-16
fix(cli): mirror paste path for transcript copy on local terminals

andrei-kiparuk:
## Summary

- Transcript copy (`/copy`, drag-select auto-copy, Ctrl+C on selection) only called `tea.SetClipboard` (OSC 52).
- Terminals that do not bridge OSC 52 to the OS pasteboard (for example Cursor and VS Code integrated terminals on macOS) showed **"copied to clipboard"** while `pbpaste` stayed empty.
- Paste already reads through `github.com/atotto/clipboard`; this change batches the same `clipboard.WriteAll` fallback ahead of OSC 52 so local copy matches paste behavior while SSH/tmux setups that honor OSC 52 keep working.

## Root cause

`copyToClipboard` in `internal/cli/transcript.go` emitted OSC 52 only. `copySelectionWithNotice` always shows the success hint without verifying the pasteboard updated.

## Test plan

- [x] `go test ./internal/cli/ -count=1`
- [x] Manual: drag-select in Reasonix TUI inside Cursor terminal, then Cmd+V confirms text is on the system pasteboard
- [x] `/copy 1` and Ctrl+C selection copy use the same path

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/cli/copy_export_test.go
+++ b/internal/cli/copy_export_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 
 	tea "charm.land/bubbletea/v2"
+	"github.com/atotto/clipboard"
 
 	"reasonix/internal/agent"
 	"reasonix/internal/control"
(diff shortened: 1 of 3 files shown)


SivanCola:
Thank you @andrei-kiparuk for identifying the OSC 52-only clipboard gap and proposing the native clipboard fallback. The hardened integration landed in #6632: local copy verifies the OS clipboard write, SSH avoids writing the remote host clipboard, and failures fall back to OSC 52 without claiming native success.

Your contribution is now permanently credited by merged PR #6664 in CHANGELOG.md and both README acknowledgments. Its preserved commit also carries your public `Co-authored-by` trailer, so GitHub can record you in the repository contributor graph.

Verified on current `main-v2` with focused clipboard tests repeated 20 times, the full `internal/cli` suite, race coverage, a real Cursor integrated-terminal macOS pasteboard round trip, and deterministic local/SSH/OSC 52 routing checks.

#6560 is superseded by #6632, so I am closing this PR with credit preserved. The runtime fix is not in v1.17.15 and will ship in the next release.

PR #6573, opened 2026-07-16
fix(desktop): stop same-process lease rebuild thrash; keep stale-turn watchdog alive

### LINKED ISSUE #6568: [Bug]: deferredRebuildLoop retries forever when session lease is held by the same process's active controller
### Problem

When a tab's controller is actively running (AI is responding), and the frontend triggers a tab rebuild (e.g. manual page refresh, tab switch race), the rebuild attempts to acquire a session lease via `tab.ensureSessionLease(path)`. This fails because the **same process** already holds the lease for that session path.

The failure causes:
1. `setTabStartupError` sets `StartupErrLeaseHeld=true`
2. `scheduleDeferredStartupBuild(tab.ID)` schedules a retry in 2 seconds
3. The `deferredRebuildLoop` fires after 2 seconds — fails again — reschedules
4. **Infinite loop**: the startup error banner pops up every ~2 seconds indefinitely

### To Reproduce

1. Open a session and send a message (controller becomes active, holds session lease)
2. Trigger a tab rebuild (e.g. manually refresh the page while the AI is still responding)
3. Observe: every ~2 seconds the topbar shows "startup error: this session is already open in another Reasonix window"
4. The error never goes away on its own — even after the AI finishes responding

### Root Cause

`desktop/tabs.go:3022-3046` — when `ensureSessionLease` fails with `ErrSessionLeaseHeld`, the code treats it the same whether the lease is held by another process (should retry) or by the same process (should stop — the active controller will release it naturally).

The issue is in `internal/agent/session_lease.go:75`:
```go
if _, loaded := sessionLeaseOwners.LoadOrStore(path, ownerID); loaded {
    return nil, ErrSessionLeaseHeld  // does not distinguish self vs foreign
}
```

`TryAcquireSessionLease` does not distinguish between "held by another controller in this process" and "held by another process entirely". While `deferredRebuildLeaseLooksFree` (in `desktop/deferred_rebuild.go:245`) attempts this distinction, its probe uses `TryAcquireSessionLease` which already fails at the global map check before PID analysis.

### Code Locations

- `desktop/tabs.go:3022` — `ensureSessionLease` failure triggers retry scheduling
- `desktop/deferred_rebuild.go:190-220` — `retryDeferredStartupBuild` does not check lease owner identity
- `desktop/deferred_rebuild.go:245-274` — `deferredRebuildLeaseLooksFree` attempts to distinguish but `TryAcquireSessionLease` fails before PID check
- `internal/agent/session_lease.go:75` — `TryAcquireSessionLease` returns `ErrSessionLeaseHeld` without owner info

### Suggested Fix

In `retryDeferredStartupBuild`, after catching `ErrSessionLeaseHeld`, check whether the lease holder is this process. If yes, clear the deferred rebuild (the running controller will release the lease when the turn completes). Only keep retrying when another process holds the lease.

This requires either:
- Exposing the lease holder's PID from `TryAcquireSessionLease`, or
- Adding a separate `IsSessionLeaseHeldByOtherProcess(path)` function that probes the lock file without the in-memory map check

### LINKED ISSUE #6569: [Bug]: Desktop event stream has no heartbeat or polling fallback — UI freezes when Wails EventsOn drops
### Problem

The desktop frontend real-time event stream depends solely on Wails `runtime.EventsOn("agent:event", cb)`. When this event channel disconnects for any reason (long-running session, system sleep/resume, GPU process restart, etc.), the frontend has no mechanism to detect the disconnection:

1. No heartbeat/ping mechanism to confirm the event stream is alive
2. No polling fallback (e.g. `setInterval` + `app.MetaForTab`) as a safety net
3. The UI freezes on whatever state it had before disconnection — only a manual page refresh restores updates

### To Reproduce

1. Open a session and send a message so the AI starts responding
2. Let the event stream drop (long wait or system sleep/resume)
3. Observe: token/text no longer streams in real-time
4. Manual page refresh: updates resume

### Code Locations

- `desktop/frontend/src/lib/bridge.ts:487-489` — `onEvent` via `window.runtime.EventsOn(EVENT_CHANNEL, cb)`, no keepalive
- `desktop/frontend/src/lib/useController.ts:2088-2130` — event handler consumes events but never checks for stream health

### Suggested Fix

Frontend: add a periodic heartbeat check. If no event received within a threshold (e.g. 30s), switch to polling via `app.MetaForTab(tabId)`.

Backend (optional): emit a periodic ping event on the event channel so the frontend can detect timeout.

erphenheimer:
## Summary

Two desktop reliability fixes:

### #6568 — deferred startup rebuild thrash
When a session lease was held by **this same process** (active controller / detached runtime), deferred startup rebuild treated the lease as free, ran `rebuildStartupTabLocked`, cleared `StartupErr`, failed `ensureSessionLease` again, and rescheduled. The topbar “session already open…” banner flashed every ~2s forever.

- Prefer `attachExistingSessionRuntime` before probing/rebuilding.
- Same-process hold with `Ctrl == nil` waits quietly after attach fails.
- Settings refresh path (`Ctrl != nil`) still proceeds so the tab’s own lease can no-op.

### #6569 — stale-turn watchdog dies after one fire
The frontend stale-turn reconciler used a one-shot `setTimeout`. Activity timestamps advance during streaming, so the timer often fired while healthy and **never rescheduled** after a later quiet drop (sleep/resume, Wails EventsOn drop, etc.).

- Switch the watchdog to `setInterval` while `running && turnActive`.
- Scope: keeps status reconcile alive after quiet. Does **not** re-subscribe EventsOn or re-stream mid-turn tokens; that remains a follow-up if needed.

## Changes

- `desktop/deferred_rebuild.go`: attach-first + same-process wait for startup retries
- `desktop/tabs_order_test.go`: `TestDeferredStartupRetryDoesNotRebuildWhileSameProcessHoldsLease`
- `desktop/frontend/src/lib/useController.ts`: interval stale-turn watchdog

## Cache impact

- Cache-impact: none — desktop lease retry / frontend reconcile only; no system prompt, memory prefix, provider request shape, or skill index changes.
- Cache-guard: N/A
- System-prompt-review: N/A

## Test plan

- [x] `go test ./desktop -run "TestDeferredStartupRetry|TestTabAndCtrlByIDRecoversStartupLease|TestBuildTabControllerBlocksWhenSessionLeaseHeld" -count=1`
- [x] `pnpm exec tsx src/__tests__/use-controller-meta.test.ts` (stale-turn helpers)
- [ ] CI full matrix on PR
- [ ] Manual: hold session lease in-process → deferred startup no longer flashes banner every 2s
- [ ] Manual: long turn / sleep-resume → quiet stream still reconciles running state without full page refresh

Fixes #6568
Fixes #6569

### DIFF
--- a/desktop/deferred_rebuild.go
+++ b/desktop/deferred_rebuild.go
@@ -239,6 +239,13 @@ func (a *App) retryDeferredStartupBuild(tabID string, tab *WorkspaceTab) {
 		a.clearDeferredRebuild(tabID)
 		return
 	}
+	// Attach first when another same-process tab/detached runtime owns this
+	// session; looksFree would otherwise wait forever under the same-process
+	// hold (#6568).
+	if path := a.deferredRebuildSessionPath(tab); path != "" && a.attachExistingSessionRuntime(tab, path, a.ctx) {
+		a.clearDeferredRebuild(tabID)
+		return
+	}
 	if !a.deferredRebuildLeaseLooksFree(tab) {
 		return
 	}
@@ -331,6 +338,10 @@ func (a *App) tryRecoverStartupLeaseHeldTab(tab *WorkspaceTab) bool {
 	if !a.tabHasRetryableStartupLeaseError(tab) {
 		return a.controllerForTab(tab) != nil
 	}
+	if path := a.deferredRebuildSessionPath(tab); path != "" && a.attachExistingSessionRuntime(tab, path, a.ctx) {
+		a.clearDeferredRebuild(tab.ID)
+		return a.controllerForTab(tab) != nil
+	}
 	if !a.deferredRebuildLeaseLooksFree(tab) {
 		return false
 	}
@@ -347,13 +358,10 @@ func (a *App) tryRecoverStartupLeaseHeldTab(tab *WorkspaceTab) bool {
 	return false
 }
 
-// deferredRebuildLeaseLooksFree cheaply probes whether the tab's session lease
-// could be acquired right now, without touching tab.sessionLease (only the
-// serialized rebuild paths may mutate that). The probe path can lag the
-// reconciled path the rebuild will use; a stale answer either re-defers on the
-// next tick or lets the rebuild fail back into the pending set, so a mismatch
-// only delays the retry.
-func (a *App) deferredRebuildLeaseLooksFree(tab *WorkspaceTab) bool {
+func (a *App) deferredRebuildSessionPath(tab *WorkspaceTab) string {
+	if tab == nil {
+		return ""
+	}
 	a.mu.RLock()
 	ctrl := tab.Ctrl
 	path := strings.TrimSpace(tab.SessionPath)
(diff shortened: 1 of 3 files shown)


erphenheimer:
关闭此 PR。上游 `main-v2` 已通过 `e4272b88`（7月5日）和 `2edf7fda`（7月7日）更彻底地修复了 #6568 和 #6569 的问题，我们的 PR 晚了近两周且不再需要。

Closing this PR. The upstream `main-v2` has already addressed #6568 and #6569 more comprehensively via `e4272b88` (Jul 5) and `2edf7fda` (Jul 7). Our PR was submitted nearly two weeks later and is no longer needed.

PR #4516, opened 2026-06-15
Fix desktop provider resolution across project configs

yuanyuanlove:
## Summary
- merge user and project `[[providers]]` by provider name so project configs do not drop global custom providers
- canonicalize desktop `provider_access` handling across add/remove/settings display and fallback resolution
- fail cleanly when no accessible model fallback exists instead of silently keeping a stale tab model

Fixes #3476.

## Testing
- `go test ./internal/config -run TestLoadForRootMergesProvidersAcrossUserAndProjectConfigs`
- `go test -run 'TestBuildTabControllerErrorsWhenNoAccessibleFallbackExists|TestRebuildKeepsExistingControllerWhenNoAccessibleFallbackExists|TestSettingsTreatsLegacyExplicitProviderAccessAsAdded|TestBuildTabControllerFallsBackWhenSavedTabProviderAccessWasRemoved|TestSetTokenModeFallsBackWhenTabProviderAccessWasRemoved' ./...` (from `desktop/`)

PR #4492, opened 2026-06-15
Fix desktop model switcher provider scope merge / 修复桌面模型切换器配置作用域合并

SivanCola:
## Summary
- Merge `[[providers]]` across user and project TOML sources by provider name, matching the existing plugin merge behavior.
- Merge desktop `provider_access` across config scopes so project desktop settings do not hide account-level providers from the chat model switcher.
- Add a regression test for a workspace project config that previously hid user-configured MiniMax models.

Fixes #4486.

## Verification
- `go test ./internal/config`
- `go test ./internal/config ./internal/control ./internal/boot` with isolated user config
- `go test .` in the desktop module

PR #6601, opened 2026-07-17
fix: project-only provider_access must not hide user-level providers

ttmouse:
## Summary

When the user config (`~/.reasonix/config.toml`) does not declare `desktop.provider_access` (implicit "allow all"), a project-level `reasonix.toml` that defines `provider_access` currently restricts the desktop model switcher to only those listed providers — hiding the user-level providers that exist outside the project-scoped access list.

**Root cause:** `mergeTOMLProviderAccess` merges `provider_access` as a union across config sources. When only the project config declares it (user config does not), the union equals the project's list alone. The user's default "allow all" is lost.

**Fix:** Track whether the user-level config explicitly declared `provider_access`. If only the project declares it (user did not), skip the restriction so all configured providers remain visible in the model switcher.

## Verification

- `TestMergeTOMLProviderAccessProjectOnlyDoesNotRestrictWhenUserUndefined` — NEW: project-only restriction is correctly skipped
- `TestMergeTOMLProviderAccessUserAndProjectMergeAsUnion` — NEW: both user and project declare → union merge unchanged
- `TestMergeTOMLProviderAccessNoDeclarations` — NEW: neither declares → no restriction
- `TestMergeTOMLProviderAccessUserDeclaresEmptyList` — NEW: user declares `[]`, project has list → project list takes effect (user explicitly opted into per-project control)
- `TestModelsForTabKeepsUserProvidersWithProjectConfig` — EXISTING PASS: user + project both declare → union unchanged
- `go test ./internal/config/...` — all pass
- `go build ./...` — all pass

## Cache impact

Cache-impact: none — only a small logic change in config loading, no system prompt or provider changes
Cache-guard: existing config tests and desktop ModelsForTab tests cover the behavior
System-prompt-review: N/A

## Related PRs

- PR #4492 (merged) — Added `mergeTOMLProviderAccess` with union logic, but left the gap when only the project declares the list.
- PR #4516 (open) — Related provider access resolution changes in `desktop/app.go`, does not overlap with this fix.

---
🤖 This contribution was prepared with Reasonix (DeepSeek fork of Claude Code), no additional plugins installed.

### DIFF
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -519,6 +519,7 @@ func mergeTOMLProviderAccess(paths []string) ([]string, bool, error) {
 	var merged []string
 	seen := map[string]bool{}
 	saw := false
+	userDeclared := false
 	for _, path := range paths {
 		if _, err := os.Stat(path); err != nil {
 			continue
@@ -532,6 +533,9 @@ func mergeTOMLProviderAccess(paths []string) ([]string, bool, error) {
 			continue
 		}
 		saw = true
+		if isUserConfigPath(path) {
+			userDeclared = true
+		}
 		for _, name := range f.Desktop.ProviderAccess {
 			name = strings.TrimSpace(name)
 			if name == "" || seen[name] {
@@ -540,6 +544,12 @@ func mergeTOMLProviderAccess(paths []string) ([]string, bool, error) {
 			seen[name] = true
 			merged = append(merged, name)
 		}
+	}
+	// If only the project config declared provider_access (not the user), the
+	// user's implicit "allow all" takes precedence: project-level restrictions
+	// must not hide account-level providers from the desktop model switcher.
+	if saw && !userDeclared {
+		return nil, false, nil
 	}
 	return merged, saw, nil
 }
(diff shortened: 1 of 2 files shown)


esengine:
Thanks — the diagnosis is right and the fix is the correct shape. An undeclared user-level `provider_access` means "allow all", and unioning it with a project-only declaration silently narrows the switcher to whatever the project happens to name. That's exactly the case #4492's union was meant to prevent and didn't.

I've landed it as #7395 rather than merging this branch: `mergeTOMLProviderAccess` was reworked after you opened this (`statConfigPath`, `meta.IsDefined` on the declaration, explicit-empty preservation), so the diff no longer applies. Your approach and the `userDeclared` guard carried over as-is, and you're credited as co-author on the commit.

One semantic call I made on top of your version: `TestMergeTOMLProviderAccessPreservesExplicitEmpty` now writes to the user config path. Explicit-empty ("I removed every provider") stays meaningful where it can be expressed — the user config — while a project's empty list no longer blanks the switcher, which follows from the same rule.

Closing this in favour of #7395. Thanks for the careful root-cause write-up — it made the port trivial.

Issue #6607, opened 2026-07-17
[Bug]: 长 Goal 任务中途卡死、切换会话卡顿 + 内容丢失


Patrickpoix:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

 Long Goal sessions freeze mid-execution + session switching loses content

### 标题
[desktop] Long Goal sessions freeze mid-execution + session switching loses content

### 描述
Reasonix Desktop v1.17.13 / Windows 11 23H2 / AMD Ryzen 7 7840H / 48GB RAM

#### Bug 1：长 Goal 任务中途卡死
开启 Long Goal（长目标/交付模式）后，任务运行一段时间（大约 10–30 分钟）后停止响应：
- 没有错误提示
- 没有 progress 更新
- 会话无法自动恢复
- 必须强制关闭桌面端进程才能恢复
- 浏览器 WebSocket/webview 连接断开后无法重连

初步观察似乎与以下情况相关：
- 大量子代理并行时（subagent skill review/explore）
- 上下文累积到一定 token 量后（约 50K+）
- 代理通过 proxy（llm-retry-proxy）转发时频繁遇到 502/524 重试

#### Bug 2：切换会话卡顿 + 内容丢失
- 在多个会话之间切换时，桌面端有明显卡顿（2–5 秒无响应）
- 切换后部分历史消息不显示或显示不完整
- 偶尔会出现「新建会话」复用了旧会话的上下文
- 关闭会话后重新打开，部分消息丢失

#### 日志 / 截图
（可补充桌面端日志或截图）

#### 环境信息
- Reasonix 版本: desktop-v1.17.13
- OS: Windows 11 23H2
- 硬件: AMD Ryzen 7 7840H, 48GB RAM
- 供应商: AIHub（通过 llm-retry-proxy 代理转发, 19 key 轮换）
- 模式: 目标模式 / 交付模式
- 会话模型: aihub/gpt-5.6-sol + aihub/gpt-5.6-terra/luna

### Steps to reproduce

1. 新建一个会话，选择 aihub/gpt-5.6-sol（planner）+ terra/luna（executor）
2. 输入一个需要多步骤的 long goal 任务
3. 等待 10–30 分钟，观察是否卡死
4. 切换 2–3 个不同会话，观察卡顿和内容完整性

### OS / platform

Windows11

### Relevant logs or output

```shell

```

SivanCola:
[...]

Issue #6346, opened 2026-07-12
[Bug]: 有的时候无端的卡死，没有任何反应，只能强行结束


baosfeng:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

 1.17.10

### What happened?

有的时候运行一个任务，几十分钟几百分钟没有任何反应，只能强行结束，结束之后输出

<img width="761" height="221" alt="Image" src="https://github.com/user-attachments/assets/2908e1a4-cff5-45f0-ba32-3004844dc7eb" />

background task may be stalled: task-4 — still running after 15m0s with no visible output; inspect with wait/bash_output or stop with kill_shell

再次开始的时候，又好像失智了，完全忘记之前的所有行为了，完全不理解上下文了

### Steps to reproduce

<img width="761" height="221" alt="Image" src="https://github.com/user-attachments/assets/d5ab165d-cfac-49ec-a4c1-c84f2d35f300" />

### OS / platform

MacOS 26.5.2

### Relevant logs or output

```shell

```

SivanCola:
A likely mechanism for this freeze was found while investigating #6607 and is fixed in PR #6613 (pending merge): the provider retry loop read a non-OK response body with **no read deadline**. When an upstream gateway/proxy returns 502/524 and then stalls the connection half-open, the turn blocks indefinitely with no error and no progress — the "background task may be stalled" warning in your screenshot is the watchdog noticing the wedged run, and force-kill was indeed the only way out. The read is now bounded by a hard deadline, so the attempt fails fast and retries visibly.

If your setup involves a proxy or a gateway-fronted provider, this is very likely the same defect. Please re-test once the fix ships in a release; if it still reproduces afterwards, fresh desktop logs would help.

**中文**:调查 #6607 时定位到本 issue 的高概率同因:重试循环读取错误响应体无超时,网关 502/524 半开连接下整轮无限阻塞(截图里的 stalled 告警正是看门狗发现了卡死的运行)。已在 PR #6613 修复,待合并发版后请再验证;若仍复现请补桌面端日志。

esengine:
Update — the fix referenced above has shipped, so this is ready to re-test.

#6613 (merge `dae65e25`) is in **`desktop-v1.17.15`** and every release since, including `desktop-v1.18.0`. For your case that is the retry loop reading a non-OK response body with no deadline: a gateway 502/524 that then stalls the connection half-open no longer blocks the turn indefinitely.

[...]

Issue #5760, opened 2026-07-02
[Bug]: 模型回复卡死 win desktop


SauronSkywalker:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.14.1

### What happened?

<img width="2560" height="1504" alt="Image" src="https://github.com/user-attachments/assets/46e27dda-bde5-4c4b-8456-943514a02594" />
目前就是这种情况，卡在这里我也不知道他是不是在处理什么东西或者单纯的只是卡住了，没有token输出。按停止也没反应。感觉可能需要在右侧工作区加一个新标签，显示客户端启动的cmd或者powershell等终端，看一下命令行在不在干活，或者卡在哪里了。
As things stand now, it's just stuck like this—I can't tell whether it's churning away on something or simply hung up with no tokens coming out. Hitting stop does absolutely nothing. It makes me think there ought to be a new tab over in the right-hand panel showing whatever terminal the client has spun up—be it cmd, PowerShell, or what have you—so we can peek at the command line and see if it's actually doing any work, or exactly where it's gotten bogged down.

### Steps to reproduce

随机出现卡死情况，就是正常对话中出现的，不清楚具体触发条件
The app occasionally hangs mid-conversation. No clear repro steps yet

### OS / platform

版本	Windows 11 家庭版 中文版 版本号	24H2 安装日期	‎2025/‎10/‎28 操作系统版本	26100.6899 序列号	425102C325A5090525 体验	Windows 功能体验包 1000.26100.253.0

### Relevant logs or output

```shell

```

jkqjkq:
我也出现了这个问题，不知道是程序的问题，还是模型就是慢

SivanCola:
已合入 `main-v2`：#5920（merge commit `a2eebf8f`）。

这次 PR 修复了一大类可能表现为“对话卡住、Stop 不响应、切换/关闭会话后 runtime 卡死”的锁顺序问题：桌面 controller snapshot/close、模型/设置 rebuild、session 删除/解绑、serve 切换、stdio MCP wait、bot gateway shutdown 等路径都做了锁范围收窄、超时等待和重入保护。

验证：
- #5920 的 CI 通过 `desktop`、多平台 `test`、`race`、CodeQL、cache-impact。
- 本地复核跑过 desktop focused/race、`internal/acp`、`internal/config`、`internal/bot`、`internal/botruntime` 等相关测试。

[...]

Issue #5337, opened 2026-06-26
[Bug]: 在多个会话来回切换会出现延迟加载（5s+）


SaltFish001:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.12.0

### What happened?

我平时需要在4-5个会话之间切换，会出现切换后显示是一个全新的空会话，需要等好一会才会加载出来。
会影响体验，以及如果在加载过程中输入了东西，会导致权限 思考程度被还原为默认值。还会导致执行状态丢失，会话中是在运行，显示是停止状态

https://github.com/user-attachments/assets/07a935c3-1fc8-4442-b9bd-b1b69e448170

### Steps to reproduce

1. 创建多个会话
2. 在多个会话中执行任务
3. 来回切换

### OS / platform

macOS 26.5.1

### Relevant logs or output

```shell

```

SivanCola:
Two of the mechanisms behind this were found while investigating #6607 and are fixed in PR #6613 (pending merge):

1. **5s+ delayed loading on switch**: every switch paid a full transcript serialize + digest of the outgoing session just to discover there was nothing to write — the cost grows with session size, which matches "4-5 sessions, gets slower". Unchanged sessions now skip that work entirely.
2. **switch briefly showing an empty/new session**: a stale activation completing after a newer click could delete the visible tab's cached state; stale completions are now dropped (last click wins).

The "inputs during loading reset permissions/effort" part is a separate surface and stays open. Please re-test the lag and empty-session symptoms once the fix ships.

**中文**:本 issue 的两个机制已随 #6607 调查定位并在 PR #6613 修复:① 切换时对外流会话做全量序列化+digest(会话越大越慢,与 4-5 个会话来回切的场景吻合),无变更时现已完全跳过;② 过期激活结果晚到会清掉可见 tab 的缓存状态导致"显示全新空会话",现丢弃过期结果。"加载中输入导致权限/思考程度重置"属另一表面,保持跟踪。发版后请再验证前两项。

esengine:
Update — the fix referenced above has shipped, so this is ready to re-test.

#6613 (merge `dae65e25`) is in **`desktop-v1.17.15`** and every release since, including `desktop-v1.18.0`. For your case that is the switch-time work: an unchanged session no longer serializes and digests its whole transcript on every tab switch, which is what scaled the delay with transcript size.

[...]

PR #6613, opened 2026-07-17
Fix long-run freeze on stalled error bodies and session-switch lag/content loss / 修复重试卡死、切换会话卡顿与内容丢失

SivanCola:
Addresses #6607 (also the mechanism behind #6346 / #5760 for the freeze half, and #5337 / #5315 / #5185 for the switch-lag / content-loss half).

Issue #6607 reports two symptom clusters on desktop v1.17.13: (1) long Goal runs freezing mid-execution with no error and no progress, correlated with 502/524 retry storms through a proxy, and (2) session switching lagging 2–5s, transcripts displaying incompletely, and messages lost after close/reopen. Code tracing found four independent defects; this PR fixes all four, each with a mutation-verified regression test.

## 1. Freeze: unbounded error-body read in the provider retry loop — `fix(provider)`

`SendWithRetry`'s non-OK path drained the response body with no read deadline (`internal/provider/retry.go`). A half-open 524 — headers sent, body stalled — is exactly what proxies produce under load, and:

- `http.Client` deliberately has no `Timeout` (streams must live long);
- `ResponseHeaderTimeout` (120s) stops applying once headers arrive;
- the stream idle watchdog only wraps the success-path SSE body;
- `RetryNotify` fires only before the next backoff, so the blocked read produced **no user-visible event at all**.

The turn blocked indefinitely with no error and no progress. Cancel would unblock it (body reads are context-aware), but during the silent window the webview transport can drop, so Cancel never arrives — matching "must force-kill the process".

**Fix:** `readErrorBody` arms a hard deadline (`errorBodyReadTimeout`, 10s) that closes the body to unblock the read; both the 4 KiB error snippet read and the drain are covered. `TestSendWithRetryUnblocksStalledErrorBody` drives a stalled 502 through the loop and fails via watchdog if it wedges.

## 2. Content loss: event-log tail repair permanently discarded bytes — `fix(agent)`

`repairSessionEventLogTail` `os.Truncate`d everything past the last cleanly-replayed record — the only place in the codebase that permanently discards persisted transcript bytes. Replay stops not only at a torn final record but also at a **buried** undecodable or out-of-order record (the shape two interleaved writers leave); every intact turn past that point was then silently destroyed on the next save. That is "reopened session lost messages".

**Fix:** the discarded range is appended to a `<id>.events.jsonl.damaged` salvage sidecar (one-line JSON incident header + raw bytes) before truncation, via the `store` path authority, and registered in `SessionSidecarFiles` so all five delete surfaces clean it up. The name deliberately does **not** end in `.jsonl`: older binaries scanning a shared session directory classify non-excluded `.jsonl` files as primary transcripts and would resurrect the damaged bytes as a phantom session. Salvage is best-effort and never blocks the repair (a full disk both tears tails and fails salvage).

## 3. Switch lag: every defensive snapshot paid a full serialize + digest — `perf(agent)`

Desktop switch/close/prune paths snapshot the outgoing session synchronously on the awaited bound call (`SetActiveTab`, `keepOnlyVisibleTab` — once per hidden tab, twice per pruned runtime). The existing `upToDate` skip only triggered **after** serializing and digesting the whole transcript and probing the disk; on 50K+-token sessions that is the reported seconds of unresponsiveness, multiplied on Windows by AV interference on file opens.

**Fix:** snapshot-mode saves short-circuit when in-memory bookkeeping proves a no-op (same path + version + rewriteVersion, ledger revision known, no pending load-time repair or log damage). Two invariants are locked in by existing tests:

- the check runs **under the save locks**, so a saver that waited on a concurrent writer re-evaluates against the state it must persist (`TestSaveSnapshotCapturesContentUnderSaveLock`);
- only **save-verified** baselines arm the fast path — a load-adopted baseline can pair the transcript with a stale ledger, and the first save after a load must run in full to heal it (`TestSameContentSaveHealsStaleLedgerDigest`). New coverage: `TestSnapshotUpToDateFastPath`.

## 4. Wrong surface after rapid clicks: stale topic activation applied unconditionally — `fix(desktop)`

`activateTopic` applied its result unconditionally after awaiting the backend. When a newer navigation completed first, the stale completion flipped the visible tab away from the user's last click **and** ran the single-surface prune that deletes every other tab's cached state — blanking the transcript the user was looking at. The App-level navigation queue serializes queued topic clicks, but its `latest()` guard only covers App-level effects; the `setActiveTabId` + state prune inside `activateTopic` were unguarded.

**Fix:** anchor each activation to the existing `activeNavigationSeqRef` epoch (the mechanism rewind/fork already use) and return without applying when superseded. New test: `activate-topic-stale.test.tsx` (registered in the frontend test chain).

## Not fixed here (needs reproduction)

The "new session occasionally reuses the old session's context" sub-symptom has three candidate paths (blank-tab controller reuse, legacy topic→session fallback, detached-runtime reattach by key). All are densely guarded (build generations, lease keys, superseded checks) and `blankTabMatchesTargetLocked` already rejects any non-system message; changing them blind risks the protected chat workflows. Tracked for follow-up once the reporter provides logs.

## Backward compatibility

| Change | Old-data / old-binary behavior | Conclusion |
| --- | --- | --- |
| New `.events.jsonl.damaged` sidecar | Old binaries ignore it (name does not end in `.jsonl`, so discovery skips it); old delete paths leave it behind (bounded, rare, forensic-only) | Safe |
| `SessionSidecarFiles` gains one entry | Additive; all delete surfaces consume the helper | Safe |
| `saveVerified` on `sessionPersistState` | In-memory only, never serialized | Safe |
| Snapshot fast path | Skips only writes that were already no-ops byte-for-byte; heal/repair paths (`ledgerStale`, damaged log, load-adopted baselines) still run in full | Safe |
| `readErrorBody` | Error-path only; success-path streaming and request serialization untouched | Safe |

## Verification

- `go test ./internal/provider/ ./internal/agent/ ./internal/store/ ./internal/control/` — all pass.
- `cd desktop && go test ./...` — all pass.
- Frontend: `tsc --noEmit` clean; `activate-topic-stale.test.tsx` (new), `tab-switch-hydration.test.tsx` (75), `open-topic-coalescing.test.tsx` (15), `pending-prompt-stale-status.test.tsx` (78) — all pass.
- Root `go test ./...` — one pre-existing machine-local failure (`TestMouseDragReleaseAutoCopies`, OSC 52 clipboard fallback on a headless host) which fails identically on clean `origin/main-v2`; unrelated.
- Every behavioral fix mutation-verified: disabling the fix makes its regression test fail (retry watchdog, salvage sidecar, fast-path scribble probe, stale-activation guard).

Cache-impact: none — `internal/provider/retry.go` changes only the non-OK error-path body read; request construction, serialization, streaming success path, and tool schemas are untouched, so provider-visible bytes are identical.
Cache-guard: `./scripts/cache-guard.sh` passes (`TestReleaseCacheHitGuard`); `go test ./internal/provider/` passes including the new stalled-body regression test.


SivanCola:
Updated this PR in cdd17602 (3 commits) addressing four cross-module review findings against the original head 2e1cb14a. All four were confirmed against live code; each fix is mutation-verified (disabling the fix makes its regression test fail).

**[P1] `doctor redact` skipped the new salvage sidecar** — `redactSessionCandidate` did not recognize `.events.jsonl.damaged`, so secrets preserved in salvaged tail bytes survived a scrub that reported success. Fixed in `internal/doctor/session_redact.go`: the suffix is now a candidate with lease mapping, and the scrub **deletes** the sidecar outright — the bytes are undecodable by definition, so no format-aware masking can prove them clean (same rationale as the existing torn-tail compaction precedent). Dry-run reports without deleting; leased sessions are skipped. Tests: `TestRedactSessionsRemovesDamagedSalvageSidecar`, `TestRedactSessionsSkipsLeasedDamagedSalvage`.

**[P1] Desktop trash left the sidecar in the live directory** — `sessionTrashArtifacts` is an explicit list and did not include the sidecar, so delete left raw session bytes behind and purge never removed them. Fixed in `desktop/sessions.go`: one entry covers delete, restore, and purge symmetrically (all three flows consume the same list). Test: `TestRestoreTrashedSessionFile` now asserts the sidecar leaves the live dir on delete, lands in trash, and comes back on restore.

**[P1] Last-click-wins guard missed the production navigation queue** — correct: `App.enqueueNavigation` serializes clicks, so a queued click never ran `activateTopic` and never advanced the controller epoch; the running stale activation passed the guard and pruned the newer surface's state. Fixed by exposing `noteNavigationIntent` from `useController` and calling it in `App.enqueueNavigation` at enqueue time, so any queued click immediately invalidates the running activation's destructive apply. The regression test now drives the **real** `enqueueNavigationRequest` queue (first activation held mid-flight, second queued) and a source-level assertion locks the App.tsx wiring. Reproduced your expected-fail shape before the fix: the stale activation flipped the visible tab.

**[P2] Repaired sessions never re-armed the fast path** — `normalizedDirty` / `eventLogDamaged` were never cleared after a successful save persisted the repair. Fixed in `internal/agent/save.go`: a save-verified baseline clears both flags (and the obsolete `rawMessages` capture); `checkSnapshotWrite` re-loads on-disk state and consults that object's flags, so nothing reads the live session's copies post-save. Test: `TestRepairedSessionArmsFastPath` covers both flags.

Verification: `go vet ./...` clean; full suites for `internal/agent`, `internal/doctor`, `internal/store`, `internal/provider`, `internal/control` pass; full `desktop` module passes; `tsc --noEmit` clean; frontend protected-workflow tests pass (activate-topic-stale 6, tab-switch-hydration 75, open-topic-coalescing 15, pending-prompt-stale-status 78, new-session-load-race 15).

SivanCola:
Updated in dda12222: one more lifecycle registration for the salvage sidecar, found in review follow-up.

`migrateSessionSidecars` (the overlong-name rename path) is yet another explicit sidecar list and did not include `SessionEventLogDamaged` — a renamed transcript would orphan the sidecar under the retired stem, holding raw session bytes that no delete, trash, redact, or bundle flow could reach (they all derive paths from the live transcript name). Fixed with one migration pair; `TestReconcileOverlongMigratesDiagnosticSidecars` now seeds the sidecar and asserts it migrates with the rename and leaves nothing behind (mutation-verified).

Closing sweep across every `os.Rename` site that can relocate session files: the sidecar is now registered in all six explicit lists — `SessionSidecarFiles` (delete), `sessionTrashArtifacts` (desktop trash/restore/purge), `redactSessionCandidate` (doctor redact), `sessionBundleArtifacts` (doctor bundle), `migrateSessionSidecars` (overlong rename), and discovery exclusion by name. Remaining rename sites are the v0.x legacy import (pre-sidecar data) and non-session subsystems (repair/updater/metrics). Full `internal/agent` suite passes.

PR #6618, opened 2026-07-17
fix(agent): honour DeepSeek finish_reason=stop on reasoning-only final answer

### LINKED ISSUE #6617: [Bug]: Agent keeps "thinking" after DeepSeek finishes (finish_reason=stop + reasoning-only treated as no answer)
### Version line
v2 - Go rewrite (1.x), main-v2 (active development)

### Exact version
main-v2 @ f590a66e (analyzed from source on the current default branch; reproducible on main-v2)

### What happened?
On DeepSeek thinking models (e.g. `deepseek-v4-pro`), the agent frequently keeps "thinking" / re-issuing requests **after the model has already finished** a simple task. The model streams a long `reasoning_content`, then returns `finish_reason="stop"` with an **empty `content`** block - i.e. it has explicitly signalled completion, and its answer lives in the reasoning stream that was already streamed to the user.

But the empty-final-answer retry (`internal/agent/agent.go`, `hasVisibleFinalAnswer`) only checks `strings.TrimSpace(text) != ""`, ignoring both `reasoning` and `usage.FinishReason`. So it misclassifies the finished turn as "no answer", injects a synthetic "provide a concise visible answer" user message, and forces DeepSeek to think again - the "still thinking after the task is done" / "shows finished but is still reasoning" symptom. After 3 such rounds it errors with `model finished without a visible final answer 3 times`; and because `emptyFinalBlocks` only resets when a tool is called, a model alternating reasoning-only and tool-call turns can loop indefinitely.

Expected: when the model returns `finish_reason="stop"` with non-empty reasoning, the host should accept the turn as complete (the reasoning was already streamed to the user), not override the stop signal and retry.

Note: `emptyFinalNoticeDetail` itself logs `finish=%s` in this path, showing the model returns `finish=stop` here - yet the retry still fires.

### Steps to reproduce
1. Configure a DeepSeek thinking model (e.g. `deepseek-v4-pro`) via the OpenAI-compatible provider (`reasoning_protocol = "deepseek"`).
2. Ask a simple question where the model emits `reasoning_content` and finishes with `finish_reason="stop"` + empty `content`.
3. Observe the agent re-issue a request carrying the synthetic "provide a concise visible answer" user message; DeepSeek thinks again; it may repeat up to 3x then error with `model finished without a visible final answer 3 times`, or persist if interleaved with tool calls.

Unit-level repro (added in the linked PR): `TestRunAcceptsReasoningOnlyFinalWhenModelStopped` replays exactly this - `ChunkReasoning` + `ChunkUsage{FinishReason:"stop"}` + `ChunkDone`, no content - and demonstrates both the bug and the fix.

### OS / platform
Linux (source-level analysis; the provider/model finish-reason behavior is OS-independent)

### Relevant logs or output
`emptyFinalNoticeDetail` emits a notice like:
```
empty final answer blocked: deepseek returned no visible answer text (finish=stop, reasoning=N chars); retrying
```
The `finish=stop` in the log confirms the model signalled completion, yet the retry still fires.

xianyu-sheng:
## Summary

DeepSeek thinking mode occasionally streams a long `reasoning_content` and then finishes with `finish_reason="stop"` but an empty `content` block - the model has explicitly signalled completion and its answer lives in the reasoning stream that was already streamed to the user. The empty-final-answer retry only checked text content (`hasVisibleFinalAnswer`, `internal/agent/agent.go`), so it misclassified such a finished turn as "no answer", overrode the model's stop signal, and forced another expensive thinking round (the user-visible "still thinking after the task is done" symptom), up to 3 times - or indefinitely when interleaved with tool calls, since `emptyFinalBlocks` only resets when a tool is called.

This guards the retry with `reasoningOnlyFinishHonoured(usage, reasoning)` so a `finish_reason="stop"` + non-empty `reasoning` turn is accepted as a final answer (falls through to `return nil`) instead of being retried.

Closes #6617

## Verification

- New test `TestRunAcceptsReasoningOnlyFinalWhenModelStopped` (`internal/agent/empty_final_test.go`) replays the exact DeepSeek sequence - `ChunkReasoning` + `ChunkUsage{FinishReason:"stop"}` + `ChunkDone`, no content - and asserts the turn is accepted: 1 provider call, no synthetic "visible answer" retry message injected.
- Existing `TestRunRetriesReasoningOnlyFinalAnswer` and `TestRunStopsAfterRepeatedEmptyFinalAnswers` still pass unchanged - `scriptedProvider` emits no usage chunk, so `usage` is `nil` and `reasoningOnlyFinishHonoured` returns `false`, preserving the current retry-then-error behaviour when there is genuinely no answer.
- `go test ./internal/agent/ ./internal/provider/openai/` - ok (full suites, ~29s).
- `go vet ./internal/agent/ ./internal/provider/openai/` - clean.
- `gofmt -l internal/agent/agent.go internal/agent/empty_final_test.go` - clean.

## Cache impact

Cache-impact: low - change is to the empty-final-answer retry control flow only; it does not mutate the system-prompt prefix, tool schemas, memory prefix, output styles, or provider request serialization that form the cache-stable surface. The provider-visible request shape is unchanged.
Cache-guard: TestRunAcceptsReasoningOnlyFinalWhenModelStopped covers the new accept path; TestRunRetriesReasoningOnlyFinalAnswer + TestRunStopsAfterRepeatedEmptyFinalAnswers guard the unchanged retry path; full internal/agent + internal/provider/openai suites pass and go vet is clean.
System-prompt-review: N/A

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -1502,15 +1502,15 @@ func (a *Agent) Run(ctx context.Context, input string) (runErr error) {
 				continue
 			}
 			if !hasVisibleFinalAnswer(text) {
-				// DeepSeek thinking mode frequently streams a long reasoning_content
-				// and then finishes with finish_reason="stop" but an empty content
+				// DeepSeek thinking mode can stream a long reasoning_content and
+				// then finish with finish_reason="stop" but an empty content
 				// block: the model has explicitly signalled completion and its
 				// reasoning was already streamed to the user. Retrying here overrides
 				// that stop signal and forces another expensive thinking round (the
 				// "still thinking after the task is done" symptom), so honour the
 				// stop when reasoning carried the substance of the answer and treat
 				// the turn as a final answer instead of retrying.
-				if !reasoningOnlyFinishHonoured(usage, reasoning) {
+				if !reasoningOnlyFinishHonoured(a.prov, usage, reasoning) {
 					emptyFinalBlocks++
 					if emptyFinalBlocks >= maxEmptyFinalBlocks {
 						return fmt.Errorf("model finished without a visible final answer %d times", emptyFinalBlocks)
@@ -2434,11 +2434,20 @@ func hasVisibleFinalAnswer(text string) bool {
 
 // reasoningOnlyFinishHonoured reports whether the model finished with a stop
 // signal but placed its answer in the reasoning stream rather than the content
-// block. DeepSeek thinking mode does this routinely: it streams a long
+// block. DeepSeek thinking mode does this occasionally: it streams a long
 // reasoning_content, then returns finish_reason="stop" with an empty content.
 // The model has signalled completion, so the host accepts the turn instead of
 // retrying and forcing another expensive thinking round.
-func reasoningOnlyFinishHonoured(u *provider.Usage, reasoning string) bool {
+//
+// The accept is scoped to DeepSeek thinking mode (ToolCallReasoningPolicy):
+// for other providers a reasoning-only turn keeps the empty-final retry
+// safety net — local <think>-tag models often recover a visible answer on
+// the second attempt, and a gateway that mislabels truncation as "stop"
+// must not have a degenerate turn committed as the final answer.
+func reasoningOnlyFinishHonoured(p provider.Provider, u *provider.Usage, reasoning string) bool {
+	if !provider.RequiresToolCallReasoning(p) {
+		return false
+	}
 	if u == nil || u.FinishReason != "stop" {
 		return false
 	}
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -1502,14 +1502,24 @@ func (a *Agent) Run(ctx context.Context, input string) (runErr error) {
 				continue
 			}
 			if !hasVisibleFinalAnswer(text) {
-				emptyFinalBlocks++
-				if emptyFinalBlocks >= maxEmptyFinalBlocks {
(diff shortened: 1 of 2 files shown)


SivanCola:
Pushed a supplemental commit c9016d6 (via maintainer edit) scoping the reasoning-only accept to DeepSeek thinking mode.

As written, the guard accepted any reasoning-only `finish_reason="stop"` turn from **any** provider, which silently widened the fix beyond the #6617 symptom: local `<think>`-tag models (ollama/vLLM) lost the empty-final retry safety net that often recovers a visible answer on the second attempt, and a gateway that mislabels truncation as `"stop"` could commit a degenerate turn as the final answer. The commit gates the accept on the existing `provider.RequiresToolCallReasoning` capability (true exactly for DeepSeek thinking mode — explicit `reasoning_protocol="deepseek"` or an official DeepSeek base URL), and adds `TestRunRetriesReasoningOnlyStopWithoutDeepSeekPolicy` pinning the scope. Note: DeepSeek routed through a third-party gateway now needs `reasoning_protocol = "deepseek"` on that provider entry to benefit.

Live verification against the real DeepSeek API (v4-pro/v4-flash, 27 calls): the reasoning-only stop is occasional rather than routine; truncation surfaces as `finish_reason="length"` (correctly still retried); and `finish_reason:"stop"` + `usage` arrive in the same final SSE chunk, so the `usage.FinishReason` check is live on the real wire. All suites pass.

(PR body wording updated accordingly: "routinely" → "occasionally".)

PR #6518, opened 2026-07-15
test(config): isolate Windows user config writes

lemonhall:
## Summary

- isolate Windows user config paths in two write-capable config tests
- add a guard that fails if a test path escapes its temporary root
- prevent tests from overwriting `%APPDATA%\reasonix\config.toml`

## Verification

- `go test ./internal/config -run 'Test(SaveToScopesUserAndProjectFiles|ProjectConfigCannotOverrideCredentialStoreMode)$' -count=1 -v`
- `go test ./internal/config -count=1`
- `go test ./internal/control -count=1`
- `go test ./internal/provider/openai -run 'Test(BuildRequest.*Image|ImageURLDetail)' -count=1 -v`
- `go vet ./...`
- `go build ./...`

## Cache impact

Cache-impact: none; test-only config path isolation.
Cache-guard: `go test ./internal/config -count=1`
System-prompt-review: N/A

PR #6639, opened 2026-07-18
test(config): isolate test user state / 隔离测试用户状态

### LINKED ISSUE #6640: [Bug]: 测试会覆盖用户级配置文件
### Version line

### Exact version

git `40ef98d`

### What happened?

具体的信息在 https://github.com/esengine/DeepSeek-Reasonix/pull/6639

### Steps to reproduce

`make test`

### OS / platform

Arch Linux

ffff2004:
## Summary

- Add a shared test-only guard that redirects HOME, USERPROFILE, XDG_CONFIG_HOME, and AppData to a disposable directory.
- Clear inherited REASONIX_HOME, REASONIX_STATE_HOME, and REASONIX_CACHE_HOME in state-writing test packages.
- Preserve sandboxed helper subprocess behavior and add Windows/default-path containment assertions.
- Use the locked session snapshot API in a steer fallback test, fixing a race exposed by the required full-repository race job.

## Root cause

Test isolation was duplicated and incomplete. Several package test binaries could inherit explicit Reasonix paths or platform-specific user paths and write to caller-owned configuration, credentials, session, trust, or cache state. A polling test also read the session message slice directly while a background turn could append to it.

Fixes #6640.

## Verification

- `make test` with external Reasonix home/state/cache sentinels; the configuration hash was unchanged and no unexpected files were created.
- `go test -p=1 ./... -count=1` with external home/state/cache sentinels.
- `go test -race ./internal/testenv ./internal/config ./internal/cli -count=1`.
- `go test -race ./internal/control -count=1`; related steer tests also passed 20 repeated race runs.
- `go vet ./...`.
- `./scripts/check-cache-impact.sh`.

## Cache and prompt impact

Cache-impact: none - only test isolation changes; no runtime prompt, tool registration, or cache behavior changes.
Cache-guard: `./scripts/check-cache-impact.sh` and sentinel-backed `make test`.
System-prompt-review: reviewed by @SivanCola; prompt-sensitive paths contain test-only files and prompt assembly is unchanged.

No runtime persistence format changes.

## Credits

Keeps and adapts the Windows config-path isolation and path-containment coverage from #6518 by @lemonhall.

### DIFF
--- a/internal/boot/main_test.go
+++ b/internal/boot/main_test.go
@@ -1,0 +1,11 @@
+package boot
+
+import (
+	"testing"
+
+	"reasonix/internal/testenv"
+)
+
+func TestMain(m *testing.M) {
+	testenv.RunWithIsolatedUserState(m)
+}
--- a/internal/bot/main_test.go
+++ b/internal/bot/main_test.go
@@ -1,0 +1,11 @@
+package bot
+
+import (
+	"testing"
+
+	"reasonix/internal/testenv"
+)
+
+func TestMain(m *testing.M) {
+	testenv.RunWithIsolatedUserState(m)
+}
--- a/internal/botruntime/main_test.go
+++ b/internal/botruntime/main_test.go
@@ -1,0 +1,11 @@
+package botruntime
+
+import (
+	"testing"
+
+	"reasonix/internal/testenv"
+)
+
+func TestMain(m *testing.M) {
+	testenv.RunWithIsolatedUserState(m)
+}
--- a/internal/cli/chat_tui_test.go
+++ b/internal/cli/chat_tui_test.go
@@ -36,6 +36,25 @@ const tinyPNGBase64 = "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42
 func TestMain(m *testing.M) {
 	old := detectTermuxTerminal
 	detectTermuxTerminal = func() bool { return false }
+	home, err := os.MkdirTemp("", "reasonix-cli-test-home-")
+	if err != nil {
+		panic(err)
+	}
+	for key, value := range map[string]string{
+		"HOME":            home,
+		"USERPROFILE":     home,
+		"XDG_CONFIG_HOME": filepath.Join(home, ".config"),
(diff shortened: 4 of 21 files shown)


ffff2004:
又抓到一条没修复的路径，正在修复

ffff2004:
> 又抓到一条没修复的路径，正在修复

已完成

ffff2004:
The apparent overlap between the two CLI isolation changes is intentional because they enforce different boundaries:

- The package-level `TestMain` is a safety boundary for the entire CLI test binary. It ensures that tests which do not call an isolation helper—such as `TestRunThemeSubcommandSwitchesAccentAndTextarea`, the remaining reproducer—can never inherit and write the developer's real home directory.
- `isolateCLIConfigHome(t)` provides per-test isolation. Each caller gets a fresh home and working directory, preventing config and migration state from leaking between tests. It also remains correct when a test deliberately sets `REASONIX_HOME` before calling the helper.
- The change in `internal/config/edit_test.go` is separate: that test selected the wrong environment variable for `UserConfigPath()` and therefore needed its own targeted correction.

Removing the package-level isolation would reopen the host-state escape for tests that do not use the helper. Removing the per-test isolation would make helper callers share mutable config state and weaken migration tests. The repeated environment setup is therefore a small, deliberate defense-in-depth measure rather than two implementations of the same responsibility.

ffff2004:
这个 PR 修复 https://github.com/esengine/DeepSeek-Reasonix/issues/6640

SivanCola (approved):
Reviewed the original fix and the integrated test-isolation hardening.

Verified `make test` with external Reasonix home/state/cache sentinels, a serialized full-repository run, focused race coverage, `go vet ./...`, and the cache-impact guard. The sentinel configuration hash remained unchanged and no external state/cache files were created. Changes are test-only and preserve helper subprocess behavior.

SivanCola (approved):
Final-head review complete. The CI-discovered test race was corrected by reading Session.Snapshot instead of the concurrently-mutated Messages slice. Focused control race tests, all five required checks, the cache-impact guard, and CodeQL now pass. The current base advances documentation only and merges cleanly.

PR #6270, opened 2026-07-09
feat（windows-desktop）: 中断后保留流式回复的部分内容 / Preserve partial streaming response on user interrupt

SauronSkywalker:
feat（windows-desktop） 中断后保留流式回复的部分内容 / Preserve partial streaming response on user interrupt

## 问题描述 / Problem

用户中断模型的流式回复（中止按钮 / ESC 退出 ask 工具）后，刷新页面或切换会话时，已流出的回复内容完全消失。用户只能看到提问，看不到任何模型回复。

When a user interrupts a streaming model response (stop button / ESC from ask), after refreshing the page or switching sessions, the partial response disappears entirely. The user sees their question but no assistant reply.

### 1. `Agent.Run` 未处理 context.Canceled

`stream()` 中 `ctx.Done()` 路径返回 `interrupted=false`，`Agent.Run` 直接 `return err`，已流出的文本和推理内容未保存到 session。

### 2. `stripCancelledVisibleTurnMessagesAfter` 清除所有助理消息

即使消息被保存到 session，`stripCancelledVisibleTurnMessagesAfter` 也会清除所有非 User 角色的消息。

### 3. `SaveRewrite` 与 autosave 冲突

`replaceSessionAfterCancel` 调用 `SaveRewrite` 时，若 autosave 已推进文件 revision，检测到冲突后 `recoverSnapshotConflict` 可能采纳磁盘旧版本（不含助理消息）。

## 方案 / Solution

引入 `ErrTurnInterrupted` 哨兵错误，将"用户中断有部分回复的 turn"定义为系统的一个标准流程终止状态，而非异常：

```
Agent.Run 加消息到 session 内存
  ↓ return ErrTurnInterrupted（哨兵）
  ↓
Controller 哨兵处理器：saveInterruptedTranscript() + clearInFlightTurn + stopGoal
  ↓ return nil（turn 正常结束）
  ↓
defer snapshotActivityIfChanged → 文件已最新 → 跳过
```

### File 1: `internal/agent/agent.go`

| 行 / Line | 变更 / Change |
|---|---|
| +7 | 添加 `"errors"` 导入 / Added `"errors"` import |
| +46-51 | 新增 `ErrTurnInterrupted` 哨兵错误 / Added sentinel error |
| +1156-1173 | 新增 cancel handler：保存助理消息 + 双语标记 + `return ErrTurnInterrupted` / Added cancel handler: save assistant message + bilingual marker + return sentinel |

[...]

PR #6666, opened 2026-07-19
Preserve cancelled turn display history / 保留中断轮次的可见历史

### LINKED ISSUE #6260: [Bug]: 手动取消的对话再次对话会丢失部分历史记录

### LINKED ISSUE #6297: [Bug]: 桌面版在对话中按 esc 或者中断对话，再继续对话时，上下文丢失，基本场景的重大问题
### Version line

### Exact version

1.17.10

### What happened?

桌面版在对话中按 esc 或者中断对话，再继续对话时，上下文丢失，基本场景的重大问题

### Steps to reproduce

1.输入一个话题
2.在他结束前按 esc 或者中断对话
3.继续输入补充话题确认
4.明显感觉 ai 丢失了之前的对话

### OS / platform

macos 15.7.4

### Relevant logs or output

```shell

```

### LINKED ISSUE #6644: [Bug]: Desktop 手动终止会话后当前 AI 回复丢失（用户消息已保存，流式回复未落盘）
### Version line

### Exact version

v1.17.15（此问题在 v1.17.15 之前就已存在）

### What happened?

在 Desktop 中手动点击终止会话（interrupt/停止），当前正在生成的那一轮 AI 回复内容完全丢失。此行为在 v1.17.15 之前就已存在，并非本次版本的回归。

具体表现：
- **用户消息（提问）**：已正常保存到会话文件，重新打开会话能看到
- **AI 回复**：流式输出的内容（已显示在界面上的部分）在终止后全部丢失，重新打开会话时不存在

### Steps to reproduce

1. 在 Desktop 中发起一个耗时较长的提问（例如调用 MCP 工具）
2. 在 AI 正在生成回复时，点击终止按钮（红色停止）
3. 关闭会话标签页，重新从侧边栏打开该会话
4. 观察到用户的提问还在，但 AI 的回复为空/不存在

### OS / platform

Windows 11

### Relevant logs or output

根因推测：会话写入器（session writer）只在 AI 完整生成一轮后才 commit 回复内容到 .jsonl 文件。流式输出中的中间片段仅保存在内存中用于界面渲染，终止时未触发 flush/commit。

SivanCola:
## Summary

Preserve the visible output of a user-cancelled Desktop turn across planner cancellation, tab switches, reloads, and runtime detach/reattach without treating incomplete assistant, reasoning, or tool output as trusted model context.

The canonical session transcript retains the user's original message (including image attachments) and removes the incomplete assistant/tool tail. Already-rendered output is persisted separately in the existing display-only planner sidecar, together with a localized notice that explains the context boundary and asks the user to inspect the workspace before continuing or reverting.

Fixes #6260
Fixes #6297
Fixes #6644

Refs #6278
Refs #6270

## Problem and root cause

Desktop kept streaming cards only in frontend memory. After cancellation, the controller correctly removed incomplete assistant/tool messages from the provider transcript to prevent malformed tool sequences and repeated execution, but there was no durable display buffer for the output the user had already seen. Switching sessions or reloading therefore made the cancelled turn appear to disappear.

The full lifecycle had several gaps:

- cancellation during planning could happen before the executor appended the current user message;
- display buffers belonged to a transient `WorkspaceTab`, so runtime detach/reattach could split one turn across discarded wrappers;
- the shared planner display sidecar handled writes but did not remove entries on every permanent deletion path;
- executor text was appended with growing-string concatenation, causing quadratic allocation amplification for every streamed turn;
- sidecar load-modify-save serialization used only an in-process mutex, so two Reasonix processes could each publish a stale map and drop the other process's entry.

The cancellation flag was also cleared during the current turn-finalization path before Desktop could distinguish a user cancellation from other completion outcomes.

## Changes

- Add an internal cancellation marker to `TurnDone` and capture it before finalization clears the cancellation state.
- Preserve an explicit, prefix-free current user fallback (including image attachments) when planning is cancelled before the executor starts.
- Buffer planner and executor display messages in runtime-owned shared state and transfer it before sink rebinding during detach/reattach.
- Accumulate streamed display text as chunks and materialize it once; successful executor turns discard their chunks without joining them.
- Persist cancelled executor output in the existing display-only sidecar while keeping incomplete assistant/reasoning/tool output out of the canonical provider transcript.
- Guard the full planner-display load-modify-save/remove transaction with a fixed, bounded cross-process advisory lock on Unix and Windows.
- Keep failed display-only writes in an ordered in-memory queue and retry them asynchronously instead of dropping the turn after a transient lock or filesystem failure.
- Reject corrupt planner-display maps during normal recording and retire an unreadable map during destructive cleanup so deleted display data does not linger.
- Remove planner display entries during live-session cleanup and trash purge; prune only orphaned entries while retaining data for restorable trashed sessions.
- Add a stable `cancelled_turn_display` notice code with English, Simplified Chinese, and Traditional Chinese translations.
- Cover planner-stage cancellation, image preservation, reload, runtime detach/reattach, permanent deletion, trash retention, orphan pruning, allocation growth, retry behavior, corrupt data, in-process concurrency, cross-process concurrency, runtime status, and frontend localization with regression tests.

### Kept and adapted

- #6278 by @GTC2080 established the safer architecture: separate display-only cancelled history from the canonical model transcript. This PR adapts that mechanism to the current `main-v2` finalization path, adds stable notice localization, cross-process-safe persistence, and expanded end-to-end and race coverage. The adopted contribution is credited with a commit-level `Co-authored-by` trailer.

### Reviewed but not adopted

- #6270 by @SauronSkywalker explored preserving partial assistant/tool state directly in the continuation transcript. That path was not adopted because incomplete or unpaired tool state can produce invalid provider sequences or cause interrupted work to be repeated. Its reproduction, investigation, and alternative design informed this integration, so @SauronSkywalker is also credited as a repository contributor with a commit-level `Co-authored-by` trailer.

## Backward compatibility

| Field or artifact | Old-data behavior | Conclusion |
| --- | --- | --- |
| Session JSONL / canonical transcript | Format is unchanged; existing sessions load exactly as before | Backward-safe |
| `.planner-display.json` sidecar | Path and JSON schema are unchanged; old files remain readable and new writes remain readable by older Desktop versions | Backward-safe |
| `.planner-display.json.lock` | New empty advisory lock file is ignored by older versions and contains no user content | Backward-safe |
| Notice `code` | Older readers ignore the unknown optional field and can still display the persisted content | Backward-safe |
| Internal `Event.Cancelled` boolean | Existing event producers deserialize or construct the zero value (`false`) | Backward-safe |
| Cross-version Desktop use | No provider request or canonical transcript shape changes; old versions can read the same session and display files | Backward-safe |

## Cache impact

No provider-visible system prompt, tool schema, tool ordering, or request serialization changes were made. The canonical cancelled-turn transcript remains clean. Only transient compose prefixes are removed from the preserved cancelled user message, which avoids carrying stale mode instructions into later turns. `./scripts/cache-guard.sh` passes all scenarios.

## Security and privacy

- No dependency, network, credential, permission, or privilege changes.
- The display sidecar keeps its existing local history format and sensitivity boundary; temporary writes and the new lock file use restrictive permissions.
- Cross-process lock acquisition is deadline-bounded, and no `App` or controller callback lock is held while waiting.
- Planner display entries are removed when a session is permanently deleted or purged, and orphaned entries are pruned during session listing.
- Partial cancelled output is explicitly marked as display-only and is not silently promoted into trusted model context.

## Verification

- `go test ./...`
- `go vet ./...` in both Go modules
- Focused `internal/control`, file-lock, and Desktop cancellation, planner, reload, detach/reattach, deletion, pruning, corruption, and retry tests
- Focused `-race` tests for controller cancellation, runtime handoff, display buffering, sidecar persistence, and file locking
- Deterministic subprocess regression that forces two independent Reasonix processes through the former stale read-modify-write window
- Allocation regression: a 64 KiB stream split into 2,000 chunks uses 166,968 bytes and 17 allocations per operation (previous repeated concatenation reproduced about 69 MiB)
- Windows/amd64 cross-compilation for the file-lock package and Desktop test binary
- `pnpm typecheck`
- `pnpm exec tsx src/__tests__/use-controller-meta.test.ts` (124 passed)
- `./scripts/cache-guard.sh`
- `git diff --check`

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -2007,6 +2007,9 @@ func removeDesktopSessionArtifacts(path string) error {
 	if err := removeSessionDisplay(filepath.Dir(path), path); err != nil {
 		return err
 	}
+	if err := removeSessionPlannerDisplay(filepath.Dir(path), path); err != nil {
+		return err
+	}
 	if err := agent.DeleteSubagentsByParent(filepath.Dir(path), agent.BranchID(path)); err != nil {
 		return err
 	}
@@ -2385,6 +2388,7 @@ func (a *App) ListSessions() []SessionMeta {
 		}
 	}
 	_ = pruneSessionDisplays(dir, protectedDisplays)
+	_ = pruneSessionPlannerDisplays(dir, protectedDisplays)
 	titles := loadSessionTitles(dir)
 	channelRoutes := channelSessionRoutesForDir(dir)
 	active := a.activeSessionPath(dir)
--- a/desktop/frontend/src/__tests__/use-controller-meta.test.ts
+++ b/desktop/frontend/src/__tests__/use-controller-meta.test.ts
@@ -236,6 +236,11 @@ console.log("\nuse controller meta");
     "workspace lease contention uses its stable localized notice code",
   );
   eq(
+    localizedNoticeText("reworded cancelled-turn copy", "cancelled_turn_display"),
+    "This turn was interrupted. Partial output is kept for reference but is not included in the model's next-turn history; inspect the workspace before continuing or reverting changes.",
+    "cancelled turn history explains the model-context boundary",
+  );
+  eq(
     localizedNoticeText("Tool round limit reached; asking the assistant to summarize progress.", "unknown_future_code"),
     "Tool round limit reached; asking the assistant to summarize progress.",
     "an unknown notice code falls back to exact-text matching",
--- a/desktop/frontend/src/lib/useController.ts
+++ b/desktop/frontend/src/lib/useController.ts
@@ -1424,6 +1424,7 @@ const noticeCodeKeys: Record<string, DictKey> = {
   tool_budget: "notice.toolBudget",
   loop_guard: "notice.loopGuard",
   workspace_lease: "notice.workspaceLease",
+  cancelled_turn_display: "notice.cancelledTurnDisplay",
 };
 
 // localizedNoticeText localizes a notice's main copy by its stable code first,
--- a/desktop/frontend/src/locales/en.ts
+++ b/desktop/frontend/src/locales/en.ts
@@ -2360,6 +2360,7 @@ export const en = {
   "notice.toolBudget": "Tool round limit reached; asking the assistant to summarize progress.",
   "notice.loopGuard": "The assistant is not making progress; asking it to reassess the current step.",
   "notice.workspaceLease": "Another Delivery session is writing to this workspace; this session will continue automatically when it is safe.",
+  "notice.cancelledTurnDisplay": "This turn was interrupted. Partial output is kept for reference but is not included in the model's next-turn history; inspect the workspace before continuing or reverting changes.",
   "notice.contextLarge": "Context is getting large; preserving cache until cleanup is needed.",
   "notice.contextCleanupSkipped": "Context cleanup skipped for now.",
   "notice.contextCleanupPaused": "Automatic context cleanup paused because the context window is too small.",
--- a/desktop/frontend/src/locales/zh-TW.ts
+++ b/desktop/frontend/src/locales/zh-TW.ts
@@ -1580,6 +1580,7 @@ export const zhTW: Record<DictKey, string> = {
   "notice.toolBudget": "工具呼叫輪數已達上限，已要求助手總結目前進度。",
   "notice.loopGuard": "助手沒有取得有效進展，已要求它重新評估目前步驟。",
   "notice.workspaceLease": "另一個交付會話正在寫入此工作區；安全後本會話會自動繼續。",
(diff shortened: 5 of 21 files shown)


SivanCola:
Heads-up: #6667 modifies `internal/control/controller.go` near `recoverInterruptedTurn` (adjacent to the `stripTurnMessagesAfter` region this PR touches) plus `desktop/app.go` and `desktop/tabs.go`. No semantic overlap spotted — expect at most a small rebase for whichever merges second.

PR #6667, opened 2026-07-19
Bound session save lock waits / 限制会话保存锁等待时间

SivanCola:
## Summary

- bound cross-process compatibility session-save lock acquisition to five seconds on Unix and Windows
- keep retrying short-lived lock contention and save normally when the competing writer releases before the deadline
- on a true desktop shutdown timeout, persist the in-memory transcript to a process-specific recovery branch before closing controllers or releasing session leases
- reuse the existing session checkpoint, event log, branch metadata, recovery handoff, telemetry, and session-listing flows
- add regressions for bounded failure, transient lock release, durable shutdown recovery, controller re-anchoring, and desktop shutdown ordering
- relates to the true-quit/session-lock portion of #6637; it does not change close-to-background behavior or claim to fix upstream long-context 503 responses

## Verification

- `go test -race ./internal/control`
- `go test ./...`
- `go vet ./...`
- `cd desktop && go test -race ./...`
- Windows amd64 cross-compilation for `internal/agent`, `internal/control`, and the desktop module
- focused lock/recovery tests repeated 20 times
- focused desktop shutdown test repeated 10 times
- cache-impact guard

## Compatibility and safety

- no persisted session-format, configuration, provider-request, tool-schema, or frontend API changes
- ordinary snapshots retain their existing behavior; the recovery branch is used only when the final shutdown snapshot returns the bounded file-lock sentinel
- the shutdown recovery path never reacquires the stalled original file lock and includes a process-specific suffix to avoid a second blocked recovery target
- recovery depth remains capped for ordinary conflict handling; shutdown recovery saturates at the cap so data durability does not reintroduce unbounded recovery chains
- all application mutexes are released before persistence and recovery callbacks; controller snapshot serialization remains in place

## Cache impact

Cache-impact: none — session persistence and desktop shutdown only; provider-visible prompts, request bodies, tool schemas, and compaction thresholds are unchanged.

Cache-guard: `scripts/check-cache-impact.sh` reports no cache-sensitive files changed.

System-prompt-review: N/A

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -792,19 +792,22 @@ func (a *App) shutdown(context.Context) {
 	a.mu.RLock()
 	tabs := a.runtimeTabsLocked()
 	type shutdownItem struct {
-		tab  *WorkspaceTab
-		ctrl control.SessionAPI
+		tab      *WorkspaceTab
+		ctrl     control.SessionAPI
+		readOnly bool
 	}
 	items := make([]shutdownItem, 0, len(tabs))
 	for _, t := range tabs {
 		if t.Ctrl != nil {
-			items = append(items, shutdownItem{tab: t, ctrl: t.Ctrl})
+			items = append(items, shutdownItem{tab: t, ctrl: t.Ctrl, readOnly: t.ReadOnly})
 		}
 	}
 	a.mu.RUnlock()
 	for _, it := range items {
-		if err := a.snapshotTab(it.tab); err != nil {
-			slog.Warn("desktop: shutdown snapshot failed", "tab", it.tab.ID, "err", err)
+		if !it.readOnly {
+			if err := it.ctrl.SnapshotForShutdown(); err != nil {
+				slog.Warn("desktop: shutdown snapshot failed", "tab", it.tab.ID, "err", err)
+			}
 		}
 		it.ctrl.Close()
 		it.tab.releaseSessionLease()
--- a/desktop/app_autosave_test.go
+++ b/desktop/app_autosave_test.go
@@ -247,6 +247,13 @@ func TestDesktopSnapshotConflictRecoveryUpdatesTabAndProjectTree(t *testing.T) {
 	if tab.SessionPath != recoveryPath {
 		t.Fatalf("tab session path = %q, want recovery path %q", tab.SessionPath, recoveryPath)
 	}
+	saved := loadTabsFile()
+	if len(saved.Tabs) != 1 || saved.Tabs[0].ID != tab.ID {
+		t.Fatalf("saved tabs = %+v, want recovered tab %q", saved.Tabs, tab.ID)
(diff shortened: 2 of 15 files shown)


SivanCola (commented):
Review record (full verification pass before merge; GitHub does not allow the author account to formally approve, so this comment stands in as the review).

This is a well-scoped fix for the true-quit/session-lock portion of #6637. It addresses desktop shutdown wedging on a cross-process session save lock by bounding lock acquisition to a five-second non-blocking retry window on Unix and Windows, and — only when the final shutdown snapshot still times out — persisting the in-memory transcript to a process-specific recovery branch that never re-acquires the stalled original lock.

- It honors the existing invariant that UI-critical lock domains must never wait indefinitely on a cross-process file lock. Under the old blocking acquisition, a background autosave stuck on a stale `.lock` held `snapshotMu` forever, so the final shutdown snapshot deadlocked behind it; the bounded wait fixes the wedge at its root instead of papering over one call site.
- The shutdown recovery branch reuses the standard session, event-log, and branch-meta formats and the existing recovery handoff (callback → controller re-anchor → checkpoint rebind → in-flight turn-marker transplant), so session listing, resume, and the meta-driven recovery GC all handle it unchanged. Recovery depth saturates at the cap instead of exceeding it, preserving the anti-cascade policy for ordinary conflicts.
- Making the lease-backed tab path authoritative during the two-phase recovery handoff closes a real window where a concurrent layout save could overwrite the recovery anchor with the controller's stale path. The lock-free lease-key mirror keeps `App.mu` holders from waiting on lease file operations, and all five lease mutation sites update the mirror under `sessionLeaseMu`.

Test coverage is solid: bounded failure, transient lock release, the shutdown branch bypassing the held original lock, controller re-anchoring plus handoff bookkeeping, desktop shutdown ordering (durable snapshot before Close), the concurrent-layout-write window, and failed lease acquisition staying anchored to the original session.

Verified locally at c1fbd06:
- `go build ./...` and module-wide `go vet ./...` passed (root and desktop modules)
- `go test -race ./internal/agent ./internal/control` passed
- `cd desktop && go test -race ./...` passed
- trial merge with the latest `origin/main-v2` is clean; vet and focused tests pass post-merge
- `scripts/check-cache-impact.sh` reports no cache-sensitive files changed

Two non-blocking notes:
- Behavior change worth knowing: every save path (CLI/serve/ACP included) now fails after ~5s with `ErrSessionFileLockHeld` instead of waiting indefinitely when another process holds the `.lock`; the in-memory transcript is retained and the next autosave retries. Strictly better than the previous indefinite hang.
- Integration order: #6529 patches the same `Controller.snapshot` block against the old two-argument signature and will need a rebase once this lands; #6666 touches an adjacent controller region. Both are semantically complementary — no conflict in intent.

Issue #6652, opened 2026-07-18
[Bug]: mcp无法使用,功能失效,我提供的只是其中一个mcp,几乎所有mcp均有同样问题


suanfalalala:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

v1.17.15

### What happened?

最新版本配置mcp后无法正常使用

### Steps to reproduce

环境信息
组件	版本
Reasonix	v1.17.15
操作系统	Windows 11
MATLAB	R2025b
MATLAB MCP Server（二进制）	v0.11.2
MATLAB MCP Server Toolbox（MATLAB 端插件）	v0.3.0
matlab-agentic-toolkit（Reasonix 端技能插件）	v0.1.0
配置
[[plugins]]
name    = "matlab"
type    = "stdio"
command = "matlab-mcp-server.exe"
args    = ["--matlab-root", "D:\\MATLAB\\R2025b", "--matlab-session-mode", "new"]
auto_start = false

报错
plugin "matlab": read: EOF: stderr: Failed to initialize MCP Server.

根因
MATLAB MCP Server v0.11.2 在收到 initialize 请求后，先发两条通知再回复结果（MCP 规范允许）：

← notifications/tools/list_changed
← notifications/resources/list_changed
← initialize 响应

Reasonix 收到第一条通知后直接关闭连接，服务器端收到 stdin EOF 退出。

验证
脱离 Reasonix 用 .NET Process 直接握手完全正常，返回 5 tools、2 resources。

### OS / platform

win10

### Relevant logs or output

```shell
matlab: plugin "matlab": read: EOF: stderr: Failed to initialize MCP Server
```

SivanCola:
Fixed by #6669, merged into main-v2 as 4bd70c9fc1a09dfe5fbc305b2bc254447ace854c.

The stdio MCP initialization path now preserves startup notifications and a final response delivered with EOF, and it answers server-initiated JSON-RPC requests without blocking the read loop. User-configured MCP servers also return to direct add-and-connect behavior.

Verification included deterministic regressions for the reported startup ordering, the full Windows root test job, Windows Desktop tests, race tests, and the complete CI matrix. The original MATLAB environment was not directly available during verification; please reopen this issue with fresh logs if the next build still reproduces it.

Availability: the fix is merged but is not in a stable release yet. It will ship with the next release.

PR #6669, opened 2026-07-19
Streamline MCP install and launch authorization / 精简 MCP 安装与启动授权

SivanCola:
## Summary

- harden stdio MCP initialization for startup notifications, EOF-with-data, and server-initiated JSON-RPC requests
- serialize MCP reconnect, provider teardown, ordinary runtime teardown, plugin uninstall, and shutdown against live runtime work
- restore direct add-and-connect behavior for user-configured MCP servers
- honor explicit install/launch authorization for all MCP calls by default, including tools that declare `destructiveHint`
- replace the legacy session and capability trust state machine with one exact project launch grant
- remove Desktop and CLI trust, re-verification, revocation, change-review, and manual catalog-maintenance surfaces
- let valid incomplete JSON drafts return to the MCP form without weakening final submission validation

## Problem

MCP use had become more complicated than configuration warranted. User-added servers could encounter extra trust concepts and repeated tool-call confirmation, while repository-discovered servers exposed session receipts, capability snapshots, identity drift, re-verification, revocation, and catalog controls. The same subsystem also carried startup and lifecycle races that could block initialization or leave shared registries inconsistent.

## Root cause

Launch authorization, tool-call safety, package identity, capability snapshots, session receipts, and catalog state were coupled in one large trust workflow. Desktop, CLI, events, Wails bindings, plugin startup, lazy schema caching, and persisted state all had to reproduce that workflow. Separately, stdio request handling and runtime teardown were not fully serialized with shared MCP lifecycle work. The MCP permission gate also evaluated `destructiveHint` before the effective source-aware `approve` policy, so an explicit install could still trigger another confirmation.

## Fix

- answer server-initiated stdio requests outside the read loop and preserve the final message when EOF arrives with data
- keep effective project/user source precedence and shared-host lifecycle operations deterministic
- allow servers explicitly added by the user to save, connect, and expose tools without another authorization step
- treat an effective `approve` policy as the user's install/authorization decision for all calls, while preserving explicit deny and fresh destructive review under `auto`, `prompt`, and `writes`
- require one durable exact launch authorization only for repository-controlled project configuration
- replace `internal/mcptrust` with the smaller `internal/mcplaunch` state owner
- remove trust inspection, trust scopes, session receipts, capability snapshots, re-verification, revoke-trust, change-review, and catalog-refresh APIs and UI
- remove the CLI `mcp trust`, `untrust`, `verify`, and `catalog refresh` commands
- keep cached and live tool fingerprints stable while deriving ordinary read-only classification from local declarations or signed package metadata
- allow incomplete connection fields only while switching JSON drafts back to the form; final add/save remains strict

## User flow

1. A user adds or imports an MCP server.
2. Reasonix saves it, connects immediately, and permits its tools directly by default, including tools that declare `destructiveHint`.
3. Advanced users can opt into `auto`, `prompt`, or `writes`; explicit deny remains highest priority.
4. A server discovered from repository configuration asks once for **Authorize and connect** before its first launch.
5. Server details expose connection information, reconnect, delete, and tools. There is no second trust or re-verification workflow.

## Security boundaries retained

- explicit deny rules and opt-in MCP approval modes (`auto`, `prompt`, and `writes`)
- fresh confirmation for destructive tool calls whenever the effective mode is not `approve`
- sandbox and process isolation boundaries
- exact project launch authorization
- signed official-package verification and background revocation updates
- explicit local or signed authority for strict read-only execution

These controls protect launch and tool execution while making explicit install/authorization the default tool-call boundary, without turning tool-list snapshots into a second authorization system.

## Backward compatibility

| --- | --- | --- |
| `mcp-security.json` path and version | Read and written at the existing path with version 1 | Compatible |
| `launch_grants` | Existing exact workspace grants remain readable | Compatible |
| `launcher_locks` | Existing launcher locks retain their format and enforcement behavior | Compatible |
| legacy `receipts` | Raw records are preserved; an exact matching project receipt may migrate once to a launch grant, but its tool snapshot grants no tool authority | Safe migration |
| `official_denials` and `legacy_imports` | Raw JSON is preserved during read-modify-write and ignored as tool authority | No data loss |
| MCP configuration | Existing user and project configuration fields remain readable; omitted approval fields use the new source-aware runtime default | Compatible |
| provider-visible MCP tools | Tool names, schemas, normalization, fingerprints, and ordering remain stable | Cache stable |

## Cache impact

Provider-visible MCP tool schema bytes and ordering are preserved. Approval policy and the removed trust metadata remain host-side state and UI; neither is injected into the system prompt. `./scripts/cache-guard.sh` passed all scenarios at or above the required threshold on the original integration candidate, and the follow-up commits do not change provider-visible registration or schemas.

## Verification

Candidate: `36deb9152951b8089c52e78254d0b84e1332890e`

- root `go test ./...`
- focused `go test ./internal/permission ./internal/control ./internal/plugin ./internal/agent`
- frontend MCP action regression test
- frontend production and test TypeScript typechecks
- focused race tests for MCP launch, plugin, agent, authorization, reconnect, and update paths on the integration candidate
- cache guard on the integration candidate
- Windows cross-compilation for affected root and Desktop packages on the integration candidate
- browser verification of the simplified MCP settings flow and Form/JSON switching
- `git diff --check`

Reporter retesting for the original initialization reports can continue against this candidate.

Refs #6652
Refs #6657

Cache-impact: none - MCP approval ordering and host-side launch policy do not change provider-visible schemas, names, ordering, or prompt content.
Cache-guard: ./scripts/cache-guard.sh - passed all required scenarios on the integration candidate; follow-up commits do not alter provider-visible registration or schemas.
System-prompt-review: Reviewed - MCP launch and approval changes preserve generated system-prompt content and ordering.


SivanCola:
**Review summary** (full three-dot diff against `origin/main-v2`, trust-gate mechanism, and desktop reconnect lifecycle)

Both reported mechanisms check out:

- **stdio dispatch**: answering server-initiated requests such as `ping` is required by the MCP spec; the old dispatcher silently dropped them, and the final-message-with-EOF case lost the initialize response. Both are covered by deterministic regressions replaying the reported startup sequence (notifications + blocking ping before the initialize response).
- **trust recovery**: `SetSpecTrust`'s exact preflight launched a second server instance while the live one was still running, which single-instance servers reject. The new stop → preflight → reconnect/restore flow has correct failure atomicity, and the restore path cannot bypass the launch-approval gate — `LaunchAuthorized` is enforced inside the connect path, so a drifted binary still cannot start.

**Supplemental commit** fdd17fb: moved server-request replies off the read loop into a bounded reply queue owned by a dedicated writer goroutine. `readLoop` is the only goroutine draining stdout; letting it block on `writeMu` behind a jammed stdin pipe could deadlock both pipes once the server also blocks writing stdout, and a failed reply write killed the whole transport even with a healthy read side. Replies now drop on overflow (degrading to the pre-existing unanswered-request behavior) and reply write failures are non-terminal. Added `TestStdioReadLoopStaysLiveWhenReplyWriterIsBlocked`.

**Verified locally on the merged tree** (`origin/main-v2` + PR head):
- `go test ./internal/plugin -race -count=1` passed
- focused stdio regressions `-count=5 -race` passed
- `go test -race -run TestSetMCPTrustWorkspaceRefreshesEverySharedHostRegistry` (desktop module) passed
- `go vet`, `gofmt`, `GOOS=windows GOARCH=amd64 go test -c ./internal/plugin` passed

**Merge follow-up**: keep #6652 / #6657 open until the reporters re-test in their real environments (Windows 11 + MATLAB R2025b; existing chrome-devtools configuration), per the release acceptance list. Note the exact #6652 mechanism is not yet reporter-confirmed — the released v1.17.15 read loop dropped notifications rather than closing the connection, so the duplicate-launch preflight and the dropped server `ping` are the plausible culprits this PR addresses.

SivanCola:
**Follow-up review round**: a second review pass confirmed a P1 concurrency defect on the previous head, now fixed in f540eef.

**[P1] Trust preflight raced MCP enable/disable/reconnect.** `SetMCPTrust` held `runtimeRebuildMu` and snapshotted per-tab enabled state before the preflight, but `SetMCPServerEnabled`, `ReconnectMCPServer`, and the other MCP mutation entry points did not take that lock. Deterministic repro: a disable issued during a blocked workspace preflight returned in under a millisecond, and the trust reconnect then re-registered the server's tools from the stale snapshot — re-exposing tools the user had just disabled and desynchronizing the provider-visible registry. Fix: every Wails-exposed MCP mutation (trust inspect/set, add/update/remove, reconnect, auth cleanup, enable/disable, tier, catalog revocation apply) now serializes through `lockMCPMutation` on `runtimeRebuildMu`, with the documented lock order `runtimeRebuildMu → App.mu → Host/Registry`. Two new deterministic regressions use a TCP start gate to hold or fail the preflight launch: `TestSetMCPTrustSerializesConcurrentDisable` and `TestSetMCPTrustRestoresConnectionAfterPreflightFailure`.

**[P3]** The `transport` interface comment still claimed server-initiated requests are ignored; updated to match the implemented behavior (notifications ignored, `ping` answered, unsupported requests get JSON-RPC method-not-found).

**Verified on the new head** (nested-lock audit found no path that re-enters `runtimeRebuildMu` or calls another locked entry point):
- desktop `go test ./...` passed
- the three MCP trust regressions with `-race` passed, concurrency pair stable at `-count=3`
- `go test ./internal/plugin -race`, `go build ./...`, `go vet`, `gofmt` clean

**Related PR ordering**: #6033 (persistent MCP disable / per-server state control) edits the same Desktop MCP entry points and is currently conflicting. Recommend merging this PR first and rebasing #6033 onto the unified lifecycle lock.

SivanCola:
**Third review round**: one more P1 found and fixed in fb5abf8.

**[P1] Plugin uninstall bypassed the MCP lifecycle lock.** `RemovePlugin` disconnected the uninstalled plugin's MCP servers through its `OnDisconnect` callback before `rebuild()` acquired `runtimeRebuildMu`. Racing a workspace trust preflight, the unlocked disconnect could interleave so the trust reconnect relaunched the just-removed server from its stale snapshot, leaving an orphaned process attached to the shared Host. Fix: `RemovePlugin` now holds the lifecycle lock across the whole uninstall and calls `rebuildSettingLocked("plugins")` so the rebuild does not re-acquire the lock.

**Sweep**: with this change, a desktop-wide audit of `ConnectMCPServer` / `DisconnectMCPServer` / `UnregisterMCPServerTools` call sites confirms every live-MCP mutation entry point now runs under `lockMCPMutation`.

**Verified on the new head**:
- `TestRemovePluginSerializesWithTrustPreflight` fails against the unlocked implementation and passes with the lock (mutation-tested)
- all four MCP lifecycle regressions pass with `-race`
- desktop `go test ./...` passed; `go build ./...`, `go vet`, `gofmt` clean

SivanCola:
**Fourth review round**: two more P1s confirmed and fixed in e62262c.

**[P1] Stale active-work check across the lifecycle-lock wait.** `RemovePlugin` checked for active runtime work before waiting on the MCP lifecycle lock (a wait that can last a 30s trust preflight). Work starting mid-wait was only detected by the rebuild after the plugin had already been deleted — the user got a failure for an uninstall that had happened, and a running turn could lose its MCP mid-flight. Fix: the check is repeated under the lifecycle lock and the active tab's turn gate, the gate is held through the uninstall and rebuild (`rebuildSettingTurnLocked` is the turn-gate-free split of `rebuildSettingLocked`), and the removal now fails before deleting anything.

**[P1] Global uninstall only cleaned the active tab.** The uninstall's `OnDisconnect` only called the active tab's `DisconnectMCPServer`: sibling tabs on the shared Host kept provider-visible tools backed by the closed client, and other workspaces kept running the uninstalled server. Fix: `disconnectMCPServerAllRuntimes` sweeps every visible and detached runtime tab — `DisconnectMCPServer` stops the shared client once per Host and drops the stale tool prefix from every other controller's registry.

**Test gap addressed**: the previous serialization test only proved lock ordering with a nonexistent plugin. Two new regressions exercise the real uninstall flow against an installed plugin package (native manifest + `pluginpkg` state): `TestRemovePluginRechecksActiveWorkUnderLock` (background job injected during the lock wait; asserts the guard fires and the plugin is still installed) and `TestRemovePluginDisconnectsEveryRuntime` (asserts the shared client is gone, all three registries dropped the tool, and the plugin root/state are removed). Both fail against the previous implementation (mutation-tested).

**Verified on the new head**: full `TestSetMCPTrust*`/`TestRemovePlugin*` set with `-race`, desktop `go test ./...`, `go build ./...`, `go vet`, `gofmt` — all clean.

SivanCola:
**Fifth review round**: both P1s confirmed and fixed in c24ab5b with the suggested unified runtime gate.

**[P1] Global uninstall gated only the active tab.** `RemovePlugin` held the active tab's turn gate while disconnecting every visible and detached runtime — a background job on a sibling tab let the uninstall complete and stripped the running work of its plugin MCP.

**[P1] Trust reverify's idle check raced the disconnect.** `SetMCPTrust` checked sharing controllers for active work without holding their turn gates; a turn starting between the check and `UnregisterMCPServerTools`/`DisconnectMCPServer` lost its MCP tools mid-flight.

**Fix — `lockRuntimeTurnGates`** (the suggested helper): snapshots the affected visible and detached tabs under `runtimeRebuildMu`, locks each `turnStartMu` in stable tab-ID order, re-verifies active work under the gates (reading `tab.Ctrl` under `App.mu` so detached runtimes count — `controllerForTab` would return nil for them), and releases in reverse order. `RemovePlugin` gates all runtimes across the uninstall and rebuild; `SetMCPTrust` gates every tab sharing the Host across preflight, restore, and reconnect. A lock-site audit confirms no path acquires `turnStartMu` before `runtimeRebuildMu`, so the multi-gate order introduces no deadlock.

**Regressions** (all three fail against the previous implementation — mutation-tested):
- `TestRemovePluginRejectsBusySiblingRuntime` — busy sibling fails the removal before anything is deleted; plugin, shared client, and registries intact
- `TestRemovePluginRejectsBusyDetachedRuntime` — detached runtimes count in the same sweep
- `TestSetMCPTrustWaitsForSiblingTurnGate` — reverify blocks on a held sibling turn gate (preflight never launches, client stays connected), then fails via the gated re-check once the turn's work is visible; no registry loses tools

**Verified on the new head**: full `TestSetMCPTrust*`/`TestRemovePlugin*` set with `-race -count=3`, desktop `go test ./...`, `go build ./...`, `go vet`, `gofmt` — all clean.

SivanCola:
**Sixth review round**: the snapshot-stability P1 is confirmed and fixed in 3c7cc36 with a unified work-admission barrier.

**[P1] The gate snapshot was not stable.** `lockRuntimeTurnGates` assumed the runtime set could not change under `runtimeRebuildMu`, but `OpenTopicTab` inserts tabs under `App.mu` only and `startTabControllerBuild` attaches controllers from an unguarded goroutine — a runtime appearing after the snapshot could start work mid-uninstall, and during the 30s preflight a late build could attach to the same shared Host (single-instance relaunch, unseen registry). Re-enumerating after acquiring gates would just move the race, as the review noted.

**Fix — `runtimeAdmissionMu`, a work-admission RW barrier.** `lockRuntimeTurnGates` takes the write side before its snapshot; turn admission (`beginTabTurn` through `finishTabTurnStart`) and the async controller build (`buildTabControllerWithContext`, covering the shared-Host attach and swap) hold the read side. The invariant is now **work admission, not tab-set stability**: tabs may still be created mid-hold, but they cannot attach a controller or admit a turn until release, so everything that can possibly be busy is in the snapshot and anything newer stays inert. `SetMCPTrust`'s affected predicate now evaluates Host membership inside the gated snapshot (and the controllers list is computed after the gates), closing the residual pre-built-map race. Deadlock safety: writers acquire the barrier only while holding `runtimeRebuildMu`, read holders never acquire `runtimeRebuildMu`, and `runtimeRebuildMu → runtimeAdmissionMu → turnStartMu → App.mu` is acquired in one direction only.

**Regressions on the real paths** (both fail with the barrier acquisitions removed — mutation-tested):
- `TestRemovePluginBlocksLateTurnAdmission` — a tab added after the snapshot goes through the real `beginTabTurn`; it must stay blocked while the uninstall holds the barrier and only be admitted after the uninstall completed
- `TestSetMCPTrustBlocksLateControllerBuild` — a tab created during the held preflight runs the real `startTabControllerBuild`; the build must not complete until the reverify releases the barrier, then completes with the shared client reconnected

**Verified on the new head**: full `TestSetMCPTrust*`/`TestRemovePlugin*` with `-race -count=3`, desktop `go test ./...`, `go build ./...`, `go vet`, `gofmt` — all clean.

SivanCola:
**Seventh review round**: three P1 defects and one P2 recovery gap were confirmed and fixed in c1ba7da3.

**[P1] Ordinary teardown bypassed a pending turn admission.** `CloseTab`, session/topic deletion, visible-tab pruning, and workspace removal used the read side of `runtimeAdmissionMu`. A turn-start token already holding that same side could still be handed a controller while teardown closed it. `TestCloseTabWaitsForPendingTurnAdmission` failed against the previous head because `CloseTab` returned and closed the controller before the admission was released. Fix: all controller-destroying teardown paths now use the exclusive runtime mutation order `runtimeRebuildMu -> runtimeAdmissionMu -> sessionRemovalMu`, matching MCP/provider lifecycle writers.

**[P1] Stored `uvx --from` locks produced invalid argv.** The package locator index points to the value in `uvx --from package command`; inserting `--offline` there produced `--from --offline package command`. `TestStoredUVXFromLauncherLockKeepsFromValueAdjacent` failed against the previous implementation with that exact argv. Fix: insert the enforcement flag before the separated `--from package` pair while retaining the resolved package in the canonical identity.

**[P1] Revoking project MCP trust did not stop the authorized process.** The connected path removed receipts and updated tool classification only, leaving the shared client and provider-visible tools alive until restart. Fix: while all shared runtime turn gates are held, revocation unregisters the server from every controller and disconnects the shared client immediately.

**[P2] Revocation hid its own recovery action.** An intentional disconnect creates no failed launch attempt, so the configured project MCP fell back to a generic deferred row without **Authorize and connect**. Fix: retain an explicit launch-approval-required Host status after revocation; the same regression verifies the process is gone, every registry is clean, and the re-authorization action remains available.

**Verification on exact head c1ba7da31abdcc89e38b80070dbd0d57000b61d5**:

- root and Desktop `go test -race ./...`
- root and Desktop `go vet ./...` and `go build ./...`
- frontend `pnpm test:all` and production build
- `./scripts/cache-guard.sh` (all scenarios 92-95% tail average, 90% threshold)
- root Darwin/Linux/Windows amd64+arm64 cross-builds and Desktop Windows amd64+arm64 test compilation
- GitHub: 17/17 checks passed, including Linux race, Windows root/Desktop tests and installer build, CodeQL, lint, coverage, and cache impact

No unresolved review threads or external review blockers remain. Keep #6652 and #6657 open until the reporters validate the installable build in their real environments.

PR #5943, opened 2026-07-04
feat(config): support per-model context_window in model_overrides / 支持为单个模型配置独立的上下文窗口

HaoyueQin:
Closes #5915. Closes #4460  .

### Summary

This PR adds a `context_window` field to `ProviderModelOverride`, enabling each model under a shared provider entry to declare its own context window size. The desktop UI (Settings → Models → Access → Configure) now includes an **"Advanced Settings"** toggle in the model picker for per-model configuration.

### Changes (10 files, +167 / −16)

**Data model & config resolution:**
- `internal/config/config.go` — added `ContextWindow int` to `ProviderModelOverride`; `applyModelOverride()` now overrides the entry's `ContextWindow` when the per-model override is `> 0`. Added struct doc comment.
- `internal/config/render.go` — `renderModelOverride()` renders `context_window`; `modelOverrideEmpty()` updated to include `ContextWindow <= 0`; TOML comment updated.

**Desktop backend bridge:**
- `desktop/settings_app.go` — added `ContextWindow int` to `ProviderModelOverrideView`; `providerModelOverridesForView()` and `providerModelOverridesForSave()` handle the new field. Added struct doc comment.

**TypeScript types:**
- `desktop/frontend/src/lib/types.ts` — added optional `contextWindow?: number` to `ProviderModelOverrideView`.

**Desktop UI (Settings → Models → Access → Configure):**
- `desktop/frontend/src/components/SettingsPanel.tsx`:
  - Added an **"Advanced Settings"** / **"退出高级设置"** toggle button next to "Select All" / "Clear" in the model candidate picker.
  - When enabled, each selected model row expands to show a per-model context window input. Default value = the provider's current `context_window` setting. Placeholder shows the provider-level value as reference.
  - The provider-level context window field in the collapsible advanced section below is **hidden** when per-model mode is active, replaced with an informational hint.
  - When exiting advanced mode, the provider-level context window is **restored** to its original value (captured on entry).
  - Per-model values are persisted via `modelOverrides` in `buildModelOverridesForSave()`, which merges per-model context windows with any existing overrides (reasoning protocol, vision, effort).
  - Hover tooltip on the toggle button explains the feature.
  - Imported `ProviderModelOverrideView` type from `../lib/types`.

**Styling:**
- `desktop/frontend/src/styles.css` — added `.provider-model-draft__ctx`, `__ctx-label`, `__ctx-input` rules matching the existing `.provider-model-draft__vision` font size (`var(--font-control-small)`) and design tokens.

[...]

PR #6539, opened 2026-07-15
fix(desktop): prevent model list option overlap when provider has many models / 修复供应商模型较多时设置页面模型列表选项重叠

HaoyueQin:
Fixes #5563. Fixes #5585. Fixes #5785. Fixes #6480. Fixes #6723.

---

## Summary

Fixes model list option overlap in the desktop settings when a provider (e.g., SiliconFlow, Volcengine) returns many models — the CSS grid container for model candidates had an undersized `min-height` per option card and no overflow protection, causing option cards to visually overlap, especially when model names wrapped to multiple lines.

### Root cause

`.provider-model-draft__option` had `min-height: 32px`, which is smaller than the actual minimum content height (~46px: checkbox + model name + 4px gap + vision checkbox row + 10px padding). The grid container `.provider-model-draft__list` used `max-height: 220px` with `overflow: auto`, so with many models the tightly packed rows allowed content to bleed into adjacent cards. Long model names that wrapped to 2+ lines made the overlap worse since the card height grew beyond 32px but had no explicit clipping boundary — the vision checkbox could be partially or fully covered by the next card.

This affected four independent user reports across different providers (SiliconFlow, Volcengine, and others), all with the same underlying cause. A previous attempt ( #5943 only increasing `max-height` from 220px to 280px and adding bottom padding) merely reduced the probability of truncation without fixing the real issue — the cards could still overflow their boundaries.

### Changes

**1 file changed:** `desktop/frontend/src/styles.css` (+3, −2)

**`.provider-model-draft__list`:**
- `padding-right: 2px` → `padding: 0 2px 6px 0` — bottom padding prevents the last grid row from touching the container edge
- Added `grid-auto-rows: min-content` — each grid row sizes to the tallest option card in that row, preventing vertical overlap between rows regardless of how many lines a model name wraps to

**`.provider-model-draft__option`:**
- Removed `min-height: 32px` — content now determines height naturally via flexbox; the card can no longer be shorter than its actual content
- Added `overflow: hidden` — prevents text or checkbox content from bleeding into adjacent cards; long model names are handled by the existing `overflow-wrap: anywhere` on the inner `span`

### Scope

These CSS classes are shared by both `ProviderModelDraftPicker` (the "Refresh Models" draft panel shown after clicking "Refresh Models" in the provider access page) and `ProviderEditorModelPicker` (the inline model picker inside `ProviderEditor`), so both views benefit from the fix without any JSX changes.

### Closes

[...]

PR #6677, opened 2026-07-19
feat(config): support per-model context windows / 支持逐模型上下文窗口

### LINKED ISSUE #4460: [Feature]: 配置上下文窗口大小
### Version line

这种要怎么为每个模型配置不同的上下文大小

Image

### Proposed solution

_No response_

### LINKED ISSUE #5915: 支持为单个模型配置独立的上下文大小（context size）
## Problem
目前 Reasonix 的上下文大小是按"模型服务"（provider）维度配置的。但在实际使用中，一个 API 端点/中转站下往往会有多个不同厂商的模型，而这些模型的上下文窗口大小各不相同。

例如通过一个统一的 API 端口接入了 DeepSeek、Claude、Gemini 等多个厂商的模型，它们的 context size 分别是 128K、200K、1M 等。目前只能为每个模型服务单独配置上下文大小，导致需要创建多个"模型服务"条目来区分，配置非常繁琐。

## Proposed solution
支持在模型级别（而非仅服务级别）配置 context size / context window，例如：

```toml
[[providers]]
name = "shared-api"
api_key = "xxx"

  [[providers.models]]
  name = "deepseek-v4"
  context_size = 128000

  [[providers.models]]
  name = "claude-sonnet"
  context_size = 200000
```

或者在 GUI 的模型配置中增加独立的 context size 字段。

## Use case
- 统一 API 端点下包含大量不同上下文大小的模型
- 减少重复的"模型服务"配置项
- 更精确地控制 auto-compact 触发阈值和上下文用量统计

## Alternative considered
目前只能通过创建多个 provider 来区分不同 context size 的模型，但同一端点下的模型共享 endpoint，这种 workaround 不够优雅。

SivanCola:
## Summary

- add `model_overrides.<model>.context_window` as an optional per-model override for shared provider endpoints
- preserve the provider-level context window as the fallback when the override is missing or zero
- expose per-model token inputs in Desktop settings while preserving reasoning, effort, and vision overrides
- clarify that the unit is tokens, distinguish context limits from maximum output tokens, and warn non-blockingly below 16,384 tokens
- document the TOML shape and behavior in English and Chinese

Fixes #5915.
Fixes #4460.

## Why

A single OpenAI-compatible endpoint can expose models with very different context limits. Reasonix previously had only one provider-wide `context_window`, so users either accepted inaccurate compaction thresholds or duplicated provider entries that shared the same endpoint and credentials.

## Implementation

- apply a positive per-model context window during `ResolveModel()` through the existing model-override path
- render and normalize the additive TOML field without changing existing zero-value behavior
- round-trip the field through the Desktop Go/TypeScript bridge with an optional JSON property
- merge UI edits into existing model overrides instead of replacing reasoning, effort, or vision settings
- provide localized guidance for `128000` versus provider-documented `131072` values

## Compatibility

| Field / contract | Old-data or old-version behavior | Conclusion |
| --- | --- | --- |
| `ProviderModelOverride.ContextWindow` | Missing fields decode to `0`; `0` inherits the provider-wide value. | Backward-safe. |
| `model_overrides.<model>.context_window` | Older TOML readers accept the additive nested key and ignore it, then use the provider-level value. | Cross-version readable with a safe fallback. |
| Desktop `contextWindow` JSON property | The Go response omits zero values and the TypeScript property is optional; older payloads remain valid. | Backward-safe. |
| Existing reasoning / effort / vision overrides | Desktop merge logic preserves these fields while changing context limits. | No override data loss. |

## Cache impact

No provider-visible system-prompt, tool-schema, ordering, or request-serialization bytes change. An explicitly smaller per-model value can intentionally trigger compaction earlier and therefore reduce cache reuse; unset values preserve current behavior, and the UI warns for values below 16,384 tokens.

### Kept and adapted

- PR #5943 by @HaoyueQin established the core per-model `context_window` mechanism. This PR reimplements that mechanism on the latest `main-v2`, preserves it in the commit co-author trailer, and adds current-architecture UI merging, validation, compatibility coverage, documentation, and cache guidance.

### Reviewed but not adopted

- PR #6439 by @ashishexee proposes converting custom-provider `context_window = 0` into a 128K default. That changes the established meaning of zero. This PR keeps `0` as “automatic compaction disabled” and instead makes that behavior explicit while providing per-model positive overrides.

## Verification

- `./scripts/cache-guard.sh`
- `go vet ./...`
- `go test ./...`
- `pnpm --dir desktop/frontend exec tsx src/__tests__/provider-model-refresh.test.ts` — 22 passed
- Codex in-app browser: verified localized token guidance, the 8,192-token warning, and no horizontal overflow

Cache-impact: low - Only explicit per-model values change compaction timing; provider-visible prompt bytes remain unchanged.
Cache-guard: ./scripts/cache-guard.sh
System-prompt-review: reviewed - no system-prompt construction changes.


SivanCola:
Thanks for this well-scoped PR. The per-model override is the right mechanism, and I verified it end to end: `applyModelOverride` runs on the resolved copy in `ResolveModel`, so the override reaches every consumer (CLI boot, desktop tabs via `boot.Build`, subagents, planner, guardian, ACP) exactly where the provider-level value did. The zero-value fallback chain (missing/<=0 per-model → inherit provider value; provider-level 0 → compaction disabled) matches the current code, where `agent.Options.ContextWindow <= 0` skips compaction, and the additive TOML/JSON round-trip is backward-safe — including the legacy-decoder regression test in `render_test.go`.

One coordination note: this overlaps #6439 (`fix/ctx-window-default`), which takes the opposite position on `context_window = 0` — it backfills custom providers to 128000 at load time and clamps the desktop save to >= 128000, removing the ability to disable compaction. We've decided to keep the current "0 disables compaction" semantics that this PR documents, so #6439 will be reworked rather than merged as-is. The two PRs also edit the same three locale keys and the same `ProviderEditor` region in `SettingsPanel.tsx`, so whichever lands second will need a rebase.

Two minor, non-blocking notes:
- The new UI hint says 0 disables compaction for the provider, but official presets get `context_window` backfilled to a positive value at load (`backfillOfficialContextWindow`), so 0 effectively cannot disable compaction there. Worth a small wording tweak, or fine to leave since it only affects built-in presets.
- Not being able to express "disable compaction for one model" (per-model 0 = inherit) is a reasonable design choice and it is documented — just confirming it is intentional.

Also a heads-up that #6539 touches adjacent `.provider-model-draft__*` CSS. There is no hard conflict, but the picker cards get taller with the new per-model input, so a quick visual pass over the model list will be worthwhile after both land.

No blockers from my side once this is marked ready for review and CI stays green.

PR #6716, opened 2026-07-20
Align Remote SSH config and credentials with OpenSSH / 对齐远程 SSH 配置与凭据处理

### LINKED ISSUE #6711: [Bug]: 远程连接SSH的一些问题和建议
### Version line

### Exact version

1.17.16

### What happened?

1、客户端添加的SSH主机无法删除，点击删除按钮没有任何反应
2、连接失败，提示：

```bash
ssh: handshake failed: remote: password required but no prompt available
```
但是我的主机已经配置了密钥登录，且选择了芈月文件，提示却是要求密码。

3、请求支持密码登录，而不是强制使用密钥，做成可选吧

### Steps to reproduce

手动新增和导入都无法登录：

```bash
ssh: handshake failed: remote: password required but no prompt available
```

且添加的主机均无法删除，点击删除按钮无反应。

### OS / platform

macOS 14.x

### Relevant logs or output

```shell

```

SivanCola:
## Summary

- keep imported SSH aliases live and resolve only the selected connection target through the installed OpenSSH `ssh -G` configuration stack, including the normal user and system config files
- surface installed OpenSSH configuration errors instead of silently downgrading, while retaining the embedded-parser fallback when the `ssh` executable is unavailable
- match OpenSSH authentication semantics for ordered identities, `IdentityFile none`, and `IdentitiesOnly=yes` with configured public identities backed by `ssh-agent`
- give every interactive password/passphrase request an opaque prompt ID so an older desktop dialog cannot clear or resolve a later prompt with identical metadata
- preserve Reasonix-specific workspace, bootstrap, credential-reference, and forwarding settings when an existing alias is imported again from the desktop or CLI
- discover importable aliases without running `ssh -G` or `Match exec` once per host

## Root cause

The previous implementation used `-F ~/.ssh/config`, which excluded the system client configuration, and treated every `ssh -G` failure as a reason to use the partial parser. Alias discovery also eagerly evaluated every target. Authentication discarded `IdentityFile none` and interpreted `IdentitiesOnly` as disabling the agent entirely, although OpenSSH permits a configured public identity to select its private key from the agent. Finally, desktop prompts were identified by reusable display metadata and re-import replacement rebuilt entries without two Reasonix-owned fields.

## Backward compatibility

| Field or behavior | Existing data behavior | Conclusion |
| --- | --- | --- |
| `remote.hosts[].host` with `use_ssh_config = true` | The persisted `host` remains the lookup key. New imports keep the live alias; older resolved-host snapshots keep their saved target and explicit values instead of being redirected through `name`. | Compatible |
| Re-importing an existing alias | Refreshes the OpenSSH lookup fields while retaining workspace, `serve_install`, credential references, and forwards. | Prevents data loss |
| OpenSSH executable unavailable | Falls back to the existing embedded parser. Installed OpenSSH timeouts and invalid configurations are now surfaced. | Compatible fallback with clearer errors |
| `secretPrompt` status metadata | Adds an internal opaque `promptId`; plaintext remains absent, and the optional identity value remains only a key basename. | Additive desktop contract |

## Security

- prompt events contain only an opaque ID, host, credential kind, and optional key basename; full local identity paths and plaintext secrets are not emitted
- plaintext secrets cross a one-shot in-memory channel and are never persisted in host TOML or frontend state
- stale prompt IDs are rejected by the backend, and stale dialogs cannot clear a newer prompt in the frontend store
- `IdentitiesOnly=yes` offers only agent signers that match configured identity public keys; unrelated agent identities remain filtered
- CLI and desktop credential writes retain the existing rollback-capable lifecycle; no dependency, telemetry, sandbox, or provider-visible prompt changes are introduced

## Verification

- `env -u DEEPSEEK_API_KEY go test ./...`
- `go test -race ./internal/remote ./internal/config ./internal/cli`
- `cd desktop && go test -v -timeout 5m ./...`
- root and desktop `go vet ./...`
- `pnpm --dir desktop/frontend test:remote`
- `pnpm --dir desktop/frontend test:typecheck`
- Windows amd64/arm64 compilation of root Remote SSH, CLI, desktop, and update-helper packages
- real local OpenSSH `ssh -G` evaluation of a `Match exec` config
- integration authentication through a configured public identity whose matching private key exists only in `ssh-agent`
- `scripts/check-cache-impact.sh`: no cache-sensitive prompt/tool files changed

## Cache impact

Cache-impact: none - Remote SSH resolution, authentication, credential lifecycle, tests, and desktop UI do not change provider-visible prompts, tool schemas, or request serialization.
Cache-guard: changed paths are outside the cache-sensitive prompt and tool surfaces checked by `scripts/check-cache-impact.sh`.
Cache-evidence: the updated diff remains limited to Remote SSH, credential lifecycle, tests, and desktop UI contracts.
System-prompt-review: N/A

## Known boundary

The transport remains the Go SSH client. Advanced system-OpenSSH-only features such as arbitrary `ProxyCommand`, PKCS#11 or FIDO providers, GSSAPI, and ControlMaster are not claimed as equivalent in this PR.

Fixes #6711


SivanCola:
Addressed the follow-up review findings in `7960c71be`:

- selected aliases now use the normal OpenSSH `ssh -G` user + system config stack, while installed-client errors are surfaced and alias discovery stays lazy
- `IdentityFile none` and `IdentitiesOnly=yes` now follow OpenSSH semantics, including configured public identities backed only by `ssh-agent`
- each desktop password/passphrase request has a backend-validated opaque prompt ID, preventing stale dialogs from resolving or clearing a later prompt
- desktop and CLI re-imports retain workspace, bootstrap policy, credential references, and forwards

Added focused regression coverage for each case. All current CI checks are green, including macOS/Linux/Windows tests, race, desktop and desktop-windows, lint, coverage, CodeQL, cache-impact, and govulncheck. The PR is ready for re-review.

PR #6765, opened 2026-07-21
fix(acp): defer available_commands_update after session/new response

ghost:
## Summary

- Clients like agentic.nvim silently drop the available_commands_update notification when it arrives before the session/new response completes, causing the client's tool palette to remain empty until the next session update.

## Verification

- agentic.nvim tool palette populates correctly after session/new
- go test ./internal/acp/ ./internal/control/
- go vet ./...

## Cache impact

Cache-impact: `none`
Cache-guard: `none`
System-prompt-review: `none`

### DIFF
--- a/internal/acp/service.go
+++ b/internal/acp/service.go
@@ -651,7 +651,11 @@ func (s *service) sessionNew(ctx context.Context, raw json.RawMessage) (any, err
 	s.mu.Lock()
 	s.sessions[id] = sess
 	s.mu.Unlock()
-	s.sendAvailableCommands(sess)
+	// Fire in a new goroutine so the session/new response reaches the client
+	// first. Clients like agentic.nvim register their session subscriber in the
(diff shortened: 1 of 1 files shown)


SivanCola:
Thank you @ssfdust for reporting the agentic.nvim behavior and auditing the additional session lifecycle paths.

This work was consolidated into #6803, which has now been merged into `main-v2` as `14e97c9`. The integration keeps the broader session-lifecycle coverage and applies deterministic ordering to `session/new`, `session/load`, and `session/resume`. It uses a transport-owned post-response hook instead of goroutine deferral because scheduler timing cannot guarantee wire order, while preserving the existing config-rebuild notification order.

Your contribution is preserved in the merged commit with commit-level `Co-authored-by` attribution. Closing this PR as superseded by #6803. Thank you for the contribution.

PR #6859, opened 2026-07-23
feat(cli): add redacted machine interfaces / 增加脱敏机器接口

mchenziyi:
# Reasonix PR: redacted machine interfaces

## Summary

This PR adds stable, scriptable, read-only, and redacted CLI interfaces for terminals, CI, monitoring dashboards, desktop clients, and automation:

- `reasonix session list --json [--dir SESSION_DIR]`
- `reasonix session show|status <machine-session-id> --json [--dir SESSION_DIR]`
- `reasonix session recovery [<machine-session-id>] --json [--dir SESSION_DIR]`
- `reasonix hook list|status --json [--project-root PATH] [--home-dir PATH]`
- `reasonix task list|show [<task-id>] --json [--dir SESSION_DIR] [--session MACHINE_SESSION_ID]`
- `reasonix run --events-jsonl <task>`

All new JSON responses contain `schema_version: 1`. Lists use deterministic ordering and encode empty collections as `[]`. Argument errors exit with status 2; state/query errors exit with status 1 and return stable machine error codes such as `invalid_argument`, `session_not_found`, `task_not_found`, `task_ambiguous`, and `recovery_state_unavailable`.

## Redaction boundary

- Session responses contain deterministic opaque IDs, UTC timestamps, scope, turn counts, lifecycle state, and recovery state. They do not expose transcript content, preview/title, model-bearing file names, absolute paths, PID, or hostname.
- Hook responses contain event, matcher, scope, trust/activity state, and source statistics. They do not expose commands, context files, settings paths, or raw parse input.
- Task responses contain task/subagent ID, opaque parent session ID, kind, lifecycle timestamps, and artifact completeness. They do not expose labels, output, paths, arguments, or mutation receipt content.
- Recovery responses contain state, task/failure/pending/in-flight counts, and timestamps. They do not expose tool, subject, arguments, failure excerpts, or proposals.
- `--events-jsonl` is a dedicated content-free contract. It omits prompts, tool arguments/results, reasoning, approval text, compaction summaries, and guardian rationale; the final record is `kind: "run_done"`.

## Lifecycle semantics

- A live background task is `running` only while its parent session owns a live lease.
- A persisted `running` job is normalized to `interrupted` only when the current process holds the parent transcript write lease and no live in-process manager owns it; observers without that lease defer the record without rewriting or publishing a tombstone.
- `finished_at` remains empty while a task is running.
- `artifact_complete=true` is emitted only for a terminal task whose persisted artifact exists and completed without an artifact error.
- Runtime model, effort, skill, and work-mode switches reject foreground work, pending prompts, or background jobs before building a replacement controller.
- `--events-jsonl` cannot be combined with `--output-format`; `stream-json` remains the richer, content-bearing UI transport.

## Compatibility

| Contract | Old-data / old-reader behavior | Conclusion |
| --- | --- | --- |
| Existing session/job JSON fields | No fields were removed or retyped; old records still decode with zero values for absent fields. | Backward-safe |
| Optional job `ownerId` metadata | Old records omit it and take the conservative cold-recovery path; old readers ignore the unknown field. It is excluded from machine projections. | Backward-safe |
| New `interrupted` job status | Older readers already treat every value other than `running` as terminal, so they do not resurrect or cancel it. | Backward-safe |
| Legacy persisted `running` metadata | A lease-owning runtime normalizes it to `interrupted`, sets a terminal timestamp, and clears artifact completeness; non-owners leave it untouched. | Automatically repaired |
| Machine session IDs | Only the unreleased schema-v1 machine projection changed; persisted branch IDs and existing resume/doctor interfaces remain unchanged. | No existing-data migration |
| Cross-version coexistence | New metadata remains valid JSON for older versions; an unknown terminal status fails safe instead of consuming running capacity. | Backward-safe |

Within schema version 1, consumers must ignore unknown fields; fields will not be removed or retyped.

## Implementation notes

- Session queries reuse `agent.ListSessionOrder`, branch metadata, leases, and recovery sidecars.
- Hook queries reuse read-only `hook.Inspect`; they never execute hooks or modify trust.
- Task queries reuse jobs artifact metadata and SubagentStore metadata. Start metadata is content-free; a private owner token protects same-process rebuilds, while the parent transcript lease authorizes cross-process cold recovery. Neither ownership detail is included in machine projections.
- Recovery queries reuse `recovery.LoadSnapshot` and the session in-flight marker.
- No provider-visible prompt, tool schema, request serialization, or cache prefix changes are introduced.

The complete schema and CLI usage contract is documented in `docs/CLI.md` and `docs/CLI.zh-CN.md`.

## Verification

- `go test ./...` (all deterministic packages passed; the invalid local `DEEPSEEK_API_KEY` probe passed with the variable unset)
- `go test -race ./internal/jobs ./internal/agent ./internal/bot ./internal/remote/workbench/runtime`
- `go test -race ./internal/cli ./internal/serve`
- `go vet ./internal/jobs ./internal/cli ./internal/serve ./internal/i18n`
- `./scripts/cache-guard.sh`
- real binary smoke checks returned valid schema-v1 JSON for empty session/task state and redacted hook status

Provider-specific task execution was not run with a valid live credential, so that optional end-to-end probe is not claimed as complete.

Cache-impact: none - `internal/boot/boot.go` only injects a runtime-local jobs ownership callback; provider-visible prompt, tool schema order, request serialization, and cache prefix bytes are unchanged.
Cache-guard: `./scripts/cache-guard.sh` passed all cache scenarios; targeted, race, and full Go test suites passed.
System-prompt-review: SivanCola reviewed the boot diff and confirmed it does not alter system prompt assembly or provider-visible content.


internal/bot/gateway.go
```diff
@@ -2123,10 +2211,18 @@ func (gw *BotGateway) getOrCreateSession(ctx context.Context, key string, msg In
 		gw.logger.Error("build controller failed", "err", err)
 		return nil
 	}
+	leases := control.NewSessionLeaseKeeper()
```
SivanCola, line 2214: `boot.Build` creates the shared subagent store and runs `CleanupStaleRunning` before this path attempts `Rebind(profile.sessionPath)`. A failed attach can therefore rewrite live `running` metadata owned by another process. Moving only the target lease acquisition earlier is not sufficient because cleanup scans the entire `sessionDir/subagents` directory; gate each entry by proven `ParentSession` ownership, or remove global cleanup from per-controller Build. Please add a cross-process regression covering a live subagent owned by another session.

internal/agent/session_lease.go
```diff
@@ -194,6 +194,18 @@ func SessionLeaseHeldByOtherRuntime(path string) bool {
...
 
+// SessionLeaseHeldByCurrentRuntime reports whether this process currently owns
+// path's session lease. It is intentionally a registry-only query: callers use
+// it to authorize ownership-sensitive repairs without probing or disturbing the
+// cross-process lock.
+func SessionLeaseHeldByCurrentRuntime(path string) bool {
+	if strings.TrimSpace(path) == "" {
+		return false
+	}
+	_, ok := sessionLeaseOwners.Load(canonicalSessionSavePath(path))
```
SivanCola: `SessionLeaseHeldByCurrentRuntime` is registry-only, while `Release` unlocks the OS lease before deleting the registry entry. During that handoff window another process can acquire the real lease while this process still authorizes `Manager.loadSessionArtifacts` to rewrite `running` metadata. Revoke the current generation before unlocking, or pass a generation-bound lease proof to the manager, and add a controlled handoff regression.

internal/cli/hook_machine.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	inspection := hook.Inspect(hook.LoadOptions{
+		ProjectRoot: options.projectRoot,
+		HomeDir:     options.homeDir,
+		Trusted:     hook.IsTrusted(options.projectRoot, options.homeDir),
+	})
+	if operation == "list" {
+		items := make([]machineHook, 0, len(inspection.Entries))
+		for _, entry := range inspection.Entries {
+			status := "active"
```
SivanCola: `Inspect` intentionally includes missing commands and unknown events, while invalid regex matchers on tool events never fire. This path nevertheless reports every trusted entry as `active`. Return a stable `invalid`/`inactive` status or diagnostic issue codes that match `Load`/`MatchesTool`, and cover an empty command, unknown event, and invalid `PreToolUse` matcher. Non-tool events ignore matchers and should remain executable.

SivanCola (changes requested):
Thanks—this is directionally useful, but three ownership/contract issues still need to be addressed before merge:

- scope stale-subagent cleanup to proven parent-session ownership;
- make repair authorization generation-bound across lease handoff;
- report hook executability truthfully through the machine interface.

Please add deterministic cross-process, lease-handoff, and invalid-hook regressions. Once these are addressed and CI remains green, this should be ready for another review.

SivanCola (approved):
Re-reviewed the latest head after the requested ownership and machine-contract fixes. The full local, race, desktop, cache, latest-base integration, and GitHub CI evidence is green. No remaining blocking findings from this review.

internal/bot/gateway.go
SivanCola: Addressed in 47f76e6e. `CleanupStaleRunning` now groups records by proven `ParentSession`, acquires each parent transcript lease before rewriting, re-reads metadata while holding that lease, and skips records without parent proof. `TestSubagentStoreCleanupStaleRunningSkipsForeignLiveParent` exercises a real helper process holding another session lease and verifies the child remains `running` until release. Full, race, desktop, latest-base integration, and all 16 GitHub checks passed; PR #6859 has merged as 16d71aa4.

internal/agent/session_lease.go
SivanCola: Addressed in 47f76e6e. Session lease reservations and active ownership generations are now tracked separately, and `Release` revokes the exact active generation with `CompareAndDelete` before unlocking the OS lease. That closes the handoff window without allowing a stale release to evict a successor. `TestSessionLeaseReleaseRevokesRepairAuthorizationBeforeUnlock` covers the controlled ordering, alongside stale-generation regressions. Full and race suites plus all 16 GitHub checks passed; PR #6859 has merged as 16d71aa4.

internal/cli/hook_machine.go
SivanCola: Addressed in 47f76e6e and hardened further in 7335fea5. The machine projection now reports `invalid` for unknown events, empty or unusable command/context sources, invalid tool-event matchers, and non-regular/unreadable plugin context files, while non-tool event matchers remain ignored and executable. CLI and capability diagnostics share `hook.ContextFileUsable`, with focused regressions for each contract edge. Full, race, desktop, cache, and all 16 GitHub checks passed; PR #6859 has merged as 16d71aa4.

PR #6892, opened 2026-07-24
Preserve MCP config scopes and precedence / 统一 MCP 配置作用域与优先级

SivanCola:
## Summary

- Treat MCP servers declared by project `reasonix.toml` and `.mcp.json` as trusted configuration with no separate launch confirmation.
- Persist explicit Desktop, CLI, TUI, and cc-switch installs to the user-global `config.toml`.
- Keep project MCP edits in their owning project file and expose project/global/plugin-managed sources consistently in Desktop and CLI.
- Resolve duplicate names deterministically: project `reasonix.toml` > project `.mcp.json` > global `config.toml` > legacy/plugin supplements.
- Make edit, authentication cleanup, enable/disable, and removal operate on the effective declaration's exact source and workspace.
- Reveal and cache-first register the next lower-priority declaration after removing an override.

This supersedes #6889 with the complete source-aware persistence and duplicate-name behavior.

## Root cause

Project scope was coupled to launch trust, while runtime config merging and persistence were handled by different paths. The merged entry could be correct for execution but lose its owning-file context during Desktop or CLI mutations. Some CLI and controller paths also used the process working directory instead of the session workspace.

## User impact

Installing an MCP server is now a one-step global operation that works across projects. A project can override the global server simply by declaring the same name in project config. Project declarations remain project-local, require no extra confirmation, and are available to the main agent, Planner, and eligible sub-agents through the same effective runtime configuration.

Removing a project override does not delete the global installation. The next declaration becomes effective immediately and is restored from cache or as an on-demand connection surface without eagerly starting a process.

## Cache impact

Project MCP tools can surface earlier because configured project servers no longer wait for launch confirmation. Tool names, schemas, and registration ordering are unchanged. Duplicate replacement preserves stable list position, and fallback restoration remains cache-first.

`./scripts/cache-guard.sh` passes all scenarios with tail averages of 92–95 against the threshold of 90.

## Backward compatibility

| --- | --- | --- |
| Global `config.toml` | Existing MCP entries load unchanged; explicit installs now consistently target this file. | Compatible |
| Project `reasonix.toml` | Existing entries load unchanged and edits stay in the same file. | Compatible |
| Project `.mcp.json` | Existing Claude-compatible entries load unchanged and unknown fields are preserved on edits. | Compatible |
| Duplicate MCP names | No data is migrated or deleted; the effective order becomes project `reasonix.toml`, project `.mcp.json`, then global config. Older versions can still read every file but may resolve the old order. | Safe intentional behavior change |
| Activation state | The existing source/workspace identity schema is retained; CLI now supplies the correct workspace fingerprint. | Compatible |
| Desktop `ServerView` JSON | Source fields remain optional and older payloads fall back to the configured/global presentation. | Compatible |
| Launch grants | Existing grant state remains readable; ordinary configured project servers no longer require it. Explicitly gated specs keep the compatibility path. | Compatible |
| Older Wails clients | `AuthorizeAndConnectMCPServer` remains available and reconnects the effective entry. | Compatible |

## Verification

- `go test ./...`
- `go vet ./...`
- `cd desktop/frontend && pnpm exec tsx src/__tests__/capabilities-panel-actions.test.ts`
- `cd desktop/frontend && pnpm run typecheck`
- `./scripts/cache-guard.sh`
- `git diff --check`

Cache-impact: medium - Project-configured MCP tools can surface earlier; tool names, schemas, and stable ordering remain unchanged.
Cache-guard: ./scripts/cache-guard.sh passes all scenarios with tail averages of 92-95 against the threshold of 90.
System-prompt-review: MCP trust and provenance changes were reviewed; no system-prompt text or provider-visible schema bytes changed.


SivanCola:
Superseded by #6889.

All changes from this duplicate PR were preserved in #6889 at 58b8da77, including project/global MCP trust and presentation, source-aware persistence, workspace activation, duplicate-name precedence, fallback restoration, tests, and documentation. The commit history and authorship are identical, so attribution is preserved there.

Closing this duplicate to keep review and CI on the original PR.

PR #6905, opened 2026-07-24
fix(session): prevent data loss with fsync, file locking, and lock-ti…

### LINKED ISSUE #6873: [Bug]: 吞聊天记录
### Version line

### Exact version

1.17.19

### What happened?

有时候一个项目并行几个对话（并行的时候可能是调用量大？卡顿非常严重）有的聊天记录会被吞掉 重启软件也没有恢复 

### Steps to reproduce

高负载场景并行多个对话

### OS / platform

Windows 11

### Relevant logs or output

```shell

```

myipanta:
…meout recovery / 修复会话数据丢失：fsync、文件锁、锁超时恢复

- Enable fsync for event log appends so crash between append and checkpoint no longer loses the last saved messages. Fixes #6873.
- Add filelock-protected updateSessionDisplays to prevent concurrent read-modify-write races on .display.json. Fixes #6873.
- When SaveSnapshot cannot acquire the cross-process file lock, write a recovery branch instead of silently discarding unsaved state. Previously this caused legacy session data loss on session switch (#6770) and high-load parallel tab data loss (#6873).

Closes: #6873 #6770

## Summary

-

## Verification

-

## Cache impact

Cache-impact: TODO
Cache-guard: TODO
System-prompt-review: N/A

### DIFF
--- a/desktop/sessions.go
+++ b/desktop/sessions.go
@@ -1175,17 +1175,43 @@ func saveOrRemoveSessionDisplays(dir string, m sessionDisplayMap) error {
 	return saveSessionDisplays(dir, m)
 }
 
+// updateSessionDisplays locks the .display.json sidecar, calls mutate on the
+// loaded map, and writes back when mutate returns true. The lock prevents
+// concurrent read-modify-write races across goroutines (#6873).
+func updateSessionDisplays(dir string, mutate func(sessionDisplayMap) bool) error {
+	if strings.TrimSpace(dir) == "" {
+		return errors.New("display directory is empty")
+	}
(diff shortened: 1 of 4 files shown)


SivanCola:
Thank you @myipanta. #8053 records and credits the fsync, display-locking, and save-timeout recovery work from this PR. Current main-v2 already contains the broader production implementations: flushed WAL append/replace paths, file-locked display updates, and visible recovery branches on save/shutdown conflict. The integration therefore avoids duplicating those changes and adapts only the remaining current sidecar durability gap.

PR #6908, opened 2026-07-24
fix(provider): gate current DeepSeek images with per-model opt-in / DeepSeek 当前模型默认禁图并支持按模型放行

### LINKED ISSUE #6682: [Bug]: Sending images fails with HTTP 400: unknown variant 'image_url', expected 'text'
### Version line

### Exact version

v1.17.15

### What happened?

When attempting to send an image (by pasting or referencing a local file) through Reasonix, the request fails with an HTTP 400 error. The error message indicates that the JSON payload sent to the provider (DeepSeek) is malformed and cannot be deserialized.

### Steps to reproduce

1. Start a new conversation.
2. Attempt to send an image as part of a message (e.g., paste an image in the TUI or reference a local image file).
3. Observe the error.

Image

### OS / platform

Windows 11

### Relevant logs or output

```shell

```

myipanta:
Fixes #6682.

## Summary

- Keep current official DeepSeek models text-only by default. Provider-wide `vision = true` is ignored on `*.deepseek.com`, preventing current V4 requests from sending unsupported `image_url` content parts.
- Preserve a forward-compatible escape hatch: a concrete model explicitly enabled through `vision_models` or `model_overrides.<model>.vision = true` can send image parts on the official endpoint.
- Carry that model-scoped capability provenance through provider assembly, so the provider can distinguish a deliberate opt-in from stale provider-wide configuration.
- Preserve explicit image support for custom DeepSeek-compatible gateways; `reasoning_protocol = "deepseek"` alone does not disable vision.
- Remove the original text fallback that could copy resolved image data URLs and large base64 payloads into message content.

## Official API evidence

- The current [Chat Completion schema](https://api-docs.deepseek.com/api/create-chat-completion/) defines user and tool message `content` as strings and lists only `deepseek-v4-flash` and `deepseek-v4-pro`.
- The official [GitHub Copilot integration guide](https://api-docs.deepseek.com/quick_start/agent_integrations/github_copilot/) states that DeepSeek V4 is text-only and delegates screenshots to a separate vision proxy model.
- The official [change log](https://api-docs.deepseek.com/updates/) shows that new model generations continue to use the same `https://api.deepseek.com` base URL. Capability therefore cannot be permanently inferred from the hostname alone.

## Verification

- Focused config, OpenAI provider, and boot assembly regressions passed.
- `env -u DEEPSEEK_API_KEY go test ./...` passed on the merged tree.
- `./scripts/cache-guard.sh` passed; every case remained above the 90% threshold.
- `cd desktop && go test . -run "^(TestMetaImageInputCapability|TestMetaForTabImageInputCapability|TestMetaForTabReportsImageInputCapability)$" -count=1` passed.
- `pnpm --dir desktop/frontend exec tsx src/__tests__/composer-image-capability.test.tsx` passed: 15/15.
- Merged and verified against `main-v2` at `0ac122a3a2319ec743335fb012e4bdf896408b9b`.

The configured live DeepSeek credential was unavailable, so verification used deterministic request-shape tests instead of a paid live call.

## Cache impact

- Normal text-only traffic is byte-stable. Current official DeepSeek requests with image attachments serialize identically to the same text-only request.
- Only an explicitly enabled model-scoped vision capability adds image parts; this is intentional request content, not dynamic prefix metadata.
- No system prompt, tool schema, memory prefix, compaction behavior, or tool ordering changed.

## Backward compatibility

| Field / behavior | Behavior after this change | Conclusion |
| --- | --- | --- |
| Existing provider-wide `vision = true` on an official DeepSeek endpoint | Still loads, but current official models remain text-only and attachment refs remain available to OCR or vision MCP tools. | Safe automatic repair; no migration required. |
| Existing `vision_models` or model override | A selected model explicitly marked as vision-capable is now honored even on the official endpoint. | Forward-compatible opt-in using existing fields. |
| Custom DeepSeek-compatible gateway | Explicit vision support remains enabled. | Preserved. |
| Config, session, and wire persistence formats | No fields or enum values changed. The new provenance flag is internal and not persisted. | Fully compatible. |

Cache-impact: none - Normal text-only traffic is unchanged; only an explicitly enabled model-scoped vision capability adds image parts.
Cache-guard: ./scripts/cache-guard.sh passed; provider request tests compare default text-only bytes and explicit image-part serialization.
System-prompt-review: SivanCola reviewed the boot assembly diff and confirmed that no system prompt bytes, tool schemas, memory prefix, compaction behavior, or ordering changed.


SivanCola:
Maintainer follow-up is now complete at `10019010` (behavior commit `893879c5`).

Final behavior:
- current official DeepSeek models remain text-only by default, matching the official Chat Completion schema and V4 integration documentation;
- stale provider-wide `vision=true` is still ignored on official endpoints;
- future models explicitly enabled through `vision_models` or `model_overrides.<model>.vision=true` can use image parts without changing the hostname;
- custom DeepSeek-compatible gateways retain explicit vision support;
- attachment paths remain available to OCR or vision MCP tools.

Verified after merging `main-v2` at `0ac122a3`:
- focused config, provider, and boot tests passed;
- `env -u DEEPSEEK_API_KEY go test ./...` passed;
- `./scripts/cache-guard.sh` passed;
- focused Desktop tests passed;
- frontend composer image tests passed (15/15).

The live DeepSeek credential was unavailable, so provider behavior was verified with deterministic serialized-request tests.

PR #6909, opened 2026-07-24
fix(desktop): clear stale session data when reusing blank tab / 修复新建会话时展示旧工具调用记录的问题

xiaoshuai1024:
## Summary

Reusing an existing blank tab could leave tool-fold cards from the previous
frontend session visible. A late history request for the same tab ID could then
restore those orphaned cards after the backend had already selected a blank
session.

The same asynchronous boundary also allowed a tab-bar click to be overtaken by
an older blank/topic/session navigation completion. That could leave the
frontend on one tab while the backend considered another tab active.

## Root Cause

`ensureBlankTab` used `isNewTab` as the hydration reset flag. When
`EnsureBlankTab` reused an ID that was already present in frontend state,
`isNewTab` was false, so the loader kept the old transcript and could join or
accept an in-flight history request from the prior session.

Navigation and tab-switch queues also tracked last-click ownership separately.
Controller helpers advanced their navigation epoch only when execution began,
so a click waiting in either queue did not invalidate an already-running
operation. Late metadata, hydration, and backend focus changes could therefore
apply after a newer user intent.

## Fix

- Remove the redundant `NewSessionForTab` call.
- Invalidate checkpoint and session hydration work when a blank tab is opened.
- Force reset hydration using the authoritative `sessionPath` returned in
  `TabMeta`, matching blank-surface behavior.
- Capture one shared navigation intent at enqueue time and propagate it through
  tab, topic, blank, saved-session, channel-session, and Delivery navigation.
- Reject stale UI, hydration, error, and tab-metadata updates.
- If a stale backend navigation already activated its abandoned result,
  reassert the currently visible tab so frontend and backend focus converge.
- Re-check the visible tab after every asynchronous reassertion so a third
  navigation cannot be overwritten by an older repair.
- Add deterministic regressions that force old history and backend navigation
  calls to finish after newer tab intents.

## Compatibility

| --- | --- | --- |
| Persisted tabs and sessions | Formats and read/write paths are unchanged. | Compatible |
| Wails JSON and frontend APIs | No bound method or payload shape changes. Optional sequence arguments are frontend-local. | Compatible |
| Configuration and identifiers | No config, enum, path, topic ID, or session ID changes. | Compatible |

## Verification

PR head `e424a780`:

```text
new-session-load-race    28 passed, 0 failed
use-controller-meta     130 passed, 0 failed
tab-switch-hydration     79 passed, 0 failed
activate-topic-stale      8 passed, 0 failed
open-topic-coalescing    15 passed, 0 failed
app-chrome-tabs          83 passed, 0 failed
send-failed              66 passed, 0 failed
focused total           409 passed, 0 failed
tsc --noEmit                passed
git diff --check            passed
```

The same code tree was merged locally into `main-v2` at `b907a6b0` without
conflicts. The merged-state focused matrix passed 419 tests, including the 10
additional `send-failed` assertions on the newer base. TypeScript checking,
`desktop && go test ./...`, and `git diff --check` also passed in that merged
state.

The history-only author metadata rewrite from `7817880d` to `e424a780`
preserved the exact source tree (`37f32498`).

Cache impact: none. This changes desktop/frontend navigation state and tests
only; provider-visible prompts, tool schemas, request serialization, and cache
keys are unchanged.

Security impact: none. No dependencies, network access, file access, command
execution, credentials, permissions, or sandbox boundaries changed.

### DIFF
--- a/desktop/frontend/src/App.tsx
+++ b/desktop/frontend/src/App.tsx
@@ -408,12 +408,13 @@ type SidebarImConnection = {
   allowlistUsers: string[];
   allowlistMatched: boolean;
 };
-type DesktopNavigationInput =
+type DesktopNavigationIntent =
   | { kind: "topic"; scope: string; workspaceRoot: string; topicId: string; sessionPath?: string }
   | { kind: "blank"; scope: string; workspaceRoot: string }
   | { kind: "delivery-worktree"; workspaceRoot: string }
   | { kind: "sidebar-im"; connection: SidebarImConnection }
   | { kind: "resume-session"; session: SessionMeta };
+type DesktopNavigationInput = DesktopNavigationIntent & { navigationIntentSeq: number };
 type PendingDesktopNavigationRequest = PendingNavigationRequest<DesktopNavigationInput>;
 type SidebarImTopicSource = {
   platform: SidebarImPlatform;
@@ -1080,6 +1081,7 @@ export default function App() {
     openTopicSession,
     activateTopic,
     noteNavigationIntent,
+    isNavigationIntentCurrent,
     syncActiveTab,
     ensureBlankTab,
     ensureBlankSurface,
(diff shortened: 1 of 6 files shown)


xiaoshuai1024:
Thanks Sivan — appreciate you cleaning this up. I'd missed that NewSessionForTab could no-op on an already-blank session without invalidating local requests; the reset-hydration approach is the cleaner fix anyway. Sorry for the extra work on this one. Let me know if there's anything else you'd like me to pick up.

Issue #6932, opened 2026-07-25
[Bug]: Job artifact path traversal allows writes outside the intended temp directory


GG-Feng:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

main-v2

## Summary

Job artifact paths can escape the intended temporary artifact directory because `parentSession` and `kind` are used in filesystem path construction without rejecting path traversal components.

This allows artifact log files to be created outside the job temp root, subject to the filesystem permissions of the running process.

## Affected Code

`internal/jobs/jobs.go`

`StartForSession` trims `parentSession`, but does not reject `..`, `/`, or `\` path components. It also builds the job ID directly from `kind`:

```go
parentSession = strings.TrimSpace(parentSession)
id := fmt.Sprintf("%s-%d", kind, m.seq)
logPath, metaPath, file, artifactErr := m.openArtifactLocked(parentSession, id)
```

`artifactDirLocked` then joins `parentSession` with `m.tempRoot`, and `openArtifactLocked` creates the resulting artifact file with `os.Create`.

## Reproduction

I reproduced this through the real `jobs.Manager` code path.

Variant 1:

```text
parentSession = "../escaped_parent"
artifactPath = ...\Temp\escaped_parent\bash-1.log
```

The artifact was created outside the `reasonix-jobs-*` temp root.

Variant 2:

```text
kind = "../../leaked_kind"
id = ../../leaked_kind-1
artifactPath = ...\Temp\leaked_kind-1.log
```

This also wrote the artifact outside the intended temp root.

## Expected Behavior

Job artifacts should always be created inside the configured artifact/temp directory.

## Actual Behavior

Path traversal components in `parentSession` or `kind` can cause artifact files to be created outside the intended directory.

## Impact

If an attacker can influence these values, they may be able to write artifact files outside the job artifact directory, subject to process filesystem permissions.

## Suggested Fix

- Validate `parentSession` and `kind` with a strict allowlist, such as `^[A-Za-z0-9_-]+$`.
- Reject path separators and `..` components.
- After constructing artifact paths, verify that the final path remains inside the intended root using `filepath.Rel` or an equivalent containment check.

### Steps to reproduce

1. Create or trigger a job through `jobs.Manager.StartForSession`.
2. Use a `parentSession` value containing a parent-directory component, for example `../escaped_parent`.
3. Let the job write an artifact log.
[...]

PR #6937, opened 2026-07-25
fix(jobs): preserve trusted session path semantics / 保持可信 session 路径语义

PaulDing98:
Refs #6932. The caller-influenced `StartForSession` vulnerability is closed by #6936.

## Summary / 摘要

`SetActiveSessionPath` receives transcript paths already selected by the store/controller flow. This PR adds defense-in-depth syntax validation without reinterpreting those trusted paths: control characters are rejected, while valid absolute, relative, `..`, and host-specific path spellings retain their normal filesystem semantics.

`SetActiveSessionPath` 接收由 store/controller 流程选定的 transcript 路径。本 PR 增加防御性语法校验，但不重新解释这些可信路径：拒绝控制字符，同时保留合法绝对路径、相对路径、`..` 及当前平台路径写法的正常文件系统语义。

## Fix / 修复

- Reject NUL and other control characters before a transcript path reaches artifact migration or loading.
- Preserve trusted path spelling instead of treating lexical `..` or Unix backslash characters as containment violations.
- Keep active-session notice routing correct even when persistent path validation fails.
- Clear stale `artifactDirs` and `loaded` bindings after a rejected rebinding, so future jobs cannot keep writing to an older transcript with the same session id.
- Preserve the legacy empty-path behavior: update the active session only, without warning or filesystem work.
- Fix the CI race in `TestStartForSession_AcceptsValidInput` by holding the worker behind a cancellable channel gate and snapshotting `Job` fields under `Job.mu`. Apply the same lock discipline to the sibling session-path test.
- Merge the latest `main-v2` baseline without changing the net PR scope.

## Security boundary / 安全边界

This helper is deliberately not a trusted-root containment check and does not make arbitrary caller-controlled absolute paths safe. Lexical rejection of `..` cannot provide containment while arbitrary absolute paths remain valid, and it breaks legitimate callers such as headless `--resume ../sessions/a.jsonl`.

If `SetActiveSessionPath` ever accepts an untrusted path, the API must also receive an authoritative trusted root and verify the resolved relationship with `filepath.Rel` or an equivalent containment mechanism.

该 helper 不属于 trusted-root containment 校验，也不能保证任意调用方控制的绝对路径安全。在仍接受任意绝对路径的情况下，单纯拒绝 `..` 无法提供 containment，反而会破坏 `--resume ../sessions/a.jsonl` 等合法调用。

## Files / 文件

| File | Change |
|:--|:--|
| `internal/jobs/jobs.go` | Validate trusted path syntax while preserving host path semantics; keep active routing correct and clear stale bindings on rejection. |
| `internal/jobs/jobs_security_test.go` | Preserve #6936 security coverage while synchronizing concurrent `Job` field reads. |
| `internal/jobs/jobs_sessionpath_test.go` | Cover trusted `..` paths, invalid rebinding, empty paths, control characters, normal persistent artifact creation, and synchronized artifact snapshots. |

## Compatibility / 兼容性

| Surface | Behavior | Conclusion |
|:--|:--|:--|
| Existing transcript paths | Absolute, relative, `..`, and host-specific spellings remain accepted | Preserved |
| Empty session path | Updates active session only | Preserved |
| Invalid control-character path | Emits a warning, updates active routing, and removes stale persistence bindings | Fail-safe |
| Persisted artifact format | No schema or file-format changes | Backward compatible |
| Provider prompt/tool schemas | Unchanged | No cache impact |

## Verification / 验证

Against latest `main-v2` (`b907a6b09`):

- `git diff --check` - passed
- `go vet ./internal/jobs/...` - passed
- `go test -race -count=50 ./internal/jobs -run '^(TestStartForSession_AcceptsValidInput|TestSetActiveSessionPath_AcceptsValidInput)$'` - passed
- `go test -race -count=1 ./internal/jobs/...` - passed
- `go test -count=1 ./internal/jobs/...` - passed
- `go test -count=1 ./...` - passed

Cache-impact: none - no provider-visible serialization, prompt, tool schema, memory, or model-call path changed.


PaulDing98:
Race condition note (out of scope for this PR):

The CI `race` check run for this PR flagged a data race that originates from #6936 (already merged), not from the `validateTrustedSessionPath` work in this PR:

```
WARNING: DATA RACE
  Read at 0x00c0002b8928 by goroutine 984:
    internal/jobs/jobs_security_test.go:154 +0x204
  Previous write at 0x00c0002b8928 by goroutine 985:
    internal/jobs/jobs.go:506 +0xa24
  Goroutine 985 (finished) created at:
    internal/jobs/jobs.go:442 +0x106c
    internal/jobs/jobs_security_test.go:149 +0x1e9
```

Root cause: `TestStartForSession_AcceptsValidInput` (introduced in #6936) reads `j.status` without taking `j.mu`, while the run goroutine writes `j.status = st` under `j.mu`. The field has been mutated without the lock since long before #6936 — the race detector only caught it once #6936 added a test that observes the field from another goroutine without synchronization.

Fix is being prepared in a separate PR (add a locking `Status()` helper on the `Job` type and update the test to use it). The race is unrelated to `SetActiveSessionPath` or `validateTrustedSessionPath`; the path-validation hardening in this PR is independent of it.

Race evidence captured from this PR's CI run: https://github.com/esengine/DeepSeek-Reasonix/actions/runs/30275206427/job/90012258161

SivanCola:
Follow-up: the race reported above is fixed in `f321a817c`.

The test now keeps the worker in a deterministic running state with a cancellable channel gate and snapshots `Job` fields while holding `Job.mu`. The sibling session-path test was audited and updated to use the same lock discipline.

Verification:
- Focused race regression passed 50 consecutive runs.
- Full `internal/jobs` race suite passed locally.
- Full repository test suite passed locally.
- GitHub CI `race` passed: https://github.com/esengine/DeepSeek-Reasonix/actions/runs/30324504856/job/90166933896
- All other checks on the updated PR head passed.

Issue #6958, opened 2026-07-26
refactor(goal): advance() output model allows interceptMsg/notice cross-contamination between multiple FSM mechanisms


erphenheimer:
## Problem

The `goalMachine.advance()` function uses a single channel (`interceptMsg` field + `notice` return value) for three different mechanisms that can keep a goal running:

1. **Idle detection** (L321-330): sets `g.interceptMsg` when `idleTurns >= maxGoalIdleTurns`
2. **Todo incompleteness** (L272-277): sets `g.interceptMsg` when `[goal:complete]` arrives but todos remain
3. **Strict mode self-check** (L282-284): sets `g.interceptMsg` for quality check

All three exit with `notice=""` (meaning `cont=true`, goal continues).

Additionally, the **auto-turn limit** (L332-339) clears `interceptMsg` and sets a forced `notice`, which can override any of the above.

The output model has two issues:

### Issue A: Interceptors overwrite each other

All three mechanisms write to `g.interceptMsg`, but only one can survive per advance() call. If a turn triggers both idle detection AND a `[goal:complete]` intercept, the last writer wins:

```
Turn 1: Agent says [goal:complete] → formatIncompleteTodos → interceptMsg="unfinished todo A"  
Turn 2: takeIntercept() consumes "unfinished todo A" → synthetic turn runs
Turn 2: advance() → ToolNotCalled → idle detection sets interceptMsg="No tool calls..."  // OVERWRITE!
```

### Issue B: notice="" is a lossy signal

`cont = (notice == "")` conflates "continue because the goal is actively running" with "continue temporarily while showing an intercept message". The caller cannot distinguish between these states.

### Issue C: Auto-turn limit erases context

When `g.turns >= 50` fires, it unconditionally sets `g.interceptMsg = ""` on line 337, discarding whatever intercept was pending.

## Impact

- **#6956** ([goal:blocked] overridden by idle detection) is a direct symptom of Issue A
- "Agent says complete but goal continues" reports may be caused by Issue A + strict mode interaction
- Debugging intercept races is extremely difficult because the state is ephemeral (cleared on consumption)

## Suggested approach

Refactor `advance()` to return intercept as a dedicated output field instead of mutating the machine's state:

```go
type goalAdvanceResult struct {
    notice    string  // Terminal message (goal complete / blocked >=3 / limit reached)
    intercept string  // Non-terminal hint (idle reminder / incomplete todos / self-check)
    cont      bool    // (notice == "")
[...]

Issue #6956, opened 2026-07-26
fix(goal): reset idleTurns on [goal:blocked] to prevent premature intercept


erphenheimer:
## Bug: [goal:blocked] marker does not reset idle detection, causing premature "make progress" intercept

**Version:** main-v2 (recent, also present in v1.17.21)
**Severity:** Low (annoyance) — agent's explicit "I'm blocked" gets overridden by idle detection

### Description

When an agent signals `[goal:blocked:...]` in goal mode, the goal FSM records the intent to block but requires 3 consecutive blocked signals before actually stopping the loop (`internal/control/goal.go:307-310`). This is intentional design (prevents accidental stuck).

However, the idle detection mechanism (`internal/control/goal.go:321-330`) checks `g.idleTurns` to detect when the agent has gone multiple turns without making tool calls. The `GoalStatusBlocked` case does **not** reset `g.idleTurns`. So if the agent sends `[goal:blocked]` (an explicit action) without any tool calls in that turn, `idleTurns` keeps accumulating.

After 2 idle turns (`maxGoalIdleTurns = 2`), the FSM injects the intercept message:

> "No tool calls in recent turns. Either make progress with tools or signal [goal:blocked:<reason>]."

This intercept overrides the agent's already-stated intent to stay blocked, encouraging it to "make progress" — which can cause the agent to break a promise to wait (e.g., "waiting for user review").

### Root cause

`internal/control/goal.go`, `advance()` function, `GoalStatusBlocked` case (line 296-310):

```go
case GoalStatusBlocked:
    reason := cleanGoalBlockReason(in.reason)
    if sameGoalBlock(g.block, reason) {
        g.blocks++
    } else {
        g.blocks = 1
        g.block = reason
    }
    if g.blocks >= 3 {
        g.status = GoalStatusBlocked
        notice = "goal blocked: " + reason
    }
    // NOTE: g.idleTurns is NOT reset here
```

The idle detection that follows (line 321-330) fires when `notice == "" && g.interceptMsg == ""`. Since blocks < 3 leaves `notice` empty, the idle check proceeds. If no tools were called in the turn, `idleTurns` increments, and eventually the intercept fires.

### Expected behavior

When the agent explicitly sends `[goal:blocked:...]`, it should not be penalized by the idle counter. The idle detection should respect that the agent consciously chose to pause.

### Suggested fix

Add `g.idleTurns = 0` in the `GoalStatusBlocked` case:

```go
[...]

PR #6960, opened 2026-07-26
fix(goal): reset idleTurns on [goal:blocked] to prevent premature idle intercept

erphenheimer:
When the agent signals \`[goal:blocked]\`, the goal FSM increments the blocks counter but did not reset g.idleTurns. If blocks < 3 (so notice remains empty) and the agent makes no tool calls for 2 subsequent turns, the idle detection at maxGoalIdleTurns fires and sets an interceptMsg that overrides the agent's blocked signal.

**Fix:** Add \`g.idleTurns = 0\` in the \`GoalStatusBlocked\` case so that an explicit blocked signal resets the idle counter. The agent is not idling — it deliberately signalled a blocking condition.

**Fixes:** #6956

---

**Related:** The broader architectural issue of interceptMsg/notice cross-contamination between multiple FSM mechanisms is tracked in #6958.

### DIFF
--- a/internal/control/goal.go
+++ b/internal/control/goal.go
@@ -294,6 +294,7 @@ func (g *goalMachine) advance(in goalAdvanceInput) goalAdvanceResult {
 		g.interceptMsg = ""
 		notice = goalCompleteNotice
 	case GoalStatusBlocked:
+		g.idleTurns = 0
 		reason := cleanGoalBlockReason(in.reason)
 		if reason == "" {
 			reason = "blocked"
--- a/internal/control/goal_test.go
+++ b/internal/control/goal_test.go
@@ -756,6 +756,26 @@ func TestGoalRepeatedBlockedStopsAfterThreeTurns(t *testing.T) {
 	}
 }
 
+func TestGoalBlockedSignalDoesNotTriggerIdleIntercept(t *testing.T) {
+	g := &goalMachine{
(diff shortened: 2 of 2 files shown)


erphenheimer:
@SivanCola found the root cause and fixed it. CI passed. Please take a look when you have time. Thanks!

SivanCola (approved):
Approved. This is a well-scoped and worthwhile fix. Resetting idleTurns for an explicit blocked signal preserves the three-signal blocked audit without replacing its continuation prompt with an idle reminder.

I added TestGoalBlockedSignalDoesNotTriggerIdleIntercept in 6e23344c to cover the near-threshold idle state that the existing lifecycle test did not observe.

Compatibility:
- Persisted fields/formats: unchanged; old goal state remains readable and behaviorally compatible.
- Cache/provider surface: unchanged; no prompt, tool schema, or request serialization changes.
- Security surface: unchanged; no new dependencies, I/O, network, or privilege behavior.

Verified:
- go test ./internal/control -run '^TestGoalBlockedSignalDoesNotTriggerIdleIntercept$' -count=1
- go test ./internal/control -run 'Goal' -count=1
- go test -race ./internal/control -run '^TestGoalBlockedSignalDoesNotTriggerIdleIntercept$' -count=1
- go test ./...
- git diff --check

PR #6961, opened 2026-07-26
refactor(goal): bind FSM intercept outputs to Goal lifecycle / 让 FSM 拦截输出绑定 Goal 生命周期

erphenheimer:
## Summary

`advance()` previously wrote `interceptMsg` directly to `goalMachine` as a side effect. The initial version of this PR exposed `intercept` through `goalAdvanceResult`, but then wrote it back through a second lock acquisition before consuming it.

This PR now keeps intercepts as explicit FSM outputs and binds every Goal-owned turn and continuation result to a monotonic lifecycle epoch.

Two lifecycle windows are closed:

1. A synchronous Notice callback could replace the active Goal before a synthetic continuation was composed, pairing an old intercept with new Goal context.
2. A Goal could be replaced while a real or synthetic model turn was running, after which the old assistant marker could advance or complete the replacement Goal.

Continuation admission atomically captures the exact Goal text, AutoResearch task, and delivery scope used by the synthetic turn. Every real, edited, subagent, and synthetic Goal turn also carries its owning epoch into the next FSM advance. `goalMachine` validates that epoch under its lock before applying any assistant marker, so stale input and stale output both stop without modifying the replacement lifecycle.

### `internal/control/goal.go`

- Add `intercept` and `continuationEpoch` to `goalAdvanceResult`.
- Produce todo, strict self-check, and idle intercepts as local FSM outputs.
- Remove the transient `interceptMsg` field and the `setIntercept()` / `takeIntercept()` round trip.
- Invalidate pending results on Goal replacement/clear, stop, resume, state restore, rollback, and every newer `advance()` call.
- Atomically validate and capture continuation Goal context at synthetic-turn admission.
- Reject assistant output whose owning epoch no longer matches the current Goal before mutating FSM state.

### `internal/control/input.go`

- Split composition so admitted continuations use their captured Goal snapshot instead of re-reading mutable Goal state.

### `internal/control/turn_orchestrator.go`

- Consume the returned intercept directly.
- Revalidate after synchronous Notice callbacks and abort stale continuations.
- Carry captured Goal, AutoResearch task, and delivery scope through admitted synthetic turns.
- Capture lifecycle ownership for real, edited, subagent, and synthetic Goal turns and pass it into the following FSM advance.

### Tests

- Verify replacement, stop/resume, and newer FSM steps reject stale results.
- Verify a synchronous Notice callback cannot move an old intercept into a replacement Goal.
- Verify output from both ordinary Goal turns and synthetic continuations cannot advance or complete a Goal installed during runner execution.
- Update the merged #6960 blocked-signal test to assert the result-owned intercept model.
- Preserve the existing end-to-end incomplete-todo intercept behavior.

## Compatibility

| --- | --- | --- |
| Persisted Goal sidecar | `goalState` and its JSON fields are unchanged | Backward compatible |
| In-memory Goal state | `interceptMsg` was transient and never serialized | No migration required |
| Prompt/cache shape | System prompt, tool schemas, provider serialization, and composed Goal bytes are unchanged | No stable-prefix impact |

## Verification

- focused Goal lifecycle regression tests
- `./scripts/cache-guard.sh`
- `go test -count=1 ./internal/control`
- `go test -race -count=1 ./internal/control`
- `go test ./...`
- `git diff --check`

## Related

- Issue #6958 tracks the FSM output model.
- PR #6960 fixes the separate blocked-signal `idleTurns` bug and is included through the branch main-v2 merge.


erphenheimer:
@SivanCola refactored the FSM output model to make intercept flow explicit. CI passed. Please take a look when you have time. Thanks!

PR #7054, opened 2026-07-30
fix(desktop): render cleanly-paired $N$ as inline math (10–$20$ MeV)

lightfront:
## Problem

Two related inline-math classification gaps caused physics expressions to render with stray literal `$` signs:

1. **Paired bare numbers** — `10–$20$ MeV`, `$42$ elements` were classified as **currency** and escaped to `&#36;20&#36;`, rendering literally as `10–$20$ MeV`.
2. **One-sided equality** — `$=1$`, `$=-1$`, `$=x$`, `$x =$` were rejected (the one-sided relation rule covered `<`, `>`, `≠`, `≤`, `≥` but not `=`), rendering literally as `$=1$`.

Both stem from the classifier not recognising that a **paired `$...$` span is math intent**.

## Fix

Two small, targeted changes to `desktop/frontend/src/components/mathClassify.ts`:

### 1. Bare numbers in a paired span → math

The old rule treated every bare number as currency:

```ts
// Pure numbers, decimals, and percentages are too often currency...
if (/^\d+(?:\.\d+)?%?$/.test(math)) return false;
```

But **real currency is written *without* a closing `$`** (`costs $5`, `$4 and $5`). A *cleanly-paired* `$N$` is almost always a physics value. The classifier only ever sees content extracted from **paired** `$...$` spans (the Step 6 regex in `normalizeMath`), so flipping this to `return true` is safe:

| Input | `$` structure | Result | Why |
|---|---|---|---|
| `10–$20$ MeV` | paired | ✅ math | cleanly-paired `$20$` |
| `mass is $20$ MeV` | paired | ✅ math | cleanly-paired `$20$` |
| `costs $20 today` | 1 (odd) | ✅ literal | no closing `$`, never reaches classifier |
| `cost $4 and $5` | 2 (even) | ✅ literal | pairs as `$4 and $`, inner `"4 and"` has a space → rejected by the word rule |
| `price is $5 and $10` | 2 (even) | ✅ literal | same space-rule protection |

Currency protection comes from the existing space/word rule and pairing behavior, not the bare-number rule.

### 2. One-sided equality (`=1`, `x =`) → math

Extended the one-sided relation rule to include `=` on both the leading and trailing side, and allow a signed operand on the leading side (`$=-1$`, `$=+x$`). Unambiguous in a paired span — prose never wraps a bare `=N` in dollar signs.

## Changes

- `desktop/frontend/src/components/mathClassify.ts` — flip the bare-number rule; extend the one-sided relation rule to cover `=` (leading/trailing, signed operand).
- `desktop/frontend/src/__tests__/math-golden.test.ts` — re-align classifier tests to math-first semantics; add regression coverage for the MeV range, the `$=1$` family, and the currency-protection cases.

## Verification

- `math-golden.test.ts`: **264 passed, 0 failed** (250 existing + 14 new regression tests)
- End-to-end render through `react-markdown + remark-math + rehype-katex`:
  - `10–$20$ MeV`, `mass is $20$ MeV`, `set $=1$ here` → KaTeX math
  - `cost $4 and $5`, `costs $20 today` → plain literal (no KaTeX)

### DIFF
--- a/desktop/frontend/src/__tests__/math-golden.test.ts
+++ b/desktop/frontend/src/__tests__/math-golden.test.ts
@@ -182,6 +182,17 @@ check("cost $4 and $5 stays literal (space rule)", () => normalizeMath("cost $4 
 check("price is $5 and $10 stays literal", () => normalizeMath("price is $5 and $10") === "price is &#36;5 and &#36;10");
 // Odd-count currency: a lone $ never pairs, stays literal regardless.
 check("costs $20 today (no closing $) stays literal", () => normalizeMath("costs $20 today") === "costs $20 today");
+
+console.log("\nisLikelyInlineMath — one-sided equality (=1, x =) is math");
+// Regression (user-reported): "$=1$" rendered as a literal "$=1$" because the
+// one-sided relation rule covered <, >, ≠, ≤, ≥ but not =.
+check("$=1$ → math", () => isLikelyInlineMath("=1") === true);
+check("$= 1$ → math", () => isLikelyInlineMath("= 1") === true);
+check("$=-1$ → math (signed operand)", () => isLikelyInlineMath("=-1") === true);
+check("$=x$ → math", () => isLikelyInlineMath("=x") === true);
+check("$=1.5$ → math", () => isLikelyInlineMath("=1.5") === true);
+check("$x =$ → math (trailing)", () => isLikelyInlineMath("x =") === true);
+check("$=1$ in prose stays math", () => normalizeMath("set $=1$ here") === "set $=1$ here");
 check("URL", () => isLikelyInlineMath("https://example.com") === false);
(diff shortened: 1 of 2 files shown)


esengine:
This landed, though not as this diff. `ca8a6a163` (`fix(desktop): move inline math policy to remark AST`) adapts the paired-number and one-sided-equality cases you identified.

The cases were right; the implementation moved. Deciding what counts as inline math by pattern-matching the raw string works until it meets text that merely looks like math — `10–$20$` being exactly that kind of trap — so the policy was reimplemented against the remark AST, where the parser has already decided what is a math node and what is prose. Your case analysis is what defined the behaviour it had to produce.

Leaving this open afterwards was our mistake. Closing it now and saying where it went.

Thanks for it — the failing examples were more useful than the patch, which is not a criticism of the patch.

PR #7094, opened 2026-07-31
perf(settings): batch model polling and prevent stale writes / 批量模型轮询并防止过期写入

Oumuamu4:
## Summary

Reduces model-list polling jank in Settings when many providers and hundreds of model rows are configured, while preserving failure-atomic settings rebuilds and preventing stale background writes.

- **Cache and single-flight:** keeps a 60-second model-list cache, deduplicates in-flight requests, and keys entries by every request-relevant field plus the opaque catalog fingerprint.
- **Backoff and slow-network guard:** applies bounded 1-second to 60-second retry backoff and skips opportunistic background discovery on constrained connections.
- **Large-list rendering:** uses React transitions, deferred candidates, and `content-visibility` to keep large model lists responsive.
- **Refresh cooldown:** keeps a 30-second session cooldown per opaque provider-set fingerprint.
- **Batch discovery:** `FetchAllProviderModels` performs one Wails round trip, with at most four concurrent provider requests and a 15-second timeout per provider.
- **Race-safe catalog persistence:** background discovery submits only catalog fields and an opaque expected fingerprint. The backend re-reads current state under the config and credential edit locks, applies only model-catalog fields, and skips stale snapshots.
- **Revision-aware cache invalidation:** credential/config fingerprint changes create a new frontend cache identity, including changes made by another window or the CLI.
- **Failure-atomic runtime refresh:** catalog changes are persisted in one transaction and followed by one synchronous, serialized runtime rebuild. Lease-held rebuilds retain deferred retry behavior.
- **Secret-safe cooldown identity:** persistent session cooldown keys contain only the backend HMAC fingerprint, never custom provider header values.

## Concurrency and race safety

- Reasonix credential writes through `StoreCredentialLines`, `SetCredential`, and `RemoveCredential` share a process mutex and path-derived advisory file lock.
- Catalog compare-and-apply uses lock order config then credentials, re-reads a content-derived credential revision after acquiring both locks, and keeps both through config commit.
- Provider config identity and catalog mutation are compared and applied under `LockUserConfigEdits`.
- The Go batch result map is mutex-protected and request concurrency is capped at four.
- Frontend effect generations prevent stale completions after navigation, reload, edit, or unmount.
- A deterministic channel-controlled test forces old snapshot read, same-length credential rotation, then stale apply; the stale catalog is rejected without sleeps.

## Compatibility

| Surface | Existing data or older payload | Result |
| --- | --- | --- |
| Provider TOML and credential `.env` | File formats are unchanged | Compatible |
| Wails `modelCatalogFingerprint` | Additive field; missing values normalize to empty | Compatible |
| Older payload without fingerprint | Opportunistic background auto-refresh is skipped; manual refresh remains available | Safe fallback |
| Provider request serialization | Unchanged | No provider protocol impact |

## Validation

- Focused Desktop catalog tests under `go test -race`
- Root `go test ./...` and `go vet ./...`
- Desktop `go test ./...`, `go vet ./...`, and `go build ./...`
- Frontend `pnpm test`, `pnpm typecheck`, and `pnpm build`
- `./scripts/cache-guard.sh`
- `git diff --check`

The only added dependency remains `golang.org/x/sync`, used for bounded `errgroup` batch fetching.

Cache-impact: none to provider-visible prompts or request serialization; this changes Settings model-list caching only.
Cache-guard: `./scripts/cache-guard.sh` passed locally.


SivanCola (changes requested):
Thanks — the performance direction is worthwhile, especially concurrent model discovery and cheaper large-list rendering. However, the current head introduces correctness regressions and should not merge yet.

Please address these blockers:

1. **Preserve the settings rebuild contract and deferred lease retry.**
   `applyConfigChangeWithWarning` now persists configuration, launches `rebuildSetting` in a goroutine, and immediately reports success for every caller using this shared path — not only model polling. Non-lease rebuild failures are no longer returned, lease-held failures no longer schedule `deferredRebuildWarning`, and the frontend has no listener for either `settings:rebuild-complete` or `settings:rebuild-failed`. This can leave the persisted configuration and active controller out of sync while the UI reports success. The existing `TestDeferredRebuildRetryAppliesAfterLeaseRelease` fails at this PR head, and the live Desktop CI job fails in the same area. Please retain failure atomicity and the existing deferred-retry behavior. If rebuilds remain asynchronous, the operation needs an observable completion state bound to the original tab, must surface non-lease failures, and must preserve lease retry.

2. **Implement a real time-based retry window.**
   In `providerModelCache.ts`, the first failure changes the stored backoff from 1 s to 2 s, but no failure timestamp or `retryAt` is recorded. Every later call sees `backoff.has(key)` and returns `[]` forever, so the advertised 1 s -> 60 s exponential backoff never retries. A focused reproduction still had one fetch call and an empty result after waiting beyond the first delay. Please store and compare a retry deadline, preserve the original error semantics where appropriate, and add fake-timer coverage for retry, reset, and cap behavior.

3. **Correct and invalidate the provider cache identity.**
   `ProviderView` exposes `modelsUrl`, while `cacheKey` reads `modelsURL`; therefore the model-discovery endpoint is silently omitted from the key. A focused reproduction with the same key env/base URL and two different `modelsUrl` values reused the first provider's models. `invalidateProviderCache` is also never called, so saving a new credential can reuse a stale catalog or permanent backoff state. Please use the typed `ProviderView` shape, include all request-relevant identity needed to prevent cross-provider collisions, invalidate after credential/provider changes, and add regression tests.

The global `settings-auto-refresh-at` cooldown should also either be keyed by the provider-set fingerprint described in the PR or the description should be corrected; as written, an unrelated provider change can be suppressed by another provider's cooldown.

Verified against head `4c8001eb166617fc134cb2a7be9ca610d2400f18`:

- `git diff --check` passed.
- Frontend typecheck and the existing focused provider/settings tests passed, but they do not cover the new cache implementation.
- `go test -run '^TestDeferredRebuildRetryAppliesAfterLeaseRelease$' -count=1 .` failed at the PR head.
- `go test ./...` in the Desktop module failed on the latest-base merge tree.
- The focused cache reproductions failed for both permanent backoff and `modelsUrl` collision.
- The live Desktop CI job is failing.

Please also rebase onto the current `main-v2` before the next review. Once these contracts are fixed, covered by focused tests, and Desktop CI is green, I will be happy to re-review.

SivanCola (approved):
Approved after remediation in 812a1ccab.

The update restores failure-atomic settings rebuilds and deferred lease retry, implements time-based provider model cache backoff with complete request identity and credential invalidation, batches background catalog persistence into one rebuild, and adds focused regression coverage. The branch was also merged with the current main-v2 baseline before publication.

Verified locally:
- desktop: go test ./... and go build ./...
- frontend: pnpm test, pnpm typecheck, and pnpm build
- focused cache/settings/terminal integration suites
- ./scripts/cache-guard.sh
- git diff --check

GitHub checks are green, including desktop, desktop-macos, desktop-windows, CodeQL, cache-impact, and release-note validation.

SivanCola (changes requested):
Requesting changes after re-reviewing head `16c111abb29479861ee41ca1928cc7621d764e86` against the current `main-v2`.

The performance work is worthwhile and most of the earlier remediation is sound: rebuild failure atomicity is restored, batch discovery is bounded, provider config writes use a narrow compare-and-apply API, and the frontend has a generation guard. Two credential-concurrency gaps still block merge:

1. **Make credential rotation and catalog compare-and-apply one linearizable transaction.**
   `SaveProviderModelCatalogs` reads the credentials file metadata and compares the fingerprint while holding `LockUserConfigEdits`, but `SetCredential` / `RemoveCredential` rewrite the credentials file outside that lock. A credential write can therefore land after the catalog path reads the old revision but before it saves the old request's catalog. The sequential test using a different-length key does not force this overlap. Please introduce a shared credential revision/lock that all credential writes and catalog comparison use, then add a channel-controlled test: pause catalog apply after its initial read, rotate the credential, release apply, and assert that no stale models are persisted.

2. **Include the opaque credential/config revision in the frontend model-cache identity.**
   `providerModelCache.cacheKey` ignores `modelCatalogFingerprint`. A fresh `ProviderView` after another window or the CLI rotates a credential therefore reuses the old 60-second catalog when the provider name, endpoint, and key source are unchanged. A focused reproduction fetched `old-catalog`, then called the cache with a new fingerprint and received `old-catalog` again with only one backend call. Please key/invalidate the cache and auto-refresh cooldown with an opaque revision token and add coverage for old fingerprint -> new fingerprint causing a real fetch.

Security follow-up: `SettingsPanel` currently serializes complete provider header values into the `sessionStorage` cooldown key. Please use an opaque digest/token instead of persisting raw header values; custom gateway headers may contain credentials.

Concurrency classification:
- **CONFIRMED:** external credential rotation can race catalog compare/apply because the two files use different lock domains.
- **CONFIRMED:** credential revision changes do not invalidate the frontend model cache.
- **VERIFIED SAFE:** provider config edits are compared and applied under `LockUserConfigEdits`; batch result-map writes are mutex-protected; frontend tab changes invalidate older effect generations; runtime rebuilds remain serialized and failure-atomic.

Verified on this head:
- focused Desktop catalog tests under `go test -race` passed
- provider cache tests, settings snapshot tests, and frontend typecheck passed
- cache guard passed
- `git diff --check` passed
- GitHub CI and CodeQL are green

SivanCola (approved):
Approved after remediation at ff044ad3dd486568baa53977a4db4c4ce7a52714.

This is a worthwhile Settings performance improvement: model discovery is batched and bounded, large lists render more cheaply, and catalog writes remain failure-atomic.

The previously blocking concurrency and security issues are resolved:
- Reasonix credential writes and provider catalog compare-and-apply now share a process and advisory file-lock boundary. Catalog persistence re-reads a content-derived credential revision under the lock and keeps it through commit.
- The frontend model cache and auto-refresh cooldown use the opaque catalog fingerprint, so credential rotation from another window or the CLI triggers a real fetch.
- Session cooldown keys no longer persist custom provider header values.
- Deterministic channel-controlled coverage forces an overlapping same-length credential rotation and proves the old catalog is rejected.

Concurrency classification:
- FIXED: credential rotation versus stale catalog apply.
- FIXED: credential revision versus frontend cache reuse.
- VERIFIED SAFE: narrow config compare-and-apply, batch map synchronization, effect-generation stale-result suppression, and serialized failure-atomic runtime rebuilds.

Compatibility and security:
- Provider TOML, credential file format, and provider request serialization are unchanged.
- The Wails fingerprint field remains additive, with a safe older-payload fallback.
- CodeQL, govulncheck, cache-impact, and release-note validation are green.

Verified on the current head and latest main-v2 integration:
- Root and Desktop tests, vet, and build
- Frontend full tests, typecheck, and production build
- Focused Go race and frontend cache/privacy regressions
- Cache-impact guard and git diff --check
- GitHub CI, including race and Desktop Windows

PR #7109, opened 2026-07-31
fix(agent): rate-limit missing tool-call reasoning warnings / 按配置限频工具调用思考告警

### LINKED ISSUE #7059: DeepSeek tool calls without reasoning_content — warning repeats every session despite "shown once per session"
The warning "deepseek returned tool calls without reasoning_content; continuing, but thinking context is lost on such turns (shown once per session)" re-appears in every new session, even though its text says "shown once per session".

Please either:
- Actually show it only once per session (as advertised), or
- Add a flag (e.g. `--suppress-tool-call-reasoning-warning`) to permanently suppress it when the provider inherently does not return reasoning_content on tool calls.

DeepSeek never returns reasoning_content on tool calls — this is normal API behavior, not an anomaly.

sudoviber:
Closes #7059

## Problem

DeepSeek's current Thinking Mode contract requires `reasoning_content` from tool-call assistant turns to be preserved and sent back in subsequent requests. A tool call without replayable thinking content is therefore a compatibility incident, but repeating the same notice in every session is noisy.

The previous implementation persisted only a provider name forever. That could suppress a newly broken endpoint/model configuration, hide a later regression after recovery, and store a user-visible provider label unnecessarily.

Official contract: https://api-docs.deepseek.com/zh-cn/guides/thinking_mode/

## Fix

- Track an active incident by an opaque SHA-256 fingerprint of the exact adapter configuration; raw provider names, model IDs, and endpoint URLs are never persisted.
- Rate-limit the same active incident for 24 hours, then re-arm it automatically.
- Clear the incident after a healthy tool-call turn returns replayable thinking content, so a later regression is reported again.
- Preserve event ordering with nanosecond missing/healthy watermarks so delayed cross-process transactions cannot revive a resolved incident.
- Coalesce concurrent observations of the same incident inside one process while preserving the latest event timestamp.
- Keep read-check-write atomic under the bounded cross-process lock; retry failed healthy-state writes before the next claim and fail visible while I/O remains unavailable.
- Align the OpenAI-format warning policy with known DeepSeek thinking models, including V4 Flash, while leaving non-thinking and unknown custom models unchanged unless thinking is explicitly enabled.
- Use protocol-neutral diagnostic wording because the same contract can arrive through OpenAI, Anthropic, Responses, or remote adapters.

## Compatibility

| State / reader | Behavior | Conclusion |
| --- | --- | --- |
| Existing stable release | Ignores the additive state sidecar. | Safe |
| No state file | The first incident creates v2 state with a hashed identity and timestamps. | Safe |
| Unreleased v1 `{"providers":[]}` preview state | Re-arms once, then migrates to v2 on the next write. | Safe |
| Existing v2 incident without nanosecond watermarks | Derives ordering from the existing millisecond fields and upgrades on the next write. | Safe |
| Corrupt, expired, or future-dated state | Treats it as empty and self-heals. | Safe |
| Direct `agent.New` embedder without a state directory | Keeps in-memory active-incident deduplication and healthy re-arming. | Safe |

Other model providers are not sent different requests: prompt text, request serialization, replay bytes, tool schemas, and remote protocol schemas are unchanged. The new fingerprint is local diagnostic state only, and adapters that do not opt into this warning remain unaffected.

## Verification

- `go test ./...`
- `go vet ./...`
- `golangci-lint run --timeout=5m`
- `./scripts/cache-guard.sh`

The cache guard passed all 8 scenarios with 92–95% tail averages against the 90% threshold. Tests cover exact-configuration separation, 24-hour expiry, healthy recovery and regression, stale-health and delayed-failure ordering, v1/v2 migration, future/corrupt state recovery, owner-only permissions, transient read/write failures, and concurrent same/distinct incident claims.

No credentialed live-provider call was run; the verification is deterministic adapter, agent, race, integration, lint, and cache coverage on the latest `main-v2` merge tree.

Cache-impact: low - local diagnostic incident state only; provider-visible prompts, request serialization, replay bytes, tool schemas, and remote protocol schemas are unchanged.

Cache-guard: ./scripts/cache-guard.sh passed all 8 scenarios with 92-95% tail averages against the 90% threshold.

System-prompt-review: internal/boot/boot.go only wires the shared state directory into agent options; no prompt construction or system-prompt assembly changed.


sudoviber:
CI 状态说明:`test (windows-latest)` 失败于 `TestSaveSnapshotSucceedsWhenCrossProcessFileLockReleasesBeforeDeadline`(internal/agent/save_test.go:267,会话快照跨进程文件锁测试)——与本 PR 改动无关(未触及 session save / filelock 路径)。

这是 pre-existing 的 Windows flaky:**main-v2 分支自身的 CI 也有同样失败**(run 30659196765,同一测试)。

其余全部通过:cache-impact、test (ubuntu/macos)、lint、race、coverage、desktop、CodeQL、govulncheck、Analyze。能否帮忙 rerun 该 job 或直接合并?

PR #7192, opened 2026-08-02
fix(config,repair,update,desktop): TOML-safe config writes, startup recovery, update reconciliation, session ownership and smart close

SivanCola:
feat(config,repair,update,desktop): TOML-safe config writes, startup recovery, update reconciliation, session ownership and smart close

Four coordinated changes shipped in one PR (per the plan, they land in the
same desktop preview candidate):

## 1. TOML-safe config writes with reversible repair

- encodeTOMLString/encodeTOMLKeyPart: single TOML 1.0 basic-string encoding
  entry point; never emits Go-only escapes (\\a, \\v, \\xNN) the TOML parser
  rejects; invalid UTF-8 returns an error instead of silent replacement. All
  render.go value sites converted from %q/strconv.Quote.
- validateAndWriteConfigResolved: unified write pipeline used by SaveTo,
  SaveToScope, WriteFile, WritePermissionsAllow, project incremental merges
  and migrations: parse with the production parser, semantic round-trip via
  ConfigPersistenceSnapshot, delta-field verification for incremental
  merges, original-file state-ID re-check before the atomic replace; a
  renderer that silently drops or mangles a persisted field is caught by the
  intended-config-vs-decoded comparison (normalizePersistedConfig merges
  explicit values over built-in defaults, applies accessor normalization,
  injects implicit built-in providers with clean array-of-table decoding and
  clears legacy non-persisted fields).
- ScanTOMLPathEscapes/ApplyTOMLPathEscapes: TOML lexer diagnosing Windows
  absolute paths with unescaped backslashes (invalid-escape and silent
  semantic-escape classes), token-level high-confidence repair that preserves
  UNC prefixes, refusal on ambiguous/multi-line/non-path cases; fuzzed.
- Guard/macOS startup auto-repair of the global config with repair
  transaction, timestamped backup, undo, snapshot and last-known-good
  refresh; project reasonix.toml is preview-only and requires confirmation
  with bound state IDs; desktop recovery banner with undo/open/restore
  actions and a persistent damaged-config state.

## 2. Startup recovery and config fault isolation

- config.LoadUpdateNetworkConfig: minimal updater loader reading only update
  channel/proxy/network fields, falling back to the last-known-good snapshot
  and then env proxy defaults; the updater no longer depends on a full
  config.Load, so damaged configs never block update checks or installs.
- ConfigLoadError carries the failing file path and line; a broken project
  reasonix.toml only degrades its tab (file/line/fix-preview/apply/open
  actions), a broken global config boots tabs on the recovery configuration
  with external integrations disabled and a persistent banner.

## 3. Pending-update reconciliation

- InspectPendingUpdate derives the pending-update state (none | prepared |
  probationary | failed_install | restored | active_handoff | blocked) from
  the transaction, installed release-unit record, failure marker, running
  version and OS lock; the executor (repair.ReconcilePendingUpdate) cancels
  stale prepared transactions, rolls back failed installs exactly, preserves
  probationary transactions for the healthy-start commit and fails closed on
  damaged ones; active_handoff is never acted on and settled restored
  transactions end through the verified path. Guard launch, desktop startup,
  pre-check and pre-install all reconcile; UpdateInfo carries the recovery
  view and progress emits reconciling/rolling_back/recovered instead of the
  opaque "a pending update already exists" error.

## 4. Session ownership and smart close

- SessionRuntimeIssue extended with issueId, ownerKind and a non-null
  actions array; classification distinguishes current_tab,
  current_detached, same_instance_hidden, external_process, stale_reclaimed
  and unknown via the OS lease, its metadata and the process-local runtime
  registry. ResolveSessionRuntimeIssue re-validates the issue ID, runtime
  epoch binding and lease ownership and commits retry/read_only/copy in a
  single write-locked critical section; copy clones the authoritative
  transcript (checkpoint + event log under the cross-process file lock) to a
  reserved path and discards stale clones.
- Smart close (close_behavior=smart): idle quits without probing the tray,
  active work backgrounds with a notification and tray task count, and a
  missing restore path prevents the close and keeps the window visible.
  Config schema v6: new installs default to smart (version-gated accessor),
  upgraded configs without an explicit preference explicitly keep background,
  older binaries degrade smart to background. Settings gains the three-way
  selector; the frontend renders a classified session-issue card with
  actions.

Tests cover the encoding table and fuzz, path-fix fuzzing (including a
two-process clone test with a child writer holding the cross-process lock),
the write-pipeline state-ID and semantic-drift matrix, the update
reconciliation state matrix and idempotence, session owner classification
(external via a child process), smart-close decisions on every platform, and
v6 migration with downgrade safety. Full root and desktop go test, targeted
-race, vet, frontend tests and production build pass on the current
main-v2.

---

**Cache-impact: low** — no model-facing prompt fields changed. The only
cache-sensitive path touched is internal/config/config.go (schema version 6,
close-behavior normalization and its version-gated accessor); none of these
feed the system prompt (the prompt builder reads agent/skills/tools config,
which this PR does not alter), and `reasonix.example.toml` / provider-visible
payloads are unchanged.

**Cache-guard: `scripts/cache-guard.sh` passes** (internal/agent
TestReleaseCacheHitGuard, which pins the release cache prefix, ran green
locally and in this PR's CI).

**System-prompt-review:** the system prompt is built by internal/agent from
agent/skills/tools/permissions config — none of those files or values change
in this PR. config.go only adds the smart-close schema default and keeps the
rendered prompt output byte-identical for existing configs; please confirm on
review that no prompt-affecting field sneaks through the config plumbing.


SivanCola:
Added background-job recovery controls in `545efa87b811f83b9f163ba0816cec3c6f264414`.

What changed:

- Running jobs now appear on demand in the status bar and can be stopped per tab.
- Model, effort, and work-mode blockers identify whether the active work is a running answer, pending prompt, or background job.
- Local and Remote Workbench cancellation preserve the captured tab routing, so a late completion cannot update a newly selected tab.
- An unclicked "Continue checks" action remains idle and does not block switching.

Compatibility:

| Surface | Change | Result |
| --- | --- | --- |
| Wails API | Adds `CancelJob` and `CancelJobForTab`; existing methods are unchanged | Backward compatible |
| Busy error text | Preserves the legacy blocker phrase and appends structured active-work detail | Existing string matching remains valid |
| Remote Workbench | Reuses the existing `job/cancel` protocol | No wire-schema change |
| Persistence/providers/cache | No persisted-data format, prompt, provider request, or capability-route change | Cache-neutral for this commit |

Verification completed on the combined PR tree:

- `go test ./...`
- `cd desktop && go test -race . -run 'Test(SetTokenModeRejectsBackgroundJobs|SmartClose)' -count=1`
- `cd desktop/frontend && npm run build`
- `cd desktop/frontend && npm test`
- Focused background-job, blocker-localization, and tab-routing tests

SivanCola:
## Split in progress (not superseded yet)

#7192 is being split into five focused PRs per the agreed plan. **Do not merge this PR.** The branch is retained as the read-only source head (`a0f7d4791`); no force-push.

| Area | Status |
| --- | --- |
| TOML safe persistence | Open: #7228 |
| Config recovery / isolation | Waiting on #7228 merge |
| Pending-update reconciliation | Open: #7229 |
| Session ownership recovery | Waiting on #7198 merge |
| Smart close + config v6 | Waiting on #7198 merge |
| Background jobs / mode switch | Canonical in #7198 (not re-copied) |

#7192 will be closed as **superseded** only after all five split PRs exist and the mapping is complete. #7199 rebases last after the five + #7198 land.

Issue #6259, opened 2026-07-09
[Bug]: 一直显示 deepseeker 工具缺失思考参数


baosfeng:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

 1.17.9

### What happened?

<img width="824" height="370" alt="Image" src="https://github.com/user-attachments/assets/4cef64b5-8dbb-4466-b94f-43f3c4ffc98b" />

现在更新到最新版本，经常出现这个异常

### Steps to reproduce

现在更新到最新版本，经常出现这个异常

### OS / platform

MacOS 26.5.1

### Relevant logs or output

```shell
deepseek returned 1 tool call(s) without reasoning_content; continuing, but thinking context is lost for this turn (is a gateway dropping the reasoning field?)
```

HaoyueQin:
我接opencode也会时常显示这个错误

Aar222n:
＋1，我也出现了这个问题，之前没有出现过。我是windows系统，版本号1.17.9.出错原文：deepseek returned 1 tool call(s) without reasoning_content; continuing, but thinking context is lost for this turn (is a gateway dropping the reasoning field?)

phoenixlucky:
我这边也遇到同样问题。\n\n- 当前版本：v1.17.9\n- 环境：Windows 10\n- 报错原文：\deepseek returned 1 tool call(s) without reasoning_content; continuing, but thinking context is lost for this turn (is a gateway dropping the reasoning field?)\\n\n现象：DeepSeek 返回 tool call 时没有带 \
easoning_content\，程序继续执行，但本轮 thinking context 会丢失。怀疑中间 gateway / provider 转发时丢掉了 reasoning 字段。

faxble:
+1，今天开始频繁出现此问题，版本v1.17.9，操作系统Windows11专业版 25H2 ，报错原文：
deepseek returned 1 tool call(s) without reasoning_content; continuing, but thinking context is lost for this turn (is a gateway dropping the reasoning field?)

cfh0081:
看起来和官方API接口调用要求有关：
请注意，区别于思考模式下的未进行工具调用的轮次，进行了工具调用的轮次，在后续所有请求中，必须完整回传 reasoning_content 给 API。

官方链接：
https://api-docs.deepseek.com/zh-cn/guides/thinking_mode/#%E5%B7%A5%E5%85%B7%E8%B0%83%E7%94%A8

[...]

PR #7230, opened 2026-08-02
feat(provider): append endpoint-specific guidance to missing-reasoning warn

Lxiny-zy:
## Problem

The missing-reasoning warning (introduced for #6259 / refined in #7059) fires when a DeepSeek thinking-mode tool-call turn arrives without replayable `reasoning_content`:

> `deepseek returned tool calls without replayable thinking content; continuing with degraded reasoning`
> `... Check the selected model, endpoint, and reasoning protocol.`

The generic guidance gives users nothing actionable in either common scenario:
- **Third-party relays** commonly strip `reasoning_content` and only forward the standard OpenAI `content` field — the fix is a configuration change (Model capability mode → Plain chat / OpenAI reasoning).
- **Official `api.deepseek.com`** can also omit it on tool-call turns (observed reliably with `deepseek-v4-flash`, even at max reasoning effort; see #7059) — the user should know this is an endpoint behavior with concrete workarounds, not a config error.

## Change

- New optional provider policy `provider.MissingToolCallReasoningGuidancePolicy` (`MissingToolCallReasoningGuidance() string`), following the existing optional-policy pattern in the package.
- The OpenAI adapter implements it with endpoint-specific hints:
  - `base_url` is not `api.deepseek.com` (relay) → hint to set the capability mode to Plain chat / OpenAI reasoning (`reasoning_protocol` in config.toml).
  - Official endpoint → points at the DeepSeek Thinking Mode contract (api-docs.deepseek.com/guides/thinking_mode), notes that repeated absence on tool-call turns with `deepseek-v4-flash` is endpoint behavior, and lists workarounds (`thinking=disabled` or a model that returns thinking on tool calls).
  - Empty when the warning would not fire (`RequiresToolCallReasoning()` false, e.g. thinking disabled) or for non-DeepSeek endpoints.
- `Agent.warnMissingToolCallReasoning` appends the guidance to the warn `Detail`. The warn text, rate limiting, and resolution semantics are unchanged.

## Tests

- `TestMissingToolCallReasoningGuidance` (openai, 5 cases): official endpoint → contract hint; official endpoint with thinking disabled → empty; non-DeepSeek endpoint → empty; relay with explicit `reasoning_protocol = "deepseek"` → relay hint; relay with `"none"` → empty.
- `TestRunAppendsEndpointGuidanceToMissingReasoningWarn` (agent e2e): a provider implementing the guidance policy gets its hint appended to the warn `Detail`.

## Verification

- `go test ./internal/provider/... ./internal/agent/` — pass (agent suite run multiple times; one Windows environment flake unrelated to this change, re-runs green)
- `gofmt -l` clean, `go vet` clean

Cache-impact: none — warn-event detail only; no system prompt, tool schema, memory prefix, or provider request serialization changes. (`internal/provider/openai/openai.go` and `internal/agent/agent.go` are cache-sensitive paths, but the diff adds an optional guidance string to a runtime warning, not to any provider-visible prefix.)
Cache-guard: existing agent/provider suites pass unchanged, plus the new tests above.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -1323,9 +1323,15 @@ func (a *Agent) warnMissingToolCallReasoning(calls []provider.ToolCall, reasonin
 	if a.missingReasoningWarnPendingResolveAt.IsZero() {
 		a.missingReasoningWarnStateChecked = true
 	}
+	detail := fmt.Sprintf("this round carried %d tool call(s), but the endpoint omitted the thinking content DeepSeek requires clients to replay. Check the selected model, endpoint, and reasoning protocol. Repeated broken rounds are rate-limited for this exact provider configuration for up to 24 hours; a healthy tool-call turn re-arms future regressions.", len(calls))
+	if p, ok := a.prov.(provider.MissingToolCallReasoningGuidancePolicy); ok {
+		if guidance := strings.TrimSpace(p.MissingToolCallReasoningGuidance()); guidance != "" {
+			detail += " " + guidance
+		}
+	}
 	a.sink.Emit(event.Event{Kind: event.Notice, Level: event.LevelWarn,
 		Text:   fmt.Sprintf("%s returned tool calls without replayable thinking content; continuing with degraded reasoning (shown once for this incident)", a.prov.Name()),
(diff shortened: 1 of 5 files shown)


Lxiny-zy:
Superseded by #7259 (silently recover missing tool-call reasoning), which removes the user-visible warning entirely — endpoint guidance is deferred to a future doctor/config diagnostic. Thanks for reviewing; happy to have contributed to the discussion.

PR #7231, opened 2026-08-02
fix(agent): require consecutive healthy turns to clear missing-reasoning warn

Lxiny-zy:
## Problem

The 24h missing-reasoning rate limit (#7059 → #7109) is re-armed by **a single** healthy tool-call turn. Endpoints such as `deepseek-v4-flash` intermittently return replayable thinking on tool-call turns: most tool rounds omit `reasoning_content`, but one occasionally carries it. Each such turn resolves the incident, so the next missing round warns again — users on the official DeepSeek API with `deepseek-v4-flash` see the warning on **nearly every tool call**, defeating the rate limit entirely (observed state file: `warnedAt` … `lastResolvedAt` 6 minutes later, repeated every round).

## Change

Resolve an incident only after **`missingReasoningWarnResolveStreak` (3) consecutive healthy tool-call turns**:

- `missingReasoningIncident` gains `ResolveStreak`; state schema moves to **v3**.
- `resolveAt` increments the streak and records `LastResolvedAtUnixNano` only when the streak reaches 3; it still reports "state recorded" (not "threshold reached"), so callers never retry and double-count a healthy observation.
- `persistClaimAt` resets the streak on every missing observation — an intermittent healthy turn can never accumulate into a resolve.
- `load` accepts v2 documents (streak 0), so an upgrade keeps the existing incident and its active cooldown instead of losing it and re-warning.

Behavior matrix:

| Pattern | Before | After |
| --- | --- | --- |
| Intermittent health (flash) | warns on nearly every tool call | silent for the 24h cooldown |
| Sustained regression | one warning per 24h | unchanged |
| True recovery (3+ healthy turns) | re-armed | re-armed (future regressions still visible) |

## Tests

- `TestMissingReasoningWarnStateHealthyTurnRearmsRegression`: one healthy turn must **not** re-arm; three consecutive healthy turns must.
- `TestMissingReasoningWarnStateIntermittentHealthDoesNotRearmEveryRound`: missing observations between healthy turns reset the streak; the incident stays quiet through the cooldown.
- `TestMissingReasoningWarnStateDelayedFailureCannotReviveResolvedIncident` / `TestMissingReasoningWarnStateFutureLastMissingSelfHeals`: updated to the three-turn resolve semantics; watermark ordering guarantees preserved.
- `TestHealthyToolCallReasoningRearmsFutureRegression` (e2e): single healthy turn keeps the cooldown; three consecutive healthy turns re-arm a later regression.
- `TestMissingReasoningWarnStateLoadsV2IncidentWithoutNanosecondField` / `...LegacyPreviewRearmsAndMigrates`: v2/v1 upgrade paths still pass (migration now targets v3).

## Verification

- `go test -count=1 ./internal/agent/` — pass (full package)
- `gofmt -l internal/agent/` clean, `go vet ./internal/agent/` clean, `go build ./cmd/reasonix` pass

Cache-impact: none — diagnostic rate-limit state only; no provider-visible system prompt, tool schema, memory prefix, or request serialization changes. (`internal/agent/reasoning_warn_state.go` is not on the cache-sensitive path list.)
Cache-guard: existing agent suite passes unchanged, plus the streak regression tests above.

### DIFF
--- a/internal/agent/loop_e2e_test.go
+++ b/internal/agent/loop_e2e_test.go
@@ -561,6 +561,23 @@ func TestHealthyToolCallReasoningRearmsFutureRegression(t *testing.T) {
 	if got := run(healthy); got != 0 {
 		t.Fatalf("healthy turn warnings = %d, want 0", got)
 	}
+	// A single intermittent healthy turn must not re-arm the cooldown
+	// (deepseek-v4-flash returns thinking only occasionally; re-arming on one
+	// healthy turn would re-warn on every tool call).
+	if got := run(missing); got != 0 {
+		t.Fatalf("regression after a single healthy turn warnings = %d, want 0", got)
+	}
+	// Three consecutive healthy turns resolve the incident, so a later
+	// regression warns again.
+	if got := run(healthy); got != 0 {
+		t.Fatalf("healthy turn warnings = %d, want 0", got)
+	}
+	if got := run(healthy); got != 0 {
+		t.Fatalf("healthy turn warnings = %d, want 0", got)
+	}
+	if got := run(healthy); got != 0 {
+		t.Fatalf("healthy turn warnings = %d, want 0", got)
+	}
 	if got := run(missing); got != 1 {
 		t.Fatalf("post-recovery regression warnings = %d, want 1", got)
 	}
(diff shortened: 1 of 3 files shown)


Lxiny-zy:
Superseded by #7259, which adopts the anti-flapping streak idea (reimplemented with omitempty fields on the v2 state document) and credits this PR's author as Co-authored-by in commit 2a2d0e674. Closing as resolved upstream.

Issue #7221, opened 2026-08-02
[Bug]: 添加自定义供应商拉取模型后界面崩溃


superj8888:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.19.2

### What happened?

[react]

Error: Minified React error #310; visit https://react.dev/errors/310 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at updateWorkInProgressHook (http://wails.localhost/assets/vendor-react-BitsYX4p.js:1:79114)
    at Object.useDeferredValue (http://wails.localhost/assets/vendor-react-BitsYX4p.js:1:97508)
    at e.useDeferredValue (http://wails.localhost/assets/vendor-react-BitsYX4p.js:1:7192)
    at ProviderEditorModelPicker (http://wails.localhost/assets/SettingsPanel-BOgYwYwt.js:2:211851)
    at renderWithHooks (http://wails.localhost/assets/vendor-react-BitsYX4p.js:1:77552)
    at updateFunctionComponent (http://wails.localhost/assets/vendor-react-BitsYX4p.js:1:105364)
    at updateSimpleMemoComponent (http://wails.localhost/assets/vendor-react-BitsYX4p.js:1:103183)
    at beginWork (http://wails.localhost/assets/vendor-react-BitsYX4p.js:1:123695)
    at performUnitOfWork (http://wails.localhost/assets/vendor-react-BitsYX4p.js:1:173227)
    at workLoopSync (http://wails.localhost/assets/vendor-react-BitsYX4p.js:1:173084)

at ProviderEditorModelPicker (http://wails.localhost/assets/SettingsPanel-BOgYwYwt.js:2:211445)
    at div (<anonymous>)
    at ProviderEditor (http://wails.localhost/assets/SettingsPanel-BOgYwYwt.js:2:214505)
    at div (<anonymous>)
    at AddProviderPanel (http://wails.localhost/assets/SettingsPanel-BOgYwYwt.js:2:197419)
    at div (<anonymous>)
    at div (<anonymous>)
    at section (<anonymous>)
    at SettingsSection (http://wails.localhost/assets/SettingsPanel-BOgYwYwt.js:2:85141)
    at ProvidersSection (http://wails.localhost/assets/SettingsPanel-BOgYwYwt.js:2:186904)
    at ModelsSection (http://wails.localhost/assets/SettingsPanel-BOgYwYwt.js:2:175614)
[...]

Issue #7189, opened 2026-08-02
[Bug]: 自定义模型


sauyler:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

 v1.19.1

### What happened?

自定义模型完成，复制完模型名称后报错

### Steps to reproduce

1. 输入完url、key后。
2. 在粘贴deepseek/deepseek-v4-flash报错。

### OS / platform

win11

### Relevant logs or output

```shell
[react]

Error: Minified React error #310; visit https://react.dev/errors/310 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at updateWorkInProgressHook (http://wails.localhost/assets/vendor-react-BitsYX4p.js:1:79114)
    at Object.useDeferredValue (http://wails.localhost/assets/vendor-react-BitsYX4p.js:1:97508)
    at e.useDeferredValue (http://wails.localhost/assets/vendor-react-BitsYX4p.js:1:7192)
    at ProviderEditorModelPicker (http://wails.localhost/assets/SettingsPanel-DNDndJL-.js:2:209934)
    at renderWithHooks (http://wails.localhost/assets/vendor-react-BitsYX4p.js:1:77552)
    at updateFunctionComponent (http://wails.localhost/assets/vendor-react-BitsYX4p.js:1:105364)
    at updateSimpleMemoComponent (http://wails.localhost/assets/vendor-react-BitsYX4p.js:1:103183)
    at beginWork (http://wails.localhost/assets/vendor-react-BitsYX4p.js:1:123695)
    at performUnitOfWork (http://wails.localhost/assets/vendor-react-BitsYX4p.js:1:173227)
    at workLoopSync (http://wails.localhost/assets/vendor-react-BitsYX4p.js:1:173084)

at ProviderEditorModelPicker (http://wails.localhost/assets/SettingsPanel-DNDndJL-.js:2:209528)
    at div (<anonymous>)
    at ProviderEditor (http://wails.localhost/assets/SettingsPanel-DNDndJL-.js:2:212588)
    at div (<anonymous>)
    at AddProviderPanel (http://wails.localhost/assets/SettingsPanel-DNDndJL-.js:2:195502)
    at div (<anonymous>)
    at div (<anonymous>)
    at section (<anonymous>)
    at SettingsSection (http://wails.localhost/assets/SettingsPanel-DNDndJL-.js:2:83224)
[...]

PR #7239, opened 2026-08-02
fix: add null guards for model/provider fields to prevent React crash

ardiannurcahya:
## Problem

The model picker crashes when `ModelInfo.model` or `ModelInfo.provider` is null/undefined, and similarly when `ProviderView.name` or `ProviderView.baseUrl` is missing. This affects ModelSwitcher.tsx (line 83) and SettingsPanel.tsx (the canonicalOfficialProviderName name.trim() call chain in providerGroupID).

Refs #7189, #7221

## Fix

Added nullish coalescing guards at every call site that calls .toLowerCase() or .trim() on potentially-null model/provider/name fields:

**ModelSwitcher.tsx:**
- m.model.toLowerCase() -> (m.model || '').toLowerCase()
- m.provider.toLowerCase() -> (m.provider || '').toLowerCase()
- m.provider in groupBy map -> m.provider || ''

**SettingsPanel.tsx:**
- canonicalOfficialProviderName: name.trim() -> (name || '').trim()
- providerGroupID: p.name -> p.name || 'unknown'
- providerGroupLabel: return p.name -> return p.name || ''
- providerGroupDescription: return p.baseUrl -> return p.baseUrl || ''

## Testing

Verified by inspecting that all .toLowerCase(), .trim(), and .name/.baseUrl return expressions in the affected functions are now guarded.

### DIFF
--- a/desktop/frontend/src/components/ModelSwitcher.tsx
+++ b/desktop/frontend/src/components/ModelSwitcher.tsx
@@ -80,7 +80,7 @@ export function ModelSwitcher({
   const keyword = query.trim().toLowerCase();
   const filtered = useMemo(
     () => keyword
-      ? models.filter((m) => m.model.toLowerCase().includes(keyword) || m.provider.toLowerCase().includes(keyword))
+      ? models.filter((m) => (m.model ?? '').toLowerCase().includes(keyword) || (m.provider ?? '').toLowerCase().includes(keyword))
       : models,
     [models, keyword],
   );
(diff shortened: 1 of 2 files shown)


esengine:
Thanks for this — and for tracing it to specific call sites instead of just reporting the crash.

Two things worth separating here.

The crash in #7189 / #7221 turned out not to be the null fields. It was hook order: `ProviderEditorModelPicker` returned early on an empty candidate list *above* its `useDeferredValue` call, so an async model fetch took one render from four hooks to five and React #310 fired. That is fixed in `5dfbbe00e`.

But the guards you added stand on their own. A malformed or legacy provider catalog really can carry null display fields, and trusting the non-null TypeScript bridge contract at runtime is a genuine gap — the crash just happened to have a different cause. That part was adapted into `4a05f32d8`, which normalizes model and provider display metadata at the component ingestion boundaries rather than at each call site: same protection, one place to maintain as new call sites appear.

So the work landed, just not as this diff. Closing it as absorbed rather than leaving it open against code that has since moved underneath it. Both issues you referenced are closed as well; the fix ships in the next release.

Appreciated. Null-safety at boundaries is the kind of thing that is easy to skip and expensive later.

PR #7262, opened 2026-08-03
feat(cli): add Ctrl+Insert copy and Shift+Insert paste to the chat TUI

rongbc:
## Summary

Add the classic terminal clipboard keys to the chat TUI:

- **Ctrl+Insert** copies an active transcript or composer selection to the native clipboard and arms the copied notice, mirroring Ctrl+C without its destructive side effects. With no active selection it is a strict no-op — it never clears the input, cancels an in-flight request, or quits.
- **Shift+Insert** pastes through the same async native-clipboard read path as right-click, so it works whether the terminal delivers bracketed paste or forwards the key sequence (e.g. the kitty keyboard protocol), and SSH sessions get the same remote hint without ever reading the remote host's clipboard.

The selection lives in-app because Reasonix owns the mouse, so the terminal's own selection copy would see an empty selection — binding the key explicitly is what makes it work.

## Verification

- `go test ./internal/cli/ -run 'TestShiftInsert|TestCtrlInsert|TestMouseRightClick' -count=1` — ok
- `gofmt -l internal/cli/chat_tui.go internal/cli/chat_tui_test.go` — no output (clean)
- New tests: `TestShiftInsertPastesClipboardText`, `TestShiftInsertPasteOverSSHDoesNotReadRemoteClipboard`, `TestCtrlInsertCopiesTranscriptSelection`, `TestCtrlInsertCopiesComposerSelection`, `TestCtrlInsertWithoutSelectionIsNoOp`

## Cache impact

Cache-impact: none — changes are limited to `internal/cli/chat_tui.go` and its tests; no provider, tool, boot, config, memory, or output-style code is touched.
Cache-guard: none needed — no cache-sensitive paths are modified.
System-prompt-review: N/A

### DIFF
--- a/internal/cli/chat_tui.go
+++ b/internal/cli/chat_tui.go
@@ -1089,14 +1089,14 @@ func (m chatTUI) update(msg tea.Msg) (tea.Model, tea.Cmd) {
 
 	case tea.KeyPressMsg:
 		// Any keystroke dismisses a finished selection (copy is a right-click),
-		// with a few exceptions: Ctrl/Super/Meta+C copies the selection, the
-		// paste shortcuts keep it so the async clipboard result can replace
-		// it, and Left/Right collapse it to its ordered start/end.
+		// with a few exceptions: Ctrl/Super/Meta+C and Ctrl+Insert copy the
+		// selection, the paste shortcuts keep it so the async clipboard result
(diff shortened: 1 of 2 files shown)


SivanCola:

Kept: Ctrl+Insert copy without Ctrl+C side effects; Shift+Insert native-clipboard paste (SSH-safe).

Issue #7215, opened 2026-08-02
[Feature Request] Add a way to suppress the "session changed on disk; unsaved local transcript was saved as a conflict copy" notice forwarded to IM bot chats


Shelby0313:
# [Feature Request] Add a way to suppress the "session changed on disk; unsaved local transcript was saved as a conflict copy" notice forwarded to IM bot chats

## Summary

When Reasonix runs as a CLI-hosted bot gateway (`reasonix bot start --channels weixin`), every turn that hits a session snapshot divergence emits a **Warn-level Notice** which the bot gateway **unconditionally forwards** to the chat:

```
⚠️ session changed on disk; unsaved local transcript was saved as a conflict copy
```

There is currently **no config option to suppress or downgrade this notice**. For single-process CLI-hosted bot deployments (the recommended WeChat/QQ/Feishu setup), these conflicts are transient and non-destructive (the transcript is preserved via the recovery branch), so the message is pure noise for the end user — but it cannot be turned off.

## Environment

- Reasonix v1.19.1 (CLI gateway only; desktop bot connections disabled; no desktop app running)
- `reasonix bot start --channels weixin --dir <workspace>`
- Single `reasonix.exe` process (verified — no second instance/desktop app)

## Where it comes from (source)

1. **Emitted** in `internal/control/controller.go` (~line 3752), in `recoverSnapshotConflict`:

   ```go
   c.sink.Emit(event.Event{Kind: event.Notice, Level: event.LevelWarn,
       Text: "session changed on disk; unsaved local transcript was saved as a conflict copy"})
   ```

2. **Forwarded unconditionally** in `internal/bot/render.go` (~lines 195–201):

   ```go
   case event.Notice:
       if e.Level == event.LevelWarn {

Shelby0313:
Running v1.19.1 CLI-hosted (reasonix bot start --channels weixin, single process, desktop bot connections disabled), this Warn notice is forwarded to WeChat on every turn that hits a session snapshot divergence — which happens routinely, so the end user sees "⚠️ session changed on disk; unsaved local transcript was saved as a conflict copy" repeatedly in the WeChat chat, with no config option to suppress it.

[...]

PR #7344, opened 2026-08-03
Fix CLI recovery session proliferation and resume clarity / 修复 CLI 恢复会话膨胀与识别

SivanCola:
## Summary

- group conflict-recovery copies with their source session across interactive CLI resume surfaces
- label recovery copies with a short parent session ID while keeping `--continue` strictly chronological
- automatically move only covered, untouched, idle, unleased recovery copies to the recoverable session trash
- stage every recovery artifact invisibly and publish the complete Desktop trash item atomically
- keep `/resume <n>` bound to the list the user saw instead of mutating it during execution

## Problem

CLI conflict-recovery branches can accumulate after save conflicts. The resume picker previously flattened them into ordinary sessions, so users could not tell which entry was the original conversation. At the same time, non-interactive `--continue` promises to resume the most recent session and must not reuse presentation ordering that can place an older recovery leaf first.

The first recovery-trash implementation also published the primary transcript before its authoritative event log and sidecars had moved. A process crash, or a concurrent Desktop restore/purge, could therefore strand a split live/trash artifact set. Numeric `/resume` had a second intent race: it ran GC after completion had assigned an index, so removing an earlier row could retarget the number.

## Fix

Interactive resume ordering now keeps each recovery family together, preserves whole families at the picker cap, and puts the newest visible leaf first within each family. Recovery entries carry a localized badge containing the short parent ID. `--continue` bypasses that presentation ordering and selects directly from the chronological session listing.

CLI startup and resume presentation flows run conservative recovery hygiene. A branch is eligible only when:

- its metadata identifies it as a recovery branch
- its transcript still matches the original recovery digest
- its parent contains all branch content
- it has been idle for at least 24 hours
- no runtime holds its session lease

Before moving anything, cleanup acquires the parent removal guard, then the branch removal guard, and revalidates both idle time and parent coverage.

### Atomic recovery-trash publication

1. Create a non-session-shaped directory under `.trash`.
2. Durably write `.recovery-trash-pending.json` before the first destructive rename.
3. Move the transcript, authoritative event log, all owned sidecars, checkpoint/job directories, telemetry, and sub-agent artifacts into that invisible stage.
4. Publish the complete directory with one same-filesystem rename.
5. Write the existing Desktop `.trash-meta.json`, clear the pending marker, and release/remove the session lock sidecars.

`ReconcileCleanupPending` now completes both pre-publication and post-publication stages on startup. It also retains the earlier `recovery-trash:<item>` reconciliation path so an intermediate build cannot strand its marker.

Incomplete stages have neither a session-shaped directory name nor Desktop trash metadata, so older runtimes ignore them. Older runtimes can observe the entry only after the complete directory has been published.

Numeric `/resume <n>` no longer runs GC between displaying/completing the index and resolving it. Bare `/resume` still cleans before constructing the picker, and startup keeps the ordinary automatic sweep.

## Scope

This PR improves ordinary CLI resume and recovery-copy hygiene. It does not change bot notice forwarding or the conflict-generation path reported in #7215; that issue remains separate.

## Concurrency and data safety

- **Verified safe:** parent and branch save/lease locks stay held across final coverage validation, staging, and atomic publication.
- **Verified safe:** Desktop sees either no entry or one complete transcript/sidecar set; restore/purge cannot interleave with partial artifact movement.
- **Verified safe:** active CLI/Desktop sessions are skipped through the shared lease registry and cross-process lease probe.
- **Verified safe:** a continued, diverged, corrupt, missing-parent, or fresh recovery branch is preserved.
- **Verified safe:** process crashes before or after publication are repaired from the durable staging marker.
- **Verified safe:** numbered resume execution does not mutate its own index mapping.

## Compatibility audit

| Surface | Old-data / old-runtime behavior | Conclusion |
| --- | --- | --- |
| Session JSONL and branch metadata | No schema fields or identifiers change; sessions without recovery metadata keep their existing ordering and behavior. | Backward compatible |
| Existing Desktop trash entries | The `.trash/<item>/<session>.jsonl` and `.trash-meta.json` contract is unchanged. | Backward compatible |
| New recovery staging directory | Its name is not a session transcript and it has no Desktop trash metadata until complete; older runtimes ignore it. | Cross-version safe |
| New staging marker | New runtimes reconcile it; older runtimes do not inspect the invisible staging directory. | Cross-version safe |
| Earlier `recovery-trash:<item>` marker | The compatibility reconciler remains in place. | Upgrade safe |
| CLI translations | Adds one catalogue string in English, Simplified Chinese, and Traditional Chinese; no persisted data changes. | Backward compatible |

## Verification

- `go test ./internal/agent ./internal/cli`
- `go test -race ./internal/agent ./internal/cli`
- `go vet ./...`
- `go test ./...`
- deterministic crash recovery test: transcript staged while sidecars remain live, then startup reconciliation publishes one complete Desktop trash entry
- deterministic numeric-resume test: an eligible earlier recovery row cannot shift an already-completed index
- existing intermediate-marker reconciliation test remains green

## Cache impact

Cache-impact: none — the changes affect session listing, local persistence, and CLI selection only. No system prompt, tool schema/order, provider request serialization, compaction input, or model-visible prefix changes.

Cache-guard: the changed files are outside prompt/cache construction paths; root, CLI, agent, race, vet, and Desktop suites pass.

System-prompt-review: no system-prompt content or assembly path changed.


276408029:
终于修了这个，希望还能降低冲突几率，加油

PR #7373, opened 2026-08-04
Unified extension kernel and Extension Protocol v1 / 统一扩展内核与 Extension Protocol v1

SivanCola:
# Unified extension kernel and Extension Protocol v1 / 统一扩展内核与 Extension Protocol v1

## Summary

Lands the unified extension kernel and Extension Protocol v1 in one PR with
12 independently green commits: immutable runtime snapshots, fail-atomic
reload across CLI/Desktop/ACP, Plugin Manifest v1 (prompts, themes,
full-trust code runtimes), the stable JSON-RPC sidecar protocol,
interceptor + replacement-strategy dispatch, a streaming provider adapter
wired into first-boot model resolution and every model picker, structured
UI on all three frontends, the stdlib-only Go SDK with a runnable example,
and bilingual docs.

With no v1 extension installed, the system prompt, tool schemas, provider
request serialization, and prompt-cache prefix hash are byte-identical to
the pre-refactor runtime (committed goldens, unregenerated).

## Stage commits

1. `41a5c82b3` Baseline goldens + legacy compat gates (test-only).
2. `e024cbff5` internal/extension kernel: contribution catalog, builder, immutable snapshot, resource adapters.
3. `455592d9e` boot.BuildRuntime / boot.Rebuild; CLI `/reload`, Desktop "Reload Runtime", ACP `_reasonix.io/session/reloadExtensions` — one fail-atomic semantic.
4. `727ac2760` Plugin Manifest v1 (strict parse, merge rules), prompts, plugin themes (fallback-preserve IDs), FULL TRUST install display.
5. `0e90e2d1c` Extension Protocol v1 wire schema + generator (canonical JSON Schema, `sha256:41d3978a31b184c576241a86ed96893bf083fec6972d2988492766791a0e1d48`, `-check` enforced) + sidecar lifecycle (handshake, bounded shutdown, crash supervision, content refs).
6. `8a57881a2` Interceptor dispatch + replacement strategies at all 17 points (chain first, slot owner last; permission allow overrides host deny, audited).
7. `71ac1a4c9` Streaming provider adapter: seq-buffered chunks with lastSeq freeze, cancel, crash-without-fallback, `provider:<ref>` claim-gated conflicts.
8. `ae8a7fc0f` Structured UI: uihub, `extension_surface`/`extension_status` events (enum tail), TUI/Desktop/ACP rendering with text fallbacks, `/<plugin>:<action>` everywhere.
9. `0b5131553` sdk/go (stdlib only, zero requires) + generated SDK DTOs + fullsidecar example + 17 host↔SDK conformance tests.
10. `8ed68b316` Bilingual docs, CI sdk job (three-OS), sdk/go tag convention, final compat audit.
11. `4a2655185` Extension providers resolve on first boot and in all model pickers (preflight before model resolution, single Manager, executor/planner/guardian/subagent all on the merged resolver).
12. `b8bc976d1` ACP discovers non-current extension models; startup `host/ui/request` waits on a readiness gate instead of erroring; deterministic regression for the ContractEntries AB-BA deadlock fix.

## Cache impact

- No v1 extension installed: prompt, tool schemas, provider request, and
  prefix hash are byte-identical to the pre-refactor goldens
  (`internal/boot/testdata/golden`, `TestGoldenBaselineNoExtensions`).
- Every dispatcher point is nil-gated; zero behavior change without sidecars.
- `context.prepare` / `provider.request` replacements are ephemeral per
  request; the frozen snapshot never absorbs per-turn dynamic data
  (`TestBootSnapshotImmutableAcrossDispatch`,
  `TestProviderRequestReplacementCacheEphemerality`).
- A reload that changes prompt/tools yields exactly one attributable cold
  miss on the first post-swap request; a no-op reload leaves the prefix
  hash unchanged. `scripts/cache-guard.sh`: 8/8 pass (tail_avg 92–95,
  threshold 90), identical values to the pre-refactor baseline.
- Bundle budget: initial raw JS+CSS 2250→2265 KiB (extension UI surface);
  all gzip budgets (user-visible transfer cost) unchanged and passing.

## Full-trust threat model

- Code extensions run outside the sandbox with the unfiltered inherited
  environment; they can read the full session/environment, bypass
  permissions, and operate the machine. `permission.decision=allow`
  overrides a host deny (audited and surfaced).
- Authorization = install/update/replace/`--link` via the plugin flow; only
  `plugin-packages.json` state can launch a runtime (project configuration
  can never declare one). No second confirmation; `--link` keeps trusting
  changed content.
- Disclosure: FULL TRUST block in the install preview, `plugin show`,
  capability diagnostics, and the Desktop installer (command, intercepts,
  replacement slots, provider/UI capabilities).
- The handshake rejects any capability beyond the manifest
  (`capability_not_declared`); replacements are re-validated per point;
  sidecar output is credential-redacted before UI/logs/errors; crashes fail
  explicitly — never a silent fallback to another model or strategy.

## Persistence compatibility

| Artifact | Old-version behavior | New-version behavior |
| --- | --- | --- |
| `reasonix-plugin.json` (no apiVersion) | legacy parse, unknown fields ignored | unchanged (compat tests) |
| `reasonix-plugin.json` (v1) | parsed as legacy; `runtime` ignored | strict parse; unknown fields are errors |
| `plugin-packages.json` | schema v1 | unchanged; no new required fields |
| session JSONL | append format | unchanged; extension state isolated to `<session>.extensions.json` (ignored by old versions) |
| `desktop-theme-state.json` | v1→v2 migration; unknown IDs cleared | `plugin:` IDs preserved across uninstall; base-style rule unchanged |
| install plan JSON | old consumers ignore new fields | omitempty `RuntimePlanInfo` + prompt/theme counts only |
| provider refs | `plugin/...` unknown → model unavailable | resolved via the merged resolver |
| event kinds | unknown kinds dropped (TUI/Desktop); Notice/Ask fallback (ACP) | `extension_surface`/`extension_status` appended at enum end |
| ACP `_meta` | unknown metadata ignored | `sessionReloadExtensions`, `extensionSurface` entries |

## Concurrency audit

- Each turn pins one snapshot generation for the turn, tool batch, and
  compaction; reload swaps atomically only after the new build fully
  succeeds (session/approvals/goal/recovery migrated); the old controller
  and sidecars close only after the swap (`RuntimeSet.CloseIfGeneration`
  blocks stale cleanup).
- UI surfaces carry (plugin, surface, session, generation); stale-generation
  publishes dropped, stale requests cancelled.
- Provider streams: 1-based seq buffering, lastSeq freeze, gap→interrupted;
  crash drains buffered chunks then fails; no fallback.
- `ContractEntries` runs per-tool callbacks after releasing the registry
  lock (AB-BA fix) with a channel-controlled deterministic regression test.
- Startup `host/ui/request` gated on controller readiness with
  build-failure and cancellation unblocks (deterministic tests).
- `go test -race` green: extension/*, agent, control, boot, eventwire,
  remote, plugin, tool, desktop runtime.

## Verification matrix

| Gate | Result |
| --- | --- |
| Root `go test ./...` | PASS |
| `go test -race` (extension/*, agent, control, boot, eventwire, remote, plugin, tool, desktop) | PASS |
| `cd desktop && go test .` (+ `-race`) | PASS |
| Desktop frontend `typecheck`, full test suite, production build | PASS |
| `cd sdk/go && go test ./...` (+ `-race`, zero-require guard) | PASS |
| `cmd/extension-protocol-gen -check`, `cmd/remote-protocol-gen -check` | up to date |
| `scripts/cache-guard.sh` | 8/8 PASS |
| Six-platform cross-compile (darwin/linux/windows × amd64/arm64) | PASS |
| First-boot `default_model="plugin/..."` E2E through `Controller.RunTurn` | PASS |
| Host↔SDK conformance (17 tests: handshake, capability subset, intercepts, strategy, provider stream/cancel/usage, >64 KiB content refs, UI round-trips, timeout, crash, shutdown) | PASS (`-race`) |
| Legacy manifest / plugin state / session JSONL / theme state round-trips | PASS |
| Real-credential sidecar provider scenarios | NOT RUN (no credentials used); covered by fake-provider E2E + conformance instead |

## Notes

- ACP vendor method is `_reasonix.io/session/reloadExtensions` (ACP v1
  reserves bare names for the core protocol), advertised in initialize
  `_meta` alongside `extensionSurface`.
- A plugin provider can be the initial model of a fresh boot: sidecars
  preflight before model resolution (single Manager, exactly one
  `StartPackages` per build, fail-closed cleanup).

## Rollback

Each stage commit is independently green; revert from the tip in reverse
order (`git revert b8bc976d1 4a2655185 8ed68b316 ...`). Stages 1–2 are
additive; reverting 3–12 restores the pre-refactor assembly path while
keeping the golden gates that prove it.

## Review follow-ups (e473e4f7b)

- Sidecar writes are now bounded: rpcwire writes honor the caller context
  (abort fails only that request) and an opt-in MaxWriteStall
  (WriteStallError fails the connection); the sidecar client defaults to a
  10s bound and its supervisor kills the process tree on any unexpected
  connection end. Deterministic no-reader/full-pipe tests included.
- Golden baseline pins `[tools.search] engine="native"` (the grep
  description — and with it tool schemas/request/cache prefix — varied
  with rg on PATH, failing rg-less CI). Only that description changed in
  the regenerated artifacts; verified with rg excluded from PATH.
- providerext out-of-order buffer capped at a 256-seq window (interrupts
  on overflow), Windows-only syscall.Kill split behind build tags, sdk CI
  job pins shell: bash, the PR race sweep covers extension/boot/control/
  tool plus a desktop race step, SA1012/QF1007 lint items fixed.

Cache-impact: none — no-extension prompt/tools/request bytes are golden-pinned identical (only the fixture-pinned grep description line changed, from "ripgrep-backed" to the native wording); replacements are per-request ephemeral; no-op reload keeps the prefix hash.
Cache-guard: ./scripts/cache-guard.sh — 8/8 pass (tail_avg 92–95, threshold 90), identical to the pre-refactor baseline.
System-prompt-review: SivanCola — the golden-pinned no-extension system prompt is byte-identical; system_prompt.build replacement runs only for installed full-trust extensions and lands in the frozen snapshot hash.

## Review follow-ups, round 2

- rpcwire writes are now serialized through exactly one writer goroutine
  per connection with a bounded queue: frames are atomic/ordered by
  construction, pre-start cancellation drops the frame, the independent
  stall watchdog (not preemptable by per-call timeouts) fails wedged
  transports with WriteStallError, and the sidecar supervisor reaps the
  process tree on unexpected ends. Deterministic gated-writer and
  two-phase wedge tests.
- Windows PID probe uses golang.org/x/sys/windows with
  ERROR_ACCESS_DENIED-only-means-alive semantics.
- Deferred-reload failure text is credential-redacted before logs and UI
  (CodeQL #280), with a credential-shaped test.
- gofmt / SA1012 mechanical items; branch rebased onto latest main-v2.

## Review follow-ups, round 3

- rpcwire: closed connections can no longer report dropped writes as
  success — terminalError is never nil, the writer drains the queue
  forever after close answering every late frame with the terminal error,
  and the writer/watchdog start lazily so a never-served Conn leaks no
  goroutine. New post-EOF Notify/Request regression (20x).
- Session-save test compares typed dispatch.SessionPayload fields (raw
  JSON substring failed on Windows paths).
- deferredReloadFailedText is now the single production formatter for
  both log and tab warning, so the tested helper is the production path.
- Branch re-merged with the latest main-v2 (17 commits, no conflicts).

## Review follow-ups, round 4 (base sync)

- Merged the latest main-v2 (goal-runtime, /docs, send-as-message,
  usage-stats, planner degrade, narrowed settings columns, Responses API
  items). Extension actions still resolve before the new send-as-message
  fallback; planner degrade keeps plugin-ref resolution.
- Golden baselines regenerated for three upstream provider-visible builtin
  additions (docs, read_subagent_result, update_goal) — upstream
  cache-prefix growth, not extension drift.
- Race fix: the sink wraps (stats recorder, goalUsageTee) now compose
  before the extension UI hub closes over the sink, closing a real race a
  preflight UI publish exposed.
- CI: docs-impact gate needs a Documentation-impact field — added below.

Documentation-impact: updated - new EXTENSIONS.md/.zh-CN.md and EXTENSION_PROTOCOL.md/.zh-CN.md companions, PLUGIN_PACKAGES Manifest v1 chapter (en/zh), CLI /reload entry, ACP reloadExtensions + extensionSurface negotiation, CAPABILITY_DIAGNOSTICS v1 additive fields, THEME_PACK plugin kind + fallback-preserve, RELEASING sdk/go tag convention, and the generated EXTENSION_PROTOCOL.generated.md index.

## Documentation

New (bilingual): docs/EXTENSIONS.md/.zh-CN.md (capability map, install/manage,
reload semantics, dev guide, compatibility, full-trust security model),
docs/EXTENSION_PROTOCOL.md/.zh-CN.md (prose companion: transport, lifecycle,
content refs, interception, streaming providers, structured UI, errors, v1
stability contract), docs/EXTENSION_PROTOCOL.generated.md (generated
method/event/limit/error index), sdk/go/README.md.

Updated (bilingual): PLUGIN_PACKAGES (Manifest v1 chapter incl. FULL TRUST
and --link risk), CAPABILITY_DIAGNOSTICS (additive v1 fields), CLI
(/reload), ACP (reloadExtensions + extensionSurface negotiation), THEME_PACK
(plugin themes + fallback-preserve), RELEASING (sdk/go/vX.Y.Z tag
convention).

## Plan conformance (vs. the original implementation plan)

Fully implemented: the kernel pipeline and immutable snapshot, fail-atomic
reload with one shared semantic across all three frontends, all Manifest v1
parse/validate/display rules, the 16-method protocol with all limits, the
17-point interceptor decision matrix (permission override with audit,
timeout tiers, cache ephemerality proofs), the streaming provider adapter
(including first-boot default_model="plugin/..." and three-frontend
discovery/switching, crash without fallback), structured UI with fallbacks,
the stdlib-only SDK + example + 17 host<->SDK conformance tests, and every
acceptance gate in the plan's verification section (measured, not assumed).

Deliberate deviations (declared and verified):

| Plan | Shipped | Why |
| --- | --- | --- |
| ACP method reasonix/session/reloadExtensions | _reasonix.io/session/reloadExtensions | ACP v1 reserves bare names for the core protocol |
| Generator emits host DTOs | Host DTOs are the handwritten source; generator emits schema/hash/SDK DTOs/docs | Mirrors the existing remote protocolgen exactly |
| 2-3 real-credential sidecar scenarios | Not executed; fake-provider E2E + conformance instead | The plan allows recording this when no safe credential exists |
| Bundle budget | raw 2250->2270 KiB (gzip budgets untouched) | Extension UI surface; noted in the script and above |

Two deliberate narrowings:

- Plugin file changes do not get an fs watcher "needs reload" marker; the
  reload entry points are /reload, Desktop "Reload Runtime", and the ACP
  method (plugin uninstall/disable keeps using the runtime mutation lock +
  deferred rebuild). The protocol's resources/changed notification is
  host->extension and is implemented.
- The session_policy strategy rules/vetoes load/save/rotate and adjusts
  the phase payload, but deep transcript/meta rewriting is intentionally
  confined to the payload surface (host keeps its atomic file format;
  plugin-private state lives in the sidecar extensions file).

Branch note: the plan's 10 stage commits became 12 (stage 11 wires providers
into production paths, stage 12 closes review items), plus review-round and
base-merge commits; stage boundaries were never squashed, so point rollbacks
stay possible.


SivanCola:
Implementation update (current `main-v2`)

- Rebased through merge on `main-v2@1300dddd5` (Task Monitor merge included); current PR head: `1de1b9934`.
- Added a bilingual Extensions & Sidecars section to the public docs site, with MCP-vs-extension guidance, trust/caching semantics, installation, reload, doctor, and source references.
- Added README discovery links, expanded the maintained extension docs, and added an installable Go starter extension with tests and EN/ZH instructions.
- Added embedded-product-doc and website contract tests so the developer path remains searchable and complete.
- Resolved the combined Task Monitor + Extension UI bundle-budget overlap from measured production output; all bounded assets remain within the updated limits.

Local validation on the merged head is green for root tests/vet, SDK tests/vet and Windows cross-build, frontend tests and production build, site tests and 42-page build, Desktop vet/build, the focused race suite, cache/docs contracts, and an isolated starter install/doctor flow. The intermittent macOS Desktop TempDir cleanup `bad file descriptor` failure also reproduces on an untouched `main-v2@1300dddd5` worktree.

Final CI for `1de1b9934`: **23 successful, 0 failing, 0 pending**. This includes all root and SDK platforms, race, coverage, lint, site, cache/docs impact guards, Desktop macOS/Windows, and CodeQL.

PR #7375, opened 2026-08-04
Preserve parallel subagent results / 保留并行子 Agent 完整结果

SivanCola:
## Summary

- Keep every persisted `parallel_tasks` and `fleet` child answer in its own transcript instead of concatenating all full reports into one truncation-prone tool result.
- Return fair bounded previews plus stable `Subagent reference` values, and add the read-only `read_subagent_result` pager for lossless UTF-8 retrieval.
- Route ordinary parallel tasks through the shared task scheduler while preserving read-only, cancellation, Plan mode, and no-tool direct-answer behavior.
- Restrict result reads to completed children in the current conversation lineage and workspace, with no new setting or recovery step.

## Issues

Related discussion: https://github.com/esengine/DeepSeek-Reasonix/discussions/7240

## Verification

- `go test ./... -count=1`
- `go test -race ./internal/agent -count=1`
- `./scripts/cache-guard.sh` (all scenarios passed; scores 92-95 against a minimum of 90)
- `git diff --check`
- Concurrency audit: **VERIFIED SAFE**. Persisted children use unique refs/files, result reads share the existing per-ref store lock with terminal saves, and cancellation/status accounting remains scheduler-owned.

## Compatibility

| Field or surface | Old-data behavior | Conclusion |
| --- | --- | --- |
| Existing subagent metadata and transcripts | Formats and enum values are unchanged; old records continue to load normally. | Compatible |
| `ProfileExecSpec.AllowNoTools` | Internal, non-persisted field; the zero value preserves the existing fail-closed behavior. | Compatible |
| New result reader | Additive tool over the existing transcript format; older binaries simply do not expose the reader. | Compatible |
| Headless sessions without a persisted parent | Existing ephemeral behavior is retained; they receive bounded previews but cannot mint durable refs. | Compatible, documented limitation |

## Documentation impact

Documentation-impact: updated - Updated the English and Chinese guide and tool-contract documentation, plus the specification and changelog, for bounded previews and lossless subagent result retrieval.

## Cache impact

Cache-impact: low — one additive static read-only tool schema changes the provider-visible prefix once after upgrade; result refs never enter schemas and subsequent bytes remain stable.
Cache-guard: `./scripts/cache-guard.sh` passed all scenarios.
System-prompt-review: Reviewed by @SivanCola — system-prompt bytes are unchanged; the additive `read_subagent_result` schema is static, canonically serialized, and contains no dynamic refs or workspace data.


esengine:
Review status: direction looks right (bounded previews + stable subagent references instead of concatenated truncation-prone results). The only gate is the cache-impact check: the PR touches internal/agent result shaping, so the body needs literal `Cache-impact:` and `Cache-guard:` lines (see scripts/check-cache-impact.sh) describing how the new preview/reference format affects the executor's context bytes across turns — since you designed the result shape, that analysis is yours to state. Add the two lines and CI should go green; happy to merge after that.

PR #7397, opened 2026-08-04
feat(cli): add /voice dictation via OpenAI-compatible transcription

scrappylabsai:
## Summary

- Adds `/voice` (alias `/dictate`): dictate into the composer. Audio is captured locally and sent to an OpenAI-compatible `POST /v1/audio/transcriptions` endpoint, so a local server (Speaches, LocalAI, vLLM, any faster-whisper wrapper) and a hosted provider (OpenAI, Groq, Fireworks) differ only in config, not code.
- **Push-to-talk where the terminal supports it**: hold <kbd>SPACE</kbd> to talk, release to stop, <kbd>Enter</kbd> sends. Voice mode **stays armed after sending**, so the next hold dictates the next turn — a whole conversation can run hands-free. <kbd>Esc</kbd> leaves the mode.
- Holding a key needs key *release* events, which only arrive under the Kitty keyboard protocol. Support is detected at runtime from `KeyboardEnhancementsMsg`, and `/voice` **falls back to a toggle** elsewhere rather than binding a key that could never be let go. `no_push_to_talk` forces the toggle.
- Auth reuses the existing credential machinery via `api_key_env`, so secrets stay out of the config file, plus a `headers` map for gateways with their own scheme. A configured-but-unset credential fails **by name** rather than surfacing as a bare `401`.
- There is deliberately **no default `url`** — guessing localhost fails confusingly, and baking in a vendor is a poor default for self-hosted users. Unset gives an actionable error.
- **Setup discoverability** is the hard part of a feature like this, so `/voice` with no endpoint configured prints a copy-pasteable `[voice]` block (hosted *and* local variants) and points at the docs — rather than naming the missing key and leaving the reader to guess what a valid endpoint looks like:

  ```
  voice needs a speech-to-text endpoint. Add to reasonix.toml:

    [voice]
    enabled     = true
    url         = "https://api.openai.com/v1/audio/transcriptions"
    model       = "whisper-1"
    api_key_env = "OPENAI_API_KEY"

  Or a local server, no key needed:

    url   = "http://127.0.0.1:8000/v1/audio/transcriptions"
    model = "Systran/faster-whisper-large-v3"

  Any OpenAI-compatible endpoint works. Full guide: docs/VOICE.md
  ```

- Docs: `docs/VOICE.md` and `docs/VOICE.zh-CN.md`, linked from both READMEs. Commented `[voice]` block added to `reasonix.example.toml`.

### Design note: three things the key handling has to get right

Each is covered by a test, and each is silent if you get it wrong:

1. **Auto-repeat.** A held key keeps emitting press events. Without filtering on `Key.IsRepeat`, every repeat tears down and restarts the recording — so a hold captures *nothing*.
2. **The space bar stringifies as `"space"`, not `" "`.** `Key.String()` explicitly excludes the space text and falls through to `Keystroke()`. A literal `" "` comparison never matches, and push-to-talk is inert with no error.
3. **`Enter` must fall through to the normal submit path** — except while a hold is still in flight (Enter pressed before the key came up), where the transcript has to land first or the turn is sent without the words just spoken.

While armed, <kbd>SPACE</kbd> is the talk key; other keys still reach the composer, and a modal surface (approval prompt, rewind picker, clear confirmation) keeps ownership of its keys so <kbd>SPACE</kbd> cannot be swallowed away from it.

### Design note: why the whole buffer is re-sent

While recording, the client re-sends the **entire accumulated buffer** each tick and replaces the composer with the response, rather than stitching fragments. That keeps the client simple and lets the model revise earlier words as context arrives (`"Out of voice"` → `"Add a voice command"` once the sentence continues).

This is affordable because Whisper-family encoders always pad to a fixed 30s window, so 1s and 30s cost about the same. Measured end-to-end over HTTP on one consumer GPU running faster-whisper large-v3-turbo:

| Buffer | 1s | 5s | 10s | 20s | 30s |
|---|---|---|---|---|---|
| Latency | 0.128s | 0.138s | 0.150s | 0.188s | 0.297s |

That flatness is a property of the encoder, **not of ASR in general** — an autoregressive speech model measured 0.43s → 4.35s across the same range on the same hardware, which would fall behind a 500ms cadence after ~2s of speech. So the client is guarded to stay correct on any backend:

- partials are clamped to a **30s tail window**; the final pass sends the complete recording
- **single-flight**: a tick arriving while a request is pending is dropped, never queued, so a slow endpoint cannot build a backlog
- responses carry the buffer length they were built from and are **discarded if a newer partial already rendered**
- the poll interval is floored (200ms) and a recording stops itself at `max_seconds`
- a warmup request is sent on start (cold calls measured 569ms vs ~150ms warm)

Raw PCM is buffered and a WAV header synthesized per request, because a WAV written by a still-running process carries wrong RIFF/`data` sizes until it closes.

Capture uses `arecord` on Linux and `sox` on macOS, fully overridable with `record_cmd` (must emit raw s16le/16kHz/mono on stdout) for other platforms or devices.

## Issues

None open on this topic; searched before implementing.

## Verification

- `go build ./cmd/reasonix` — clean.
- `go vet ./...` — clean (the `make hooks` pre-push gate).
- `gofmt -l` — clean.
- `go test ./internal/cli/ ./internal/config/ ./internal/i18n/` — pass, including `TestCatalogsComplete` (new strings added to `en`, `zh`, and `zh-TW`).
- New tests in `internal/cli/voice_test.go`:
  - `TestTranscribeRequestShape` — pins what goes on the wire against an `httptest` stub: `Authorization: Bearer`, custom headers, all form fields (`model`, `response_format`, `language`, `prompt`, `temperature`), and that a user-supplied header cannot clobber `Content-Type` and break the multipart encoding.
  - `TestTranscribeMissingCredential` — a configured-but-unset key errors by name.
  - `TestTranscribeRequiresURL` — unset `url` is a config error, not a dial attempt.
  - `TestWavFromPCMHeader` — RIFF/`data` sizes and format fields.
  - `TestSnapshotTailWindow` — partials take the **newest** audio, not the head.
  - `TestSingleFlightClaim` — claim/release semantics.
  - `TestTranscribeLive` — opt-in (`REASONIX_VOICE_PCM`), runs against a real endpoint and **fails if a warm pass exceeds the poll interval**, which is the precise condition under which partials fall behind.
  - `TestVoiceTalkKeyMatchesSpaceBar` — pins `"space"` against the real bubbletea key types.
  - `TestPushToTalkIgnoresAutoRepeat` — a repeat must not restart the capture.
  - `TestPushToTalkEnterSubmitsAndStaysArmed` / `TestPushToTalkEnterMidHoldCommitsFirst` — Enter submits, except mid-hold.
  - `TestPushToTalkYieldsToModals` — a modal keeps its keys.
  - `TestPushToTalkRequiresTerminalSupport` — never claims the mode without release events.
- Manual: real-time push-to-talk dictation over several turns against a local faster-whisper endpoint, 130–152ms per partial. Verified in **kitty** and **Ghostty**, where key releases are delivered.
- Manual: the toggle fallback, verified in **WezTerm**, which does not deliver key releases. (An earlier revision of this PR credited that WezTerm session as push-to-talk; it was the fallback. Instrumenting the key stream showed WezTerm sending no release events at all — and with `enable_kitty_keyboard = true` it answers the capability query claiming event-type support while still sending none. That measurement is what motivated evidence-based arming and the stuck-hold backstop below.)
- Manual: typing prose while voice mode is armed — spaces land in order and do not open the mic; a deliberate hold does.

## Robustness of the hold

A capability handshake is a claim, not a guarantee, and treating it as one is how `/voice` could
arm a hold that nothing was able to end:

- **Arming waits for evidence.** `KeyboardEnhancementsMsg` reporting event-type support is not
  sufficient; a release must actually have arrived. Typing `/voice` produces several, so this
  costs nothing on a terminal that works.
- **A stuck hold is bounded.** If a hold runs 45s before any release of the talk key has ever been
  seen, the release is not coming: the transcript is committed and dictation drops to the toggle
  for the rest of the session. Retired as soon as one release is observed, so a genuinely long
  hold on a working terminal is unaffected.
- **The fallback is announced.** Silently changing what the keys do is indistinguishable from a
  bug — that is exactly how the terminal issue above presented.
- **A tap of SPACE is still a space.** Voice mode stays armed between turns, so binding SPACE
  outright would make the composer untypable. The key becomes the talk key only when held past
  `ptt_hold_ms` (default 1s). The space is inserted on the press rather than the release, because
  deferring it reorders fast typing, and is retracted if the hold promotes.

## Cache impact

Cache-impact: none — no provider-visible surface is touched. `/voice` is a local TUI command whose output lands in the composer as ordinary user text; it adds no tool, no schema, and no system prompt, memory prefix, output style, or skill index change. Its HTTP client is separate from `internal/provider` and only ever calls `/v1/audio/transcriptions`.
Cache-guard: not applicable — no prompt or tool prefix is constructed. Coverage for the new code is the `internal/cli/voice_test.go` suite above; existing cache guards are unaffected and still pass under `go vet ./...` / `go test`.
System-prompt-review: Requested from CODEOWNERS (@SivanCola, @esengine) — flagged because `internal/config/config.go` is on the system-prompt-sensitive list. The change there is a single additive `Voice VoiceConfig` field (tagged `toml:"voice"`) on the `Config` struct; the new type lives in its own file (`internal/config/voice.go`) and nothing in this PR reads, constructs, or emits the system prompt, memory prefix, output style, or skill index. Please confirm that reading is correct before merge — as an outside contributor I cannot self-approve this gate.

## Documentation impact

Documentation-impact: updated - adds `docs/VOICE.md` and `docs/VOICE.zh-CN.md` (setup for any OpenAI-compatible transcription endpoint, hosted or local; the push-to-talk key model including the `ptt_hold_ms` hold threshold; which terminals actually deliver key releases, measured; and what happens when one does not), and links both from `README.md` and `README.zh-CN.md`. `reasonix.example.toml` gains a commented `[voice]` block. No existing documentation statement is invalidated: `/voice` is additive and off by default.


scrappylabsai:
Updated: `/voice` is now **push-to-talk** where the terminal supports it — hold <kbd>SPACE</kbd> to talk, release to stop, <kbd>Enter</kbd> sends, and voice mode stays armed so the next hold dictates the next turn. A whole conversation runs hands-free. Terminals without key-release events fall back to the previous toggle behaviour automatically.

Three subtle things the key handling needs, now each covered by a test: filtering `Key.IsRepeat` (a held key auto-repeats, and unfiltered every repeat restarts the recording so the hold captures nothing), matching the space bar as `"space"` rather than `" "` (`Key.String()` excludes the space text and falls through to `Keystroke()`, so the literal comparison is silently inert), and letting <kbd>Enter</kbd> fall through to the normal submit path except while a hold is still in flight.

Driven from real use rather than theory — it started as a toggle, and dictating with it for a while made the push-to-talk loop the obvious shape.

Docs updated in `docs/VOICE.md` and `docs/VOICE.zh-CN.md`. `go vet ./...`, `gofmt`, and the package tests are clean.

Thanks for building Reasonix — it's a genuinely nice codebase to work in, and the slash-command and i18n layers made this a small, contained change. 🐋

scrappylabsai:
Updated. Rebased onto current `main-v2`, and the push-to-talk path is now robust against a
terminal that claims key-release support without delivering it.

That case turned out to be real, not theoretical. Instrumenting the key stream showed WezTerm
sending no release events at all, and with `enable_kitty_keyboard = true` it answers the
capability query claiming event-type support while still sending none — so a hold armed and
nothing could end it. An earlier revision of this PR credited a WezTerm session as push-to-talk
testing; it was the toggle fallback. That is corrected in the description.

What changed as a result:

- **Arming waits for evidence**, not for the terminal's claim: a key release must actually have
  arrived. Typing `/voice` produces several, so this is free on a terminal that works.
- **A stuck hold is bounded.** A hold running 45s before any release of the talk key has ever been
  seen commits its transcript and drops to the toggle for the session, instead of recording to
  `max_seconds` with no way to stop. Retired once one release is observed, so a long hold on a
  working terminal is untouched.
- **The fallback says so.** Downgrading in silence is indistinguishable from a bug — which is
  exactly how this presented.
- **A tap of SPACE is still a space.** Voice mode stays armed between turns, so binding SPACE
  outright left the composer untypable. It becomes the talk key only when held past `ptt_hold_ms`
  (default 1s). The space is inserted on the *press* rather than the release, since deferring it
  reorders fast typing, and is retracted if the hold promotes to dictation.
- The recorder is now reaped when its output closes; previously each hold left a zombie for the
  life of the session.
- `Key.IsRepeat` is reported **false** by every terminal measured (kitty, Ghostty, WezTerm), so the
  guard that actually filters auto-repeat is the already-holding check. Comment and test corrected
  rather than left asserting something no terminal does.

`docs/VOICE.md` and `docs/VOICE.zh-CN.md` gain a measured table of which terminals deliver key
releases. Still one commit; `go vet ./...`, the test suites, and both impact guards pass.

scrappylabsai:
Thanks for approving the workflow runs — that surfaced two things.

**Lint (mine, fixed):** 9 `errcheck` hits, all `binary.Write` into a `bytes.Buffer` in the WAV
header builder. Now routed through one explicit `_ =` helper with a note that a `bytes.Buffer`
write cannot fail. Verified locally with the pinned `golangci-lint v2.12.2` against `.golangci.yml`:
**0 issues**.

**`test (ubuntu-latest)`:** `TestUnknownSlashCommandDoesNotStartTurn` failed, and I have not been
able to reproduce it. On this branch it passes 30/30 consecutive runs, and the full `./internal/...`
suite is green locally. The diff is additive — one `builtinSlashSpecs` entry plus new files — and
the `/voice` key handling returns immediately when dictation is inactive, so I can't find a path
from it to that assertion. Happy to dig further if you can point me at something; otherwise a
re-run would tell us more than I can from here.

Nothing else changed in this push.

SivanCola:
Closing as **not planned** for this feedback-convergence round.

**Disposition:** deferred / out of scope
**Reason:** `/voice` dictation is a new product surface (provider transcription, permissions, UX), not a CLI stability/input-contract fix. The frozen 30-day plan explicitly keeps voice out of Integration A–E.

Please reopen or re-target against a future product track if still desired.

scrappylabsai:
Makes sense — thanks for drawing the line explicitly rather than just closing it; "new product surface, not an input-contract fix" is a fair read.

The branch stays up at `scrappylabsai/reasonix` → `feat/voice` (plus a `voice/silence-gate` follow-on that kills a silence hallucination) if a voice track ever opens. Reasonix was a genuinely nice codebase to work in either way.

PR #7405, opened 2026-08-04
fix(desktop): lifecycle-safe subagent cleanup / 修复子代理清理竞态

### LINKED ISSUE #7399: [Bug]: 启动错误：this session is already open in another Reasonix window or still running in the background; close the other window or open a copy
### Version line

### Exact version

1.19.5

### What happened?

Image

### Steps to reproduce

1、存在多个对话，1.19.3还没问题，升级到1.19.5继续多会话执行，突然有一个会话顶部报错，无法继续输入内容。
2、升级到1.19.5时，使用内置的升级报错。手动下载安装包升级的。
3、关闭Reasonix后，再重启开启问题暂时解决。

### OS / platform

Windows 11

### Relevant logs or output

```shell

```

zdjmrq:
## Summary

Fixes the false desktop startup error from #7399 without weakening real cross-process session-lease protection.

- Keep the bounded `ErrSessionLeaseHeld` startup retry (50 ms x 2 retries) as defense against short transient holders.
- Run stale-subagent cleanup on every controller build; do not cache a permanent "already swept" result.
- Add a desktop parent-session liveness probe before cleanup tries the durable parent lease.
- Scope the probe per build: an initial build may repair its own crash leftovers, while concurrent builds skip parents owned or being built by other live tabs.
- Wire the probe into all seven production desktop `boot.Build` call sites. CLI, serve, and ACP keep the existing lease-only behavior through the optional nil callback.

## Root cause

Desktop tabs share one session directory. `CleanupStaleRunning` scans that directory and briefly leases each stale subagent's parent session. A concurrent tab could attempt its startup bind during that short lease window and surface a false "already open in another Reasonix window" error.

The earlier process-wide `sync.Map` mitigation had two correctness problems:

1. It represented both "cleanup in progress" and "cleanup completed" with the same boolean marker, so a concurrent build could return success while the actual sweep later failed.
2. A successful marker lived for the rest of the desktop process, so stale `running` metadata created by another CLI/window crash later in the same process lifetime was never repaired.

## Concurrency guarantees

- A build never treats another live desktop parent as stale before probing its durable lease.
- An initial build does not classify its own unbound tab as a competitor, so its pre-existing crash leftovers are still marked interrupted.
- Cleanup errors are returned by every caller; there is no shared completion marker that can hide an in-flight failure.
- Later controller builds rescan, so crash leftovers created after startup remain recoverable.
- A real holder in another process remains protected by the existing session lease and still produces the same user-facing error after the bounded retry window.

## Compatibility

| Surface | Result |
| --- | --- |
| Persisted subagent metadata | Schema and decoding are unchanged. Existing `running` records retain the same interrupted-recovery behavior. |
| Desktop/CLI coexistence | Later external crash leftovers can be repaired by a subsequent desktop build. |
| Non-desktop frontends | The new boot option is optional; nil preserves lease-only cleanup. |
| Provider-visible prompt/tools | Unchanged. No prompt, tool schema, model input, or cache key changes. |

## Related work

The desktop parent-liveness probe is adapted from #7419 by @siwenyu666-beep. The permanent sweep-once map from the earlier version of this PR is intentionally not retained because it hid concurrent failures and prevented later recovery. The contributor is credited in the commit trailer.

## Verification

On the exact PR head:

- `go test ./internal/agent ./internal/boot -run 'Test(SubagentStoreCleanupStaleRunning|NewSubagentStore)' -count=20`
- `go test -race ./internal/agent ./internal/boot -run 'Test(SubagentStoreCleanupStaleRunning|NewSubagentStore)' -count=5`
- `cd desktop && go test . -run 'Test(SessionParentLive|CleanupStaleRunningWithDesktopParentProbe|WithSessionLeaseContentionRetry)' -count=20`
- `cd desktop && go test -race . -run 'Test(SessionParentLive|CleanupStaleRunningWithDesktopParentProbe|WithSessionLeaseContentionRetry)' -count=5`
- `go test ./...`
- `go vet ./internal/agent ./internal/boot`
- `cd desktop && go vet .`

The final branch also merged cleanly in a local integration check against current `main-v2` (`5635adf7b`). The focused tests, race tests, root full suite, and both vet commands passed on the combined tree.

The desktop full suite was attempted separately and reached its 10-minute package timeout in the pre-existing `TestWorkspaceGitBranch` while waiting for `git init`; that isolated test immediately passed on rerun (`go test . -run '^TestWorkspaceGitBranch$'`). No impacted test failed.

Fixes #7399

Documentation-impact: none - this changes internal lifecycle coordination only; existing user documentation remains correct.

Cache-impact: low - lifecycle cleanup and lease coordination only; no provider-visible content or cache key changes.

Cache-guard: focused, race, full root, and vet coverage passed on the PR head and the latest-base integration tree.

System-prompt-review: no system prompt, tool schema, configuration, or model input content changed.


zdjmrq:
## CI status for reviewers

- **Cache impact guard**: ✅ passed (PR body updated with the three required lines; re-ran on `edited`).
- **CI / CodeQL**: ⏸️ `action_required` — this PR comes from a first-time contributor fork, so GitHub waits for a maintainer to approve the run. The workflow itself has not started yet; nothing is failing.
- The change also touches `internal/boot/*`, which the guard flags as cache/system-prompt-sensitive by directory. A detailed note about the exact scope of the `boot.go` change and its impact has been added to the PR body (see "更详细的说明" section).

Local verification already done on this change (all green):
- `go test ./internal/boot/...` (full) — pass
- `go test ./internal/agent/...` (full) — pass
- desktop `go test . -run "Lease|SessionRuntime|TabBuild|Deferred|Startup|Resume|Rebind|Contention"` — pass
- Two new regression tests: red before the fix (the desktop one reproduced the exact reported error message), green after
- `go test -race` could not run locally (no cgo/gcc on this machine) — CI will cover it

Could a maintainer approve the CI run? Thank you!

Issue #7391, opened 2026-08-04
[Bug]: 更新失败


wmzw1984:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

Reasonix / v1.19.4

### What happened?

软件更新
当前版本：v1.19.4
更新失败：update recovery: could not safely finish the previous update: reconcile pending update: pending update target is outside the current Guard installation

Reasonix 仅检查最新正式版本。

### Steps to reproduce

桌面版本在1.19后无法自动更新，只能在官网下载覆盖安装

### OS / platform

win10

### Relevant logs or output

```shell

```

lukechenchn:
确实   我两个电脑都更新失败

lihaorangood:
在Mac电脑上同样更新失败 更新失败信息如下:
更新失败：prepare update: a pending update already exists

lukechenchn:
windows报错相同  更新失败：prepare update: a pending update already exists

alexzshl:
将报错发给reasonix，他会尝试删除旧的更新文件，然后重启软件。
如果要自己处理，大概位于 ~/AppData/Roaming/reasonix/repair 下的 pending-update.json 和 pending-update.json.lock

zhangyongqiang123:
将pending-update.json备份以下，比如pending-update.jsonbak然后重新下载更新，就不报错了

esengine:
The update dead-end this reports was addressed in two steps: #7366 (v1.19.5) quarantines unusable pending-update debris, and #7376 (merged, next release) recovers valid transactions from the superseded flat install layout and reconciles before download/install so the fail-once/retry loop is gone. If you're currently stuck, download the latest installer once manually from the Releases page — in-app updates work again from there. The remaining edge (fully foreign markers) is tracked with a fix in review (#7431) via #7407/#7416.

SivanCola:
补充：此前评论中提到的“剩余 foreign marker 边角”已随 #7431（`f966dc37f9f5`）合入，并在 **v1.21.1** 正式发布。

过期/不可用的 `pending-update.json` 会自动隔离为 `pending-update.json.unusable-*`，无需再手动删除即可继续应用内更新。

- CLI：https://github.com/esengine/DeepSeek-Reasonix/releases/tag/v1.21.1
- Desktop：https://github.com/esengine/DeepSeek-Reasonix/releases/tag/desktop-v1.21.1
[...]

PR #7431, opened 2026-08-04
fix(updater): quarantine stale self-describing transactions at reconcile

Lxiny-zy:
## Problem

Users on Windows and macOS hit a permanent updater dead end (reported the same day in #7391, #7416, #7407):

```
更新失败：update recovery: could not safely finish the previous update: reconcile pending update: pending update target is outside the current Guard installation
```

Every in-app update fails; reinstalling does not help because the marker lives in the state directory (`%APPDATA%\reasonix\repair\pending-update.json`), not the install directory. The only escape is manually deleting the marker.

## Root cause

`classifyPendingUpdate` returns `pendingUpdateActionable` with a **nil** transaction when the bytes are self-describing but fail environment-relative validation (`validateUpdateTransaction`: launcher and target directories no longer match — the install layout moved to `versions\<version>\`, or the old install directory is gone).

`ReconcilePendingUpdate`'s `case tx == nil` then **failed permanently without touching anything**, and since reconciliation runs before installation, users never reached any other handling. The marker stayed forever and blocked every future update.

## Fix

Both updater entry points now quarantine this disposition — a marker **rename** to `pending-update.json.unusable-<timestamp>`, never a delete:

- `ReconcilePendingUpdate` (`case tx == nil`): quarantines and reports `Pending: true, Cleared: true`, so installation proceeds.
- `ensureNoPendingUpdate` (preparation): applies the same quarantine, so any caller that reaches prepare without reconciliation (or where reconcile was skipped) gets the same exit instead of "a pending update already exists" forever.

Safety:
- The transaction target and all rollback material in the repair directory are **untouched** — the guard from 0bad5567 ("rather than risking real rollback material") is preserved; only the marker moves, and it remains recoverable by hand from the `.unusable-*` file.
- The marker is global and the desktop app is single-instance per `REASONIX_HOME`, so "a genuine transaction observed from the wrong install" can only occur after that install's layout stopped matching; nothing live loses state.

## Tests

- `TestReconcilePendingFileUpdateQuarantinesForeignTarget`: prepares a file update with a matching launcher, then moves the launcher to a different directory (simulating the layout migration), reconciles, and asserts: no error, `Pending && Cleared`, exactly one `.unusable-*` quarantine file, target file and every rollback backup untouched.
- `TestPrepareFileUpdateQuarantinesForeignTarget`: same layout migration, then a new `PrepareFileUpdate` succeeds (marker quarantined first) with exactly one quarantine file.
- Existing prepare rejection test (`pending update already exists`) still passes — it covers a *valid* transaction, which is still refused.

## Verification

- `go test ./internal/repair/` — pass
- `gofmt -l internal/repair/` clean, `go vet ./internal/repair/` clean

Cache-impact: none — `internal/repair` is not on the cache-sensitive path list; no provider-visible system prompt, tool schema, memory prefix, or request serialization changes.
Cache-guard: existing repair suite passes unchanged, plus the two new quarantine regression tests.


esengine:
Review: the direction is right and the reasoning in the comments is exactly the quarantine philosophy #7366 established — with the versioned-layout migration, "not valid for this installation" became a permanent state, so the deliberate refusal I kept in #7366 now blocks forever and quarantining is correct. One sequencing note: #7376 just merged and takes the complementary path first — it *recovers* valid transactions that merely belong to the superseded flat layout (finishing the update instead of shelving it) and proves the running process is the active versioned desktop before archiving anything. Please rebase on current main-v2 so your quarantine becomes the fallback for what recovery can't handle (the truly foreign/unusable markers); the two changes together close the whole #7391/#7416/#7407 class. Happy to merge promptly once rebased and green.

esengine:
(follow-up) #7376 is merged now — rebase when ready and I'll merge promptly.

Lxiny-zy:
Rebased onto current main-v2 (467a46f8, includes #7376). The quarantine here now serves as the fallback for markers that #7376's recovery cannot handle, closing the #7391/#7416/#7407 class together.

Local verification before CI:
- `go test -count=1 ./internal/repair/ ./internal/cli/ ./internal/boot/` — pass
- `go vet ./internal/repair/` + `gofmt -l internal/repair/` — clean
- `go build ./cmd/reasonix` — pass
- check-cache-impact.sh / check-docs-impact.sh — pass (repair is not in the sensitive lists)

The CI and CodeQL workflows are waiting for approval (`action_required`) as this is a fork PR. Could you approve them when convenient? Thanks.

Lxiny-zy:
Hi @esengine — just a gentle bump. The branch is rebased on current main-v2 (467a46f8) and local checks are green; CI and CodeQL are still waiting on approval (`action_required`) as this is a fork PR. Whenever you have a moment to approve the run, that would be great. Thanks!

SivanCola (approved):
Reviewed at head 908eeb2dd after merging current main-v2 ee8a85ec. The stale-marker quarantine path is now serialized under the pending lock with marker digest revalidation, and only explicit foreign-install validation failures are quarantined. Other validation failures remain fail-closed. Verified: go test -count=1 ./internal/repair, go test -race -count=1 ./internal/repair, go test -count=1 ./..., go vet ./internal/repair, gofmt, git diff --check; required label/cache/docs checks pass.

PR #7443, opened 2026-08-04
fix(desktop): persist provider draft across settings panel close/reopen

### LINKED ISSUE #4549: [Bug]: Clicking off the settings panel closes it
### Version line

### Exact version

v1.6.0-rc.1

### What happened?

Clicking off the settings panel closes it and no fields save, absolutely infuriated when setting a provider, typing hall the stuff out, changing from 100 models loaded to the few I want, and then clicking off of the modal and everything closing.

I'm using an older version so it may be fixed, but the newer versions are even more bug ridden. Slow your roll, you're vibing too fast.

### Steps to reproduce

Use the app

### OS / platform

Windows 11

### Relevant logs or output

```shell

```

ashishexee:
Closes #4549

### Problem

When the settings panel is open and a user is filling out the provider editor (14+ fields: name, kind, base URL, API key, model list, vision models, headers, extra body, context window, reasoning protocol, etc.), a stray click on the backdrop or an accidental Esc press closes the panel immediately. The panel uses conditional rendering (`{settingsTarget !== null && <SettingsPanel .../>}`), so closing unmounts the entire component tree, destroying all `useState` local state. Half an hour of work is gone with no confirmation.

### Solution

Two changes, matching the repo owner's scope in the issue:

**1. Provider draft survives close/reopen (priority 1)**
Lifted all `ProviderEditor` form fields + `modelDrafts` + `addProviderMode` from local `useState` into a `settingsDrafts` namespace on the existing Zustand overlay store (`overlays.ts`). The store is a module-level singleton — it survives panel unmount. On remount, `ProviderEditor` detects a matching stored draft (keyed by `editingProviderName`) and restores it. The draft is cleared only on **(a) successful save** (gated on `apply()` returning `true`) or **(b) explicit discard** (Cancel button, or "Discard draft" in the confirmation dialog).

**2. Dirty guard on close (priority 2)**
Backdrop click (`onMouseDown`) and the Esc keydown listener now go through `guardedClose()`. It reads the store draft with `getState()` (no subscription, no re-renders) and compares against `EMPTY_PROVIDER_DRAFT` via `JSON.stringify`. If there are unsaved edits → a `ConfirmDialog` appears with "Discard draft" (danger) / "Keep editing". If clean → closes silently. The header X button still closes directly (intentional user action).

### Files changed

- `desktop/frontend/src/store/overlays.ts` — new `SettingsProviderDraft`, `ProviderModelDraft`, `SettingsDrafts`, `EMPTY_PROVIDER_DRAFT` types; `settingsDrafts` field + `setSettingsDrafts` setter on the overlay store
- `desktop/frontend/src/components/SettingsPanel.tsx` — dirty guard hook, `useConfirmDialog` integration, `ProviderEditor` reads/writes store draft, `ProvidersSection` uses store for model drafts and add-provider mode, caller `.then()` chains gate draft clearing on `apply()` success
- `desktop/frontend/src/locales/en.ts` — 4 new i18n keys
- `desktop/frontend/src/locales/zh.ts` — same (Simplified Chinese)
- `desktop/frontend/src/locales/zh-TW.ts` — same (Traditional Chinese)

### Design decisions

- **Store, not CSS-mount.** Zustand store is simpler than keeping all tabs mounted with `display: none`, avoids pausing timers/managing focus, and adds no cross-file changes to CapabilitiesPanel/MemoryPanel.
- **"Discard / Keep Editing" dialog, no "Save".** Triggering save from a close-interceptor is an architectural trap — validation errors have nowhere to render since the modal is trying to unmount. Users keep editing and click the native Save button.
- **Draft dies with the process.** No `zustand/persist` middleware — plain in-memory store. API keys in the draft stay in JS heap only.
- **Esc in confirm dialog is safe.** `ConfirmDialog` registers its Esc listener at `capture: true` with `stopPropagation()`, so the settings panel's global Esc handler never sees it.

### Edge cases handled

- **Edit vs Create:** `editingProviderName` in the stored draft matches `initial?.name`. A draft for provider "openai-proxy" won't restore when editing "deepseek-gateway".
- **Save failure:** Callers gate `.then((ok) => { if (!ok) return; ... })` on `apply()`'s boolean result — draft is only cleared on success.
- **Model picker draft:** The curated model list after fetching (selected / vision models) is stored alongside the form, not lost.

Cache-impact: none — frontend-only changes, no cache-sensitive Go paths touched.
System-prompt-review: n/a — no config/memory/skill/boot Go code changes.

Documentation-impact: none - no docs/*.md files changed; new UI behavior (dirty-close guard) is self-evident from the confirmation dialog and requires no documentation updates.


maxitoon:
Reviewed head `d6eedd4b80c387efb9083980857df4b150b7893b` against the [owner's scope](https://github.com/esengine/DeepSeek-Reasonix/issues/4549#issuecomment-5176228166). The draft-retention direction makes sense, but these source-level blockers remain:

1. **Blocking — curated selection is not restored.** [`SettingsPanel.tsx:6191–6194`](https://github.com/esengine/DeepSeek-Reasonix/blob/d6eedd4b80c387efb9083980857df4b150b7893b/desktop/frontend/src/components/SettingsPanel.tsx#L6191-L6194) initialises `models` from `modelCandidates`, ignoring the separately stored `models`. Fetching a list, deselecting entries, then remounting restores all candidates. Restore the selected text from `resolvedDraft.models` (including an intentionally empty selection); retain candidates separately.
2. **Blocking — populated is not dirty.** The [global-empty comparison](https://github.com/esengine/DeepSeek-Reasonix/blob/d6eedd4b80c387efb9083980857df4b150b7893b/desktop/frontend/src/components/SettingsPanel.tsx#L155-L170), combined with [unconditional editor sync](https://github.com/esengine/DeepSeek-Reasonix/blob/d6eedd4b80c387efb9083980857df4b150b7893b/desktop/frontend/src/components/SettingsPanel.tsx#L6241-L6270), flags an untouched existing provider as dirty. Please scope the retained draft to its owning flow and compare with its initial baseline/explicit dirty state, so read-only visits close silently.
3. **High — draft/key transitions need a defined owner.** At [`5052–5066`](https://github.com/esengine/DeepSeek-Reasonix/blob/d6eedd4b80c387efb9083980857df4b150b7893b/desktop/frontend/src/components/SettingsPanel.tsx#L5052-L5066), switching custom → official retains the custom draft; official/preset completion does not clear it, while opening a preset conflict mounts an existing editor that overwrites the singleton. In-memory key retention is not itself a leak, but retain/clear/replace semantics need explicit coverage.

**Current-main correction:** `main-v2` is now `4d4700a39af7fcc695581e08930325246bbf0952`: settings uses a full-window [`ManagementPageShell` with `onBack`](https://github.com/esengine/DeepSeek-Reasonix/blob/4d4700a39af7fcc695581e08930325246bbf0952/desktop/frontend/src/components/SettingsPanel.tsx#L404-L406), not this PR's backdrop modal. Please reconcile the draft lifetime with that surface rather than restoring obsolete modal handlers; maintainer confirmation of the current dismissal contract would help. GitHub reports merge conflicts. #8637 overlaps URL initialisation/save semantics; #5707 overlaps key-save failure handling. #9563 is **merged**, and its `noProxy` field is already on main, so the draft shape must accommodate it.

Minimal acceptance checklist:
- [ ] Update against current main and confirm the intended navigation/dismissal contract.
- [ ] Test close/reopen with a trimmed **and empty** model selection; untouched versus dirty flow; discard versus keep-editing.
- [ ] Test save failure retention, successful-save/discard clearing, and custom/official/existing-provider transitions without cross-flow draft/key reuse.
- [ ] Add focused behavioural coverage and run frontend typecheck/tests on the updated head.

Verification here is static only: seven source assertions confirm the patterns above, not behavioural correctness. The PR changes no test files; reported successful checks are only `label`, `cache-impact`, and `docs-impact`. No frontend or Windows runtime result is claimed.

Issue #7265, opened 2026-08-03
[Feature]: option to keep thinking section collapsed while streaming (desktop)


Sneexy:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### What problem does this solve?

On the desktop app, the collapsible "thinking" section is force-expanded while the response is streaming, regardless of the expand_thinking = false setting. The setting only takes effect after generation finishes.

This is caused by the frontend logic in Message.tsx (from PR #3396):
ProcessCard defaultOpen={item.streaming || defaultExpanded}
//                         ^ always open while streaming    ^ setting only applies after done

For users who prefer a clean reading experience (especially when using a fast model like deepseek-v4-flash, where reasoning streams continuously on every tool-call round), the live-scrolling thinking text is distracting and there is currently no way to hide it during generation.

### Proposed solution

Requested behavior

Add an option (or change the existing one) so that the thinking section can be fully collapsed during streaming too, for example:

expand_thinking = false → thinking stays collapsed both during and after streaming, or
a new config like show_thinking_during_stream = false, or
make the expand state purely follow the user's default (defaultExpanded) without the item.streaming override.

lzxb:
Same problem here, and it's more than a reading-experience issue — it causes real UI lag.

With a chatty reasoning model (e.g. deepseek-v4-flash emitting reasoning on every tool-call round), the force-expanded thinking section re-renders on every token: thousands of DOM nodes plus the GSAP collapse animation trigger layout on each chunk, and the transcript becomes visibly janky within a few turns. The only workarounds right now are manually collapsing the section once per turn (works thanks to the `userOverridden` flag) or switching to a non-reasoning model.

From the current main-v2 code:
- `TurnCollapse` force-expands while `hasRunningWork` (`setOpen(true)`), ignoring the user's collapse preference until the turn ends.
- The old `expand_thinking` config still exists in the Go config/bridge layer (`desktop/settings_app.go`) but no longer has any frontend consumer after the TurnCollapse refactor, so there's no setting that affects this at all.
- `truncateStreamingReasoning` / `reasoningDisplay="hide"` already exist but are only wired to the internal `displayMode = "compact"` path — not exposed as a user option.

So a `show_thinking_during_stream = false` option (or making the streaming auto-open respect the fold preference) would fix both the distraction and the performance problem. Happy to help test a PR.

bferanmi806-sketch:
I’d like to work on this.

[...]

PR #7518, opened 2026-08-04
feat(desktop): add three-state process fold preference

bferanmi806-sketch:
## Summary

References #7265.

Documentation-impact: none - this changes an existing in-app setting and its locale strings; no standalone documentation covers the old two-state behavior.

- Add explicit `auto`, `collapsed`, and `expanded` work-process-fold preferences.
- Keep collapsed folds closed during streaming and after completion when answer content is visible outside the fold.
- Preserve the safety rule that completed process-only turns stay visible, including protection against manually hiding the only visible content.
- Apply live preference changes to existing folds, add the collapsed settings option, and update English, Simplified Chinese, and Traditional Chinese labels.
- Extend transcript regression coverage for streaming, completion, live changes, stable transition state, and process-only turns.

## Verification

Passed:

- `corepack pnpm exec tsx src/__tests__/transcript-process-fold.test.ts`
- `corepack pnpm exec tsx src/__tests__/transcript-fold-preference.test.tsx`
- `corepack pnpm exec tsx src/__tests__/settings-refresh-snapshot.test.tsx`
- `corepack pnpm typecheck`
- `corepack pnpm lint:hooks`
- `corepack pnpm check:css`
- `corepack pnpm exec vite build`
- `corepack pnpm exec node scripts/check-bundle-budget.mjs`

The full frontend test command was also attempted. Its lifecycle wrapper cannot resolve a `pnpm` shim in this environment, so the exact package `test` command was run directly; it reached one existing failure in `pending-prompt-stale-status.test.tsx` (`expected "plan-zombie", got undefined`), reproduced when that test is run in isolation. That test path does not import the modified transcript/settings modules.


bferanmi806-sketch:
This was fun to work on thanks for the opportunity to contribute. I’m happy to help with any follow-up changes, testing, or related issues if needed.

bferanmi806-sketch:
Thanks for taking a look. Just to clarify, this PR addresses the specific gap mentioned in #7265: even when the thinking section is configured to default collapsed, it still force-opens while work is streaming.

The branch is now updated with the latest `main-v2`, the merge conflict is resolved, and the focused checks pass. The implementation preserves the existing `auto` behavior while adding an explicit `collapsed` option for users who want the process fold to remain closed during streaming.

Happy to make any adjustments if you would prefer a narrower approach.

PR #7587, opened 2026-08-05
feat(cli): allow hiding per-turn usage receipts

azhonglee:
## What changed

- add `[ui].show_turn_usage` with a backward-compatible default of `true`
- hide TUI per-request token/cost receipt blocks when configured `false`
- keep token accounting and live status updates active while receipts are hidden
- document the setting and cover config round-tripping plus TUI behavior

## Why

Long TUI sessions can accumulate noisy per-request usage bands. Users can now opt out of those transcript entries without losing usage accounting.

## Validation

- `go test ./internal/config ./internal/cli`
- `go build ./cmd/reasonix`
- `go test ./...` (all packages passed except the pre-existing `cmd/reasonix-legacy-migrator` `TestMigrateFlatInstallToVersioned` failure; reproduced unchanged on `main-v2`)
- `git diff --check`

## Impact declarations

Cache-impact: none - presentation-only UI config; no system prompt, tool schema, provider request, or cache-prefix serialization changes.
Cache-guard: `go test ./internal/config ./internal/cli` covers config defaults, round-tripping, hidden receipt behavior, and retained usage accounting.
System-prompt-review: Reviewed `internal/config/config.go`; only `UIConfig` and its default changed, while `DefaultSystemPrompt` and prompt assembly remain unchanged.
Documentation-impact: updated - documented `ui.show_turn_usage` in the Guide, Configuration Paths, SPEC, example config, and changelog.


SivanCola:

Kept: `[ui].show_turn_usage` (default true) hides transcript receipts without disabling usage accounting.

PR #7600, opened 2026-08-05
fix(bash): restore OS account home as subprocess HOME

### LINKED ISSUE #7331: [Bug]: CLI tools report "not logged in" when Reasonix runs with an overridden HOME (service/supervised deployments)
## What happened

When the Reasonix process runs with `HOME` pointed at a config directory (e.g. a systemd unit or a launcher that forces `HOME` for config isolation), every bash-tool subprocess inherits that `HOME`. CLIs that resolve credentials via `/root` — `gh` (`~/.config/gh`) and similar — then look in the wrong place and report "not logged in", even though the credentials exist under the OS account's real home.

## Expected vs actual

- Expected: tool subprocesses see the account's real home (or at least a documented/configurable behavior), so CLIs keep working regardless of Reasonix's own `HOME`.
- Actual: subprocess `HOME` equals the overridden one; credentials are invisible.

## Steps to reproduce

1. `HOME=/some/config/dir reasonix serve ...` (or any launcher that overrides `HOME`).
2. In a session, run `gh auth status`.
3. Result: "not logged in", although `/root/.config/gh/hosts.yml` exists under the account home. Works fine when Reasonix runs with a normal `HOME`.

## Proposed fix (we already run this in our fork)

In the bash tool's subprocess environment construction, point `HOME` at the OS account home from the user database (`os/user`, not `/root`) when that directory exists, and keep the inherited value untouched otherwise (no `HOME` in the parent env, account home missing on disk, or already equal). Only the bash tool's env is affected; Reasonix's own config isolation (`REASONIX_HOME` etc.) is unchanged.

Reference implementation: GOGV/DeepSeek-Reasonix-Fork commit `b3324cc` (feat(bash): restore the OS account home as subprocess HOME), with unit tests.

**Question to maintainers:** acceptable as default behavior, or should it sit behind a config flag (e.g. `[tools.shell] restore_home = true`)? Happy to open a PR with the patch either way.

## Environment

- Reasonix 1.19.3 (also reproduced on 1.17.21)
- go1.26.5, Linux x86_64, reasonix serve under systemd with `Environment=HOME=/root/.config/reasonix`

skyzhao1223:
## Summary

- Service/supervised launches often set \`HOME\` to a config directory for Reasonix isolation.
- Bash tool subprocesses inherited that value, so CLIs like \`gh\` looked under the wrong home and reported \"not logged in\".
- Restore the OS account home from the user database for bash env when that directory exists; leave inherited \`HOME\` untouched otherwise. Reasonix's own \`REASONIX_HOME\` / config isolation is unchanged.

## Issues

Fixes #7331

## Verification

- \`go test ./internal/tool/builtin/ -run 'TestRestoreHomeEnv|TestBashCommandEnv'\`

## Documentation impact

Documentation-impact: none - subprocess environment fix; existing docs about REASONIX_HOME / config isolation remain correct.

## Cache impact

Cache-impact: none — bash subprocess environment only.
Cache-guard: N/A
System-prompt-review: N/A

### DIFF
--- a/internal/tool/builtin/bash.go
+++ b/internal/tool/builtin/bash.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"os/user"
 	"path/filepath"
 	"runtime"
 	"strings"
@@ -470,7 +471,7 @@ func commandPreview(cmd string) string {
 }
 
 func bashCommandEnv(ctx context.Context) []string {
-	env := secrets.ProcessEnv()
+	env := restoreHostHomeEnv(secrets.ProcessEnv())
 	if runtime.GOOS == "windows" {
 		return env
 	}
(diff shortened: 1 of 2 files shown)


SivanCola:
Closing as **not adopted** for the CLI stability batch.

Reasonix keeps subprocess `HOME` as inherited. Automatic account-home rewrite (#7600) can hide service/supervisor misconfiguration and fights `REASONIX_HOME` isolation.

Instead, Integration PR A (#7663):
- leaves bash/subprocess `HOME` untouched
- has `reasonix doctor` warn when process HOME differs from the OS account home while `REASONIX_HOME` is unset, and recommends real account HOME for services plus `REASONIX_HOME` for Reasonix isolation

Thank you for the investigation — the doctor guidance lands via #7663.

PR #7601, opened 2026-08-05
fix(config): timeout legacy keyring lookups on startup

### LINKED ISSUE #7507: [Bug]: CLI hangs forever (zero output) when session D-Bus is unhealthy — startup bus call has no timeout
### Version line

### Exact version
reasonix v1.19.6 (also reproduced on v1.19.4)

### What happened?
Every execution path — `reasonix run --auto "task"`, `reasonix -p "task"`, and the bare interactive TUI — hangs forever with **zero output** (~0% CPU, stuck process) when the user D-Bus session bus is unhealthy. Metadata commands (`--version`, `--help`, `session list --json`, `hook list --json`, `session recovery --json`) all respond fine.

Root cause (isolated via /proc inspection, no strace): the native CLI binary opens an ESTABLISHED unix-socket connection to the user session bus (`$XDG_RUNTIME_DIR/bus`, dbus-broker) at startup, then parks **all 11 threads in futex_do_wait** with the tokio event loop in ep_poll — waiting for a synchronous D-Bus reply that never arrives. In our environment the session bus was unhealthy: `gnome-keyring-daemon` was stuck in an error loop (`failed to read or parse keyring: login.keyring`, `Too many open files`, `couldn't create credential: Internal error`) and never answered the call. The CLI has no config kill-switch for D-Bus integration (notifications are already disabled; secrets come from `api_key_env`/.env, not the keyring), and there is no timeout on the startup bus interaction, so a broken/unresponsive service on the bus blocks the whole binary indefinitely.

Expected behavior: fail fast. If the session bus (or a service on it) is unavailable/unresponsive, reasonix should skip D-Bus integration and proceed, or surface a clear error — never block startup forever on a synchronous call with no timeout.

### Steps to reproduce
1. Environment where the session D-Bus is present but a service is unresponsive (our case: gnome-keyring-daemon in an error loop on the same bus; a broken/never-answering activatable service should also trigger it).
2. `reasonix run --auto "Reply PROBE_OK"` → hangs with zero output indefinitely (also `reasonix -p`, bare `reasonix` TUI).
3. Now force the bus away: `DBUS_SESSION_BUS_ADDRESS=unix:path=/tmp/nonexistent_dbus_bus reasonix run --auto "Reply PROBE_OK"` → **works** (exit 0, correct answer, ~1-2s, full `--metrics`/`--output-format json` output). The binary degrades gracefully only when the connect itself fails; it does not degrade when the connect succeeds but the call never returns.
4. Note: `DBUS_SESSION_BUS_ADDRESS=` (empty string) does NOT work — libdbus treats it as unset, rediscoveres the real bus, hangs again. A dead path is required.

### OS / platform
Fedora 44 x86_64, headless agent host (no desktop app installed), Go-rebuilt npm-distributed CLI (`@reasonix/cli-linux-x64`)

### Relevant logs or output
```
# hung process state (native binary):
#   node wrapper in do_wait; native binary 11 threads, ALL futex_do_wait except
#   event loop tid in ep_poll; fd 6 = u_str ESTAB → dbus-broker on /run/user/1001/bus
# journalctl --user (same window):
gnome-keyring-daemon[20782]: couldn't create credential for collection: Internal error
gnome-keyring-daemon[20782]: failed to read or parse keyring: /home/hermes/.local/share/keyrings/login.keyring
gnome-keyring-daemon[20782]: problem reading keyring: .../login.keyring: Failed to open file "...": Too many open files
gnome-keyring-daemon[20782]: couldn't create credential: Internal error

# with workaround (DBUS_SESSION_BUS_ADDRESS=unix:path=/tmp/nonexistent_dbus_bus):
$ reasonix run --auto --metrics m.json --output-format json 'Reply with exactly: PROBE_OK'
{"type":"result","subtype":"success","is_error":false,"duration_ms":1304,"num_turns":1,
 "result":"PROBE_OK","session_id":"20260804-161934.623488730-deepseek-v4-flash",
 "total_cost":0.00005399,"currency":"USD",
 "usage":{"input_tokens":13272,"output_tokens":17,"cache_read_input_tokens":13184,"cache_creation_input_tokens":88}}
```

### Suggested fix directions
1. Wrap the startup D-Bus interaction in a timeout; on failure/timeout, log and continue without D-Bus integration (the binary already has a working no-dbus fallback — proven by the dead-address workaround).
2. Honor an explicit opt-out (env var like `REASONIX_NO_DBUS=1` or a config key) so headless/agent hosts can skip the bus entirely.
3. Investigate whether the blocking call targets a specific activatable service — a never-answering activatable service is a general hazard for any caller.

skyzhao1223:
## Summary

- Startup legacy-credential migration calls into the OS keyring / Secret Service with no deadline.
- An unhealthy D-Bus session bus (stuck gnome-keyring, etc.) could hang every \`reasonix\` execution path with zero output.
- Bound each legacy keyring probe with a short timeout and fail closed (treat as missing) so the CLI can start.

## Issues

Fixes #7507

## Verification

- \`go test ./internal/config/ -run TestLegacyKeyringCredentialValueWithTimeout\`

## Documentation impact

Documentation-impact: none - internal startup resilience; no user-facing CLI/config surface change.

## Cache impact

Cache-impact: none — credential migration path only.
Cache-guard: N/A
System-prompt-review: N/A

### DIFF
--- a/internal/config/credentials.go
+++ b/internal/config/credentials.go
@@ -8,6 +8,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/joho/godotenv"
 
@@ -63,7 +64,37 @@ var credentialSourceTracker = struct {
 var userCredentialEditMu sync.Mutex
 
 var storedCredentialValueLookup = storedCredentialValue
-var legacyKeyringCredentialValueLookup = legacyKeyringCredentialValue
+var legacyKeyringCredentialValueImpl = legacyKeyringCredentialValue
+var legacyKeyringCredentialValueLookup = legacyKeyringCredentialValueWithTimeout
+
+// legacyKeyringLookupTimeout bounds Secret Service / platform keyring probes
+// during legacy credential migration. An unhealthy session bus (stuck
+// gnome-keyring, etc.) must not hang CLI startup indefinitely (#7507).
+var legacyKeyringLookupTimeout = 2 * time.Second
+
+// legacyKeyringCredentialValueWithTimeout wraps the platform keyring lookup so
+// a stuck D-Bus reply fails closed (treat as missing) instead of blocking the
+// process. A hung probe may leave a short-lived goroutine until the bus
+// answers; that is preferable to never starting.
+func legacyKeyringCredentialValueWithTimeout(key string) (string, bool) {
+	type result struct {
+		value string
+		ok    bool
+	}
+	ch := make(chan result, 1)
+	go func() {
+		value, ok := legacyKeyringCredentialValueImpl(key)
+		ch <- result{value: value, ok: ok}
+	}()
+	timer := time.NewTimer(legacyKeyringLookupTimeout)
+	defer timer.Stop()
+	select {
+	case r := <-ch:
+		return r.value, r.ok
+	case <-timer.C:
+		return "", false
+	}
+}
 
 // CredentialResolver resolves credentials repeatedly for one caller-owned view
 // build. It keeps expensive global credential-store lookups bounded to one per
--- a/internal/config/credentials_keyring_timeout_test.go
+++ b/internal/config/credentials_keyring_timeout_test.go
@@ -1,0 +1,52 @@
+package config
+
+import (
+	"testing"
+	"time"
+)
+
+func TestLegacyKeyringCredentialValueWithTimeout(t *testing.T) {
+	oldImpl := legacyKeyringCredentialValueImpl
+	oldTimeout := legacyKeyringLookupTimeout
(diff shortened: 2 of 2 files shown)


SivanCola:
Superseded by #7663 (merge `84988e8cdb5a33087228bb8b0a82fef38eff8d06`).

Kept/adapted (@skyzhao1223): bounded legacy keyring probes under a shared 1s budget with four-state outcomes (reworked as private in-process CallWithContext; no external helper).

Thank you for the contribution. The landed work was re-implemented on current `main-v2`, verified with focused package tests, and credited in the integration PR.

PR #7606, opened 2026-08-05
fix(cli): only Ctrl+C aborts a running turn

rixzkiye:
Esc was both "close the menu" and "cancel the run", so a stray Esc — easy
to bump while reaching for arrows or backspace — could kill a streaming
turn. Ctrl+C requires two keys by design, so it's the deliberate gesture
for abort.

This makes Esc abort nothing while a turn is running. The user has to
press Ctrl+C to actually cancel; a stray Esc can't kill the turn.

### Behavior

- Esc while a turn is running: only un-sends a just-sent bubble (server
  has not replied yet). Never calls `ctrl.Cancel()` on a streaming turn.
  The completion menu still swallows Esc to close itself, as before.
- Esc while idle: unchanged — clears non-empty input, double-Esc opens
  the rewind picker.
- Ctrl+C while running: unchanged — first press calls `ctrl.Cancel()`,
  second press within the timeout shuts down.
- Footer hint during a run switches from "Esc cancels" to "Ctrl+C
  cancels" in en, zh, and zh-tw.

### Diff stat

```
 internal/cli/chat_tui.go              | 28 ++++++++-------
 internal/cli/chat_tui_test.go         | 71 +++++++++++++++++++++++++++++-----
 internal/i18n/messages_en.go          |  2 +-
 internal/i18n/messages_zh.go          |  2 +-
 internal/i18n/messages_zh_tw.go       |  2 +-
```

### Tests

- `TestEscDoesNotCancelRunningTurn` (table-driven, completion menu
  open / no completion menu) asserts the turn survives Esc.
- `TestCtrlCCancelsRunningTurnWithCompletionOpen` replaces the prior Esc
  regression test with the Ctrl+C counterpart, so the original report
  (completion menu swallowed the cancel key) is still covered — just for
  the key that still cancels.

### Verification

- `go build ./...`
- `go vet ./internal/cli/... ./internal/i18n/...`
- `go test ./internal/cli/ -short -count=1` → PASS (5.1s)

### DIFF
--- a/internal/cli/chat_tui.go
+++ b/internal/cli/chat_tui.go
@@ -1267,10 +1267,11 @@ func (m chatTUI) update(msg tea.Msg) (tea.Model, tea.Cmd) {
 				m.acceptCompletion()
 				return m, nil
 			case "esc":
+				// Esc always dismisses the completion menu. Cancel/abort during a
+				// running turn is Ctrl+C's job (see the main Esc handler); the
+				// main handler intentionally no longer cancels a running turn
+				// so a stray Esc can't kill it.
 				m.completion = completion{}
-				if m.state == tuiRunning {
-					break // a turn is running — also cancel it via the main Esc handler
-				}
 				return m, nil
 			}
 		}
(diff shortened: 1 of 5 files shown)


SivanCola:
Closing as **not planned**.

**Disposition:** not adopted
**Related:** Integration B https://github.com/esengine/DeepSeek-Reasonix/pull/7690 (merge `835c827de`)

Esc keeps existing cancel / back-out semantics for running turns and overlays. Restricting abort to Ctrl+C only would be a breaking interaction change for users who rely on Esc during streaming; B explicitly preserved that contract.

Thank you @rixzkiye for the report and discussion.

PR #7608, opened 2026-08-05
fix(cli): let run --resume accept machine session IDs

### LINKED ISSUE #7429: One-shot run cannot resume from redacted machine session ID
## Environment

- Reasonix: v1.19.4
- macOS arm64
- One-shot automation mode: `reasonix run --events-jsonl`

## Problem

The redacted machine API returns a stable `machine-session-id`, but one-shot resume only accepts a session file path. This makes deterministic unattended resume difficult without depending on Reasonix's private on-disk layout.

## Reproduction

1. Run a one-shot task with `reasonix run --events-jsonl "<task>"`.
2. Read the terminal `run_done.sessionId` value, for example `session_<opaque>`.
3. Confirm the ID is valid with `reasonix session show session_<opaque> --json`.
4. Try `reasonix run --resume session_<opaque> --copy "<repair task>"`.
5. The command exits 1 with `open session_<opaque>: no such file or directory`.
6. Repeat step 4 with the exact session JSONL file path. The copied resume succeeds.

No real session IDs or local paths are included here.

## Expected

Please provide one automation-safe option:

- allow `reasonix run --resume <machine-session-id>`, or
- add a redacted resolver that returns a reusable one-shot resume handle/path, or
- include a reusable one-shot resume handle in the terminal `run_done` event.

An opaque handle would preserve the current privacy goal better than requiring callers to scan the private session-store layout.

## Additional observation

Interactive/query resume accepts machine IDs, while the v1.19.4 one-shot contract accepts only `PATH`. Making that distinction explicit in CLI errors would also help automation wrappers fail with a clearer diagnosis.

skyzhao1223:
## Summary

- One-shot `reasonix run --resume` treated its value as a filesystem path only, so opaque machine session IDs (`session_<hex>` from `--events-jsonl` / `session show --json`) failed with `open …: no such file or directory`.
- Route `run --resume` through `resolveSessionQuery`, and teach that helper to map machine session IDs with the same HMAC identity used by the machine session API (interactive `--resume` benefits too).
- Update CLI docs so one-shot resume is no longer documented as PATH-only.

## Issues

Fixes #7429

## Verification

```
go test ./internal/cli/ -count=1 -run 'TestResolveSessionQuery'
```

## Documentation impact

Documentation-impact: updated - `docs/CLI.md` and `docs/CLI.zh-CN.md` now document that `run --resume` accepts a path, session ID, or machine session ID.

## Cache impact

Cache-impact: none - CLI resume path resolution only; no provider prompt or tool schema change.
Cache-guard: N/A
System-prompt-review: N/A

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -138,7 +138,7 @@ echo "explain this code" | reasonix run
 
 `reasonix run` keeps the normal streamed terminal presentation unless `-p` or a
 structured output format is selected. It also accepts `--model`, `--profile`,
-`--max-steps`, `--effort`, `--dir`, `--add-dir`, `--continue`, `--resume PATH`,
+`--max-steps`, `--effort`, `--dir`, `--add-dir`, `--continue`, `--resume QUERY`,
 `--copy`, `--allowed-tools`, `--permission-mode`, and `--auto` / `-y` (an alias
 for `--permission-mode auto`).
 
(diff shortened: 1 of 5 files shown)


SivanCola:

Kept: `run --resume` resolves opaque machine session IDs; CLI docs use QUERY.

PR #7762, opened 2026-08-06
perf(cli): cache per-block wrapped transcript lines

innocarpe:
## Summary

Second TUI performance fix (follow-up to #7745): cache per-block wrapped transcript lines in `internal/cli/` so dirty frames no longer re-wrap the ENTIRE transcript (spinner ticks, stream updates, transcriptDirty paths were O(transcript) lipgloss wraps per frame).

**Equivalence gate first** — a randomized differential test (`transcript_wrap_test.go`) compares whole-transcript wrap vs per-block wrap across ANSI/unicode/trailing-newline/unbalanced-SGR inputs. It exposed a real divergence: lipgloss pads blocks to the widest line, so per-block results can be shorter than whole-transcript results — fixed by re-padding each cached block to its widest line (byte-identical output for the actual renderers, which all close their SGR at block ends).

**Cache**: `wrapCache map[int]string` keyed by block index, invalidated on transcript mutation (append/update/remove and reflow paths) and dropped on width change. Join-only per frame; wrap only changed blocks.

## Issues

None — performance, no issue report.

## Verification

- Equivalence test: randomized differential passes for all-balanced blocks; divergence case documented (widest-line padding) and handled
- `go test ./internal/cli/` — full suite pass
- `go vet` / `gofmt -l` / `go build ./cmd/reasonix` — clean

## Documentation impact

Documentation-impact: none - internal rendering; no behavior change.

## Cache impact

Cache-impact: none - TUI rendering only; no prompt/tool surface touched.
Cache-guard: N/A
System-prompt-review: N/A

### DIFF
--- a/internal/cli/chat_tui.go
+++ b/internal/cli/chat_tui.go
@@ -218,8 +218,16 @@ type chatTUI struct {
 	// cards, and replay bundles are regenerated after a resize.
 	transcriptSources []transcriptSource
 	wrappedLines      []string // transcript wrapped to viewport width (rendered each frame)
-	viewport          viewport.Model
-	sel               selection
+	// wrapCache caches each block's width-wrapped rendering, keyed by block
+	// index, so a dirty frame re-wraps only changed blocks instead of the
+	// whole transcript (see wrapTranscriptBlocks in transcript.go). Entries
+	// carry the block content they were produced from, so a stale entry is
+	// detected by comparison and can never be served; the cache is dropped
+	// wholesale when the wrap width changes.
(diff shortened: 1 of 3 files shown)


innocarpe:
Closing as duplicate — main-v2 already merged the same optimization (#6978: syncWrappedLines + invalidateWrapFrom, suffix-only re-wrap). My per-block wrapCache covers the same hot path; keeping both would add complexity for no gain. The differential equivalence test (transcript_wrap_test.go) remains useful if a future rework wants it.

PR #7862, opened 2026-08-07
build(lint): make function size and complexity a gate, not a report

esengine:
`scripts/agent-complexity-report.sh` ran funlen and cyclop with
`issues-exit-code` forced to 0, and stated its own intent in the summary it
printed:

> Future work: block only new violations relative to this baseline.

That future is buildable now — repolint's ratchet already carries five rules —
so this does it and deletes the report.

## Why measure it in repolint instead of keeping golangci-lint

- One set of semantics: weight is the **excess over the limit**, budgeted per
  file and capped repo-wide, exactly like `essay` and `file-size`. Swapping a
  120-line function for a 400-line one still trips the gate.
- Scope widens from `internal/agent` to the whole tree.
- No parsing of another tool's output.

Thresholds are unchanged from the monitor: **120 lines, complexity 30, tests
excluded**. Cyclomatic count is one plus each branch point; a bare `default`
adds nothing, which the tests pin.

## What it surfaces

192 findings. The tail is not marginal:

| function | lines | complexity |
|---|---:|---:|
| `build` | 1932 | 309 |
| `chatTUI.update` | 1005 | 252 |
| `RenderTOMLForScope` | 754 | 152 |
| `RenderTOMLProjectDelta` | 434 | 148 |

Recorded ceilings: `function-size` 9121 excess lines, `complexity` 2050 excess
branch points. Those four functions are now frozen — they cannot gain a line or
a branch.

## Verified

A synthetic 41-branch function is rejected by both rules with the exact excess:

```
zz_probe.go:3: [complexity] zzProbe has cyclomatic complexity 41, over 30
zz_probe.go:3: [function-size] zzProbe is 122 lines, over the 120-line limit
repolint: repo total for complexity is 2061, above the 2050 ceiling
```

Plus `gofmt`, `go build ./...`, `go vet`, repolint's own tests (seven new ones
covering the default-clause rule, short-circuit operators, weight-as-excess,
and receiver naming), and `actionlint` on the workflow.

Documentation-impact: none - CI tooling only; no tool, contract, or behavior
changes.


esengine:
Rebased onto main-v2 (post-#7882) and resolved the conflicts: kept main's `src.file == nil` guard in `run()` with `checkComplexity` running after `checkComments`, and regenerated `baseline.json` against the current tree in a clean worktree — the first regeneration was accidentally done against a stale `origin/main-v2` and missed the new `internal/agent/execute_batch.go` (319-line, complexity-73 `executeBatch` carried over from agent.go in #7882), which would have turned the gate red right after merge. Now baselined (1951 findings / 552 files).

PR #7561, opened 2026-08-05
fix(desktop): make local paths inside backtick code spans clickable

zdjmrq:
## Summary

- AI replies frequently quote local file paths in backticks
  (`` `D:\x\y.md` ``). `remarkLocalPathLinks` (#7465) only visited plain
  text nodes, so code spans rendered as inert styled `<code>` — clicking
  did nothing.
- The plugin now also converts `inlineCode` nodes whose entire content is
  exactly one local path into a `file:///` link, keeping the code span as
  the label (renders as `<a><code>D:\x\y.md</code></a>`) and routing
  clicks to the native `OpenLocalPath` opener.
- Whole-span literal matching (no markdown text-node unescaping) covers:
  - drive paths with unescaped spaces — `` `D:\a\b c.md` `` opens the
    real path, fulfilling the backtick workaround for spaces promised in
    #7426;
  - file:/// URLs (percent-escapes preserved);
  - UNC shares with the literal `\\` prefix (`` `\\nas\share\x.md` ``).
- Trailing sentence punctuation and a whole trailing full-width paren
  group (`` `D:\x\y.md（已生成）` ``) are trimmed for the href while the
  label keeps the original spelling, mirroring the text-node matcher.
- Conservative: command-like prefixes (`cd D:\x`), literal escaped-space
  pairs (`\ `), Windows-illegal characters (`<>"|?*`), relative paths and
  fenced code blocks are left untouched.
- The local-path suites were not part of CI; added `test:local-paths` and
  a ci.yml step so the feature is guarded against regressions.

## Issues

Refs #7426

## Verification

- `pnpm test:local-paths` — local-path-links 60 passed, 0 failed (8 new
  inlineCode cases: spaces, UNC, full-width parens, punctuation trimming,
  command/escaped-space refusal); local-path-click-e2e 20 passed, 0 failed
  (2 new headless click assertions incl. a space round-trip:
  `file:///D:/a/b%20c/report%20final.md` → `OpenLocalPath("D:/a/b c/report final.md")`)
- Related suites all green: github-link-rendering 28/28, ref-token 15/15,
  markdown-image-proxy 7/7, session-export, approval-modal-file-reference 55/55
- `tsc --noEmit` clean

## Documentation impact

Documentation-impact: none - bug fix confined to desktop chat markdown rendering; no user-facing docs describe path-click behavior (#7465 shipped without doc changes for the same reason).

## Cache impact

Cache-impact: none - rendering-only frontend change, no system prompt or tool surface churn.
[...]

PR #7875, opened 2026-08-07
fix(cli,desktop): make links in chat transcripts clickable (issue #7867)

zdjmrq:
## What
Fixes the second point of #7867: links in conversation transcripts could not be opened by clicking.

- **TUI** (`internal/cli`): markdown links, bare URLs, URLs inside backtick code spans (the boxed, highlighted form from the issue) and URLs in user messages now render as OSC 8 hyperlinks bracketed by zero-width markers, and open in the system browser on **double-click**. Single click keeps selection; long URLs split across wrapped rows stay clickable on every row; copied text stays clean (markers stripped); destinations are sanitized (scheme allowlist + control-character scrub + click-time re-validation).
- **Desktop** (`desktop/frontend`): assistant markdown links were already clickable; this adds the two missing cases - bare URLs in plain-text **user messages**, and bare URLs inside **inline code spans** - both opened via the same scheme-allowlisted openExternal path.
- **Windows open fix**: replaced `rundll32 url.dll,FileProtocolHandler` with `explorer.exe`, which newer Windows 11 builds do not silently block.

## Tests
- TUI: `internal/cli/linkify_test.go` (17 cases) - rendering, sanitization/injection, cross-row marker parsing, hit-testing, double-click / slow-click / plain-click, copy stripping.
- Desktop: `plain-text-links.test.tsx` (15 cases) - user-message and code-span linkification, punctuation trimming; existing local-path-links (51), github-link-rendering (28), local-path-click-e2e (10) stay green; tsc --noEmit and vite build pass.

Documentation-impact: none - user-facing behavior only (links become clickable); no documented interface, config, or workflow changes, so existing docs remain correct.

### DIFF
--- a/desktop/frontend/src/__tests__/plain-text-links.test.tsx
+++ b/desktop/frontend/src/__tests__/plain-text-links.test.tsx
@@ -1,0 +1,91 @@
+// Run: tsx src/__tests__/plain-text-links.test.tsx
+
+import { Fragment, createElement } from "react";
+import { renderToStaticMarkup } from "react-dom/server";
+import { linkifyPlainText } from "../components/plainTextLinks";
+
+let passed = 0;
+let failed = 0;
+
+function ok(value: unknown, label: string) {
+  eq(Boolean(value), true, label);
+}
+
+function eq(actual: unknown, expected: unknown, label: string) {
+  if (JSON.stringify(actual) === JSON.stringify(expected)) {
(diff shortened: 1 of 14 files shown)


zdjmrq:
Related prior work: PR #7561 (also from us) already fixed the sibling problem - **local file paths** inside backtick code spans were not clickable. Together the inline-code format is now fully covered: local paths (#7561) and URLs (#7875).

Issue #7492, opened 2026-08-04
[Feature]: Add opt-in JSONL trajectory recording to E2E benchmarks


ten4762:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### What problem does this solve?

The current E2E benchmark reports final verifier results, token usage, cache metrics, and cost, but it does not preserve the Agent’s execution trajectory.
This makes it difficult to understand why a task succeeded or failed, including:
- Which tools were selected and with what arguments
- Which tool calls failed
- Whether the Agent recovered through retries
- How Planner and Executor phases were related
- At which step the run became unsuccessful

As a result, evaluation is mainly outcome-based rather than process-aware.

### Proposed solution

Add an opt-in JSONL trajectory recorder implemented as an event.Sink decorator.
The recorder would observe the existing typed event stream and write one structured record per event, including:

- Event type and sequence number

- Planner/Executor source

- Tool name, arguments, result status, error, and duration

- Retry information

- Usage and turn-completion metadata

The E2E benchmark would generate one trajectory file per task and expose basic statistics such as tool-call count, tool failures, and retries.
The default behavior and existing public APIs would remain unchanged. Full reasoning and conversation content would not be recorded by default to limit data size and avoid exposing sensitive information.

esengine:
Implemented as a three-PR stack:

- #7882 (merged) — `event.Tool` gains absolute `StartedAt`/`EndedAt` (unix ms) on tool results, so gaps between calls (model thinking/streaming) become measurable; zero for calls that never ran.
- #7886 — `internal/trajectory.Recorder`, an `event.Sink` decorator exactly as proposed here: one `{schema_version, seq, ts, event}` JSONL record per event reusing the `eventwire` contract, flushed per line, optional sink capabilities (readiness audits, protocol recovery, turn completion) recorded and forwarded. Exposed as `reasonix run --trajectory PATH`. (Supersedes #7884, auto-closed on base-branch deletion.)
- #7885 — `e2ebench -trajectories DIR` keeps one file per suite task and digests it into a tools-vs-model time-attribution line plus a per-result JSON summary.

One deliberate deviation from the proposal: the default here is *full content* (args, outputs, reasoning) rather than content-free, because the content-free automation surface already exists as `--events-jsonl` and the point of a trajectory is explaining *why* a run went the way it did. It is opt-in, file-based, and documented as transcript-equivalent in sensitivity.

esengine:
[...]

PR #7886, opened 2026-08-07
feat(cli): record a run's event trajectory with --trajectory

esengine:
Implements the trajectory recorder proposed in #7492: an opt-in, per-run JSONL log of the full typed event stream, so a run's tool selection, timing, thinking, and decision points can be replayed and analyzed offline instead of judging only the outcome. Stacked on #7882 (tool-result `StartedAt`/`EndedAt`), whose bounds flow straight into the records.

## Design

- **`internal/trajectory.Recorder`** — an `event.Sink` decorator modeled on `metricsSink`: every event is appended as one record `{schema_version, seq, ts, event}` where `event` reuses the shared `eventwire` contract (one serialization surface for desktop/ACP/serve/trajectory), `seq` orders records, and `ts` is the recorder's unix-ms observation clock (injectable for tests). Optional sink capabilities are recorded *and* forwarded, so wrapping never severs the chain: readiness audits (snake_case mirror), protocol-recovery observations, turn completions.
- **Durability** — flushed per line: a killed run keeps every completed record; a partially written last line is skippable. Recording failures never block forwarding — the first error is kept and reported by `Close`; late background-job events after `Close` are forwarded but not recorded.
- **`reasonix run --trajectory PATH`** — wired between metrics accumulation and notifications. Unlike `--events-jsonl` (deliberately content-free), the trajectory includes prompts, tool arguments, and reasoning — documented as transcript-equivalent in sensitivity. The run-sink assembly moved out of `runAgent` into `buildRunSink` (`run_sink.go`) to keep `cli.go` inside the file-size ratchet.

With #7882's bounds plus per-record `ts`, the gaps between a `tool_result` and the next `tool_dispatch` attribute wall-clock to model thinking/streaming vs. tool execution.

## Tests

`internal/trajectory`: ordered/timestamped records with tool bounds intact round-trip; capability records forward to the inner sink; post-Close events forward without recording; unwritable path fails `New`. Docs updated (CLI.md + zh-CN).

Cache-impact: none - local-only sink decoration and CLI wiring; nothing enters provider requests
Cache-guard: go test ./internal/trajectory/ ./internal/cli/; recorder only observes events already emitted
Documentation-impact: updated - docs/CLI.md and docs/CLI.zh-CN.md gain a Trajectory recording section for --trajectory

### DIFF
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -157,6 +157,22 @@ reasonix run --ablate evidence,planner --metrics run.json "fix the failing test"
 This is a measurement tool, not a tuning knob: switching a subsystem off makes
 Reasonix worse at the work it was added for.
 
+### Trajectory recording
+
+`--trajectory PATH` appends the run's full event stream — tool dispatches and
+results with absolute start/end times, reasoning, retries, readiness and
+recovery decisions — as one timestamped, sequenced JSONL record per event, so
+a run can be replayed and its time attributed offline (tool execution vs. the
+model thinking between calls). Records reuse the shared `eventwire` JSON
+contract under an `event` key, wrapped in `schema_version`, `seq`, and `ts`
+(unix ms). Every completed line survives a killed run. Unlike `--events-jsonl`,
+the file contains prompts, tool arguments, and reasoning: treat it with the
+same care as a session transcript.
+
+```sh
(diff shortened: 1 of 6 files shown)


esengine:
Supersedes #7884, which was auto-closed when its stacked base branch (#7882's) was deleted on merge. Same commit, now based on main-v2.

PR #7889, opened 2026-08-07
perf(event): coalesce streaming deltas before they reach a frontend

esengine:
Targets the front half of the streaming pipeline (provider chunk → Go agent → event.Sink → Wails EventsEmit → WebView): the agent emitted one sink event per provider chunk, so a 100–300 chunk/sec stream became 100–300 EventsEmit/sec across the Go→WebView bridge (and equally many SSE frames / terminal redraw triggers). The desktop's `asyncRuntimeEmitter` already prevents the bridge from blocking the agent, but every chunk still paid serialization + IPC + a JS event dispatch. The frontend's rAF batching only absorbs the back half.

## Design

`event.Coalesce(inner, window)` — a sink decorator installed by `control.New` around every frontend sink, so desktop, CLI, serve, and ACP inherit it (behavior in the controller, not a frontend, per REASONIX conventions). It also covers the desktop rebuild paths that pass a raw `tabEventSink` without going through `desktopControllerSink`.

- **Merges only pure deltas**: consecutive `Text`/`Reasoning` events whose only payload is `Text`. The purity check compares against a zero probe via `reflect.DeepEqual`, so a future field added to `event.Event` can never be silently merged away; a companion test pins the ordering behavior.
- **Latency**: the first delta of a burst forwards immediately — time-to-first-token is unchanged. Under load the rate caps at one merged event per 25ms (`event.DefaultStreamDeltaWindow`, ~40Hz — above display refresh perception, an order of magnitude below fast-provider chunk rates).
- **Ordering**: any non-delta event (tool dispatch, message, usage, stream_attempt, …) flushes the buffer ahead of itself; kind switches flush; a 16KB cap bounds merged payload size. The coalescer's mutex serializes timer flushes with passthroughs, so downstream sinks keep the serial-Emit contract.
- **Capabilities**: readiness audits, protocol recovery, and turn completion flush-then-forward, so wrapping never severs the optional-capability chain.

Frontend contract is unchanged: consumers already treat text/reasoning as appendable deltas (`base.text + delta`), so a merged delta is indistinguishable from a fast provider sending bigger chunks.

Placed outside the `goalUsageTee` so a caller-provided tee is never double-wrapped; Usage events pass through untouched either way.

## Tests

`internal/event`: leading-edge immediacy, burst merge + barrier flush ordering, kind-switch flush, timer flush, byte-cap flush, capability flush-then-forward, non-pure delta passthrough, disabled/nil construction. Full `internal/event` + `internal/control` + `internal/agent` suites pass.

Cache-impact: none - sink-side event shaping only; provider requests and the prompt prefix are untouched
Cache-guard: go test ./internal/event/ -run TestCoalesce; deltas merge after the agent has already accumulated the full text for session/provider state
Documentation-impact: none - internal event pipeline behavior; the delta contract observed by consumers is unchanged

### DIFF
--- a/internal/control/controller.go
+++ b/internal/control/controller.go
@@ -531,10 +531,9 @@ type Options struct {
 	SessionTemp *sessiontemp.Manager
 }
 
-// New builds a Controller. A nil Sink is replaced with event.Discard. When the
-// caller did not already provide a goalUsageTee (see NewGoalUsageTee), the
-// public sink is wrapped in one so billable usage can be accounted to Goal
-// budgets; frontends observe the same forwarded stream either way.
+// New builds a Controller. A nil Sink becomes event.Discard; unless the caller
+// already provided a goalUsageTee (NewGoalUsageTee), the sink is wrapped in one
+// so billable usage can be accounted to Goal budgets.
 func New(opts Options) *Controller {
 	sink := opts.Sink
 	if nilutil.IsNil(sink) {
@@ -545,6 +544,7 @@ func New(opts Options) *Controller {
 		usageTee = NewGoalUsageTee(sink).(*goalUsageTee)
 		sink = usageTee
 	}
+	sink = event.Coalesce(sink, event.DefaultStreamDeltaWindow)
 	pluginCtx := opts.PluginCtx
 	if pluginCtx == nil {
 		pluginCtx = context.Background()
--- a/internal/event/coalesce.go
+++ b/internal/event/coalesce.go
@@ -1,0 +1,161 @@
+package event
+
+import (
+	"reflect"
+	"strings"
+	"sync"
+	"time"
+
+	"reasonix/internal/evidence"
+	"reasonix/internal/nilutil"
+)
+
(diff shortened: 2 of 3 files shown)


esengine:
Force-push fixes a self-deadlock CI caught on Windows (`TestRecoveryReviseBlocksPlanTransition` timed out): the coalescer held its mutex while forwarding, so a downstream sink callback that synchronously re-entered the controller (`ResolveRecovery` → decision receipt → `sink.Emit`) re-locked the same mutex. Forwarding now drains an outbound FIFO queue with a single drainer and never calls `inner.Emit` under the lock — a re-entrant Emit enqueues and returns, order and serial emission preserved. Regression test: `TestCoalesceReentrantEmitDoesNotDeadlock`; `internal/event` passes with `-race`.

esengine:
Window tightened 25ms → 16ms (`DefaultStreamDeltaWindow`): at most one merged batch per display frame under load, per maintainer's spec. Leading-edge forwarding is unchanged, so TTFT is unaffected.

PR #7890, opened 2026-08-07
perf(frontend): fold a frame's stream deltas into one reducer pass per tab

esengine:
Companion to #7889 (Go-side coalescing), removing the remaining per-delta cost inside the WebView. The rAF batch already confined token deltas to one flush per animation frame, but the flush still dispatched each delta individually: N deltas in a frame paid N reducer calls, N live-object allocations + string concats (`base.text + delta`), and N `liveStore` subscriber notifications — batched timing, unbatched data.

## Design

`coalesceStreamDeltas` groups the frame's deltas by tab and merges **consecutive same-kind** runs into ordered segments — deliberately not per-kind buckets, because segment order is semantic: the reasoning→text boundary drives `reasoningComplete`, and a trailing reasoning segment must reopen it. Each tab then receives a single `stream_batch` action:

```
reasoning 我 | 需要 | 检查 | 一下      →  [ reasoning: "我需要检查一下",
text 发现 | 问题 | 如下                     text: "发现问题如下" ]
```

- **One reducer pass, one notification per tab per frame**: `applyStreamBatch` folds the segments via `applyLiveSegments`, which replicates per-delta semantics exactly (text completes reasoning first; reasoning reopens it, stamps `reasoningStartedAt` on the first non-empty delta, clears `reasoningCompletedAt`). The per-delta `text`/`reasoning` event path now delegates to the same fold, so there is a single source of truth.
- **Preamble parity**: `stream_batch` mirrors `applyEvent`'s preamble for delta events — `discardTurn` swallows it, a pending user bubble flushes first, the retry indicator clears — and joins the existing delta-only `bump()` skip so whole-App renders stay unaffected.
- Tabs are independent state machines, so per-tab grouping reorders nothing observable; within a tab, segment order is the arrival order. Empty deltas are kept (an empty text delta still completes live reasoning). `completeLiveReasoning` moved to `streamDeltaBatch.ts` alongside the fold.

Net effect: `N deltas/frame → N reducer → N notify` becomes `N deltas/frame → 1 reducer → 1 notify` per tab.

## Tests

`stream-delta-batch.test.ts`: segment merging across an interleaved second tab, reasoning→text→reasoning ordering, legacy `reasoning`-field deltas, reducer equivalence against per-delta dispatch (text/reasoning/`reasoningComplete`), boundary completion timestamping, and preamble parity (discard + retry-clear). `tsc --noEmit`, eslint, and the adjacent `use-controller-*` / reasoning-display suites pass.

Cache-impact: none - frontend-only rendering path; nothing provider-visible changes
Cache-guard: tsx src/__tests__/stream-delta-batch.test.ts (reducer-equivalence assertions)
Documentation-impact: none - internal frontend performance work with no behavior change


esengine:
Force-push fixes the repolint file-size gate: `useController.ts` was 1 line over its ratchet budget from the new import. The flush loop moved into `streamDeltaBatch.ts` (`dispatchMergedDeltas`), so `useController.ts` nets one line below baseline instead.

PR #7892, opened 2026-08-07
perf(frontend): parse streaming Markdown per completed block, not per tick

esengine:
Third layer of the streaming-pipeline work (#7889 Go-side coalescing, #7890 one reducer pass per frame). The existing Markdown streaming optimizations are real — tiered commit intervals (50/150/300ms), a plain-text `appendData` tail above 8KB, stable-section splitting above ~12KB — but they leave a counter-intuitive gap: **the most common 1–8KB answer re-parses the entire growing document through ReactMarkdown + remark/rehype/normalizeMath up to 20×/sec, and below the 8KB threshold the tail mechanism doesn't exist at all**, so small answers get the most parses and none of the cushioning. Parse count scales with stream duration (a 30s answer ≈ 600 full parses), not with content.

## Changes

- `streamingCommitTarget(text)`: the parsed prefix now advances only to the **last completed block** — a blank line outside fences and display math. The in-progress block rides the existing `StreamingMarkdownTail` (`appendData`, zero parsing) from byte zero, not just above 8KB.
- A stream now costs **one parse per completed block plus the existing single finalization pass**, instead of one full parse per interval tick for the stream's whole duration. Tail growth alone schedules nothing.
- **Code streams don't regress**: an unclosed ``` fence or `76479` keeps the whole text parsed, so streaming code keeps live highlighting exactly as today; a blank line inside a closed fence is not a boundary.
- The tier interval (50/150/300ms) remains as the cadence ceiling for boundary commits; the ≥8KB path also benefits (commits happen on block completion instead of unconditionally every 150/300ms). The prefix only ever advances monotonically, and end-of-stream finalization (immediate below 8KB, idle-scheduled above) is unchanged.

Fidelity tradeoff, stated: the in-progress block shows raw Markdown syntax until it completes — the same tradeoff the ≥8KB tail already made, now applied uniformly. In exchange, new tokens appear with **lower** latency than before (per-frame `appendData` instead of waiting for the next 50ms parse commit).

## Tests

`mermaid-rendering.test.tsx` (68 passing): `streamingCommitTarget` boundary/fence/math/closed-fence cases, the hook's new contract (in-progress block schedules no parse, completed block commits to its boundary, tail growth schedules nothing, short-stream finalization immediate), plus the existing interval-tier and section-splitting assertions. `tsc`, eslint, render-optimization and reasoning-display suites pass.

Cache-impact: none - frontend-only rendering path; nothing provider-visible changes
Cache-guard: tsx src/__tests__/mermaid-rendering.test.tsx (commit-target and hook-contract assertions)
Documentation-impact: none - internal frontend performance work; visible behavior change is limited to when in-flight markdown styling appears, which docs do not describe

### DIFF
--- a/desktop/frontend/src/__tests__/mermaid-rendering.test.tsx
+++ b/desktop/frontend/src/__tests__/mermaid-rendering.test.tsx
@@ -9,7 +9,7 @@ import { act } from "react";
 import { createRoot } from "react-dom/client";
 import { renderToStaticMarkup } from "react-dom/server";
 import ReactMarkdown from "react-markdown";
-import { splitStableMarkdownSections, streamingMarkdownCommitInterval, useRenderedMarkdownText } from "../components/Markdown";
+import { splitStableMarkdownSections, streamingCommitTarget, streamingMarkdownCommitInterval, useRenderedMarkdownText } from "../components/Markdown";
 import MermaidDiagram from "../components/MermaidDiagram";
 import {
   __setMermaidPanZoomFactoryForTest,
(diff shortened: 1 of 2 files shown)


esengine:
Extended the promotion boundaries per maintainer's spec: a **closed fence** and **closed display math** now promote immediately (previously they waited for the next blank line), a **partial heading line** completes the paragraph before it (headings interrupt paragraphs, and the `#{1,6} ` prefix match is stable as the line grows), and a **terminated heading** promotes itself as a complete single-line block. All boundaries require terminated lines except the pre-heading case. 71 assertions passing.

PR #7899, opened 2026-08-07
feat(bench): model-round decomposition, per-solved efficiency, and A/B compare mode

esengine:
Closes the measurement gap in the agent-latency equation: T_total ≈ T_planner + ΣT_model + ΣT_tool-critical-path + T_retry + T_compaction. For a coding agent the dominant term is usually **how many sequential model round-trips a task takes**, not local overhead — but the trajectory digest only reported tools vs model in aggregate.

## Changes

`summarizeTrajectory` now decomposes the model side:

- **`model_rounds`** — one round per gap from the turn start (or a batch's last top-level `tool_result`) to the next top-level `tool_dispatch` (the first, possibly partial, dispatch marks when thinking ends and the call starts streaming), plus the final answer segment. Same-batch dispatches don't split rounds; subagent calls (`parentId`) stay excluded so parent rounds aren't fragmented.
- **`model_gap_total_ms` / `model_gap_p95_ms`** — where ΣT_model actually goes, per round.
- **`retries` / `compactions`** — the T_retry / T_compaction terms, counted off the same stream.
- The markdown report's time-attribution line gains `**model rounds** N (avg gap X)`.

With this, a recorded suite answers directly: how many rounds did each solve take, how expensive is a round, and how much of the total was retry/compaction tax — the denominators needed before attacking the known round-trip taxes (serial `complete_step` sign-offs, final-readiness blocks, missing-reasoning retries, dependency-barrier redo rounds).

## Tests

`TestSummarizeTrajectoryDecomposesModelRounds`: two tool rounds + final answer from a fixture with a same-batch second dispatch (no round split), an interleaved subagent call (ignored), and a retry event; exact totals and P95 asserted.

Cache-impact: none - harness-side offline analysis of recorded trajectories
Cache-guard: go test ./cmd/e2ebench/ -run TestSummarizeTrajectory
Documentation-impact: none - report format addition on an opt-in benchmark flag documented at the flag level already

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -46,6 +46,7 @@ decoder. The task ID is the directory name; tasks run in sorted ID order.
 | Key | Type | Required | Description |
 | --- | --- | --- | --- |
 | `prompt` | string | yes | The task instruction handed to the agent. |
+| `class` | string | no | Task class label (e.g. `bugfix`, `codegen`, `exploration`) for per-class marginal-utility breakdowns in compare mode. |
 | `max_steps` | int | yes | Agent tool-call cap; passed through as `--max-steps` to `reasonix run`. |
 | `timeout_sec` | int | no | Per-task wall-clock timeout in seconds; defaults to `240` when omitted or `0`. |
 
@@ -114,6 +115,7 @@ own outcome).
 | `-out` | *(stdout)* | Write the markdown report here. |
 | `-json` | *(none)* | Write the JSON report here (optional). |
 | `-trajectories` | *(none)* | Suite mode: write one `<task-id>.trajectory.jsonl` per task into this directory (the agent's full event stream with timestamps — see `reasonix run --trajectory`). The report gains a time-attribution line (tools vs. model) and each JSON result a `trajectory` digest. |
+| `-force-planner` | `false` | Suite mode: prefix each prompt with a plan-first directive so the two-model turn engages regardless of the planner gate. Use for the "with planner" arm of an A/B; results carry `plan_forced` so arms are only comparable with equal forcing. |
 | `-budget` | `800000` | Abort once total tokens cross this (`0` = no cap). Remaining tasks are reported as skipped. |
 
 Diff-mode flags:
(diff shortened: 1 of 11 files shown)


esengine:
Added the per-solved efficiency line the maintainer asked for: `**Per solved task:** model requests X · tool calls Y · wall Z · model rounds R` — totals across every accounted run (failures included) divided by accounted solves, matching the existing cost/tokens-per-solved convention. `Steps` counts every billed provider call (subagents, retries, router), so **Model Requests per Solved Task** is exactly the "how much model work buys one solve" number; `model rounds` joins it when `-trajectories` was on. Tokens/solved and cost/solved were already reported.

esengine:
Extended for the planner A/B (maintainer's priority): the metrics mirror now decodes `usage_by_source`, the report adds `**Requests by source:** executor N (T tok) · planner N (T tok) · …`, and a new mode renders the experiment readout directly:

```sh
go run ./cmd/e2ebench -trajectories t-a -json control.json
go run ./cmd/e2ebench -trajectories t-b -json ablated.json -ablate planner
go run ./cmd/e2ebench -mode compare control.json ablated.json
```

producing a per-solved delta table (solve rate, model requests, **planner requests**, model rounds, tool calls, tokens, wall seconds, cost) under the same accounted-solves convention. If the planner arm's extra requests/wall don't buy a solve-rate delta, the table says so in one screen.

esengine:
Third layer per maintainer's direction — **marginal utility, not "does it help"**:

- Tasks gain an optional `class` label (committed suite tagged: exploration / bugfix / codegen ×2 / delegation).
- Compare mode adds `**Marginal utility (A − B):** accuracy +X.Xpp · wall/task +Y.Ys` plus a per-class table (Δ accuracy, Δ wall per class) — the exact 'planner uplift / latency cost / decision' readout.
- `-force-planner` prefixes prompts with a plan-first directive the gate's `planAndExecuteDirectives` recognizes — needed because a real A/B today showed the committed suite **never trips the planner gate** (usage_by_source: zero planner calls on all five tasks). Results carry `plan_forced` so arms are only comparable with equal forcing.

The host-owned router itself already exists (`planner_gate.go`: ExecutorOnly / Light / Full off highRisk/multiFile/structured/atomic features); this data layer is what calibrates its thresholds per class.

PR #7902, opened 2026-08-07
feat: add generation-safe runtime composition / 添加代际安全的运行时组合

SivanCola:
## Summary

- Add generation-safe runtime composition with dependency graphs, lifecycle states, effect scopes, staged subgraph rebuilds, and atomic publish/commit.
- Introduce an explicit `RuntimeOwner` per controller/session lineage. It owns the publish gate, receipt ledger, file priors, message deduplication, and host-stream cancellation.
- Reuse an owner only for rebuilds of the same lineage, so rebuilding one session cannot drain, cancel, or overwrite evidence from another session.
- Carry the owner through controller turn contexts, host and extension providers, structured UI, MCP drain handling, doctor output, and file-write compensation.
- Add deterministic regression coverage for multi-session isolation, deferred publication, draining admission, callback re-entry, late drain registration, receipt isolation and retention, repeated writes, and overlay/local file receipts.
- Keep the Runtime Manifest and Extension Protocol strictly on major v2, with refreshed generated docs and Go SDK comments.
- Make `sdk/go/examples/fullsidecar` a complete installable v2 package whose manifest declares its interceptors, system-prompt strategy, fake provider, and UI action.
- Send Manifest v2 `requires`/`provides` plus dependency schema versions during initialization, and verify the sidecar activates only the installed capability ceiling.

## Owner review follow-up

Addressed in `14264d4e`:

- Default-owner compatibility fallbacks are no longer silent. `RuntimeOwnerOrDefault` counts every fallback resolution, and CLI/desktop runtime doctor exposes `runtimeOwnerFallbacks`. Product boot still creates an explicit isolated owner in `bindRuntimeOwner`.
- Receipt recovery is explicitly process-local; crash recovery is outside this PR. `ReceiptStore` retains at most 32 generations and 256 receipts per generation. Eviction releases associated file-prior bytes and marks recovery evidence truncated, so clean rollback is never claimed after evidence loss.
- The coverage-flaky snapshotter test no longer sleeps around a wall-clock ticker. It injects deterministic poll ticks and waits for a processing acknowledgement; coverage instrumentation passes repeated runs.

## Compatibility boundary

- Native runtime manifests must declare the exact `reasonix.io/plugin/v2` value.
- Manifest v1 was never publicly released, so there is no v1 dual-read, automatic migration, or explicit v1 migration path.
- `reasonix plugin migrate <name> --to-v2` accepts only pre-extension native manifests that omit `apiVersion`; any versioned manifest, including v1 and already-v2 files, is rejected by the migration parser.
- Existing `plugin-packages.json` storage remains unchanged.
- Older Reasonix releases continue to ignore extension-only sidecars, model refs, and UI event kinds as documented.

## Concurrency and safety

- Independent cold builds receive independent runtime owners and remain Active concurrently.
- Full and subgraph rebuilds drain only the previous generation in the same owner lineage.
- Replacement generations remain unpublished until readiness, state migration, and commit succeed.
- Draining controllers reject new interactive, synchronous, and headless turns.
- Drain cancellation records expired generations, so late callback registration cancels immediately instead of leaking work; callbacks run outside registry locks to avoid re-entry deadlocks.
- Receipt retention is mutex-protected and conservatively marks truncated evidence as non-clean. Eviction callbacks run outside the receipt lock.
- Runtime-owner state stays in execution context and local memory. It does not enter system prompts, tool schemas, provider-visible request bytes, or receipt IDs containing local paths.
- Security review found no new dependency, network, credential, privilege, or persistence surface. Runtime commands remain absolute-path, exec-form sidecars with the documented full-trust environment boundary.

## Integration

- The PR contains the complete remediation instead of splitting it into follow-up PRs.
- The branch includes #7905 and current `main-v2` at `6cc0d7340`; the latest base is the PR merge-base.
- The repolint baseline is unchanged by the Owner Review fixes.

## Verification

- `go test -p 1 ./... -count=1`
- `go test -race ./internal/agent ./internal/boot ./internal/control ./internal/extension ./internal/extension/providerext ./internal/extension/uihub ./internal/tool/builtin -count=1`
- `go test -race ./internal/extension/sidecar ./internal/extension/conformance -count=1`
- `go test -race ./cmd/e2ebench -run '^TestSnapshotterCapturesWorkspaceChanges$' -count=10`
- `go test -coverprofile=coverage.out -covermode=atomic ./cmd/e2ebench -run '^TestSnapshotterCapturesWorkspaceChanges$' -count=20`
- `cd desktop && go test ./... -count=1 && go vet ./...`
- `cd sdk/go && go test ./... -count=1`
- `go vet ./...`
- `go run ./cmd/extension-protocol-gen -check -root .`
- `go run ./tools/repolint`
- `git diff --check`

## Documentation impact

Documentation-impact: updated - runtime ownership, strict v2 compatibility, staged reload behavior, generation-bound UI/provider behavior, bounded process-local recovery receipts, SDK guidance, and performance notes are documented in English and Chinese.

## Cache impact

Cache-impact: low - the runtime v2 feature intentionally supports explicit extension strategy/tool contributions, while owner/context wiring, receipt retention, and diagnostics do not alter the provider-visible stable prefix.
Runtime-plan-diagnostics: `prefixChanged` is observed from the old/new snapshot `CacheHash`; `providerChanged` independently records provider capability add/remove/reload. Provider-only backend rolls with identical prefix bytes report `prefixChanged=false, providerChanged=true`; MCP schema-shape changes use a full rebuild.
Cache-guard: byte-stable system-prompt, tool-schema, snapshot CacheHash, no-op rebuild, independent-build, and #7905 coalesced-sink regression tests.
System-prompt-review: complete - `RuntimeOwner`, generation, receipt, fallback, retention, and drain state remain execution-only; reviewers should verify that only explicit extension strategy replacement changes prompt bytes.


esengine (dismissed):
## Overall

The direction is right, and it targets the actual hard part.

Hot reload was never hard because rebuilding a config is hard — it is hard because of the **side effects still in flight from the previous generation**: provider requests already submitted, files mid-write, UI already rendered. This PR doesn't try to be cleverer about rebuilding; it reframes the problem as generation migration and then reuses established patterns correctly. `stage → ready → commit` is two-phase commit / double buffering, `EffectScope` is a structured-concurrency resource scope, `ReceiptStore` plus the reversible/irreversible split is a saga compensation log, `RuntimeOwner` is an ownership boundary. Nothing here invents new vocabulary to dress up an old problem, which is the failure mode I was expecting at this size.

Three things I want to call out as clearly right, so they don't get lost in the notes below:

- **"Irreversible effects never report rollback success."** Most systems quietly pretend the rollback was clean at exactly this point.
- **Owner reuse only along the same lineage.** This turns "rebuilding session A must not drain session B" into an ownership fact instead of a rule every caller has to remember.
- **Strict v2-only, no dual-read, no auto-migration.** The rationale (v1 was never publicly released) holds, and this kind of decisiveness on compatibility is rare. Please keep it exactly as is.

My one reservation is the complexity budget. Seven new concepts and a generation state machine is a real price, and it is justified only by "the desktop is long-lived and multi-session, so a restart interrupts a running turn". If most config changes could in practice just rebuild the controller outright, this machinery is overbought for them. I'd like the design to state explicitly **which changes must be hot-swapped and which are allowed to take the full rebuild path**, rather than defaulting every change into generation migration.

## Notes

**1. `DefaultRuntimeOwner` contradicts the value proposition of the whole design.**

The point of this PR is to make cross-session bleed impossible, but there is a package-level singleton that any un-wired call path silently falls back to — which is precisely the shape of the bug being eliminated, downgraded from "crash" to "quietly shared". Across 190 files, a missed wiring site is not something review can reliably catch. Please make it observable at minimum: have `doctor runtime` report "N calls fell back to the default owner" during this run, or make a missing owner a hard error on product boot paths. Keeping the compatibility fallback is fine; keeping it silent is not.

**2. The receipt ledger is in memory, and the scenario that needs it most is a crash.**

A saga log that isn't durable isn't a log. Today the evidence only serves in-process rebuilds; if the process dies — exactly when you most want to know which irreversible request the previous generation submitted — all of it is gone. Either honestly narrow the scope and document "crash recovery is out of scope", or persist it through the existing store layer. Related: receipts grow monotonically (one per file write, sequence never reset), so an eviction policy is needed.

**3. The reversible/irreversible classification exists but doesn't drive the drain decision.**

A single drain TTL is a magic number. If a long inference happens to be running on the old generation, it gets cut at the deadline. Given that `Irreversible` is already a first-class classification, the drain policy should be derived from it: in-flight irreversible work should either be awaited or explicitly surfaced ("a request that was already billed was cancelled"), not share one timeout with a UI refresh that can be replayed for free.

**4. The evidence has a producer but no consumer semantics.**

After honestly reporting "this cannot be recovered" — then what? Right now the answer is "it shows up in doctor". What's missing is the degradation contract: on an unrecoverable generation, does the runtime refuse the rebuild, force a full restart, or publish anyway and leave a persistent warning in the UI? Diagnostics are a display, not a decision.

**5. UI events emitted by a staged generation are dropped silently; they should be rejected explicitly.**

The docs say not to rely on UI visibility before publish, but for a sidecar author that's an implicit trap: whatever they emit during handshake/ready evaporates without a trace, which is painful to debug. An explicit error code at the protocol layer is far more reliable than a documented convention.

**6. Deterministic regression tests can't establish the safety of a concurrent state machine.**

What's covered today is the set of failures that were anticipated at design time. Real confidence for generation migration comes from randomized interleavings — arbitrary sequences of rebuild / turn / drain / failure, asserting a few invariants always hold: exactly one published generation, every effect owned by exactly one owner, a draining generation never admits a new turn. A property test here buys more than ten more hand-written scenarios.

None of the above blocks the concept; 1 and 2 are the two I'd want resolved before this lands, since both are cases where the design's central promise is weaker than it reads.

SivanCola:
Updated this PR in `14264d4e`; `5e90755e` is the lint-only follow-up.

Review blockers addressed:
- **Default owner fallback observability:** `RuntimeOwnerOrDefault` now counts compatibility fallbacks, both CLI and desktop doctor expose `runtimeOwnerFallbacks`, and product boot paths continue to construct explicit isolated owners.
- **Receipt ledger scope and retention:** the contract now explicitly states that recovery is process-local and crash recovery is out of scope. `ReceiptStore` is bounded to 32 generations and 256 receipts per generation. Eviction releases associated file priors and marks recovery evidence as truncated, so the runtime cannot incorrectly report a clean rollback after evidence has been discarded.

Additional assurance:
- The snapshotter coverage regression uses injected poll ticks and acknowledgement channels, with no timing sleeps.
- The manifest remains strict v2-only, with no dual-read or automatic migration.
- There are no provider-visible prompt, tool-schema, or request-shape changes; the cache-impact guard passed.

Verification:
- CI run `31245533131` is fully green across Linux, macOS, Windows, desktop, race, SDK, coverage, lint, CodeQL, and govulncheck jobs.
- Local root and desktop Go suites, SDK suite, runtime and sidecar race matrices, deterministic E2E snapshot coverage, frontend typecheck/lint/build, `go vet`, repository lint, `golangci-lint`, and `git diff --check` all passed.

PR #6749, opened 2026-07-21
fix(desktop/composer): don't cancel in-flight turn on ESC while IME is composing

chengoak:
## Problem

The composer's ESC handler unconditionally calls `handleCancel()` whenever a turn is running, even when the ESC keydown was fired to dismiss a Chinese/Japanese IME candidate window. On macOS this is easy to hit while replying is streaming:

1. Send a prompt and let the agent start replying.
2. Start typing a new message with a Chinese IME while the reply streams.
3. With the candidate list still open, press `ESC` to dismiss the candidate.
4. **Expected:** candidate closes, agent keeps replying.
5. **Actual:** the agent turn is cancelled mid-reply.

## Root cause

`desktop/frontend/src/components/Composer.tsx` — the same `onKeyDown` handler that gates the send-Enter chord, the mention/past-chats menu, Shift+Tab plan toggle, and paste shortcuts with `!composing` was missing that guard on the running-turn ESC branch:

```ts
// Esc interrupts the in-flight turn (matches the Stop button's hint), and
// restores the text if the server hadn't replied yet.
if (e.key === "Escape" && running) {
  e.preventDefault();
  handleCancel();
}
```

So an IME dismiss ESC on a running turn falls straight into `handleCancel()`.

## Fix

Add `!composing` to the ESC guard. The existing `isImeKeyEvent()` helper already folds together `composingRef`, `nativeEvent.isComposing`, `keyCode === 229`, and a short post-`compositionend` grace window (`IME_CONFIRM_GRACE_MS`), so `composing` is already the right signal — no new state or listeners.

```ts
if (e.key === "Escape" && running && !composing) {
  e.preventDefault();
  handleCancel();
}
```

Diff: **+5 / -1** (four lines are the comment explaining why).

## Verification

Local Wails build (`main-v2` HEAD + this patch, macOS arm64):

1. Send a prompt; agent starts streaming a reply.
2. In the composer, start typing with a Chinese IME (macOS 拼音 / 双拼).
3. Press `ESC` while the candidate window is open.
4. Result: candidate window closes; agent continues streaming without interruption. ✅

Non-IME ESC behavior is unchanged:
- `ESC` with no composition and `running` still cancels the turn as before (matches the Stop button's hint).
- Menu/past-chats ESC branch above is untouched.
- Send-Enter, Shift+Tab, prompt-history nav paths are untouched.

## Notes

- Fix is scoped to `desktop/frontend/src/components/Composer.tsx`; no changes to the Go side or bindings.
[...]

PR #7919, opened 2026-08-08
fix(desktop/composer): don't cancel in-flight turn on ESC while IME is composing

chengoak:
## Problem

The composer's ESC handler used to call `handleCancel()` whenever a turn was running, even when the ESC keydown was fired to dismiss a Chinese/Japanese IME candidate window. On macOS this is easy to hit while a reply is streaming:

1. Send a prompt and let the agent start replying.
2. Start typing a new message with a Chinese IME while the reply streams.
3. With the candidate list still open, press `ESC` to dismiss the candidate.
4. Expected: the candidate window closes and the agent keeps replying.
5. Actual: the agent turn is cancelled mid-reply.

## Root cause

`desktop/frontend/src/components/Composer.tsx` already computes a shared `composing` signal for send-Enter, menus, Shift+Tab, paste shortcuts, and prompt-history navigation, but the running-turn ESC branch was deciding cancellation inline. The past-chat search input is also a sibling of the textarea, so its keydown handler did not go through the main composer IME guard either.

## Fix

The first commit adds the missing IME guard to the running-turn ESC path.

The supplemental commits centralize and reuse the keyboard policy:

- `desktop/frontend/src/lib/composerKeyboard.ts` now owns `isImeKeyEvent()` for active composition ref, `nativeEvent.isComposing`, legacy `keyCode === 229`, and the short post-`compositionend` grace window.
- `composerEscapeAction()` decides whether Escape should cancel the running turn or pass through to the IME/browser.
- `composerMenuKeyAction()` decides whether menu navigation keys should be handled or pass through during IME composition.
- `Composer.tsx` now asks the helpers for the running-turn Escape action and past-chat search menu-key action instead of duplicating those conditions inline.
- The past-chat search input tracks its own composition state before handling Escape, Enter, Tab, or arrow navigation keys.

## Verification

Manual Wails check from the original PR:

1. Send a prompt; agent starts streaming a reply.
2. In the composer, start typing with a Chinese IME.
3. Press `ESC` while the candidate window is open.
4. Result: candidate window closes; agent continues streaming without interruption.

Automated checks added/rerun after the supplemental commits:

- `go run ./tools/repolint`
- `pnpm --dir desktop/frontend exec tsx src/__tests__/composer-keyboard.test.ts`
- `pnpm --dir desktop/frontend exec tsx src/__tests__/composer-goal-toggle.test.tsx`
- `pnpm --dir desktop/frontend exec tsc --noEmit`
- `pnpm --dir desktop/frontend exec eslint src/components/Composer.tsx src/lib/composerKeyboard.ts src/__tests__/composer-keyboard.test.ts src/__tests__/composer-goal-toggle.test.tsx`
- `git diff --check`

The regression coverage now includes:

- running turn + active composition + Escape passes through
- `nativeEvent.isComposing === true`
- legacy `keyCode === 229`
- ordinary non-IME Escape still cancels a running turn
- the 100ms post-`compositionend` grace-window boundary
- menu navigation keys pass through during IME composition while ordinary Escape/Enter still remain menu-handled

## Documentation impact

Documentation-impact: none - this is a chat-composer behavior guard and focused keyboard regression coverage. No public API, documented keybinding surface, or user-visible setting changes.

## Supersedes

This PR supersedes #6749, which was closed to restart the review queue. The original runtime fix was directionally correct; this version now also carries the maintainer-requested regression coverage and keyboard-policy cleanup.

### DIFF
--- a/desktop/frontend/src/__tests__/composer-goal-toggle.test.tsx
+++ b/desktop/frontend/src/__tests__/composer-goal-toggle.test.tsx
@@ -2179,13 +2179,6 @@ console.log("\ncomposer goal toggle");
   const sessionSearch = document.querySelector(".slashmenu__search") as HTMLInputElement | null;
   if (!sessionSearch) throw new Error("recent-session search did not render");
   await act(async () => {
-    const imeEscape = new KeyboardEvent("keydown", { key: "Escape", bubbles: true, cancelable: true });
-    Object.defineProperty(imeEscape, "isComposing", { value: true });
-    sessionSearch.dispatchEvent(imeEscape);
-    await flushTimers();
-  });
-  ok(Boolean(document.querySelector(".slashmenu__search")), "IME Escape keeps the direct recent-session picker open");
-  await act(async () => {
     sessionSearch.dispatchEvent(new KeyboardEvent("keydown", { key: "Escape", bubbles: true, cancelable: true }));
     await flushTimers();
   });
--- a/desktop/frontend/src/__tests__/composer-goal-toggle.test.tsx
+++ b/desktop/frontend/src/__tests__/composer-goal-toggle.test.tsx
@@ -2179,6 +2179,13 @@ console.log("\ncomposer goal toggle");
   const sessionSearch = document.querySelector(".slashmenu__search") as HTMLInputElement | null;
   if (!sessionSearch) throw new Error("recent-session search did not render");
   await act(async () => {
+    const imeEscape = new KeyboardEvent("keydown", { key: "Escape", bubbles: true, cancelable: true });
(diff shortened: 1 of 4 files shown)


chengoak:
@SivanCola @esengine ping — when you have a moment, would appreciate a review on this small IME/ESC guard fix. Supersedes #6749 (which got stuck on `docs-impact` CI + recurring `mergeable: UNKNOWN`); the patch is functionally identical but the PR body now declares `Documentation-impact: none` up front and the comment was tightened to 3 lines per `REASONIX.md`. CI is clean, mergeable is stable. Happy to split or rebase if helpful.

PR #7922, opened 2026-08-08
feat(bench): workload-stratified e2e task corpus, first tranche

esengine:
**Stacked on #7917** (uses its `-task` filter and class-aware KPI readouts); rebase onto main-v2 once that merges.

## Why

The dashboards got ahead of the racetrack: the committed suite was five toy tasks, and the planner A/B proved none of them ever route past ExecutorOnly. Measurement precision kept improving while the workload stayed unrepresentative of what actually makes the agent slow.

## What

The full stratified corpus — two tranches totalling 44 new tasks reach the 46-task target in every class (plus 3 legacy smoke tasks):

| Class | Count | Example tasks |
|---|---|
| `atomic-bugfix` | `bugfix-tail-slice`, `bugfix-store-roundtrip` — short anchored fixes, route ExecutorOnly by design |
| `repo-exploration` | `explore-last-stage`, `explore-unreachable` — multi-file reading, invented-token answers so they can't be guessed |
| `multi-file-bugfix` | `mfix-timeout-units`, `mfix-record-roundtrip` — one bug spanning two files; prompts naturally engage the planner gate |
| `refactor` | `refactor-extract-validation`, `refactor-limits-module` — behavior-preserving, structure asserted via inspect/source checks |
| `failing-test-diagnosis` | `diagnose-float-total`, `diagnose-version-sort` — unittest red → fix the source; test files checksummed so fixing the test cannot pass |
| `api-integration` | `integrate-ratelimiter`, `integrate-kvstore` — use a provided local package per its README (single-transaction semantics audited via commit log) |
| `ambiguous` | `ambiguous-robust-stats`, `ambiguous-parser-formats` — underspecified asks; graders accept the defensible core only |
| `long-horizon` | `longh-wordfreq`, `longh-ledger` — multi-requirement specs, planner-depth full |

Existing tasks join the taxonomy (`fix-add-bug` → `atomic-bugfix`, `compaction` → `repo-exploration`); fizzbuzz/palindrome/subagent-delegation stay as legacy smoke tasks.

## Grader discipline

- Every task validated in both directions before commit: `verify.sh` fails on the pristine seed and passes on a reference solution.
- Graders judge artifacts and tracebacks, never exit-code style or other behavior the prompt leaves unspecified (two first-run failures were grader opinion and were fixed here).
- Python graders pin a throwaway `PYTHONPYCACHEPREFIX`: macOS system Python caches bytecode centrally keyed by absolute path, so a same-size same-mtime-second edit would otherwise execute stale bytecode while tracebacks display the new source. Documented in the README's verify.sh contract.

## Real-run shakedown (8 tasks, one per class, deepseek-v4-flash)

6/8 solved — and the corpus immediately differentiates where the toy suite couldn't: the two failures were `ambiguous` (1m27s) and `long-horizon` (3m39s, 26 model requests). Solved/hour drops from ~364 (toy suite) to ~52; TTCS p90 34.8s; calls/round 1.3.

## Testing

- `bash validate` sweep: 16/16 graders fail-on-seed and pass-on-reference
- 8-task real-provider shakedown above
- `go test ./cmd/e2ebench/` unaffected (no harness code in this PR beyond the stacked base)


esengine:
Tranche 2 pushed: the corpus now hits the full 46-task stratification — atomic-bugfix 8/8, repo-exploration 6/6, multi-file-bugfix 8/8, refactor 6/6, failing-test-diagnosis 6/6, api-integration 4/4, ambiguous 4/4, long-horizon 4/4 (+3 legacy smoke). All 35 non-legacy graders validated in both directions (fail on pristine seed, pass on a reference solution). The README stratification table now shows targets met.

PR #7941, opened 2026-08-08
feat(bench): live flight-recorder dashboard (e2ebench -mode serve)

esengine:
Stacked on #7937 (uses its outcome series). Live local dashboard for e2ebench runs.

## What

- `e2ebench -mode serve -trajectories <dir> [-suite <root>] [-addr host:port]`: serves a self-contained dark "flight recorder" page polling `/api/state` every 2s while a bench run writes trajectories.
- The run is a strip of task segments (done / running / queued); each task a recessed per-round trace. Verified-win and regression cells extend up/down off the baseline - polarity is carried by geometry because the green/red pair alone fails CVD validation (deutan dE 4.1).
- Incremental renderer: keyed DOM, cells pop in, numbers tween, hover state survives polling; custom cursor-following tooltips; click a run segment to jump to its task card.
- Replay engine: play back a finished (or partial) run by wall-clock timestamp with a scrubber, 30/120/600x speed, and a live-tail escape hatch.
- `summarizeTrajectory` splits into `scanTrajectoryFile` + `finish` so the dashboard can read the raw per-round series; outcome rounds carry their observation timestamp (additive JSONL field `t`).

## Verification

- `go vet ./cmd/e2ebench/` and `go test ./cmd/e2ebench/` pass; page exercised against the recorded 49-task run (live view, replay, tooltips, jump-to-card) with headless-Chromium screenshots in light/dark.

Documentation-impact: none - e2ebench is a developer benchmarking tool; its flags are self-documented via -h and not covered by docs/*.md

### DIFF
--- a/cmd/e2ebench/main.go
+++ b/cmd/e2ebench/main.go
@@ -188,7 +188,8 @@ func main() {
 		fmt.Fprintf(flag.CommandLine.Output(), "  %[1]s -profile delivery\n", strings.Replace(flag.CommandLine.Name(), "e2ebench", "go run ./cmd/e2ebench", 1))
 	}
 
-	mode := flag.String("mode", "suite", "suite | diff | swebench | compare | traj")
+	mode := flag.String("mode", "suite", "suite | diff | swebench | compare | traj | serve")
+	addr := flag.String("addr", "127.0.0.1:7480", "serve mode: live dashboard listen address")
 	subset := flag.String("subset", "benchmarks/swebench/subset.json", "swebench mode: instance subset file")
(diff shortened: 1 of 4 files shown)


esengine:
Self-review notes (no code change needed): CSP-free local page, all state derived from /api/state; the renderer is keyed so polling cannot destroy hover/replay state; replay bounds guard division by zero when a run has a single round; task IDs come from local trajectory filenames (trusted input for a localhost dev tool). One deliberate trade-off documented here: the dashboard does not yet render the debt/TTFDC fields added later in the stack — follow-up once the enforcement experiments need them on screen.

PR #7943, opened 2026-08-08
feat(bench): per-round cognition telemetry in trajectory digests

esengine:
Stacked on #7941. PR 1 of the cognition-efficiency series: pure measurement, no runtime change.

## Why

The 49-task tail decomposed into pure model-generation time (0 retries, 0 compactions, ~99% cache), but the trajectory digest could not say what a 29s round bought. Usage events have carried `reasoningTokens`/`completionTokens` all along - this joins them per round so slow rounds become attributable and the numbers backfill over every already-recorded trajectory.

## What

- Each classified model round now emits a `roundDigest`: gap ms, tool ms, executor reasoning/completion/prompt tokens, and the round's actions. Executor-only - planner and subagent usage stay in their own buckets.
- Slow-round census: rounds with a gap >= 8s accumulate count/gap/reasoning totals per run.
- `delegation` becomes its own round classification (task/parallel_tasks/fleet/research). Previously such rounds classified as `mutation`, hiding an 82s research call inside a bugfix's round ledger. Counted productive, so waste metrics keep their meaning.
- One new report line in suite reports and `-mode traj`:

  `**Cognition** (49 recorded runs): **reasoning** 38,720 tok · **completion** 108,773 tok (806 reasoning/solved) · **output rate** p50 138 · p90 393 tok/s · **slow rounds** (>=8s) 11 = 22% of model time, 3,852 reasoning tok · **delegation** 2 rounds (1m23s in subagents)`

  (real backfilled output over the recorded 49-task run)
- `intervals.go` extracted from `trajectory.go` (interval math, unchanged) to stay under the file-size ceiling.

## Verification

- `go test ./cmd/e2ebench/` (new fixture test: per-round token join, subagent exclusion, delegation classification, slow census, render line) - all pass
- `gofmt`, `go vet`, `repolint` clean
- Backfill exercised against the full recorded 49-task trajectory directory

Documentation-impact: none - e2ebench is a developer benchmarking tool; the new report line and JSON fields are self-describing and not covered by docs/*.md

### DIFF
--- a/cmd/e2ebench/cognition.go
+++ b/cmd/e2ebench/cognition.go
@@ -1,0 +1,98 @@
+package main
+
+import "fmt"
+
+// roundDigest is one model round's cost/outcome line: the gap that preceded
+// the batch, the executor thinking that gap bought, and what the round did.
+type roundDigest struct {
+	Index            int      `json:"i"`
+	Outcome          string   `json:"outcome"`
+	GapMs            int64    `json:"gap_ms"`
+	ToolMs           int64    `json:"tool_ms,omitempty"`
+	ReasoningTokens  int64    `json:"reasoning_tokens,omitempty"`
+	CompletionTokens int64    `json:"completion_tokens,omitempty"`
+	PromptTokens     int64    `json:"prompt_tokens,omitempty"`
+	Actions          []string `json:"actions,omitempty"`
+}
+
+// slowRoundGapMs is the census threshold: a model gap this long is a large
+// cognition purchase worth itemizing (p90-tail rounds sit well above it).
+const slowRoundGapMs = 8000
+
+// recordRound books one classified round into both the outcome tallies and
+// the per-round cognition digest. A nil batch is a finalization round.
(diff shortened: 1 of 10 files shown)


esengine:
Self-review pass over the stack found one real defect here, now fixed in the amended commit:

**Slow-round census counted recovery/compaction-tainted gaps.** A 30s provider-retry gap would have entered the census as a "slow thinking round" carrying near-zero reasoning tokens, misattributing adapter flakiness as cognition. The census now requires the gap to be untainted (`!gap.tainted && !gap.compaction`); the round's gap itself stays booked in the round digest and outcome tallies, so nothing is hidden — it just no longer counts as a cognition purchase. Covered by TestSlowRoundCensusExcludesRecoveryGaps.

Also verified during review: executor-only token attribution (planner/subagent/sidecar usage excluded before accumulation), digest/actions ordering via the batch's dispatch-order names slice, and the delegation round class staying inside productiveOutcomes so existing waste metrics keep their meaning.

PR #7946, opened 2026-08-08
feat(agent): shadow admission control for research delegation

esengine:
Stacked on #7943. PR 2 of the cognition-efficiency series: shadow only, nothing is blocked.

## Why

In the recorded 49-task run, one `research` delegation on a config-format bugfix cost 82.1s of subagent time plus ~35s of surrounding model gaps - about 65% of the slowest task's wall clock, on a turn that needed no external information. Before enforcing anything, this PR measures how often a conservative admission boundary would have refused such calls, and what it would have reclaimed.

## What

- `delegationAdmission` (internal/agent/delegation_admission.go): a pure verdict for each gated delegation call. Local mutation-shaped turns (via `taskintent.Classify`, judging the same bounded classifier text the delivery gates already store) get `deny - local_fix_no_external_need` unless the user explicitly asked for research, the call's arguments cite an http(s) source, or the turn is not a local fix. Gated set: `research` only, per the data.
- New content-free sink capability `RecordDelegationAdmission` (tool/verdict/reason/intent enums, never query text), forwarded through every capability-forwarding decorator and appended to trajectories as `delegation_admission` records.
- Verdicts are recorded post-batch alongside the existing shadow observers (`applyBatchGuards`); zero behavior change, and the run-loop stays exactly its former size (the turn text is reused, not re-stored).
- e2ebench joins verdicts with delegation tool durations (order-independent: the verdict record lands after the tool result) and reports:

  `**Delegation admission** (shadow): N gated calls · would deny M (P%) · subagent time behind denied calls Xs`

## Verification

- New tests: verdict table (deny / user-requested / external-source / non-local-intent), gated-tools-only emission, trajectory record round-trip, e2ebench join + render.
- `go test` across e2ebench, agent, trajectory, event, notify, telemetry, stats, control, cli, boot, tool/builtin - all pass; `gofmt`/`go vet`/`repolint` clean.

Cache-impact: none - the provider-visible prefix and all prompt bytes are unchanged; the shadow only observes tool calls and emits sink audits
Cache-guard: existing progress-guard and storm tests pin injected texts; TestObserveDelegationAdmissionRecordsOnlyGatedTools proves the shadow emits audits without touching results
Documentation-impact: none - shadow measurement is invisible in the product; the trajectory record and e2ebench report line are developer-facing and not covered by docs/*.md

### DIFF
--- a/cmd/e2ebench/cognition.go
+++ b/cmd/e2ebench/cognition.go
@@ -43,6 +43,31 @@ func (t *trajScan) recordRound(outcome string, gap gapInfo, b *toolBatch) {
 	}
 }
 
+// renderDelegationAdmission aggregates the shadow admission verdicts: how many
+// expensive delegation calls a local-fix boundary would have refused, and the
+// subagent time those refusals would have reclaimed.
+func renderDelegationAdmission(results []result) string {
+	calls, denies := 0, 0
+	var deniedMs int64
+	for _, r := range results {
+		if r.Trajectory == nil {
+			continue
+		}
+		calls += r.Trajectory.DelegationCalls
+		denies += r.Trajectory.DelegationDenies
+		deniedMs += r.Trajectory.DeniedDelegationMs
+	}
+	if calls == 0 {
+		return ""
+	}
+	line := fmt.Sprintf("**Delegation admission** (shadow): **%d** gated calls · **would deny** %d (%s)",
+		calls, denies, pct(denies, calls))
+	if deniedMs > 0 {
+		line += fmt.Sprintf(" · **subagent time behind denied calls** %s", dur(deniedMs))
+	}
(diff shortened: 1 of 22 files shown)


esengine:
Self-review notes (no code change needed): the verdict is judged against the bounded classifier text the delivery gates already store, so admission and gates read the same words; audits are content-free enums; blocked-by-permission calls still record a verdict, which is intentional (the shadow measures the model's intent to delegate, not execution). Live-chain validated in a real `reasonix run`: an explicit research request produced `{allow, user_requested, mutation}` through telemetry→notify→trajectory. Known limitation, by design: the marker list for user_requested is a narrow V1 heuristic and can only produce false allows (never false denies matter in shadow mode). Rebased onto the amended #7943.

PR #7947, opened 2026-08-08
feat(evidence): verification-debt telemetry in the outcome shadow

esengine:
Stacked on #7946. PR 3 of the cognition-efficiency series: telemetry only, no behavior change.

## Why

The one failed task in the recorded 49-run suite spent 75s of model time and 0.5s in tools - mutations without ever asking the environment whether the hypothesis held. Solution stall cannot see this (it also fires on healthy runs); verification debt names it precisely: how long does a mutation sit unverified.

## What

- `OutcomeTracker` gains a debt state machine: a mutation opens debt, each round without a discriminating observation ages it (mutation round included), a discriminating observation settles it. Per-round samples now carry `discriminating` and `debt_age`.
- **Discriminating is a two-tier classifier by design**: `IsDeliveryVerificationCommand` (unchanged, still owns readiness), OR any command mentioning the basename of a file mutated this turn - the deterministic hook for repro scripts and targeted runs. The recorded corpus proves the strict tier alone cannot clear debt on 43/49 runs that verify via ad-hoc scripts, so a debt-based policy built on it would misfire on tasks that pass today.
- Trajectory `outcome_progress` records carry both fields (additive keys).
- e2ebench condenses per run: `ttfdc_ms` (run start to first discriminating observation; zero = never) and `debt_age_max`, and the Outcome-shadow report line gains: `**discriminating checks** in K/N runs (TTFDC p50 Xs) · **verification debt max** D rounds`.
- The conservative runtime policy (two blind mutations -> "run the cheapest discriminating check" nudge) deliberately waits for this shadow data.

## Verification

- New tests: debt lifecycle (open/age/settle via repro-script and via delivery verification), trajectory field round-trip, e2ebench TTFDC/debt aggregation + render.
- `go test` across evidence, e2ebench, trajectory, agent, tool/builtin, boot, cli - all pass; `gofmt`/`go vet`/`repolint` clean.

Cache-impact: none - no prompt bytes change; the shadow only observes receipts and records samples
Cache-guard: TestOutcomeShadowRecordsEveryRoundWithoutTouchingGuards plus progress-guard tests pin all injected texts; the debt fields ride the existing sample channel
Documentation-impact: none - shadow telemetry is invisible in the product; trajectory fields and the e2ebench report line are developer-facing and not covered by docs/*.md

### DIFF
--- a/cmd/e2ebench/cognition.go
+++ b/cmd/e2ebench/cognition.go
@@ -43,6 +43,31 @@ func (t *trajScan) recordRound(outcome string, gap gapInfo, b *toolBatch) {
 	}
 }
 
+// renderDelegationAdmission aggregates the shadow admission verdicts: how many
+// expensive delegation calls a local-fix boundary would have refused, and the
+// subagent time those refusals would have reclaimed.
+func renderDelegationAdmission(results []result) string {
+	calls, denies := 0, 0
+	var deniedMs int64
+	for _, r := range results {
+		if r.Trajectory == nil {
+			continue
+		}
+		calls += r.Trajectory.DelegationCalls
+		denies += r.Trajectory.DelegationDenies
+		deniedMs += r.Trajectory.DeniedDelegationMs
+	}
+	if calls == 0 {
+		return ""
+	}
+	line := fmt.Sprintf("**Delegation admission** (shadow): **%d** gated calls · **would deny** %d (%s)",
+		calls, denies, pct(denies, calls))
+	if deniedMs > 0 {
+		line += fmt.Sprintf(" · **subagent time behind denied calls** %s", dur(deniedMs))
+	}
+	return line + "\n\n"
+}
+
 // renderCognition prices what the model's thinking bought: totals, the output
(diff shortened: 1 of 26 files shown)


esengine:
Self-review found one real defect here, now fixed in the amended commit:

**Read-only inspection of a mutated file settled verification debt.** `cat repro.py` or `grep x repro.py` mentions the mutated basename and counted as a discriminating observation — but reading a file cannot falsify a hypothesis. `commandExercisesMutation` now returns false for commands `shellsafe.CommandIsReadOnly` classifies as read-only (the same classifier the shell contract already trusts), so only commands able to execute the mutated file discriminate. Delivery-verification commands are unaffected. Covered by a new case in TestOutcomeTrackerVerificationDebtLifecycle.

Also verified: debt state resets with the turn evidence (tracker is rebuilt in resetTurnEvidence), TTFDC guards against zero timestamps, and the run-2 bench exercised the field end-to-end (41/49 runs discriminating, debt max 6).

PR #7953, opened 2026-08-08
feat(bench): verification-stress suite for the EBM mechanism experiment

esengine:
Stacked on #7952. The mechanism experiment's instrument: a suite where the EBM trigger's precision is measurable per class, not as one blended eligibility number.

## Design

| class | n | shape | what it measures |
|---|---|---|---|
| ebm-positive | 10 | scattered 2-spot bugs, cheap high-information oracle in the workdir (unittest or check script) | sensitivity: eligibility should be high; early discriminating evidence should pay |
| ebm-neutral | 5 | coherent multi-file migrations (schema rename, enum split, API rename, unit shift, config flatten) - nothing imports/runs until the whole patch lands | specificity: a nudge mid-patch must not derail correct work; the copy's "finish the minimum coherent change" clause is exactly what gets tested |
| ebm-negative | 5 | no cheap oracle: doc contracts, prod-only migration, a 45s soak as the only check, style renames, unrunnable network client | harm bound: the nudge must not shove the model into worthless or expensive checks |

Per the experiment plan, headline metrics are per-class exposure rates (positive/neutral/negative) and the mechanism chain (blind peak, debt area, extra blind mutations after EBM, TTFDC) on paired eligible trajectories - not overall wall time.

## Grader trust

Every task was validated in both directions before landing: the seeded workdir FAILS `verify.sh`, and a reference solution PASSES it. One grader case was rewritten during validation because it tested an unspecified float rounding boundary (9.995-to-cents) - graders must not punish unspecified behavior.

## Usage

```
e2ebench -suite benchmarks/verification-stress                 # arm A
e2ebench -suite benchmarks/verification-stress -policy ebm     # arm B
```

Documentation-impact: none - benchmark fixtures for a developer experiment; not covered by docs/*.md

Cache-impact: none - benchmark fixtures only; commits inherited from already-merged stack parents are content-identical to main-v2, so no prompt or tool bytes change
Cache-guard: existing progress-guard and fork payload-invariant tests pin all prompt bytes; this PR adds no runtime code

### DIFF
--- a/benchmarks/verification-stress/tasks/vs-neg-doccontract/task.toml
+++ b/benchmarks/verification-stress/tasks/vs-neg-doccontract/task.toml
@@ -1,0 +1,4 @@
+prompt = "rates.py and fees.py are public API but undocumented. Add docstrings to every public function describing parameters, return value, and the zero/negative-amount edge cases. Do not change behavior."
+class = "ebm-negative"
+max_steps = 12
+timeout_sec = 240
--- a/benchmarks/verification-stress/tasks/vs-neg-doccontract/verify.sh
+++ b/benchmarks/verification-stress/tasks/vs-neg-doccontract/verify.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+set -e
+export PYTHONPYCACHEPREFIX="$(mktemp -d)"
+python3 - <<'PY'
+import rates, fees
+for fn in (rates.convert, rates.spread, fees.wire_fee):
+    assert fn.__doc__ and len(fn.__doc__.strip()) > 20, fn.__name__
+assert fees.wire_fee.__doc__ and ("0" in fees.wire_fee.__doc__ or "zero" in fees.wire_fee.__doc__.lower())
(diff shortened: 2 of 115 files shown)


esengine:
Suite v2 after the first per-class exposure measurement (the instrument worked exactly as intended):

**v1 exposure came back inverted — 0/10 positive, 5/5 neutral, 5/5 negative.** Root cause: positive tasks shipped their oracle in the workdir and advertised it in the prompt, so the agent verified immediately and blind never reached 3 — the pathology EBM targets (verifiable-but-nobody-hands-you-the-test) never arose.

**v2**: oracle files removed from positive workdirs, assertions inlined into the hidden graders, prompts describe symptoms only. Cheap discriminating checks remain constructible (a 3-line repro settles debt via the broad classifier). All 20 graders re-validated in both directions.

**v2 measured exposure (arm A):** positive 6/10 · neutral 5/5 · negative 5/5, pass 9/10 on positive, TTFDC p50 18s → 26s — the temptation to defer verification is now real. All three experiment populations are live.

PR #7958, opened 2026-08-08
feat(agent): eligibility-point counterfactual fork harness

esengine:
Stacked on #7953. The intervention phase's core infrastructure: policy evaluation as a local causal experiment.

## Why

Policy triggering is a function of the stochastic trajectory - the same task reaches `blind>=3` on one run and peaks at 2 on the next (measured twice on this corpus). Whole-task A/B therefore mixes "did the trigger arise at all" variance into every treatment estimate. Forking control and treatment from the identical eligible state removes that layer entirely, and post-fork totals ARE Remaining-Cost-From-Eligibility - no alignment arithmetic.

## Hard constraints implemented

1. **True same-state fork**: one versioned `ForkBundle` (conversation snapshot, workspace copy, turn input, outcome-tracker seed); both arms boot from it with identical config.
2. **Pre-treatment capture**: capture arms at first eligibility and refuses under live enforcement - a treated state can never become a fork origin (test-pinned).
3. **No fake new turn**: continuation runs the normal Run path; turn-local classification computes from the same input, then the appended user message is replaced wholesale by the frozen conversation via one restore hook.
4. **Versioned bundle** from day one - the same primitive will serve the reasoning governor, delegation admission, rollback and stop-policy experiments.
5. **No deterministic-replay pretension**: pair identity (task x rep x arm) is preserved through the report; analysis is per-pair, per-class.
6. **The payload invariant, as a test**: the control fork's first provider request deep-equals the request the uninterrupted run made after the eligible round; the treatment payload differs in exactly one message by exactly the nudge suffix. This test caught a real contamination during development - an in-round capture missed the eligible round's tool results (8-vs-7 messages) - which is precisely the class of silent drift it exists to catch. Capture therefore moved to the provider wrapper, freezing the state right before the next request.

## Usage

```
# capture bundles during a baseline (arm-A) suite run
e2ebench -suite benchmarks/verification-stress -fork-capture bundles/

# control/control sanity, then control/treatment
e2ebench -mode fork -bundles bundles/ -suite benchmarks/verification-stress -arm control,control -reps 2
e2ebench -mode fork -bundles bundles/ -suite benchmarks/verification-stress -arm control,treatment -reps 3
```

Arms alternate order per rep so provider drift cannot systematically favor one arm. The report is per-pair rows plus per-class aggregation; no blended overall score by design.

## Verification

- Four new agent tests: request-payload invariant (control), single-diff treatment, enforcement-refuses-capture, workspace-copy hygiene; all existing suites (agent, evidence, e2ebench, cli, boot) pass; gofmt/vet/repolint clean.

Cache-impact: none - capture and continuation are env-gated experiment paths; with the env unset no code path changes and no prompt bytes move
Cache-guard: TestForkControlContinuationMatchesOriginalRequest pins the payload byte-contract; default-off is pinned by every existing agent test running without the env
Documentation-impact: none - env-gated experiment infrastructure plus developer-tool flags; user-facing docs are untouched until a policy graduates

### DIFF
--- a/benchmarks/verification-stress/tasks/vs-neg-doccontract/task.toml
+++ b/benchmarks/verification-stress/tasks/vs-neg-doccontract/task.toml
@@ -1,0 +1,4 @@
+prompt = "rates.py and fees.py are public API but undocumented. Add docstrings to every public function describing parameters, return value, and the zero/negative-amount edge cases. Do not change behavior."
+class = "ebm-negative"
+max_steps = 12
+timeout_sec = 240
--- a/benchmarks/verification-stress/tasks/vs-neg-doccontract/verify.sh
+++ b/benchmarks/verification-stress/tasks/vs-neg-doccontract/verify.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+set -e
+export PYTHONPYCACHEPREFIX="$(mktemp -d)"
+python3 - <<'PY'
+import rates, fees
+for fn in (rates.convert, rates.spread, fees.wire_fee):
+    assert fn.__doc__ and len(fn.__doc__.strip()) > 20, fn.__name__
+assert fees.wire_fee.__doc__ and ("0" in fees.wire_fee.__doc__ or "zero" in fees.wire_fee.__doc__.lower())
+assert rates.convert(2, 3) == 6 and fees.wire_fee(-5) == 0.0
+PY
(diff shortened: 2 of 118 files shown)


esengine:
Added a third continuation arm during the first governor probe: `low-effort` — same frozen state, no nudge, whole-continuation reduced thinking via the provider effort knob (`thinking.type`/`reasoning_effort` on DeepSeek). This makes the harness cover both intervention families (evidence nudge, cognition budget) with zero agent-side changes. First probe over 16 frozen states x 3C/3L: pass exactly preserved in 16/16 states (36/48 -> 36/48), model time down in 12/16 and reasoning tokens down in 14/16 states (sign-consistent; medians neu -3.9s/-200tok, neg -1.2s/-89, pos -0.9s/-38). Fresh-state replication pending before any conclusion graduates.

PR #7976, opened 2026-08-08
fix(cli): make number keys hot-select in every numbered menu

loserbcc:
Every modal menu in the chat TUI renders numbered rows (`❯ 1. …` via `rowLine`), but only the ask chooser and the tool-approval prompt actually accept digit presses. Everywhere else the numbers are painted on but dead — you have to arrow to the row and press Enter. The `numberKeyIndex` helper already exists, but it's only wired into the MCP manager's detail/mode stages.

This wires digit hot-select into every numbered menu, matching each menu's existing Enter semantics (same pattern Claude Code uses for its numbered prompts):

- **chooser**: digits on a multi-select question now toggle the option — previously `chooserActivate` treated a digit like Enter and committed the answer without selecting anything; single-select keeps select-and-advance
- **/clear confirm**: `1` clears, `2` cancels
- **/copy picker**: digit copies that part
- **resume picker** (legacy list path): digit resumes that session
- **rewind**: stage 0 digit jumps to that turn; stage 1 digit applies that scope (parity with the existing `b/c/d/f/s/u` letter shortcuts)
- **skill picker**: sources list, detail actions, and delete confirm
- **MCP manager**: confirm-remove and confirm-clear-auth stages

Searchable pickers (quick picker, resume search, skill search) are deliberately untouched — digits there remain query text, matching Claude Code's searchable menus.

Out-of-range digits are ignored (no accidental activation). New tests in `internal/cli/digit_select_test.go` cover the helper, single/multi chooser behavior, and one representative menu per pattern.

Documentation-impact: none - behavior change only makes already-rendered menu numbers interactive; no documented workflow, flag, or config changes, and docs/ describes none of these menus key-by-key.

### DIFF
--- a/internal/cli/chooser.go
+++ b/internal/cli/chooser.go
@@ -138,11 +138,16 @@ func (m chatTUI) handleChooserKey(msg tea.KeyPressMsg) (tea.Model, tea.Cmd) {
 	case "enter":
 		return m.chooserActivate(c.cursor)
 	default:
-		// number keys 1..9 jump to / pick an option
-		if s := msg.String(); len(s) == 1 && s[0] >= '1' && s[0] <= '9' {
-			if idx := int(s[0] - '1'); idx < len(q.Options) {
-				return m.chooserActivate(idx)
+		// number keys 1..9 pick an option: toggle it in a multi-select
+		// question, select-and-advance in a single-select one.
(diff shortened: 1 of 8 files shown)


loserbcc:
Superseded by #7977 — same change, refiled from our ScrappyLabs account. Sorry for the noise.

PR #8023, opened 2026-08-09
fix(control): prevent stale Plan and Goal results after reasoning-only stops

Nath-Vikky:
## Summary

- Extend #7817's visible-final contract to Controller Plan and active Goal workflows.
- Prevent a DeepSeek reasoning-only stop from making Plan approval, Goal evaluation, or Stop hooks reuse visible text from an earlier turn.
- Recover a current visible result with a bounded, finalization-only repair; fail closed when the current result cannot be established.

## Problem

DeepSeek thinking mode can finish with non-empty reasoning, empty visible `content`, and `finish_reason=stop`. Ordinary chat intentionally accepts that shape in one request, as established by #6618.

Controller semantic consumers previously used `lastAssistantText`, which scanned the entire transcript backwards. When the current assistant message had empty content, the scan skipped it and returned an older answer. Deterministic pre-fix regressions showed that this could:

- approve and seed an earlier Plan even though the current reasoning described a different plan;
- pass an earlier completion claim to the Goal evaluator and transition the current Goal incorrectly; or
- send stale text to a Stop hook.

#7817 protects agent-level subagents, tool-enabled planners, Guardian, and Delivery. Controller Plan/Goal consumers remained outside that contract.

### Scoped visible-final recovery

- Plan mode and active Goal turns opt in through a per-Run context contract.
- The contract lives in the current `turnRuntime` and is reset with the whole turn; it does not leak into later ordinary runs.
- A qualifying reasoning-only stop starts a finalization-only repair, capped at two additional samples.
- A repair response that attempts tools is recorded with paired blocked results before registry resolution, preview, permission, or execution. It cannot repeat work or side effects.
- Steer input arriving between repair rounds is deferred. It is consumed only after a clean visible repair returns control to a normal tool-capable round.

Construction-time `Options.RequireVisibleFinal` keeps its existing generic retry semantics for subagents, planners, and Guardian. This PR's stricter finalization-only behavior applies only to the Controller's run-scoped Plan/Goal contract.

### Current-turn result selection

- Plan approval reads only the current turn's terminal visible assistant result.
- Goal evaluation reads the captured execution-stage final from the current foreground workflow.
- Stop hooks may read current-turn visible text or a current tool preamble, but never fall back to an earlier turn.
- Plan + Goal workflows preserve the execution final as Goal evidence rather than the preceding Plan proposal.

The boundary records the Session identity, starting message position, and an in-memory whole-message-log generation. `Replace`, `Rewrite`, and explicit rewrite invalidation make the result fail closed; local metadata updates such as tool preview, resolution, and decision receipts do not. Automatic context maintenance remains projection-only and does not invalidate the canonical Session boundary.

## Safety properties and trade-offs

- **Ordinary chat is unchanged.** A normal reasoning-only stop remains a one-request success.
- **Normal visible Plan/Goal turns add no provider calls.** Only a scoped reasoning-only stop can add repair samples, with a hard cap of two.
- **No repair-time tools execute.** Calls are paired with blocked results before preview, resolution, permission, registry execution, or `executeBatch`. Tool schemas are still read to construct the provider request; that lookup is read-only.
- **Host limits remain authoritative.** Armed recovery/grace finalization wins. Explicit `max_steps` does not receive an extra repair sample. Task-budget crossing shares the existing single finalization allowance; a repair round that crosses it pauses without adding another prompt/sample. When max-step and spend boundaries coincide, the existing task-budget classification is retained without another sample.
- **Existing structured Goal reports keep mainline semantics.** A recorder-validated, readiness-ready terminal `update_goal` report remains subject to the existing completion-wins-over-pause rule and does not use assistant-text fallback. Ordinary repair exhaustion does not call the evaluator or advance a Goal from stale text.
- **Whole-log replacement is conservatively fail-closed.** Rewind/restore, Session replacement, and the experimental fork-bundle path can yield an empty current result instead of accepting an answer after the replacement. This may reject a valid current result in that uncommon path, but it cannot accept an older one.
- **Stop-hook partial text is conditional.** Current partial/tool-preamble text is preserved when the message log is stable. Cancellation/provider-error cleanup that replaces the whole log can instead give the Stop hook an empty result; it will not fall back to stale text.
- **No persisted-state migration.** No session schema, WAL format, projection format, or stored Goal/Todo representation changes. The Extension ProviderRequest protocol gains one optional `disableServerTools` policy bit; ordinary requests omit it. Rollback is a pure revert.

The synthetic repair message is excluded from user-authored turn counting, preview, title generation, and chat display.

## Scope

Relative to `main-v2@c4fbdd1d07400180fe9b4746f2c6326e53116e28`:

- 37 files, `+2111/-137` total;
- approximately 1,493 added lines are tests; four generated protocol artifacts carry the optional request-policy field and schema hash.

The larger latest-main delta closes one atomic safety boundary introduced by server-side search: strict visible-final repair, any pressure-triggered summary before it, built-in provider wire construction/history replay, and Extension request transport all receive the same server-tool prohibition. Splitting those layers would leave a repair-time side-effect path open.

### Red-before evidence

Before the fix, deterministic Controller tests reproduced all three stale-consumption paths: an earlier dangerous Plan reached approval/Todo seeding, an earlier completion claim completed a new Goal, and a Stop hook received the previous assistant answer. A first-turn reasoning-only Plan also returned successfully without entering approval. The corresponding tests pass with the new contract.

### Focused stress and compatibility

- Agent strict repair, construction-time compatibility, tool blocking, budget/step boundaries, run isolation, and steer deferral: 50-100 shuffled repetitions passed.
- Controller Plan/Goal/Stop/current-turn boundary matrices: 30-100 shuffled repetitions passed.
- Rewrite/Replace/fork-style invalidation and Session replacement: up to 100 repetitions passed.
- Existing subagent, two-model planner, Coordinator, and Guardian visible-final behavior passed; full `internal/guardian` passed.
- Full `internal/control` reached only two Windows shell-cancellation timing failures (`TestRunShell_CancelStopsCommand` and `TestRunShell_HeredocCancelReleasesTurn`). Both failed 5/5 with the same roughly 15-second `TurnDone` timeout on pristine `c4fbdd1d`; all non-environment Control regressions passed.
- `internal/agent`, excluding the six environment-dependent cache-hit HTTP tests listed below, passed with a 15-minute timeout.

Additional gates passed:

```text
go test ./... -run '^$' -count=1 -timeout=20m
go vet ./internal/agent ./internal/control ./internal/guardian
go run ./tools/repolint
gofmt / git diff --check
```

`repolint` was clean with the current upstream baseline; this PR does not modify the baseline.

### Independent review

Two independent latest-main reviews returned **GO / no P0-P2 blocker**. They separately checked:

- `turnRuntime` lifetime/reset and compatibility with the current Agent ownership split;
- repair tool-call pairing before preview/registry/execution;
- task-budget, max-step, recovery, and grace precedence;
- construction-time versus run-scoped visible-final behavior;
- steer handling;
- Session replacement/rewrite, local metadata updates, defer ordering, early hook exits, and Goal FSM evidence selection.

### Latest-main server-tool safety

- Strict repair requests set `disableServerTools` after request interceptors and freeze it before dispatch.
- Anthropic and Responses omit server-search declarations and historical search replay; an explicit forbidden server-search chunk fails closed before Session/UI commit.
- Pressure-triggered context summaries that precede repair carry the same policy and cannot silently fall back after a forbidden search.
- Built-in wire suppression, Responses item-then-search ordering, Extension DTO/schema round-trip, ordinary-request byte stability, and complete Agent Run ordering were tested repeatedly.
- Older Extension SDKs with strict unknown-field decoding will reject a strict repair request containing the new optional field. This is a deliberate fail-closed compatibility limitation; ordinary requests omit the field and remain byte-stable.
- A fully trusted malicious provider sidecar could ignore the policy and conceal remote work as text; the host cannot detect that existing trust-boundary violation. Explicit server-search chunks are rejected.

### Environment limitations

- Full `internal/agent` encounters the repository's existing localhost cache-hit harness issue on this Windows machine: five tests return DeepSeek HTTP 502 after roughly 107-111 seconds and `TestSetSessionResetsSessionCache` then waits in `provider.SendWithRetry`. A representative test was reproduced on frozen upstream main under the same environment. The six tests are `TestCacheHitPrefixStable`, `TestCacheHitClimbsWithoutCompaction`, `TestCacheHitSurvivesTooSmallWindow`, `TestReasoningRoundTripCost`, `TestSessionAggregateCacheRate`, and `TestSetSessionResetsSessionCache`.
- `-race` was unavailable because this Windows environment has CGO disabled.
- No live commercial DeepSeek endpoint was called. Deterministic provider scripts reproduce the relevant reasoning-only wire shape and were used for repeated fault injection.

Repository CI and CodeQL still require repository-side approval for this external contribution.

## Issues

N/A - self-discovered follow-up to #7817. Related design context: #6618 and #7519.

## Documentation impact

Documentation-impact: updated - regenerated `docs/EXTENSION_PROTOCOL.generated.md` documents the optional `disableServerTools` ProviderRequest policy used by strict visible-final repair. No user-facing command or configuration changes.

## Cache impact

Cache-impact: low - ordinary turns and normal visible responses are unchanged. Only a scoped reasoning-only stop appends a synthetic repair turn, which can change the later request suffix. No system prompt, tool schema, memory prefix, or cache-stable provider prefix changes.

Cache-guard: tests verify ordinary reasoning-only chat remains one request, the scoped requirement does not leak across Runs, and the repair prompt remains synthetic.

System-prompt-review: N/A - no system prompt or cache-stable prefix changes.


Nath-Vikky:
Latest-main safety revalidation (2026-08-11):

- Merged `main-v2@15239191e` with a normal merge; current head is
  `61f6d19ab` and the PR remains mergeable.
- Independently checked the merge against #8185/#8228/#8244: Goal/Chat run
  ceilings, Goal-stuck and Recovery pauses, ContextManager/projection CAS, and
  the Plan contract are preserved.
- An armed host pause returns before repair. If the last in-budget sample itself
  is reasoning-only before a pause is armed, repair may add at most two provider
  samples, but repair-time tools are paired as blocked and never reach the
  registry, approval, or executor.
- A validated terminal `update_goal` continues to follow main's
  completion-wins-over-pause rule; stale assistant text is never used by that path.
- Focused Agent/Controller/Plan matrices passed 20 shuffled repetitions; full
  Control/Hook/PlanContract passed; all-package compile, vet, repolint, gofmt,
  and diff checks passed. Full Agent passed in 33.962 s with a temporary
  direct-loopback test setting that was not committed.
- The standard Windows run is affected by the machine's system proxy intercepting
  localhost; the same representative 502 reproduces on frozen main. Local
  `-race` is unavailable with CGO disabled, and no live provider was
  called in this round.

The PR body now records these boundaries, the exact latest-main scope
(15 files, `+1153/-85`), and the rollback/no-migration statement.

PR #8085, opened 2026-08-09
fix(agent): abort degenerate repetition loops at the stream boundary

ovuruska:
## Problem

DeepSeek-class models can degenerate into re-emitting the same sentence ("Let me run the gate scripts now.") hundreds of times instead of acting. Observed in a live session: ~300 near-identical "Let me do it" lines, no tool call, tokens burning until manual interrupt. Three compounding failures:

1. The provider's own `repetition_truncation` often misses sentence-level loops (it targets token-level repetition), so the stream runs to the output cap.
2. Nothing client-side watches the visible/reasoning text stream; the existing guards cover repeated failing tool calls (`repeat_failure_guard`), zero-evidence rounds (`progress_guard`), and premature finals (`final_readiness`), but not intra-stream text degeneration.
3. Worst: the degenerate text commits to the session, so the next attempt sees 300 "Let me do it" lines and the loop self-reinforces.

This is the same defense mature harnesses converge on (Claude Code: cycle detection + step budgets + discard-and-correct; see also DeepSeek-R1's model-card guidance on endless repetition).

## Fix

A per-stream `repetitionDetector` (new `internal/agent/repetition_guard.go`) segments text and reasoning chunks at line/sentence boundaries (ASCII enders need trailing whitespace so decimals survive; fullwidth CJK enders split unconditionally), normalizes segments, and counts repeats across a 192-segment sliding window. At 16 identical segments of >= 12 runes the attempt is cut off with `errRepetitionLoop` via a synchronous `filter` at the single chunk-receipt point in `streamWithFrozen`.

The abort rides the existing sampling-recovery rails: the attempt is discarded (never written to the session, killing self-reinforcement), and the frozen request is re-sampled under a deliberately smaller retry budget (`maxRepetitionRetries = 2`) than transport interruptions, since every degenerate attempt burns real completion tokens before tripping. Retry policy for both cases now lives in one `samplingRetryPlan` helper. On exhaustion the turn fails with a legible error and the junk stays display-only (LocalOnly recovery record).

Guard is on by default for every frontend via `agent.New`; `Options.RepetitionTripLimit` tunes it (zero = default, negative = off) and `REASONIX_REPETITION_GUARD=0` is the env escape hatch (governor precedent: config graduation waits on field experience).

## Tests

- Detector unit tests: sentence loops fed in odd-sized chunks, A/B/C cycles, CJK sentences, varied-prose and short-structural non-trips, carry bound on separator-less streams, nil/disabled behavior.
- Integration: a runaway-loop provider stops after initial + 2 retries with `repetition_loop` discard events; a degenerate-then-clean provider commits only the clean retry, and the degenerate text never appears in the session.
- Effect test at the final boundary (`internal/boot/repetition_effect_test.go`, real `boot.Build`): the next turn's provider request contains the clean retry text and none of the degenerate attempt.
- `TestCompactionHealthyWindowNeverLoops` fixture updated: it grew assistant text with 600 copies of the same sentence, which the new guard (correctly) reads as a degenerate loop. The fixture now emits distinct sentences at the same byte volume, preserving its intent (size growth, not literal repetition).

## repolint baseline

`tools/repolint/baseline.json` re-measured: repo file-size +6 (the documented `Options` field), repo function-size net -1 (the merged retry branches shrink `streamWithFrozen`'s callers more than the one added filter line grows it). No hot-path budget widened beyond `internal/agent/agent.go` +7 lines / +1 function unit; new logic lives in its own file.

Cache-impact: none - the guard filters received stream chunks only; prompt bytes, tool schemas, and the system-prefix composition are untouched, so the provider-visible prefix stays byte-identical
Cache-guard: existing prefix-shape guards (internal/agent cache_shape and cachehit e2e tests) plus TestEffectRepetitionGuardDiscardsDegenerateAttempt, which asserts the next-turn provider request through real boot.Build
System-prompt-review: boot diff is a new effect test only and agent.go changes are stream-side; no prompt-path change - reviewer: oguzvuruskaner
Documentation-impact: none - internal resilience guard; the only user-visible surface is an automatic discard-and-retry rendered by the existing stream-retry UI, and the env escape hatch is intentionally undocumented while the guard gathers field experience

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -278,14 +278,15 @@ type ToolHooks interface {
 // Agent drives a single task: a Provider, a tool Registry, and a Session wired
 // into the main loop.
 type Agent struct {
-	prov               provider.Provider
-	tools              *tool.Registry
-	session            *Session
-	sessMu             sync.Mutex // guards the session pointer for external Session()/SetSession
-	maxSteps           int
-	maxStepsKey        string
-	reasoningByteLimit int
-	maxOutputTokens    int
+	prov                provider.Provider
+	tools               *tool.Registry
+	session             *Session
+	sessMu              sync.Mutex // guards the session pointer for external Session()/SetSession
+	maxSteps            int
+	maxStepsKey         string
+	reasoningByteLimit  int
+	repetitionTripLimit int
+	maxOutputTokens     int
 	// executorHandoffGuard is enabled by Coordinator for the executor agent. The
 	// per-turn marker check in Run keeps ordinary single-model turns unaffected.
 	executorHandoffGuard bool
@@ -1034,6 +1035,10 @@ type Options struct {
 	// uses the default guard; a negative value disables only this client guard.
 	// Provider output budgets are a separate protocol/model capability.
 	ReasoningByteLimit int
+	// RepetitionTripLimit is how many identical output segments the recent
+	// stream window tolerates before the client repetition guard aborts the
+	// attempt. Zero uses the default guard; a negative value disables it.
+	RepetitionTripLimit int
 	// MaxOutputTokens overrides the provider's configured/default total output
 	// budget. Zero delegates to the provider; a negative value asks optional
 	// protocols to omit the budget (Anthropic still requires max_tokens).
@@ -1251,6 +1256,7 @@ func New(prov provider.Provider, tools *tool.Registry, session *Session, opts Op
 		maxSteps:                  opts.MaxSteps,
 		maxStepsKey:               maxStepsKey,
 		reasoningByteLimit:        reasoningByteLimit,
+		repetitionTripLimit:       resolveRepetitionTripLimit(opts.RepetitionTripLimit),
 		maxOutputTokens:           opts.MaxOutputTokens,
 		temperature:               opts.Temperature,
 		pricing:                   opts.Pricing,
@@ -2337,6 +2343,7 @@ func (a *Agent) streamWithFrozen(ctx context.Context, turn int, sink event.Sink,
 	if err != nil {
 		return streamedTurn{usage: provider.UsageWithRequestAttemptCount(ctx, nil), err: err}
 	}
+	repetition := newRepetitionDetector(a.repetitionTripLimit)
 
 	// A PostLLMCall hook rewrites the whole reasoning block, so when one is wired
 	// up we buffer reasoning silently and emit the transformed text once after the
@@ -2436,7 +2443,7 @@ func (a *Agent) streamWithFrozen(ctx context.Context, turn int, sink event.Sink,
 					partialCalls: partialCalls, maxArgChars: maxArgChars,
 				}
 			}
-			chunk = c
+			chunk = repetition.filter(c)
 		}
 		switch chunk.Type {
 		case provider.ChunkReasoning:
--- a/internal/agent/compact_loop_e2e_test.go
+++ b/internal/agent/compact_loop_e2e_test.go
@@ -119,6 +119,18 @@ func compactionsPerTurn(t *testing.T, windowTok int, blob, finalText string, tur
 	return perTurn, paused, prunes
 }
 
+// variedAnalysisText grows assistant text like a real answer: same volume as
+// the old repeated-filler blob (~20 chars per sentence), but with distinct
+// sentences so the client repetition guard (correctly) does not read it as a
+// degenerate loop.
+func variedAnalysisText(sentences int) string {
+	var b strings.Builder
+	for i := range sentences {
+		fmt.Fprintf(&b, "analysis part %04d. ", i)
+	}
+	return b.String()
+}
+
 func consecutiveCompactingTurns(perTurn []int) int {
 	worst, run := 0, 0
 	for _, n := range perTurn {
@@ -161,7 +173,7 @@ func TestCompactionPausesWhenWindowTooSmall(t *testing.T) {
 // session grows but reclaims enough headroom that it never fires on consecutive
 // turns and never trips the stuck guard.
 func TestCompactionHealthyWindowNeverLoops(t *testing.T) {
-	perTurn, paused, _ := compactionsPerTurn(t, 40000, "small tool output", strings.Repeat("analysis paragraph. ", 600), 20)
+	perTurn, paused, _ := compactionsPerTurn(t, 40000, "small tool output", variedAnalysisText(600), 20)
 
 	total := 0
 	for _, n := range perTurn {
--- a/internal/agent/repetition_guard.go
+++ b/internal/agent/repetition_guard.go
@@ -1,0 +1,206 @@
+package agent
+
+import (
+	"errors"
+	"os"
+	"strings"
+	"unicode/utf8"
+
+	"reasonix/internal/provider"
+)
+
+// The client repetition guard aborts a stream whose output degenerates into
+// re-emitting the same sentence instead of acting — a DeepSeek-class failure
+// the provider's own repetition_truncation often misses. Aborted attempts ride
+// the sampling-recovery discard path, so degenerate text never reaches the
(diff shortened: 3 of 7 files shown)


boscocp:
Thanks — this is the right layer for the problem, and the discard-before-commit design is the part I'd most want kept: killing self-reinforcement matters more than the detection itself. The positive cases work. Running your detector against a sentence flood on this branch:

```text
TestControlTextFloodIsCaught
    Run error      = output repetition loop detected; stream stopped by client guard
    provider calls = 3 (a discarded attempt is re-sampled)
```

Four things came up while I was pointing this at real local-model sessions. I may be misreading the intent on some of them, so I have written them as questions rather than as change requests — the measurements are what I ran, the conclusions are mine to be corrected on. The first two are about one constant, `repetitionSegmentMinRunes`, which looks to me like it is low enough to count Markdown fences and high enough to ignore a one-rune flood.

1. **Is a healthy run tripping the guard expected at this threshold?**

   `push` keeps any segment of 12 runes or more, and a thinking model drafting code in its reasoning channel emits Markdown fences and repeated code lines legitimately. I fed the guard the reasoning and visible text of eight complete `reasonix run` sessions against a real repository — a local gemma4-class model, one task per session, none of them degenerate, all eight producing files that I checked by hand. Seven pass. The longest one trips:

   ```text
   TestDetectorFalsePositiveOnRealTraces
       T1.1-A1.txt   13119 chars  ok
       T1.1-A2.txt   10012 chars  ok
       T1.1-B1.txt    5692 chars  ok
       T1.1-B2.txt    4492 chars  ok
       T1.2-R1.txt  195175 chars  TRIPPED on "```javascript"
       T1.2-R2.txt   31369 chars  ok
       T1.2-R3.txt   51966 chars  ok
       T1.2-R4.txt   22224 chars  ok
   ```

   One in eight is not alarming on its own. What makes me raise it is which one: the trip is on the longest session by a factor of four, and the window is a fixed 192 segments, so the chance of twelve collisions inside it grows with how long the model thinks. That points the false-positive risk at exactly the long-running sessions this guard is meant to protect.

   That session has 13 900 segments, 3 716 of which survive the 12-rune filter. The guard trips at segment 1148 — **30% in**, with exactly 12 occurrences of the fence inside the 192-segment window. The run had 70% of its work still ahead of it, and it went on to write every file the task asked for.

   The fence is not the only thing at the threshold. These are the most repeated surviving segments of that session:

   ```text
     50x  13 runes  '```javascript'
     25x  12 runes  "let's do it."
     22x  20 runes  'action: `read_file`.'
     20x  15 runes  'actually, okay.'
     18x  33 runes  '<script src="js/app.js"></script>'
   ```

   `"let's do it."` is twelve runes — a thinking model's verbal tic landing exactly on the floor. And what decides whether a code fence counts at all is the length of the language name:

   ```text
   '```javascript'  13 runes  COUNTS      appears 50x
   '```html'         7 runes  discarded   appears 33x
   '```bash'         7 runes  discarded   appears 4x
   '```'             3 runes  discarded   appears 84x
   ```

   If I am reading it right, a run that drafts HTML survives while the same run drafting JavaScript is discarded, re-sampled twice and then fails with `errRepetitionLoop`. Is that the intended trade, with the expectation that the stem tier absorbs it? If not, would excluding fenced code and structural lines from the count work — or, better, requiring that a trip coincide with an absence of progress? `progress_guard` already knows what a zero-evidence round looks like, and a degenerate loop is a stream with no tool call and no new evidence, which is what would separate it from a model revising a draft.

2. **Should the one-rune and separator-less shapes be in scope for this guard?**

   The DeepSeek-class degeneration I have on record is not sentence-shaped. One local run emitted 130 989 completion tokens over 44.5 minutes, ending in 35 197 hyphens across 37 442 newlines — segment length, one rune. `push` drops every one of them:

   ```text
   TestGuardAgainstRealHyphenFlood
       35 197 lines of "-": err=<nil> calls=1
   TestGuardAgainstRealDigitFlood
       130 000 chars of "1234567890": err=output repetition loop detected calls=3
   ```

   The digit ramp from the same run is caught — but only by accident. It has no separator at all, so it falls to the `repetitionCarryMaxBytes` block path, and the comment there is right that identical blocks count "when the stream repeats in phase". Whether they do is decided by the size of the SSE deltas the provider happens to send, which the client does not control. Same content, six delta sizes:

   ```text
   TestDigitFloodPhaseSensitivity
       chunk= 64 bytes -> tripped=true
       chunk= 63 bytes -> tripped=false
       chunk= 61 bytes -> tripped=false
       chunk=100 bytes -> tripped=true
       chunk=128 bytes -> tripped=false
       chunk=127 bytes -> tripped=false
   ```

   So the separator-less case looks like a coin flip on framing to me, and the one-rune case is never counted at all. Both are low-entropy streams. Is that deliberate scoping — the PR is explicitly about sentence-level loops, and these may be meant for `repetition_truncation` on the provider side? If they are in scope, would a distinct-character or distinct-segment ratio over the window be a reasonable addition? It would cover both without touching the sentence logic and would not depend on delta framing.

3. **Are tool-call chunks meant to be out of scope for `filter`?**

   ```go
   if (c.Type == provider.ChunkText || c.Type == provider.ChunkReasoning) && d.observe(c.Text) {
   ```

   Tool calls arrive as `ChunkToolCallStart`, `ChunkToolCallArgsDelta` and `ChunkToolCall` (`provider.go:717-719`), so a stream that degenerates into re-opening calls is not observed. I have that shape on record too: one completion streamed 161 `bash` openings, none ever completed, for roughly five minutes, after leaking a raw template marker into visible text. Reduced to a deterministic case on this branch:

   ```text
   TestReproToolCallFloodIsUnbounded
       Run error        = <nil>
       provider calls   = 1
       openings emitted = 501 of 500 streamed
   ```

   Same agent, same options and same stream length as the sentence flood above, which does trip. If this shape is meant to be covered, would counting `ToolCall.Name` per stream, or the openings themselves, fit the existing design? The trip limit would not have to be the same one.

4. **Rebase — is the August baseline still the one you want?**

   The merge base is `0bf2002f4` (2026-08-09); current `main-v2` is `d9d152b3f`, **2112 commits** ahead. `git merge-tree` does not come back clean:

   ```text
   CONFLICT (content): Merge conflict in internal/agent/agent.go
   CONFLICT (content): Merge conflict in internal/agent/run_loop.go
   CONFLICT (content): Merge conflict in tools/repolint/baseline.json
   ```

   The baseline conflict is the one I would watch: the ratchet has moved several times since August, so the August numbers probably need re-deriving rather than carrying forward. Happy to be wrong if you have already checked this against a newer base locally.

One smaller question: `repetitionGuardDisabledByEnv` is resolved once at package init, so `REASONIX_REPETITION_GUARD=0` cannot be set per process by an embedding host, and a test cannot re-enable the guard after disabling it. If the escape hatch is meant for field diagnosis that is probably fine — but it is worth a comment saying so, since the neighbouring `Options.RepetitionTripLimit` is per-agent.

Verification performed, on this PR's head (`7cc9b05f1`):

- `GOTOOLCHAIN=go1.26.6 go test ./internal/agent/ -run Repetition` passed, and `./internal/boot/ -run Repetition` passed — the PR is green on its own base
- the sentence-flood control above trips as designed (3 attempts, discarded, never committed)
- the real-trace false positive, the hyphen flood and the tool-call flood reproduce as shown
- `gofmt -l` and `git diff --check` are clean
- latest-base verification could not be completed: the merge conflicts listed in item 4

Happy to hand over the eight trajectories and the throwaway tests if they are useful — they are captures from running this harness against a real repository with a local model, which is where all four items came from.

BuGlessRB:
I made a similar detection PR, it uses less CPU, and is less likely to catch false positives, and works on any repetition type (single character, or more, it does not matter).  It's current and actively rebased. https://github.com/esengine/DeepSeek-Reasonix/pull/10431

PR #8123, opened 2026-08-10
fix(agent): snip the projection, not the canonical transcript

### LINKED ISSUE #7935: [Bug]: 反复显示 "snipped N stale tool results before compaction" 但上下文从未真正压缩
### Version line

### Exact version

v1.21.2

### What happened?

长时间会话中，日志里几乎每次工具调用之后都会出现一条提示：
```
(i) snipped 206 stale tool results (~34672 tokens est.) before compaction
(i) snipped 206 stale tool results (~34769 tokens est.) before compaction
(i) snipped 206 stale tool results (~34763 tokens est.) before compaction
(i) snipped 207 stale tool results (~34905 tokens est.) before compaction
```

Image

——看起来像是"压缩前剪掉了 206 个陈旧工具结果、节省约 34K tokens"。

但与此同时，上下文窗口面板显示：

Image

上下文窗口 766.6K / 1M，状态"即将压缩"，距压缩阈值仅 33.4K
"压缩摘要"来源占比 0%
也就是说：提示声称一直在 snipped 陈旧工具结果，但上下文实际上从未被真正压缩，仍在持续增长到接近窗口上限。snipped 提示与真实状态明显矛盾。

模型: opencode-go / deepseek-v4-flash

### Steps to reproduce

1. 开启一个长时间、高轮次会话（数千次请求、上百万 tokens）
2. 持续进行工具调用（edit_file / bash 等）
3. 观察：每次工具调用后日志都打印 (i) snipped N stale tool results (~X tokens est.) before compaction

### OS / platform

Windows 11

### Relevant logs or output

```shell

```

esengine:
Fixes #7935.

## What the report showed

A long session printed the same line after nearly every tool call:

```
(i) snipped 206 stale tool results (~34672 tokens est.) before compaction
(i) snipped 206 stale tool results (~34769 tokens est.) before compaction
(i) snipped 207 stale tool results (~34905 tokens est.) before compaction
```

while the window kept filling toward 766.6K / 1M. The reporter read this as "it keeps claiming to snip but nothing is ever compacted", and that reading turns out to be right for a reason the log never showed.

## Root cause

`snipToProjection` and preflight's free prune both rebuilt their view from the canonical transcript. Canonical is never rewritten, so:

**An existing summary is discarded.** `installPruneProjection` stores the canonical-derived view as the whole projection, so a snip pass silently undoes the compaction that produced the summary. Measured on a 12-turn tool-heavy session:

| | messages | tokens |
| --- | --- | --- |
| after compaction | 15 | 8073 |
| after one snip | 37 | 25966 |

The context grows 3.2x immediately after it was supposed to shrink, and the session climbs straight back to the compaction threshold.

**The count never settles.** `shouldMaintainToolResult` excludes results that already carry the snipped marker, but canonical never carries it, so every pass re-counts the same 206 results and re-announces the same running total. The number creeping 206 -> 207 is new stale results arriving, not progress being made.

Both symptoms are the same bug: the maintenance base.

## Fix

`maintenanceBase` returns the active projection when one is valid, canonical only when there is none. Both the snip path and preflight's prune path go through it.

The summary now survives a snip, and because already-snipped results stop qualifying, the count becomes what the pass actually reclaimed — so a settled view reports nothing at all instead of repeating itself. The notice also drops "before compaction", which implied a compaction was imminent when the session could still be far from the threshold.

## Tests

New `compact_snip_projection_test.go`:

- a snip after a compaction keeps the summary and must not grow the projection
- three consecutive snips announce themselves once, not three times
- preflight's prune path keeps the summary too

Cache-impact: low - the provider-visible projection is unchanged when no maintenance runs; where it differs it now preserves a summary that was previously dropped, which shrinks the prefix rather than perturbing a stable one
Cache-guard: go test ./internal/agent/ -run 'Snip|Prune|Compact'
Documentation-impact: none - no documented behavior, option, or output contract changes; the notice text is an informational log line

### DIFF
--- a/internal/agent/compact_projection.go
+++ b/internal/agent/compact_projection.go
@@ -670,17 +670,35 @@ func (a *Agent) runCompactionSummary(ctx context.Context, fold []provider.Messag
 }
 
 // snipToProjection builds a projection that only snips stale tool results.
+// maintenanceBase is the view tool-result maintenance must rewrite: the active
+// projection when one is valid, never the canonical transcript. Rebuilding from
+// canonical would silently discard an existing summary and undo the compaction
+// that produced it.
+func (a *Agent) maintenanceBase(msgs []provider.Message, version uint64) []provider.Message {
+	st := a.compactionState
+	if !projectionValid(st, msgs, version, a.currentPromptCacheKey()) {
+		return msgs
+	}
+	if visible := modelVisibleFromProjection(st.Projection, msgs); len(visible) > 0 {
+		return visible
+	}
+	return msgs
+}
+
 func (a *Agent) snipToProjection(ctx context.Context) error {
 	_ = ctx
-	msgs, _ := a.session.snapshotMessagesVersion()
-	snipped, st := a.applyToolResultMaintenanceView(msgs, toolResultSnip)
+	msgs, version := a.session.snapshotMessagesVersion()
+	// Snipping the projection rather than canonical also makes the count honest:
+	// already-snipped results carry the marker and no longer qualify, so this
+	// reports what this pass actually reclaimed instead of the running total.
+	snipped, st := a.applyToolResultMaintenanceView(a.maintenanceBase(msgs, version), toolResultSnip)
 	if st.Results == 0 {
 		return nil
 	}
 	ratio := a.tokPerChar()
 	saved := int(float64(st.SavedChars) * ratio)
 	a.sink.Emit(event.Event{Kind: event.Notice, Level: event.LevelInfo, Text: fmt.Sprintf(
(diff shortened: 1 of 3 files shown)


esengine:
Superseded by #8109, which landed the same root cause with a broader fix.

#8109's diagnosis is identical — `pruning rebuilt from canonical history instead of the installed projection` — and it goes further by making `ContextManager` the single request-time owner with projection-only CAS views and durable receipts. main-v2 already carries `context_maintenance_issue7935_test.go` covering this.

Closing as duplicate; no action needed here.

PR #8144, opened 2026-08-10
perf(memory): bound the prefix memory index with a recency budget

chuiwenwei:
## Summary

- 现状： MEMORY.md 索引在启动时全量折叠进缓存前缀（system prompt），每轮请求都付这份 token 成本；索引按 Name 排序、陈旧事实永不淘汰（唯一移除路径是手动 forget）——记忆越积越多，前缀成本单调上涨且无上限。
- 本 PR：新增 IndexBounded ，把进前缀的索引投影改为——陈旧事实（沿用 memoryFreshness：user/feedback 90/365 天、reference 14/45 天、其余 30/180 天）折叠成一行汇总，活跃事实按最近更新降序，1200 字符软预算（至少保留 1 行），折叠行提示 "+N more facts (M stale) — search with the memory tool"，告知模型更多记忆存在、可用 memory 工具检索。
- 磁盘 MEMORY.md 保持全量不变（reindex 照旧），只裁剪进前缀的投影；Index() 全量版保留给工具/诊断。渲染是 (记忆集, now) 的纯函数，会话内前缀字节稳定（cache-first 不变式保持），standing instructions 与召回行为零改动。

## Issues

（可留空）

## Verification

- 新增 9 个测试：recency 排序 / stale 折叠（含全 stale）/ 预算裁剪（至少保留 1 行）/ 确定性 / 组合折叠 +N more (M stale) / 零时间戳回落 CreatedAt / 折叠行不匹配 managed 索引正则 / Load 接线预算断言
- go test ./internal/memory/ ./internal/boot/ 全通过（boot golden 无漂移，testdata 无记忆 fixture）
- gofmt -l 无输出；go vet 通过
- 全量 go test -count=1 ./... ：仅 2 个已知 macOS sandbox 环境失败（与本 PR 无关，见 fix(sandbox): probe sandbox-exec usability instead of PATH presence #7807）

## Documentation impact

Documentation-impact: none - 不改记忆模型、命令或配置；索引投影是内部渲染规则，MEMORY.md 的语义描述不变。

## Cache impact

Cache-impact: low - 索引投影是 (记忆集, now) 的确定性函数、启动时快照，会话内前缀字节稳定（与现状同一规则）；仅投影规则变化（一次性），standing instructions 与 recall 未动。
Cache-guard: 聚焦 guard—— go test ./internal/boot -run TestGoldenBaseline （前缀基线）+ TestIndexBoundedDeterministic （字节稳定）+ 既有 memory 测试；scripts/cache-guard.sh 为 release 级缓存命中检查（本次未跑，改动不触 provider 请求序列化）。
System-prompt-review: esengine — 模型可见的索引列表可能缩减（折叠行保留可发现性，memory 工具仍是检索路径）；需维护者确认后合并。

### DIFF
--- a/internal/memory/index_bounded.go
+++ b/internal/memory/index_bounded.go
@@ -1,0 +1,99 @@
+package memory
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	"reasonix/internal/textutil"
+)
+
+// defaultPrefixIndexMaxChars bounds the index projection rendered into the
+// cached system-prompt prefix. The on-disk MEMORY.md stays complete; only the
+// prefix-facing projection is capped, keeping every-turn prefix tokens bounded
+// as memories accumulate. Roughly 300 tokens at 4 chars/token.
+const defaultPrefixIndexMaxChars = 1200
+
+// IndexBounded renders the index projection that loads into the cached prefix:
+// stale facts (per memoryFreshness) fold into a summary line and the rest are
+// ordered by recency, capped to a soft budget of maxChars (at least one line
+// always renders). Deterministic in the memory set and now, so the prefix
+// stays byte-stable within a session.
+func (s Store) IndexBounded(now time.Time, maxChars int) string {
+	memories := s.ListAll()
+	if len(memories) == 0 {
+		return ""
+	}
+	return renderBoundedIndex(memories, now, maxChars)
+}
+
+// maxIndexDescriptionGraphemes bounds one line's description in the prefix
+// projection so a single long fresh fact cannot consume the whole budget; the
+// on-disk index keeps descriptions whole.
+const maxIndexDescriptionGraphemes = 120
+
+// renderBoundedIndex is the deterministic core of IndexBounded, separated out
+// for direct testing with a fixed clock and memory set.
+func renderBoundedIndex(memories []Memory, now time.Time, maxChars int) string {
+	// Shadowed globals (project overrides same-name global, #7995) are
+	// represented by their project winner; the shadowed global is dropped so
+	// recency can never surface it as authoritative when the winner was cut.
+	shadowed := map[string]bool{}
+	for _, o := range FindOverrides(memories) {
+		shadowed[o.Global.ID] = true
+	}
+	active := make([]Memory, 0, len(memories))
+	folded := 0     // stale + hard-expired facts
+	suppressed := 0 // shadowed globals represented by their project winner
+	for _, m := range memories {
+		freshness := memoryFreshness(m, now)
+		if freshness == FreshnessStale || freshness == FreshnessExpired {
+			folded++
+			continue
+		}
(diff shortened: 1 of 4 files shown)


maxivillus:
Independent verification of the same idea from our deployment (host E2E, 2026-08-12):

- Unbounded index: memory block was **20.7 KB / 5 403 tok (o200k_base)** in the system prompt; with a capped projection it drops to ~4.5 KB with a truncation marker — and the on-disk `MEMORY.md` sha is unchanged (prompt-only projection, exactly the invariant in this PR).
- We also clip each line's description to a stable width (120 graphemes) before the budget applies — with your stale-folding + recency budget this still leaves long fresh descriptions to consume budget, so a per-line clip composes well with it.

Supportive of the approach; the `+N more facts (M stale)` hint is a nice touch over a bare truncation marker. Happy to drop the comment if noise.

chuiwenwei:
> Independent verification of the same idea from our deployment (host E2E, 2026-08-12):
> 
> * Unbounded index: memory block was **20.7 KB / 5 403 tok (o200k_base)** in the system prompt; with a capped projection it drops to ~4.5 KB with a truncation marker — and the on-disk `MEMORY.md` sha is unchanged (prompt-only projection, exactly the invariant in this PR).
> * We also clip each line's description to a stable width (120 graphemes) before the budget applies — with your stale-folding + recency budget this still leaves long fresh descriptions to consume budget, so a per-line clip composes well with it.
> 
> Supportive of the approach; the `+N more facts (M stale)` hint is a nice touch over a bare truncation marker. Happy to drop the comment if noise.

Thanks for the independent verification — the 20.7 KB → ~4.5 KB drop with an unchanged on-disk MEMORY.md sha is exactly the invariant this PR targets, and it's great to have real deployment numbers behind it.

I've folded the per-line description clip into the bounded projection (120 graphemes + ellipsis, on-disk index untouched) — with stale-folding + the recency budget, a single long fresh description could otherwise consume the whole budget. New test TestIndexBoundedClipsLongDescription locks the behavior.

chuiwenwei:
One design note I want to flag proactively: IndexBounded takes time.Now() for the staleness fold, so the bounded projection is time-dependent — unlike the legacy Index(), which was strictly time-invariant. A fact crossing the freshness boundary changes where it renders (active list ↔ fold line), which shifts the projection.

Scope of the impact: memory.Load runs at startup and on each memory write, but the controller keeps the composed systemPrompt as a fixed string and reuses it across turns, so the projection is stable within a session — the time dependence only appears between sessions (a fresh build uses a new now), which is expected since the memory set itself changes across sessions.

I've considered bucketizing now to the session-start time so the fold is pinned for the session's lifetime and the boundary behavior is explicit. Happy to fold that in if you'd like; otherwise the current behavior (stable within a session, freshness re-evaluated per build) seems reasonable. Flagging so the trade-off is on the record.

PR #8155, opened 2026-08-10
fix(agent): keep policy reads the recorded failure, not the text prefix

esengine:
`KeepErrors` only matched tool results starting with `error:` or `blocked:` — the shape the host formats for an *execution* failure. A tool that ran fine but reported a failure (a `go test` FAIL, a compiler error, any non-zero exit) never matched, so its evidence folded into the digest and the specific assertion was gone.

`provider.Message.ToolExecution` already carries `State`/`ExitCode`/`Verification`, is already persisted, and is already stripped before provider requests. The keep policy now reads it instead of guessing from text.

Because the failure then stays identifiable without its text prefix, a protected failure can also be *shortened* — which matters, since keeping one whole doubles the projection forever. `snipFailureResult` keeps the failure-carrying lines with two lines of context and drops the passing noise. A text-only failure has no such anchor and is still carried whole, preserving the existing `TestPruneHonorsKeepErrors` contract.

The rewrite happens on three paths — the planning pass, the re-write that follows archiving, and the fold's `kept` partition. Only the first was failure-aware at first, so the other two silently overwrote it; all three now go through `maintenanceReplacement` / `keptForProjection`.

## Measured

`benchmarks/compaction -mode=fidelity` plants facts in history, folds N generations, then asks about them against both the compacted context and a full-history control.

| | evidence buried mid-result | projection tokens |
| --- | ---: | ---: |
| before | 0 / 6 runs | 13003 |
| keep whole | 6 / 6 | 26048 (+100%) |
| this PR | 18 / 18 | 13455 (+3.5%) |

The full-history control kept the fact in every run, so this was a real compaction loss rather than a hard question. The other 13 probe classes — constraints, corrections, cross-fold reversals, negative evidence, verification freshness — were already at 100% and stay there.

## Bench changes

Probes can now be revised across generations (a supersede that lands on the far side of a fold), plus four new classes: late-planted constraint, three-step reversal chain, cross-generation distractor, and the buried-evidence probe this PR is about. Adds a `-snip` arm, `snipped_*` accounting, and aligns the harness with `boot`'s default `KeepPolicy` (it previously ran with none, which no real session does). Also fixes a scoring bug: a probe revised across folds was asked before its revision chain finished, scoring the then-correct answer as lost.

Cache-impact: none - fold projection content only; the system-prompt prefix (base prompt, tools, memory) stays byte-identical
Cache-guard: go test ./internal/agent/ -run 'KeepErrors|SnipFailure|Prune|Compact'
Documentation-impact: updated - SPEC.md and SPEC.zh-CN.md said keep-policy messages stay verbatim; they now state that failures are recognised from the recorded execution and that a recorded failure keeps its failure-carrying lines

### DIFF
--- a/benchmarks/compaction/main.go
+++ b/benchmarks/compaction/main.go
@@ -42,6 +42,7 @@ func main() {
 	window := flag.Int("window", 128_000, "context window in tokens")
 	control := flag.Bool("control", true, "fidelity: also score probes against full history")
 	arm := flag.String("arm", "full", "full | incremental: re-derive each digest from canonical, or fold the previous projection")
+	snip := flag.Bool("snip", false, "run the stale-tool-result snip pass before each fold, as the 0.6 threshold does in a real session")
 	out := flag.String("out", "", "write the JSON report here")
 	flag.Parse()
 
@@ -49,14 +50,14 @@ func main() {
 		res []genResult
 		err error
 	)
-	incremental := *arm == "incremental"
+	a := arms{incremental: *arm == "incremental", snip: *snip}
 	switch {
 	case *arm != "full" && *arm != "incremental":
 		err = fmt.Errorf("unknown arm %q", *arm)
 	case *mode == "cost":
-		res, err = runCost(*gens, *window, incremental)
+		res, err = runCost(*gens, *window, a)
 	case *mode == "fidelity":
-		res, err = runFidelity(*gens, *window, *control, incremental)
+		res, err = runFidelity(*gens, *window, *control, a)
 	default:
 		err = fmt.Errorf("unknown mode %q", *mode)
 	}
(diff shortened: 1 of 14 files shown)


esengine:
Self-review found one defect, fixed in `7f20956db`.

`keptForProjection` runs on **every** fold, so `snipFailureResult` must reach a fixed point on the first pass. It did for the common shape (a few failures in a long passing log — the shortened result falls below the line threshold and a second pass returns it unchanged), but not for a log that is failures throughout: the elision marker carries no failure marker of its own, so a second pass swallowed it and restated the count. Line count stayed at 61 while the bytes changed every fold — a projection folded twice would differ from one folded once, breaking the prompt cache for no gain.

Fixed by treating the marker as the signal that content is already shortened. `TestSnipFailureIsIdempotent` pins the fixed point for both shapes.

One behaviour change worth flagging for review: `bash` records a non-zero exit as `ShellStateFailed`, so a `grep` that matches nothing (exit 1) is now covered by `KeepErrors` too. Those results are far below the 1024-byte maintenance threshold, so the cost is negligible — and "this command found nothing" is arguably a fact the digest should not absorb either.

PR #8164, opened 2026-08-10
fix(agent): pair tool results within assistant turns

Nath-Vikky:
## Summary

- namespace synthesized OpenAI-compatible tool-call IDs per response
- pair persisted calls and results only within their owning assistant turn
- rebuild todos and session evidence without borrowing colliding IDs from other turns

## Root cause

Some OpenAI-compatible gateways omit tool-call IDs. Reasonix previously synthesized `call_<index>` independently for every response, so the first tool call in multiple turns could share `call_0`.

Todo restoration and session evidence then built transcript-wide success/failure maps keyed only by that raw ID. A result from another turn could therefore:

- revive a failed or missing `todo_write`
- advance a failed `complete_step`
- invalidate a successful read or verification command

The resulting task state or evidence decision was silent and order-dependent.

## Fix

Introduce a turn-local tool exchange walker:

- pair only an assistant's tool calls with its immediately following tool-result run
- match unique non-empty IDs within that exchange, including out-of-order results
- use positional pairing for empty or duplicate IDs only when the complete batch has no contradictory ID/name anchors
- ignore orphaned rows and fail closed on missing, interrupted, or structurally contradictory ambiguous results
- never pair across user, assistant, system, or local-only boundaries

Use that exchange model for:

- canonical todo reconstruction
- persisted path evidence
- persisted command verification and command hints

For future missing-ID responses, generate a response-scoped 96-bit namespace while leaving provider-issued IDs byte-identical.

## Compatibility and safety

- no session schema, migration, or rewrite
- no tool execution or retry changes; provider request construction is unchanged
- consumer-specific failure classifiers remain unchanged except that canonical todo rebuild rejects the exact host-generated `cancelled: context cancelled before execution` sentinel
- healthy ordered histories use a zero-allocation fast path
- malformed histories become fail-closed instead of borrowing another turn's result
- a selected synthetic ID is persisted once and replayed byte-identically within that session
- old histories are interpreted structurally when the existing rebuild/evidence paths read them

## Non-goals

This PR intentionally does not expand into:

- Desktop/Serve rendering of already-persisted or provider-issued duplicate IDs
- early streaming `ChunkToolCallStart` events whose upstream ID is still absent
- the existing conversation-rewind trigger that does not currently rebuild todo state
- broader `cancelled:` or failure-classifier semantics

The source fix prevents the known missing-ID adapter path from persisting new cross-turn collisions; UI/history cleanup can remain a separate follow-up.

Related historical UI report: #3359 (not closed by this PR).

[...]

PR #8185, opened 2026-08-10
fix(goal): bound Runs instead of no-progress pauses / 用有界 Run 替代 Goal 无进展误暂停

### LINKED ISSUE #8137: [Bug]:
### Version line

### Exact version

1.20.0

### What happened?

用了这么多的版本，目前只有1.21.2和1.19.7的goal模式让我用着比较舒服，不会中止，明明没有达到暂停条件，模型自以为是的就暂停了goal

### Steps to reproduce

用了这么多的版本，目前只有1.21.2和1.19.7的goal模式让我用着比较舒服，不会中止，明明没有达到暂停条件，模型自以为是的就暂停了goal

### OS / platform

macOS

### Relevant logs or output

```shell

```

SivanCola:
## Summary

- remove the cross-Goal-turn 4/6/10 no-progress stop condition while keeping novelty and no-progress state as observability
- add a zero-configuration 16-model-round budget for each Goal executor Run, followed by one tool-free finalization response
- promote deterministic Goal-only stuck signals to resumable `goal_stuck` pauses: 3 identical host failures/blocks or 6 successful zero-evidence tool batches
- preserve the outer 10/20/40 continuation budget as a cross-Run backstop
- expose observational `requestsUsed` through persisted Goal state, CLI, ACP, and Desktop surfaces

Fixes #8137.

## Problem and root cause

Goal mode used cross-turn host novelty as both telemetry and a termination proxy. Meaningful read/analyze work could therefore be paused by the 4/6/10 no-progress thresholds, while a single Goal Run still had no default bound on its actual model sampling rounds.

The controller also could not distinguish a host-owned Goal budget from explicit `max_steps`, planner limits, or deterministic structural loops because the agent exposed only untyped internal pause errors.

### Run boundary

- every Goal executor Run gets a default 16-round host limit unless the user explicitly configured `--max-steps`
- one round is one committed model sample; provider retry attempts remain separately observable through `Usage.RequestCount`
- after the limit, the agent allows exactly one tool-free summary response and pairs, but never executes, any attempted grace-round tool calls
- the controller still accepts validated completion or blocked reports; otherwise it persists `goal_run_budget`
- `/goal resume` starts a fresh Run budget without extending the outer 10/20/40 continuation quota
- child agents inherit half the effective parent budget, with a minimum of 5; explicit child `max_steps` still wins

### Structural stuck detection

- 3 consecutive identical host failures/blocked batches produce `goal_stuck`
- 6 consecutive successful tool batches with zero new host evidence produce `goal_stuck`
- new evidence or a different successful operation resets the relevant streak
- these hard boundaries apply only to active Goal execution; ordinary chat keeps the existing advisory behavior
- both stuck paths get one tool-free finalization opportunity before the resumable pause

### State and compatibility

- `noProgressTurns` remains observational; deprecated `noProgressLimit` continues to round-trip but no longer participates in decisions
- legacy `blocked + no_progress` sidecars migrate to `running` without extending the outer turn budget
- `requestsUsed` is optional and uses zero as the old-data default
- unknown Goal sidecar fields survive read/modify/write cycles
- new stop causes are `goal_run_budget` and `goal_stuck`; older clients continue to see a safe blocked Goal

| Contract | Old data | New reader | Previous reader | Result |
| --- | --- | --- | --- | --- |
| `requestsUsed` | missing | defaults to 0 and accumulates request usage | ignores the optional field | compatible |
| `noProgressLimit` | existing 4/6/10 values | preserved but ignored for decisions | continues to read/write it | compatible |
| legacy `no_progress` pause | blocked checkpoint | resumes as running, preserving work | unchanged on downgrade | safe migration |
| unknown sidecar fields | may come from newer peers | preserved on rewrite | n/a | no silent data loss |

## Cache impact

Cache-impact: none - ordinary provider-visible prefixes remain byte-stable; only an exhausted boundary appends a one-shot finalization turn.

- no system prompt, memory prefix, tool schema, tool order, or provider serialization changes
- the default Run budget is host context only
- dynamic finalization text is appended as a one-shot user-role message only after a boundary is exhausted
- a provider-request equality regression proves the default budget leaves the normal request prefix unchanged

Cache-guard: `TestGoalDefaultBudgetDoesNotChangeNormalRequestPrefix` plus `go test ./internal/agent` verifies request equality and the owning runtime.

System-prompt-review: Reviewed by @SivanCola - `internal/agent/task.go` only resolves inherited child budgets from host context; system prompt and tool schema bytes are unchanged.

## Concurrency, performance, and security

- Goal usage/state updates remain under the Goal mutex and stale recorder events are rejected by scope ID plus continuation epoch
- no new background goroutines or unbounded queues are introduced; structural evidence history is capped at 512 compact signatures
- normal turns add only constant-time boundary checks; provider latency is added only when the configured boundary is actually exhausted and the finalization sample is requested
- after integrating main's reasoning UI, Goal request observability brings the initial raw bundle to 2204.8 KiB; only that ratchet moved from 2203 to 2205 KiB, while gzip and per-chunk ceilings remain unchanged
- no dependency, permission, sandbox, credential, or production network surface was added
- the optional live test accepts credentials only from process environment variables and logs aggregate metrics, never response text or secrets

## Consolidation and related PRs

- supersedes #8150 by @chuiwenwei
- kept/adapted from #8150: meaningful read evidence and old-sidecar compatibility handling
- extended here: remove 4/6/10 as a decision gate, add real Run budgets, typed pause ownership, deterministic stuck boundaries, request accounting, cross-client state, and live-provider coverage
- commit attribution includes `Co-authored-by` using the source PR's public commit identity
- #8193 is now merged into `main-v2`; this head semantically integrates its live-contract turn input with the Goal structural-guard reset
- a fresh merge-tree against current heads shows #8164 remains clean, while #8023 still has content conflicts in `internal/agent/run_loop.go` and `internal/control/turn_orchestrator.go`
- #8185 should merge before #8023 because it is integrated with the current base and fixes #8137; #8023 must then rebase and preserve both contracts

## Verification

Revalidated after merging current `main-v2` at `7a1d2a78e` into head `23c5faff2`:

- `go run ./tools/repolint`
- `go test ./...`
- `go test -race ./internal/agent ./internal/control ./internal/evidence ./internal/taskcontract ./internal/sessioninbox`
- `go vet ./...`
- `cd desktop && go test . -run '^TestWorkspaceChangeDetailBoundsCheckpointSnapshot$' -count=20`
- `git diff --check`

The local Desktop suite initially hit a non-owning `bad file descriptor` flake in the unchanged workspace checkpoint test; the focused test then passed 20/20 and the full Desktop suite passed on rerun.

The first Windows Desktop CI attempt hit a non-owning autosave-idle timeout in the unchanged `TestTurnDonePersistsSession`; the test and autosave implementation are unchanged from current base, and the failed job passed on targeted rerun. All current code, lint, race, security, cache, documentation, SDK, and Desktop checks are green.

Repolint is clean after carrying forward main's unchanged `App.tsx` size baseline by eight lines.

Focused coverage includes default and explicit Run limits, the one-summary contract, Goal-only 3/6 stuck thresholds, evidence resets, child budget inheritance, outer-budget resume semantics, legacy migration, unknown-field preservation, request accounting, and provider-request prefix equality.

Live acceptance used DeepSeek V4 Flash, LongCat 2.0, and GLM 5.2 Coding Plan for normal completion, deterministic stuck, and a real 16-round exhaustion followed by resume. All scenarios completed with the expected exit cause; credentials and response bodies were not persisted.

Documentation-impact: updated - documented the Goal Run budget, resumable stop causes, request accounting, legacy migration, and observational no-progress behavior across CLI, Guide, Spec, Goal enforcement guide, and changelog.


chuiwenwei:
Thanks for the extension — glad the read-evidence part landed. Closing #8150 in favor of this.

Issue #8116, opened 2026-08-10
[Bug]: 自动压缩上下文 疯了


MimoKit:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

v1.21.5

### What happened?

自动压缩上下文疯了 说一句话压缩一次

<img width="1451" height="744" alt="Image" src="https://github.com/user-attachments/assets/dfa5ada6-4782-449c-9e05-e069d9f379b4" />

### Steps to reproduce

1

### OS / platform

win10

### Relevant logs or output

```shell

```

MimoKit:
> ### Version line
> v2 — Go rewrite (1.x), main-v2 (active development)
> 
> ### Exact version
> v1.21.5
> 
> ### What happened?
> 自动压缩上下文疯了 说一句话压缩一次
> 
[...]

PR #8205, opened 2026-08-10
fix(agent): compare the context window against tokens, not characters

esengine:
## The unit bug

`estimateTextTokens` is `max(runes, bytes/4)`. `runes > bytes/4` holds for every string, so the function always returns the rune count — one "token" per character. That is fine for the internal planning budgets it was written for (tail size, fold economics), but `estimatedPromptTokens` carried it into the comparison against `contextWindow` whenever the session had no usage calibration yet, and added a second token per CJK rune on top of it.

Measured on `main-v2`:

| sample | est | real (DeepSeek) | ratio |
|---|---|---|---|
| Chinese | 10204 | ~3060 | **3.33x** |
| zh + code | 28404 | ~8520 | **3.33x** |
| English | 18804 | ~5640 | **3.33x** |

A context 25% full measures as full, so the turn force-compacts.

## Why it looks random to users

Calibration fixes the estimate for a warm session and leaves it in place for a cold one — after resume, tab switch, fork, rewind, or against a relay that reports no usage. The same session therefore compacts at 25% or at 80% depending on when the question is asked, off one threshold. That is the "it compacts for no reason" in #8132 / #8116 (and matches the 28% figure measured in #7939, whose own root cause was fixed separately).

`/context` and the maintenance snapshot read `estimateMessagesTokens` directly, so the number shown to the user was the character count too — part of what #8191 reports as display and decision disagreeing.

## The fix

- Cold estimate = wire characters × `fallbackTokPerChar` (0.25), the ratio already documented in the same file. CJK lands at 0.75 tokens per rune against a real ~0.6, so it stays conservative without being 3x.
- `/context` (`context_report.go`) and `ContextMaintenanceSnapshot` now use `estimatedPromptTokens`, so the displayed number is the one the decision used.
- Dead after the change: `cjkRunesInMessages`, `cjkRunesIn`, `estimateToolSchemaTokens` (tool schemas are already counted by `requestCalibrationShape`).
- `estimateMessagesTokens` keeps the character unit for internal planning and now says so.

## Test change worth reviewing

`TestSharedWindowFoldUsesGuardedInputBudget` fed 50K han runes against a 100K window and asserted the summarizer input got bounded. At the corrected unit those 50K runes are ~37.5K tokens and genuinely fit, so the assertion would have passed on nothing. The input is now 200K runes (~150K tokens), which does exceed the budget — the guard is still being tested, against a case that actually needs it.

## Verification

- `TestEstimatedPromptTokensStayInTokenUnitBeforeCalibration` pins the cold estimate for Chinese and English inside a token-unit band.
- `go test ./internal/agent/ ./internal/cli/ ./internal/control/ ./internal/boot/ ./internal/serve/ ./internal/tool/builtin/` green.
- `make lint` clean for this diff (`desktop/frontend/src/App.tsx` is a pre-existing violation on `main-v2`).

Independent of #8199 (threshold collapse); they fix two different reasons for the same symptom and do not touch the same functions.

Cache-impact: none - the provider-visible prefix bytes are unchanged. The estimate only decides when maintenance runs, and it now runs later, so cache resets become less frequent.
Cache-guard: TestEstimatedPromptTokensStayInTokenUnitBeforeCalibration pins the cold-start unit; the existing calibration suite (TestCalibratedOutputBudget*) pins the warm path unchanged.
Documentation-impact: none - restores the documented meaning of context_window and compact_ratio rather than changing either.

### DIFF
--- a/internal/agent/context_report.go
+++ b/internal/agent/context_report.go
@@ -47,10 +47,10 @@ func (a *Agent) ContextReport() ContextReport {
 	}
 	if a.session != nil {
 		canonical, _ := a.session.snapshotMessagesVersion()
-		rep.CanonicalTokens = estimateMessagesTokens(provider.ModelMessages(canonical))
+		rep.CanonicalTokens = a.estimatedPromptTokens(provider.ModelMessages(canonical))
 	}
 	visible := a.modelVisibleMessages()
-	rep.ProjectionTokens = estimateMessagesTokens(provider.ModelMessages(visible))
+	rep.ProjectionTokens = a.estimatedPromptTokens(provider.ModelMessages(visible))
 	rep.Projected = rep.ProjectionTokens != rep.CanonicalTokens
 
 	if a.contextWindow > 0 {
--- a/internal/agent/context_status.go
+++ b/internal/agent/context_status.go
@@ -45,7 +45,7 @@ func (a *Agent) ContextMaintenanceSnapshot() ContextMaintenanceSnapshot {
 	}
 	for _, msg := range visible {
 		if isCompactionSummary(msg) {
-			snapshot.SummaryTokens += estimateMessagesTokens([]provider.Message{msg})
+			snapshot.SummaryTokens += a.estimatedPromptTokens([]provider.Message{msg})
 		}
 	}
 	snapshot.Headroom = max(0, snapshot.HardInputCeiling-snapshot.ProjectedTokens)
--- a/internal/agent/output_budget.go
+++ b/internal/agent/output_budget.go
@@ -200,54 +200,27 @@ func (a *Agent) calibratedPromptTokens(shape requestCalibrationShape) (int, bool
 	return 0, false
 }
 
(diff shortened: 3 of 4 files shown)


esengine:
Added a second commit: the calibrated path carried the same unit bug. When a session's CJK share rises above what calibration saw, `calibratedPromptTokens` priced the excess runes at a hard 2 tokens each — a constant chosen to match the old cold estimate. So an English-calibrated session that starts discussing Chinese read 120002 tokens for 60000 han runes worth ~36000, i.e. the 3.3x this PR removes from the cold path, still present on the warm one. Excess CJK is now priced as bytes at the same ~4 chars per token (0.75/rune), which returns 45002 for that same text — identical to the cold path.

`TestCalibratedOutputBudgetKeepsCJKConservativeFloor` grew its input to 1.2M runes for the same reason the guarded-fold test grew: at the corrected unit its old 430K leaves most of a 1M window free, so no clip would happen and the assertion would pass on nothing.

PR #8221, opened 2026-08-10
fix(control): the context gauge must measure what the trigger measures

esengine:
## Two channels for one number

`ContextSnapshot` fed every frontend's gauge from the last turn's provider usage — `PromptTokens + CompletionTokens`. Compaction decides on something else entirely:

```
modelVisibleMessages() → estimatedPromptTokens() → ContextManager.Prepare() → fold / force
```

The two disagree in three ways at once:

1. **Timing** — the gauge is last turn's snapshot; the trigger is measured before the next request. A turn that just took on a large tool result reads small while the trigger has already fired.
2. **Quantity** — the gauge includes completion tokens; the trigger only ever looks at input.
3. **Rebinding** — `LastUsage()` is nil until a session's first turn, so a rebound session reads zero. Desktop rebinds on every tab switch, fork, and recovery adopt.

Measured on a recorded 13-message session:

```
old gauge (last usage) = 0        new gauge = 27076
trigger input          = 27076    fold trigger = 800000
```

Desktop hid the zero behind a telemetry-persisted `LastUsedTokens` fallback, so the bar showed a plausible but unrelated number rather than an obviously broken one. That is the reported "上下文充足 8%" sitting directly next to "正在压缩对话": the gauge was never reporting the quantity being compared against `compact_ratio`. Reported as #8191 (and it is why #8132 / #8116 read as "8% 就压缩").

## The fix

`Agent.ContextUsedTokens()` returns exactly what `ContextManager` compares against its thresholds, and `ContextSnapshot` returns that. The estimate walks every visible message while gauges redraw constantly, so it is memoised against the three things that can change it — transcript version, projection version, calibration — rather than recomputed per repaint. `ContextMaintenanceSnapshot` was not reused for this: it also hashes the whole view, which is far too heavy for a status bar.

The desktop `LastUsedTokens` fallback is removed: a rebound session now measures its own loaded view immediately, so there is nothing to paper over.

## Test change worth reviewing

`TestContextSnapshotIncludesCompletionTokens` pinned the behaviour this PR removes — its name was the contract. It is now `TestContextSnapshotMeasuresTheTriggerInput` and asserts the gauge equals `ContextMaintenanceSnapshot().ProjectedTokens` and is *not* the last turn's 6888.

## Verification

- `TestContextUsedTokensMatchesTheTriggerInput` — gauge equals the trigger's input even with a stale tiny `lastUsage` (what a fold leaves behind).
- `TestContextUsedTokensFollowsTheTranscript` — memoisation does not go stale when the view grows.
- `TestContextUsedTokensIsZeroWithoutASession` — no session still hides the gauge.
- `go test ./internal/agent/ ./internal/control/ ./internal/cli/ ./internal/serve/ ./internal/boot/` green; `go vet ./internal/...`, desktop `go vet ./...`, and `make lint` clean (no baseline widened — the new lines were pulled back under budget instead).
- The 0 vs 27076 figures above come from a real recorded session, not a fixture.

Cache-impact: none - no provider-visible bytes change; this only alters which number the read-only gauge reports.
Cache-guard: TestContextSnapshotMeasuresTheTriggerInput pins the gauge to the trigger's own input, so the two cannot drift apart again.
Documentation-impact: none - the gauge is documented as context fill, which is what it now actually shows.

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -6682,16 +6682,11 @@ func (a *App) ContextUsageForTab(tabID string) ContextInfo {
 	if ctrl == nil {
 		return info
 	}
+	// The gauge measures the loaded view, so a rebound session reports its real
+	// fill immediately and no longer needs the persisted last-turn fallback.
 	used, window := ctrl.ContextSnapshot()
 	info.Used = used
 	info.Window = window
-	// Session rebind (project-tree switch) rebuilds the controller: the fresh
-	// executor has no per-turn usage yet, so ContextSnapshot reports used=0.
-	// Fall back to the telemetry-persisted last-used value so the status bar
-	// shows the fill percentage from the last turn instead of 0%.
-	if used == 0 && snap.Usage.LastUsedTokens > 0 {
-		info.Used = snap.Usage.LastUsedTokens
-	}
 	info.CompactRatio = ctrl.CompactRatio()
 	info.Maintenance = contextMaintenanceInfo(ctrl.ContextMaintenanceSnapshot())
 	return info
--- a/desktop/tabs.go
+++ b/desktop/tabs.go
@@ -8148,20 +8148,19 @@ func (a *App) ContextPanel(tabID string) ContextPanelInfo {
 		if sp := ctrl.SessionPath(); sp != "" {
 			tab.syncTelemetryToSession(sp)
 		}
-		used, window := ctrl.ContextSnapshot()
-		info.UsedTokens = used
+		_, window := ctrl.ContextSnapshot()
 		info.WindowTokens = window
-		// Session rebind rebuilds the controller: the fresh executor has no
-		// per-turn usage yet, so ContextSnapshot reports used=0. Fall back to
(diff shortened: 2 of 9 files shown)


esengine:
Pushed a second commit after CI caught two desktop tests.

`ContextPanel` also read `ContextSnapshot`, and it breaks one turn down into prompt/completion/reasoning/cache segments — with the gauge switched to the live view, the segments stopped summing to the total it displayed. The panel now takes its total from the same turn its segments come from; the status-bar gauge keeps the live-view measurement. Two different questions, two different numbers, each internally consistent.

The two tests that failed pinned the old rebind fallback by name. `TestContextFallbackUsesPersistedExecutorUsageAfterRebind` is now `TestContextGaugeMeasuresLiveViewAfterRebind` and asserts the gauge equals the trigger's input while the panel breakdown still comes from telemetry; the multi-attempt test keeps its `Last*` and panel assertions unchanged and only its gauge line moved.

`go test ./...` in the desktop module is green locally (70s, full suite).

PR #8264, opened 2026-08-11
fix(desktop): stabilize native transcript scrolling / 修复聊天区原生滚动跳页与卡顿

SivanCola:
## Summary

Long variable-height transcripts could feel sticky, reverse direction, jump by a virtual page, or stop following the final answer when native scrolling overlapped TanStack measurement compensation, streaming tail-follow, or resize repins. This PR makes native input a first-class scroll owner, keeps compensating writers silent until the gesture has actually settled, and replays a still-valid tail repin once after idle instead of dropping it.

It also preserves the original large-table and nested-scroll improvements while integrating the latest `main-v2` transcript selection and virtualization work.

## User-visible changes

- macOS trackpad inertia is no longer fought by virtualizer, stream, row-size, footer, or container scroll writes.
- Mouse wheels, native scrollbar drags, touch scrolling, keyboard scrolling, pen/middle-button auto-scroll, and nested table/code scrolling use the same ownership rules across macOS, Windows, and Linux webviews.
- A downward wheel over a table or code block at the physical bottom no longer accidentally disables streaming tail-follow.
- If the final token, async Markdown render, image, or measured row growth lands during a harmless downward gesture, the viewport stays still during the gesture and catches up exactly once after idle.
- A deliberate upward move remains detached from streaming tail-follow even when it stops inside the former 80px near-bottom band; queued passive repins are discarded.
- Tail-follow resumes only after the reader reaches the physical bottom or uses an explicit bottom/reset action; the same rule applies to Creation's custom scrollbar.
- Stale `scrollend`, RAF, timer, resize, row-measurement, and deferred-repin work cannot affect a newer tab generation.
- Wheel `deltaMode` values are normalized for pixel, line, and page devices before nested-scroll handoff.
- Tables and code blocks that only need horizontal overflow hand vertical scrolling back to the transcript.
- Large tables default to a bounded in-flow preview with **Expand all / Collapse** and virtualize their expanded body.
- Creation's custom scrollbar freezes its drag geometry at pointerdown, so virtual row mounts cannot change sensitivity under the pointer.
- Selection, rewind, jump, history prepend, stream follow, and custom-scrollbar semantics remain compatible.

## Implementation

- Extends `transcriptScrollSession` with explicit wheel, touch, keyboard, native-scroll, nested-scroll, and middle-button ownership.
- Propagates normalized nested-wheel direction into `useScrollManager`; only upward or already-detached intent releases auto-scroll.
- Adds a generation-aware deferred tail repin for `stream`, `container-resize`, `footer-resize`, and `row-size` owners. Rejected passive writes replay once at gesture idle using the current `scrollHeight` only if tail-follow is still valid.
- Clears deferred work on upward/manual intent, selection, explicit jump/rewind/custom-scrollbar actions, and tab/reveal generation changes.
- Distinguishes controller-owned scroll events from unowned native scroll events in `useScrollManager`.
- Latches explicit user detachment independently from geometric bottom proximity, preventing stream and resize repins after gesture settlement.
- Uses `scrollend` as a guarded hint with a fallback quiet window; stale events are rejected.
- Adds a `row-size` owner, disables anchor compensation while pinned, and frame-batches measured row growth into the tail-follow path.
- Adds a passive live-to-settled fallback for Markdown layout growth without overriding a reader who scrolled away.
- Removes unconditional full virtualizer measurement after every gesture. Measurements now run only for real width or typography invalidations, capture a stable viewport anchor, and reconcile once after layout.
- Extracts Creation scrollbar and transcript interaction wiring from `Transcript.tsx` into focused hooks.
- Adds deterministic state-machine tests and a production-build Chromium gate for variable-height scrolling, delayed jumps, short near-bottom upward gestures, harmless bottom-down growth, and Windows middle-button auto-scroll.

## Consolidation and repository contributions

- **Kept and adapted:** #8291 by @Lxiny-zy — stream-end fallback, row-size repin ownership, and the pinned row-size compensation rule. These mechanisms were integrated into #8264's native gesture ownership and generation-aware deferred replay model. Commit `2cae44a0d` includes effective Git co-author attribution.
- **Reviewed but not adopted:** #8292 by @Lxiny-zy — webview stream coalescing and streaming height-cache hygiene remain independent because they address a separate IPC/rendering scope and do not conflict with this frontend scroll-ownership fix.

## Compatibility and security

- Frontend-only DOM, scrolling, and test changes.
- No new dependency, network access, persistence, Wails API, backend protocol, provider serialization, prompt, MCP, or credential behavior.
- No transcript text, selected text, clipboard content, or deferred-repin payload is logged.
- Cache-impact: none - provider-visible prompt and tool payloads are unchanged.
- Cache-guard: existing cache-impact path guard confirms no cache-sensitive files are touched.
- System-prompt-review: N/A - no system-prompt or provider-visible prefix changes.

## Verification

- [x] `pnpm --dir desktop/frontend test:transcript`
- [x] `pnpm --dir desktop/frontend test:transcript-browser`
- [x] `pnpm --dir desktop/frontend test:typecheck`
- [x] `pnpm --dir desktop/frontend test:motion`
- [x] `pnpm --dir desktop/frontend build`
- [x] `go run ./tools/repolint`
- [x] `git diff --check`
- [ ] `pnpm --dir desktop/frontend test:bench` — scroll-path gates pass; two repository-level baseline gates remain red as described below.

Real Chromium assertions include:

- no virtualizer, stream, row-size, footer, or container writes during active wheel inertia;
- no delayed `scrollTop` jump after `scrollend`;
- harmless wheel-down plus tail-row growth remains `tail-follow`, then replays once to bottom distance 0 after idle;
- a 24px upward wheel remains in manual mode after the quiet window, and a wheel back to the physical bottom restores tail-follow;
- middle-button auto-scroll retains ownership without wheel samples;
- 20+ turn logical selection remains bounded and copies correctly;
- selection cleanup retains less than 2 MiB (latest observed 0.74 MiB).

Production bundle results:

- initial JavaScript gzip: 395.2 KiB / 400.0 KiB;
- largest initial JavaScript chunk gzip: 222.9 KiB / 280.0 KiB;
- initial raw JavaScript and CSS: 2244.7 KiB / 2246.0 KiB.

Repository-wide benchmark note: the latest branch run reports 303.8ms switch P95 (3.8ms above the 300ms gate) and 46.7 MiB retained-heap growth. The latest same-machine `main-v2` comparison reported 305ms and 46.4 MiB, respectively, so these existing benchmark failures are not attributed to this scroll change. Scroll-path evidence remains bounded: 24ms input P95, 0% DOM growth, and no observed long tasks. Native macOS Wails/WebKit and Windows Wails/WebView2 manual shell verification is still recommended before merge.

## Documentation impact

Documentation-impact: none - this repairs existing transcript scrolling behavior without changing the documented user workflow or configuration.


DUCK-Stop:
确实经常遇到这个问题，修复的很快，赞！

PR #8291, opened 2026-08-11
fix(desktop): keep transcript auto-scroll pinned through stream end

Lxiny-zy:
## Problem

Fixes the transcript auto-scroll failure reported for long answers. Answer text ≥8000 chars (STREAMING_TAIL_THRESHOLD) renders its final blocks through an async markdown worker **after** the stream settles (`MarkdownHistory.tsx` → worker parse). Nothing repins after that layout growth:

- The streaming auto-scroll effect (`Transcript.tsx`) only runs on live token changes — after `turn_done` clears the live snapshot it never fires again.
- The virtualizer's anchor compensation (`anchorTo: "end"` + `shouldAdjustScrollPositionOnItemSizeChange`) keeps the *anchor row* in place instead of following the tail.
- The subsequent bottom-state re-evaluation (`useScrollManager.updateBottomState`) sees distance ≥ 80px, flips the transcript to `manual` mode, and **permanently disables auto-scroll** — the viewport is stuck mid-transcript with only the jump-to-bottom button as an escape.

## Fix

Two coordinated changes:

1. **Pinned rows never go through anchor compensation.** New pure rule `shouldAdjustScrollOnItemSizeChange(pinned, mode)`: while pinned (tail-follow) the virtualizer does not compensate on row-size changes. Every measured row-size change while pinned is instead folded into the existing frame-batched repin — the `measureElement` wrapper calls `scheduleRepinIfWasPinned` with a new `"row-size"` scroll owner — so tail growth, async worker swaps and image loads all keep the viewport at the bottom.

2. **Stream-end fallback repin.** When the live snapshot settles (`turn_done`) and the transcript is still pinned, `scrollToBottomAfterLayout(3)` waits a few frames for synchronous markdown layout and snaps to the bottom. A user who scrolled away during the stream (`shouldRunStreamEndRepin` gates on pinned) is never yanked back.

## Tests

- 20 assertions on the new pure arbitration functions: `shouldAdjustScrollOnItemSizeChange` (all mode×pinned quadrants incl. selection modes), `shouldRunStreamEndRepin` (repin only on live→absent while pinned), and `canTranscriptScrollOwnerWrite("row-size", …)` permissions.
- Existing `scroll-manager.test.tsx` suite still passes (17 assertions).
- `tsc --noEmit` and eslint clean.

Cache-impact: none - frontend-only rendering/scroll behavior
Cache-guard: none
System-prompt-review: none - no prompt, config schema, memory, output style, or skill behavior changes
Documentation-impact: none - fixes existing transcript scroll behavior; no CLI, docs or config surface changed

### DIFF
--- a/desktop/frontend/src/__tests__/transcript-scroll-controller.test.ts
+++ b/desktop/frontend/src/__tests__/transcript-scroll-controller.test.ts
@@ -1,0 +1,61 @@
+// Run: tsx src/__tests__/transcript-scroll-controller.test.ts
+//
+// Pure-function tests for the transcript scroll arbitration rules
+// (#7420-family: pinned tail-follow must not be derailed by the virtualizer's
+// anchor compensation, and the stream-end fallback repin must never yank a
+// user who scrolled away).
+
+import { canTranscriptScrollOwnerWrite, shouldAdjustScrollOnItemSizeChange, shouldRunStreamEndRepin } from "../lib/transcriptScrollController";
+
+let passed = 0;
+let failed = 0;
+
+function ok(value: boolean, label: string) {
+  if (value) {
+    process.stdout.write(`  PASS  ${label}\n`);
+    passed += 1;
+  } else {
+    process.stdout.write(`  FAIL  ${label}\n`);
+    failed += 1;
+  }
+}
+
+// ── shouldAdjustScrollOnItemSizeChange ──────────────────────────────────────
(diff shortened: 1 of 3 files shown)


Lxiny-zy:
Superseded by #8264 (stabilize native transcript scrolling), which adapts this PR's row-size ownership, pinned measurement rule and stream-end fallback into the unified gesture controller. Thanks for the review and the Co-authored-by credit in commit 2cae44a0.

PR #8292, opened 2026-08-11
fix(desktop): reduce streaming render jitter (webview coalescing + height-cache hygiene)

Lxiny-zy:
## Problem

Streaming rendering jitter on fast model streams. Two compounding sources:

1. **Per-chunk IPC floods the webview channel.** The agent emits one event per model chunk (`internal/agent/agent.go` emits `Text`/`Reasoning` per chunk); `tabEventSink.Emit` forwards every event to Wails `runtime.EventsEmit` one by one. When the single webview event channel backs up, `EventsEmit` blocks, backlog latency spikes, and the frontend's per-frame batch drains a stale backlog in one jump instead of a smooth stream.
2. **Mid-stream heights poison the measurement cache.** Streaming rows are measured every frame (tail growth, live code highlighting); those transient heights were written into `transcriptHeightCache`, so a later re-mount of the same row key (history replay, tab switch) started from a mid-stream height and settled with a second viewport jump.

## Fix

1. **`webviewStreamCoalescer`** (new, `desktop/tab_stream_coalescer.go`): merges consecutive same-kind text/reasoning deltas for the same tab+submission into one wire event per 16ms window (one 60Hz frame — the frontend already coalesces a second time per frame, so the merged rate never exceeds the repaint rate). Event order and the reasoning→text boundary are preserved (a kind or submission switch flushes immediately). The merge sits **only on the webview exit**: bot sinks, project-tree metadata, telemetry and the display buffer keep receiving every event unchanged.
2. **Streaming rows never write the height cache.** `createTranscriptMeasureElement` accepts a `skipCacheWriteWhen` predicate; `useTranscriptRowMeasurements` matches answer rows whose `item.streaming` is still true. Settled rows keep writing their final measured height.

## Tests

Go (`tab_stream_coalescer_test.go`, all pass; full `go test ./desktop/...` green):
- window merge of consecutive text deltas
- reasoning→text boundary flushes separately
- different submissions never merge
- non-stream events (e.g. `turn_done`) flush the pending delta immediately
- window timer flushes a lone delta without further pushes
- sink-level: bot sinks still receive every event unchanged while the webview gets the merged stream

Frontend (`transcript-height-cache.test.ts`, 5 assertions; `tsc --noEmit` and eslint clean):
- predicate matching for streaming/settled/user/missing rows
- integration check that only settled rows reach the cache through the measure element

Cache-impact: none - frontend rendering path + Wails event bridge only; no prompt or tool schema bytes change
Cache-guard: none
System-prompt-review: none - no prompt, config schema, memory, output style, or skill behavior changes
Documentation-impact: none - internal rendering/IPC performance fix; no CLI, docs or config surface changed

### DIFF
--- a/desktop/frontend/src/__tests__/transcript-height-cache.test.ts
+++ b/desktop/frontend/src/__tests__/transcript-height-cache.test.ts
@@ -1,0 +1,85 @@
+// Run: tsx src/__tests__/transcript-height-cache.test.ts
+//
+// Streaming rows must never be written into the height cache: their height is
+// a moving target and cached mid-stream values would poison later estimates
+// for the same row key (re-mounts would replay a transient height).
+
+import { createTranscriptMeasureElement } from "../lib/transcriptHeightCache";
+import type { TranscriptRow } from "../lib/transcriptRows";
+
+let passed = 0;
+let failed = 0;
+
+function ok(value: boolean, label: string) {
+  if (value) {
+    process.stdout.write(`  PASS  ${label}\n`);
+    passed += 1;
+  } else {
+    process.stdout.write(`  FAIL  ${label}\n`);
+    failed += 1;
+  }
+}
+
+function isStreamingRow(row: TranscriptRow | undefined): boolean {
+  if (!row) return false;
+  if (row.kind === "answer") return row.item.streaming;
+  if (row.kind === "reasoning") return row.item.streaming && !row.item.reasoningComplete;
+  return false;
+}
+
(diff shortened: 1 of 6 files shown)


Lxiny-zy:
Superseded alongside #8291 by #8264's unified transcript scroll/rendering rework; closing to keep the PR list clean. The Go-side webview stream coalescer (per-chunk IPC merging) remains available here if it is ever wanted.

PR #8321, opened 2026-08-11
fix(grep): detect encoding from a character-safe peek

### LINKED ISSUE #8299: [Bug]: 内置grep不能正确处理部分中西混合文件编码问题
### Version line

### Exact version

1.23.0

### What happened?

grep 搜索部分中西文混合文件时无法正确处理编码问题。以下是Reasonix自己的报告：

> 诊断完成，证据链完整。以下是内置 grep 的完整情况说明。
> 
> 结论
> 内置 grep 工具（Go 重写版，docs/MIGRATING.md 说明它「匹配前会解码文件」）对无 BOM 的纯中文 UTF-8 文件做了错误的编码检测：把 UTF-8 字节按 GBK/GB18030 解码。结果：
> 
> 中文 pattern → 匹配不到（pattern 是 UTF-8，与 GBK 解码后的乱码文本比对自然失败）；
> ASCII pattern（如 1931、thesis-citation）→ 能匹配，但输出行中文全部乱码；
> 文件本身没有任何问题——read_file、外部命令行 rg 都正常。
> 证据链
> 乱码确认：内置 grep 搜 `待排印正文/02-02-第2章第2节.md` 的 1931，输出 `## 绗�簩鑺傘€€鈥滅粺涓€鎴樼嚎鈥...`。我用 PowerShell 模拟「取原文 UTF-8 字节 → 按 GBK 解码」，得到完全一致的乱码：`## 绗簩鑺傘€€鈥滅粺涓€鎴樼嚎鈥...`。反向还原也成立。
> 与 ASCII 占比强相关（全仓统计）：
> 误判失败：待排印正文 全部正文文件，ASCII 占比 0.8%–15.4%（中文为主）。
> 正常：`论文.md`（19.8%，且开头是 YAML front matter）、`研究工作/.../1994俄文档案集...md`（18.7%）、`项目治理/状态/当前占用.md`（45.9%）、`00-03-英文摘要占位稿.md`（100%）。
> 带 BOM 的 UTF-8 文件正常：`冷审计/档案/材料索引/文件索引_中文.csv`（UTF-8-sig）搜「统一战线」完美命中——说明 BOM 能帮助检测器正确识别编码。
> 阈值大致在 ASCII 占比 16% 上下（15.4% 的 03-02 未逐一验证，但 18.7% 以上正常）。

### Steps to reproduce

见上

### OS / platform

win10 19045.7548

### Relevant logs or output

```shell

```

tancheng33:
## Summary

- `grep` classifies a file from its first 8 KiB peek, and in CJK text that window usually stops mid-character: most Chinese characters are three bytes wide, so roughly two of three cut points land inside one. `Detect` reads "not valid UTF-8" as evidence of a legacy charset, and the GB18030 decoder accepts nearly any byte string, so a plain UTF-8 file was decoded as GB18030. Chinese patterns then never matched, and ASCII patterns matched but printed their lines as mojibake.
- Adds `encoding.DetectPrefix(window, truncated)`, which trims a truncated window to a character boundary before classifying it, and routes grep's peek and `read_file`'s detection sample through it.
- `read_file` already trimmed its sample to the last newline for this reason. `DetectPrefix` subsumes that and also covers a sample with no newline in it, such as a single-line CJK export or a minified bundle.

This also accounts for the ASCII-ratio correlation in the report: the more ASCII a file contains, the likelier the 8192-byte boundary lands on a character boundary and the file happens to be classified correctly. It is a probability, not a threshold, which is why the reporter's 15.4% file was ambiguous while 18.7% and above looked fine.

## Issues

Fixes #8299

## Verification

- New tests, each confirmed to fail with the trim disabled and to pass with it:
  - `TestE2EGrepUTF8CJKLargerThanPeek`: grep matches a Chinese pattern in a UTF-8 file larger than the peek, and an ASCII match keeps its surrounding Chinese readable.
  - `TestE2EGrepGBKLargerThanPeek`: a real GB18030 file over the peek still decodes, guarding against over-trimming.
  - `TestE2EReadFileSingleLineCJKLargerThanSample`: `read_file` handles a single-line CJK file larger than its 256 KiB detection sample.
  - `TestDetectPrefixTrimsTruncatedCJKWindow`, `TestDetectPrefixKeepsGB18030Detectable`, `TestTrimPartialRuneDropsOnlyTruncatedTail`.
- The fixtures assert that the detection boundary actually splits a character before running the tool, so they cannot pass by luck.
- `gofmt -l .` clean, `go vet ./...` clean, `make lint` clean (golangci-lint v2.12.2 reports 0 issues; repolint clean against the baseline, which is unchanged).
- `go test ./internal/tool/... ./internal/boot/ ./internal/fileutil/... ./internal/checkpoint/ ./internal/config/` passes.

## Documentation impact

Documentation-impact: none - detection is internal. docs/MIGRATING.md:200 already promises that grep "decodes before matching", which is the behavior this restores for UTF-8 CJK files.

## Cache impact

Cache-impact: none - no tool schema, description, registration, or other provider-visible prefix byte changes. The only change is which local byte window is used for encoding detection.
Cache-guard: the provider-visible prefix is untouched, so existing tool-schema and boot guards cover it. Ran `go test ./internal/tool/... ./internal/boot/`.
System-prompt-review: N/A

### DIFF
--- a/internal/fileutil/encoding/encoding.go
+++ b/internal/fileutil/encoding/encoding.go
@@ -70,6 +70,49 @@ func Detect(data []byte) (Kind, []byte) {
 	return LossyUTF8, data
 }
 
+// DetectPrefix classifies a stream from a leading window, where truncated says
+// more bytes follow. A truncated window is trimmed to a character boundary
+// first: Detect reads "not valid UTF-8" as a legacy charset and GB18030 accepts
+// nearly any bytes, so a window that merely stops mid-character would decode a
+// UTF-8 file into mojibake — a coin flip for 3-byte-per-character CJK text.
+func DetectPrefix(window []byte, truncated bool) Kind {
+	if truncated {
+		window = trimPartialRune(window)
+	}
+	k, _ := Detect(window)
+	return k
+}
+
+// trimPartialRune drops an incomplete trailing UTF-8 sequence. Bytes that
+// cannot begin a sequence stay, so genuinely non-UTF-8 input keeps failing
+// Detect's utf8.Valid check.
+func trimPartialRune(b []byte) []byte {
+	for i := len(b) - 1; i >= 0 && i >= len(b)-(utf8.UTFMax-1); i-- {
+		c := b[i]
+		if c&0xC0 == 0x80 {
+			continue // continuation byte — keep walking back to the lead byte
+		}
+		var width int
+		switch {
+		case c&0x80 == 0x00:
+			return b // ASCII: the window ends on a character boundary
+		case c&0xE0 == 0xC0:
+			width = 2
+		case c&0xF0 == 0xE0:
+			width = 3
+		case c&0xF8 == 0xF0:
+			width = 4
+		default:
+			return b // not a lead byte, so not a truncated sequence
+		}
+		if len(b)-i < width {
+			return b[:i]
+		}
+		return b
+	}
+	return b
+}
+
 // DetectQuick checks only for BOM prefixes in the first few bytes. This is
 // the fast path for peek-based binary rejection: BOM-prefixed files (UTF-16,
 // UTF-8 BOM) skip the NUL-byte check since 0x00 is normal in UTF-16. Returns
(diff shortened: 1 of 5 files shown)


VirgilMing:
I independently arrived at #8322 while validating #8299. I cross-ran this PR's head with the two inverse regression cases from #8322—an ASCII-only header followed by a long GB18030 line, for both `grep` and `read_file`—and both passed on this head without code changes. Since the implementations overlap, I do not plan to stack the full PR; if the extra coverage would be useful, I can send those two tests as a small PR against this head branch.

PR #8364, opened 2026-08-11
feat(tui): add /version slash command to show the installed version

### LINKED ISSUE #8289: [Feature]: Add /version command capability to tui
### Version line

I know one can determine the version of reasonix installed via `reasonix --version` from the command line.  It would also be helpful to be able to determine the version from within the tui user interface as well.  I looked whether the task bar provides this or whether there is a backslash command in the interface that can be used but was unable to find anything.

### Proposed solution

Add a /version backslash command to the tui which simply displays the version information of the reasonix interface.

onionviolet:
## Summary

Adds a `/version` slash command to the TUI that shows the same version string the CLI version flag prints (`reasonix --version`). The version value already reaches the interactive session, so the command reuses it instead of adding a second source of truth. Includes the i18n strings in all four locales and a row in the docs slash-command list.

Implements #8289.

## Issues

Fixes #8289

## Verification

- `go test ./...` passes
- `gofmt -l .` clean
- `go vet ./...` passes
- New test: `TestVersionCommandShowsInstalledVersion` (internal/cli) asserts `/version` renders `reasonix <version>`.

## Documentation impact

Documentation-impact: updated - `/version` added to the slash-commands list in `docs/GUIDE.md`.

## Cache impact

Cache-impact: none - TUI-only command; no system prompt, memory prefix, skill index, or tool surface change.
Cache-guard: TestVersionCommandShowsInstalledVersion (internal/cli).
System-prompt-review: none - no system-prompt-sensitive path touched.

### DIFF
--- a/docs/GUIDE.md
+++ b/docs/GUIDE.md
@@ -974,7 +974,7 @@ convenient.
 ## Slash commands
 
 In an interactive `reasonix` session, built-in commands (`/compact`, `/context`, `/new`, `/clear`, `/rewind`,
-`/tree`, `/branch`, `/switch`, `/todo`, `/model`, `/work-mode`, `/mcp`, `/skills`, `/hooks`,
+`/tree`, `/branch`, `/switch`, `/todo`, `/model`, `/version`, `/work-mode`, `/mcp`, `/skills`, `/hooks`,
 `/memory`, `/goal`, `/output-style`, `/sandbox`, `/language`,
 `/reasoning-language`, `/help`) run
 locally — `/help` lists them all. Built-in **skills** such as `/init`,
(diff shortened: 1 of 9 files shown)


onionviolet:
Rebased onto current `main-v2` (`e44f1fed`). The conflicts were in the slash-command registry and the three locale files, where `/work-mode` became `/preset` and `CmdWorkMode` was reworded upstream. I kept the upstream line in each case and added `CmdVersion` alongside it, so this PR does not change work-mode behavior or wording.

`go build ./...` is clean and `go test ./internal/cli/... ./internal/i18n/...` passes. CI is green. Ready for review.

PR #8407, opened 2026-08-11
fix(evidence): score a read by the question it asked, not just its path

esengine:
Split out of #8404 so the smallest fix for the most frequent complaint can land
and be reverted on its own.

The no-progress guard scored read novelty on the path alone. Measured on the
real tool loop:

| turn shape (every round returned information the turn did not have) | before |
| --- | --- |
| grep one package, ten different patterns | ordered to answer at round 7 |
| page through one long file, ten windows | ordered to answer at round 7 |

Nine of those ten rounds scored as repeats because the path had been seen. The
guard then told the model, untruthfully, that it had "produced no new evidence".

A path already read can still answer a question never asked. Both scorers now
key a read on the exact call behind the receipt as well as its paths: a new grep
pattern or the next window counts once, and only an identical call is a repeat.
It also makes grep consistent with itself — a pathless grep was already scored
by its arguments, so searching the whole tree scored better than searching one
directory.

Deliberately not in this PR: the exploration run limit still zeroes novelty
after six look-only rounds, so a read-only turn still reaches the stop tier by
arithmetic around round 12. That cliff, and the fixed 2/4/6 ladder it feeds, are
replaced in #8404.

Tests cover both scorers directly and both turn shapes through the real
`executeBatch` loop, plus the discriminator that must survive: re-reading one
path still escalates to the stop tier.

Cache-impact: none - evidence scoring only; no tool schema, prompt or provider request changes, so the system-prompt prefix stays byte-identical.
Cache-guard: `go test ./internal/boot/` (provider_request golden) and `internal/agent/cachehit_e2e_test.go`, both unchanged and green.
Documentation-impact: none - the docs describe the guard's thresholds and pause behaviour, which this change does not alter; it only stops miscounting a distinct question as a repeat.

### DIFF
--- a/internal/agent/read_novelty_guard_test.go
+++ b/internal/agent/read_novelty_guard_test.go
@@ -1,0 +1,103 @@
+package agent
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+	"testing"
+
+	"reasonix/internal/event"
+	"reasonix/internal/provider"
+	"reasonix/internal/tool"
+)
+
+// grepNoveltyProbe mirrors the real grep tool's receipt shape: read-only,
+// path-scoped, and carrying its question in an argument rather than the path.
+type grepNoveltyProbe struct{}
+
+func (grepNoveltyProbe) Name() string        { return "grep" }
+func (grepNoveltyProbe) Description() string { return "search" }
+func (grepNoveltyProbe) Schema() json.RawMessage {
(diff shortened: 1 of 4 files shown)


SivanCola:
Superseded by #8417, merged into `main-v2` as 1ac7817ff.

#8417 kept and adapted the question-keyed read novelty from this PR in both `OutcomeTracker` and `ProgressTracker`, together with the focused novelty and end-to-end guard tests. The adopted work is preserved with commit-level co-author attribution.

Thank you for isolating the false-repeat mechanism and its regression coverage. Closing this PR because the implementation is now on `main-v2`.

PR #8429, opened 2026-08-11
fix(desktop): stabilize transcript scrolling during async row growth / 修复异步行高变化导致的对话跳动

SivanCola:
## Summary

- Freeze mounted transcript row measurements while a wheel, touch, or nested-scroll gesture owns the viewport.
- Batch observed row heights until gesture idle, then remeasure once, restore the visible row anchor, and settle pinned tails across the virtualizer's layout frames.
- Distinguish explicit user intent from browser layout clamping and controller-owned follow-up scroll events, preserving both manual reading and tail-follow behavior.
- Keep row-key generation, gesture-stable measurement, deferred cleanup, and tail settling in the dedicated transcript measurement hook so the rendering component stays within its repository lint baseline.
- Extend unit and real-Chromium coverage for async growth above the viewport, physical-bottom re-engagement, repeated owned scroll events, and deferred tail growth.

## Root cause

The existing gesture lock prevented compensating `scrollTop` writes, but TanStack Virtual could still publish new row sizes and transforms while asynchronous Markdown content changed height. The viewport therefore moved even though the write gate remained silent. During tail growth, follow-up layout scroll events could also outlive the first programmatic marker and be misclassified as native user scrolling.

## User impact

Long transcripts no longer jump upward by a large distance when the user scrolls up from the bottom while Markdown, code blocks, or other variable-height content finishes mounting. Manual upward intent remains latched, while deliberate downward movement to the physical bottom restores tail-follow.

## Verification

- `pnpm test:transcript`
- `pnpm test:transcript-browser`
  - visible anchor remains `6px -> 6px` during a 1200px above-viewport row growth
  - no passive compensating writes occur during active wheel ownership
  - deferred tail growth settles at `0px` from the physical bottom
- `pnpm build`
  - initial JavaScript gzip: 402.5 KiB / 402.5 KiB
  - initial raw JavaScript and CSS: 2270.1 KiB / 2270.5 KiB
- `go run ./tools/repolint`

## Risk and compatibility

- Frontend-only change; no persisted data, Wails API, provider, prompt, cache, auth, dependency, network, or filesystem contract changes.
- Deferred measurements are scoped to the mounted transcript/tab and cleared on tab changes and unmount.
- Existing selection, jump, custom-scrollbar, middle-button, keyboard, touch, nested-scroll, streaming, and virtualization paths remain covered by the transcript suites.

Documentation-impact: none - This fixes internal transcript scrolling behavior without changing user-facing setup or documented workflows.

### DIFF
--- a/desktop/frontend/bench/transcript-scroll-stability.mjs
+++ b/desktop/frontend/bench/transcript-scroll-stability.mjs
@@ -55,8 +55,18 @@ try {
     if (!transcript) return;
     transcript.scrollTop = Math.max(0, transcript.scrollHeight - transcript.clientHeight * 2);
     window.__scrollWrites = [];
+    window.__scrollEvents = [];
     window.__scrollGestureTrace = [];
     window.__REASONIX_TRANSCRIPT_SCROLL_WRITE__ = (owner, top) => window.__scrollWrites.push({ owner, top });
+    transcript.addEventListener("scroll", () => {
+      window.__scrollEvents.push({
+        top: transcript.scrollTop,
+        height: transcript.scrollHeight,
+        mode: transcript.dataset.scrollMode,
+        gesture: transcript.dataset.scrollGesture ?? "idle",
+      });
+      window.__scrollEvents = window.__scrollEvents.slice(-20);
+    });
     new MutationObserver(() => {
       window.__scrollGestureTrace.push(transcript.dataset.scrollGesture ?? "idle");
     }).observe(transcript, { attributes: true, attributeFilter: ["data-scroll-gesture"] });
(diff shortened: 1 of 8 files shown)


SivanCola:
Superseded by #8492, which takes the ownership-model fix further: instead of freezing measurements during gestures, React Virtuoso becomes the exclusive owner of virtual measurement and anchor compensation. The competing Reasonix layers that caused double-compensation are removed entirely.

#8492 preserves the gesture-stability and tail-pinning contracts this PR introduced, and adds:
- frozen selection anchors for cross-page drag resilience
- deterministic unit and real-Chromium regression gates
- net −1383 lines by removing the scroll manager, height cache, measurement invalidation, and scroll session

Closing in favor of the maintained-library approach. Thank you for diagnosing the root cause and prototyping the measurement-freeze mitigation — that insight directly informed the final ownership model.

PR #8542, opened 2026-08-12
fix: refresh recovery lineage after repair completes

SivanCola:
## Problem

Users report seeing many duplicate/copy sessions in the project list even after upgrading to 1.24.3 with the catalog v2 migration. Video evidence shows the same session appearing dozens of times.

## Root Cause

PR #8457's recovery lineage classification logic is correct, but has a timing issue:

1. **Initial scan**:  imports all  files with 
2. **First lineage pass**:  runs, but  cannot determine the longest branch because turn counts are unknown
   - Returns  and all recovery branches stay as 
3. **Repair completes**: Background repair worker decodes JSONL and updates 
4. **No refresh**:  only updates individual session fields, never triggers lineage refresh again

Result: All recovery sessions remain permanently marked as 'diverged' → user sees all copies.

## Solution

Add periodic lineage refresh in :

1. Query directories with recently repaired recovery sessions:
   -  (repair completed)
   -  (still showing as copy)
   -  (not refreshed since scan)
2. Re-run `refreshDirectoryRecoveryLineage` for those directories
3. Update `lineage_refreshed_at` timestamp to avoid redundant work

## Schema Changes (v6)

- Add `scan_finished_at` to `catalog_directories` to track when initial scan completes
- Add `lineage_refreshed_at` to track when lineage was last refreshed
- `finishDirectoryScan` now sets `scan_finished_at`

The refresh runs every 5 seconds and processes up to 8 directories per cycle to avoid blocking the repair worker.

## Testing

- ✅ `go test ./internal/sessioncatalog/`
- ✅ `go test ./internal/tool/builtin/ ./internal/boot/`
- ✅ `go vet ./...`
- ✅ `gofmt -w .`
- ✅ Schema migration test updated for v6

## User Workaround (Immediate)

Users on 1.24.3 can manually trigger reindex once all repairs complete:

```
/doctor sessions --reindex
```

or via command line:

```cmd
reasonix-desktop.exe sessions reindex
```

This one-time reindex will correctly classify lineage because all turn counts are known by then.

---

Cache-impact: none - catalog is a disposable query projection
Documentation-impact: none - internal catalog implementation detail

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,14 @@ branch.
   **v1.24.1** only hid/reclaimed already-created covered copies and fixed Windows
   flash-window startup; **v1.24.2** stops the misclassification source and repairs
   existing user directories without rewriting authoritative JSONL/WAL/sidecar data.
+  **v1.24.3** adds periodic lineage refresh after repair completes: the initial
+  scan marks all sessions as `turns_state='unknown'`, so `promoteCanonicalLeaves`
+  cannot determine the longest branch and all recovery sessions stay marked as
+  `diverged`. After background repair finishes, a 5-second ticker re-runs lineage
+  classification for directories with recently repaired recovery sessions, correctly
+  promoting canonical branches and hiding covered copies. Users on 1.24.3 who still
+  see duplicate sessions can run `/doctor sessions --reindex` once to trigger
+  immediate lineage refresh.
 
 - Goal now runs continuously by default: the former 16-round per-Run boundary,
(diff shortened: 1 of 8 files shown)


SivanCola:
Superseded by #8599. The repair-triggered lineage reconcile and the ordinary-visibility projection from this branch were reworked on top of the Catalog v4 logical projection and are fully contained in #8599, which also folds recovery lineages into one ordinary list row and absorbs the wedged-projection guards from #8594. Closing in favor of that PR.

PR #8594, opened 2026-08-12
fix(sessioncatalog): stop the sidebar freezing on a wedged projection

yingziisme:
## Summary

- `ListTopics` hydrated each topic's sessions from **inside the open topic cursor**. The cursor pins one pooled connection and `listTopicSessions` asks for a second, so the call deadlocks whenever the pool is saturated — always in memory mode, where `projectiondb` pools a single connection (`maxOpen = 1`), and at four concurrent sidebar reads on disk (`MaxOpenConns: 4`). The frontend loads one topic page per expanded project, concurrently, so four expanded projects are enough.
- Nothing bounded the wait: the sidebar read path and the catalog goroutine's periodic metadata sync both ran on the deadline-less boot context, so one starved call silently froze every later index, reconcile, and revision bump. Observed on a running desktop: the session catalog took no write for 40 minutes (WAL byte-identical, last directory scan frozen mid-afternoon) while the other three projections kept writing normally.
- The user-visible result is a sidebar that stops moving with no error anywhere: without a revision bump the frontend never reloads a topic page, and `ListTopics` skips any topic with no indexed session row — so the session the user was actively running was missing from the tree until the app restarted.

Fix:

- Drain the topic page before hydrating sessions; no query nested inside an open cursor.
- Bound the sidebar reads (10s) and the background metadata sync (30s) so a starved pool degrades into a stale page the next revision repairs, instead of a permanent freeze.
- Merge live tabs into an uncursored topic page: a session this app is running is never hidden because the projection lags.

## Verification

- The new guards in `internal/sessioncatalog/connection_pool_test.go` were run against the parent commit (`ea9ee16e4`) in a scratch worktree and **fail there** with `context deadline exceeded` after 15s / 30s, then pass in ~0.1s with this fix:
  - `TestListTopicsHydratesSessionsOffTheTopicCursor` — memory mode, single-connection pool
  - `TestListTopicsSurvivesReadersAtThePoolLimit` — disk pool, 8 concurrent readers
  - `TestCatalogAPIsSurviveASingleConnectionPool` — the whole API surface (reconcile, index, sync, list, get, remove) on a pool of one, so any future path that holds a connection while waiting for another fails by name instead of only under production saturation
- `go test ./internal/sessioncatalog/ ./internal/tool/builtin/ ./internal/boot/`
- `cd desktop && go test . -count=1` → ok in 70s; the same selection previously hit the 10-minute deadlock timeout.
- `make lint` (golangci-lint 0 issues, repolint clean) and `go vet ./...` in both modules.
- Rebuilt and installed the desktop app locally: the session that had been missing from the sidebar is indexed on the first reconcile and ranks first in its project, and the catalog keeps writing afterwards (WAL advancing, three live pooled connections).

## Documentation impact

Documentation-impact: none - internal projection and sidebar read paths only

## Cache impact

Cache-impact: none - no prompt, tool schema, or provider request changes
Cache-guard: not applicable - no cache-sensitive file in the diff; scripts/check-cache-impact.sh reports none
System-prompt-review: N/A

### DIFF
--- a/desktop/session_catalog.go
+++ b/desktop/session_catalog.go
@@ -16,6 +16,8 @@ import (
 	"reasonix/internal/taskcatalog"
 )
 
+const sessionCatalogMetadataSyncTimeout = 30 * time.Second
+
 type SessionCatalogStatus struct {
 	State           string `json:"state"`
 	Mode            string `json:"mode"`
@@ -155,7 +157,7 @@ func (a *App) startSessionCatalog(rebuild bool) {
 			return
 		}
 		a.sessionCatalog.Store(catalog)
-		if err := a.syncSessionCatalogMetadata(ctx, catalog); err != nil && !errors.Is(err, context.Canceled) {
+		if err := a.syncSessionCatalogMetadataBounded(ctx, catalog); err != nil && !errors.Is(err, context.Canceled) {
 			slog.Warn("desktop: sync session catalog metadata", "err", err)
 		}
 		select {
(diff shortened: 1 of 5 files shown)


SivanCola:
Thanks — this is a real deadlock and the bounded-sync/read-timeout design is right. #8599 (the recovery-lineage fold) rebuilt the same area on the Catalog v4 logical projection, so rather than resolving the conflicts twice we adopted and adapted this PR's mechanism there: the bounded metadata sync and sidebar read timeouts, the live-tab merge on the uncursored first page, the drained-cursor ListTopics fix, and the connection-pool regression tests, with a commit-level Co-authored-by credit to you (now on main-v2 as f53edb599). #8599 has merged, so closing this PR as superseded. Thanks again for the diagnosis and the guards.

PR #8696, opened 2026-08-13
fix(desktop): hydrate history for a session opened mid-stream / 修复流式输出中点进会话看不到历史

yingziisme:
## Problem

Clicking into a session whose turn is already streaming renders the live turn over a blank transcript — `alignToBottom` pins those few rows to the bottom and everything above stays empty until the turn ends.

#8665 fixed the empty-surface half of this. What is left: a surface that already carries a live stream, or any row the turn has emitted, still loses its history page.

## Root cause

`shouldApplyHydratedHistory` treats "a live turn is on screen" as reason to drop the loaded page, without asking whether a history page ever landed behind it:

```ts
if (!foregroundTurnActive) return true;
return (state?.items.length ?? 0) === 0 && !hasCachedLiveTurn(state);
```

`hasCachedLiveTurn` has the same gap, so `switchTab` passes `skipHistory: true` for such a tab and the page is never even read.

## Fix

The decision becomes `replace` / `prepend` / `skip`:

- empty surface → `replace` (unchanged from #8665)
- live turn with **no** history page behind it → `prepend`, so the page lands in front of the rows the turn already streamed, deduped against them by a page-suffix / live-prefix signature match
- already-hydrated live turn → `skip`, the protection #8665 added stays intact

`hasCachedLiveTurn` gains the same precondition, so `switchTab` stops skipping the fetch for a tab that has never loaded a page.

Two existing assertions encoded the old contract and now assert the sharper one — history lands **in front of** the live turn rather than not at all:

- `tab-switch-hydration`: "cached running tab skips history hydration"
- `use-controller-history-live-race`: "late history cannot replace the live transcript"

## Verification

- `pnpm exec tsx src/__tests__/running-tab-history-hydration.test.tsx` — new suite; fails on `main-v2` (2 FAIL), passes here
- `pnpm exec tsx src/__tests__/hydrate-history-apply.test.ts` — 13 passed
- `node scripts/run-tests.mjs --keep-going` — 161/161
- `pnpm exec tsc --noEmit -p tsconfig.json` and `-p tsconfig.test.json` — clean
- `go run ./tools/repolint` — clean; `useController.ts` goes 4760 → 4758 lines, inside its budget

Documentation-impact: none - behaviour fix inside desktop hydration; no documented surface or flag changes.

### DIFF
--- a/desktop/frontend/src/__tests__/hydrate-history-apply.test.ts
+++ b/desktop/frontend/src/__tests__/hydrate-history-apply.test.ts
@@ -1,9 +1,10 @@
 // Run: tsx src/__tests__/hydrate-history-apply.test.ts
 
 import {
+  duplicateLiveItemIds,
   hasCachedLiveTurn,
+  hydratedHistoryApplyMode,
   sameSessionPlaceholderItems,
-  shouldApplyHydratedHistory,
 } from "../lib/hydrateHistoryApply";
 
 let passed = 0;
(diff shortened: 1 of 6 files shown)


SivanCola:
Superseded by https://github.com/esengine/DeepSeek-Reasonix/pull/8727.

Kept / adapted from this PR:
- `hydratedHistoryApplyMode` (`replace` / `prepend` / `skip`)
- `hasCachedLiveTurn` requiring a landed history page
- suffix/prefix signature dedupe before `history_prepend`
- the running-tab hydration suite and the sharper live-race / tab-switch assertions

Not cherry-picked as-is: current `main-v2` already has the `history_prepend` reducer, so the mechanism was re-implemented on that base instead of carrying this branch forward.

Credit: @yingziisme is a commit co-author on #8727 (`Co-authored-by` with the public GitHub noreply identity from this PR). Thank you for the mid-stream hydration design and tests.

Closing this PR to avoid a duplicate landing. Please continue any follow-up on #8727.

PR #8749, opened 2026-08-13
feat(lease): user-authorized session ownership handoff between runtimes

### LINKED ISSUE #8745: [Feature]: User-authorized session ownership handoff between runtimes (release/takeover)
### Version line

**Session ownership cannot be handed over between runtimes, which blocks remote-control / multi-device workflows.**

A session is protected by a single-writer lease (`*.jsonl.lease.json` + OS file lock). Verified behavior (v1.21.3, reproduced in the RikkaDroid investigation):

- serve `--resume` on a session held by the desktop app → **rejected**: `this session is in use by another Reasonix process (pid ...); close the other Reasonix window or process first`
- The desktop app opening a session held by serve → **rejected**: `this session is already open in another Reasonix window or still running in the background`
- When the holder **exits**, ownership auto-transfers (the desktop app re-acquires the lease of a session left by a stopped serve) ✅
- There is **no way to actively release or transfer ownership while the holder is alive** — no CLI flag, no serve endpoint, no lease-keeper API (`SessionLeaseKeeper` only has `Rebind`/`Release`; `TryAcquireSessionLease` returns `ErrSessionLeaseHeld` while the OS lock is held; v1.25's new `TryReclaimCurrentProcessSessionLease` explicitly refuses foreign leases: *"A readable info naming another live runtime: never steal it"*)

**Why this blocks a real use case:** remote control through the existing `serve` mechanism (e.g. a mobile app driving a desktop-hosted Reasonix — the RikkaDroid project). To take over a session the user is currently using on the desktop, the user must manually close the desktop tab first, or start a brand-new session. Neither is acceptable for seamless device switching: closing the tab loses context/steering state, and a new session can't continue the conversation. What's needed is an explicit, user-authorized handoff: "give this session to the other runtime".

Concrete scenarios:
1. User starts a long task on the desktop, leaves, wants to continue/monitor from the phone → phone needs to take over the same session.
2. User is chatting from the phone (via serve), arrives at the desktop, wants the desktop window to pick up the same session → desktop needs to take over from serve.
3. After a handoff, the original holder must release cleanly (no stale lease, no `already open` errors), and the new holder must be able to write immediately.

### Proposed solution

A **user-authorized session ownership transfer** mechanism, in two layers:

**1. Lease layer: explicit release + conditional re-acquire**
- Add `SessionLease.ReleaseForHandoff()` (or a keeper-level `Handoff(path)`): the current holder releases the OS lock **and** marks the lease info with a handoff flag (e.g. `"handoff": true` + `transferred_at`), so a re-acquirer can distinguish "handed off intentionally" from "stale/abandoned".
- On acquire, if the lease info carries a fresh handoff flag, treat it as releasable (the OS lock is already free; this is the safe path — the previous holder explicitly let go). Keep refusing live leases without a handoff flag, so nothing can steal an active session silently.
- Optionally: a short handoff grace window (e.g. 10–30s) during which only the lease's intended new owner (identified by a token/`writer_id` in the handoff record) may acquire, to avoid two competing runtimes racing for a freed session.

**2. serve layer: HTTP endpoints**
- `POST /release-session` `{"name": "<session name>", "to": "<optional target writer_id>"}` — releases the current serve binding for that session (or a specific named session) with the handoff flag. 204 on success; 409 if the session isn't held by this serve instance.
- `POST /takeover-session` `{"name": "<session name>", "from": "<optional source writer_id>"}` — attempts to acquire a session whose lease info carries a handoff flag for this runtime. 204 on success; 409 `session in use` if the holder is still live and no handoff flag exists.
- `GET /sessions` response: include a `heldBy` field (`"me"` / `"other:pid"` / `"free"`) so clients can render read-only vs writable state and offer a "take over" action only when appropriate.

**3. Desktop/CLI surface**
- Desktop: when opening a session held by another runtime, offer "Open read-only" and "Request handoff" instead of only erroring out. On user confirmation, send the handoff request; the holder (if it's a serve instance) can auto-accept when idle, or prompt the user on the holder side for confirmation.
- CLI: `reasonix serve --takeover <session>` (explicit user intent) vs the current implicit `--resume` behavior.

**Safety invariants to preserve:**
- No silent stealing: a live holder's lease is never broken without an explicit handoff flag.
- Handoff is atomic with respect to session writes (the releasing holder must not be mid-turn; if a turn is running, the handoff either waits for turn completion or is rejected with `409 turn running` — caller retries).
- Crash safety: if the handoff initiator dies before acquiring, the handoff flag expires after the grace window and the session returns to normal acquire semantics.

This is a follow-up formalization of the detach direction first raised in #8170 (that issue covers detaching a running foreground turn to a background job; this one covers handing session ownership between runtimes — a prerequisite for the remote-control use case that #8170's background jobs will enable).

### Alternatives considered

- **"Just close the tab / use a new session"** — current workaround; rejected because it loses context or breaks continuation for the multi-device use case.
- **Auto-steal with lease timeout** — rejectable: risk of breaking an actively-used session on the holder side; explicit handoff is safer and more predictable.
- **Fully server-mediated sessions (move state to a central daemon)** — larger architectural change; this proposal achieves the use case with a small, well-contained addition to the existing lease protocol.

Linearl:
> [!IMPORTANT]
> **Direction update 2026-09-08**: this PR's full-transfer semantic competes with the upstream mirrored-session model (`internal/serve/session_ownership.go`). Superseding it with a direction-generalized mirrored takeover — see the design comment on #8745 (https://github.com/esengine/DeepSeek-Reasonix/issues/8745#issuecomment-5578063038). Keeping this PR open until that direction is confirmed; the lease-handoff mechanics here are already absorbed upstream (`ReleaseForHandoff(targetWriterID, handoffID)`).
>
> 中文：本 PR 的「彻底转移」语义与上游 mirrored-session 模型竞争，改为在 #8745 提出方向泛化融合设计；lease 交接机制上游已吸收（双参数 ReleaseForHandoff）。方向确认后本 PR 将关闭并由基于 upstream/main-v2 的新 PR 取代。

**TL;DR**: 会话所有权交接全链路：takeover 后原 tab 停止写入，释放/接管路由完备

## Summary

Adds a **user-authorized session ownership handoff** mechanism to the session lease protocol, enabling seamless transfer of a session between runtimes (serve ↔ desktop app) without the holder closing the session first. This is a prerequisite for remote-control workflows (e.g. a mobile client driving a desktop-hosted Reasonix via `serve`).

Fixes #8745.

## Problem

A session is protected by a single-writer lease (`.jsonl.lease.json` + OS file lock). Today, when one runtime holds a session, any other runtime is rejected:

- `serve --resume` on a session held by the desktop app → `this session is in use by another Reasonix process (pid ...)`
- Desktop app opening a session held by serve → `this session is already open in another Reasonix window or still running in the background`

Ownership only transfers when the holder **exits**. There is no way to actively hand a session to another runtime while both are alive, which forces users to close tabs / start new sessions when switching devices.

## Changes

**Lease layer** (`internal/agent/session_lease.go`):
- `SessionLeaseInfo` gains `HandoffTo` + `HandoffExpiresAt` reservation fields (JSON backward-compatible, `omitempty`).
- `SessionLease.ReleaseForHandoff(targetWriterID)`: releases the OS lock (same save-drain wait as `Release`) but instead of deleting the lease info, rewrites it with a 30s reservation naming the target writer. A failed reservation write falls back to a plain release so the session never wedges.
- `TryAcquireSessionLeaseWithHandoff(path, sourceWriterID)`: acquires only under a valid, unexpired reservation; refuses wrong targets and expired reservations.
- `TryAcquireSessionLease` (plain) now refuses an **unexpired** reservation so a third runtime cannot race the transfer; expired reservations fall back to normal acquire semantics (crash-safe).

**serve layer** (`internal/serve/serve.go`):
- `POST /release-session` `{"name": "...", "to": "writer-id"}` — releases the current binding with a handoff reservation. 204 on success; 409 when the session is not held by this runtime or a turn is running (handoff never tears a mid-write session).
- `POST /takeover-session` `{"name": "...", "from": "writer-id"}` — acquires a reserved session and rebinds the controller so subsequent `/submit`/`/history` target it. 204 on success; 409 when no valid reservation exists.
- `GET /sessions` entries gain `heldBy`: `"me"` / `"other"` / omitted, so clients can render read-only vs writable state and only offer "take over" when appropriate.

**Tests** (`internal/agent/session_lease_test.go`):
- `TestSessionLeaseHandoffTransfer` — full transfer cycle: plain acquire refused during window, wrong target refused, intended target succeeds, reservation cleared, plain acquire works again.
- `TestSessionLeaseHandoffExpiry` — backdated reservation: handoff acquire refused, plain acquire not wedged.
- `TestSessionLeaseHandoffRefusesActiveHolder` — no marker → handoff acquire refuses; live holder never disturbed.

## Safety invariants preserved

- **No silent stealing**: a live holder's lease is never broken without an explicit `ReleaseForHandoff` reservation.
- **Atomic with writes**: handoff waits for in-flight saves to drain (same ABA guard as `Release`); a running turn rejects `/release-session` with 409.
- **Crash-safe**: if the handoff initiator dies before acquiring, the reservation expires after 30s and the session returns to normal acquire semantics.

## Verification

- `go build ./...` ✅
- `go test ./internal/agent/ ./internal/serve/` ✅ (including 3 new handoff tests)
- `go vet ./internal/serve/ ./internal/agent/` ✅

Cache-impact: none- handoff reuses the existing per-session lease info (added two `omitempty` fields, `HandoffTo`/`HandoffExpiresAt`); no cache key, prompt prefix, or model-tool schema changes, so request prefixes stay byte-identical and cached entries remain valid.

Cache-guard: none- no new tool/parameter that touches the cache; the new `/release-session` and `/takeover-session` endpoints are server-only state machines and do not introduce new prompt-side input. Existing cache-impact / cache-guard workflows remain green.

System-prompt-review: none- not applicable to this change - the handoff flow is a server-side lease handoff between runtimes; the model's system prompt and tool schemas are unchanged, so no system-prompt review is required.

Documentation-impact: none- session lease handoff is an internal authorization flow; existing session/lease documentation remains correct (no docs/*.md changed).

### DIFF
--- a/internal/agent/session_lease.go
+++ b/internal/agent/session_lease.go
@@ -38,6 +38,13 @@ type SessionLeaseInfo struct {
 	PID         int       `json:"pid"`
 	Hostname    string    `json:"hostname,omitempty"`
 	AcquiredAt  time.Time `json:"acquired_at"`
+	// HandoffTo, when non-empty, marks this lease as handed off by the
+	// previous holder to the named runtime (writer_id). The OS lock is free
+	// while this field is set; only the named target may acquire it, and only
+	// until HandoffExpiresAt. This is the explicit, user-authorized path for
+	// transferring session ownership between runtimes (serve <-> desktop).
+	HandoffTo        string    `json:"handoff_to,omitempty"`
+	HandoffExpiresAt time.Time `json:"handoff_expires_at,omitempty"`
 }
 
 const sessionLeaseOwnerOffset int64 = 1
@@ -48,6 +55,10 @@ func sessionLeaseOwnerBytes(b []byte) []byte {
 	copy(payload[sessionLeaseOwnerOffset:], b)
 	return payload
 }
+
+// HandoffWindow is how long a released-for-handoff lease stays reserved for
+// its intended target before falling back to normal acquire semantics.
+const HandoffWindow = 30 * time.Second
 
 type SessionLeaseError struct {
 	Path string
@@ -123,6 +134,16 @@ func TryAcquireSessionLease(path string) (*SessionLease, error) {
 	if _, loaded := sessionLeaseOwners.LoadOrStore(path, ownerID); loaded {
 		info, _ := LoadSessionLeaseInfo(path)
 		return nil, &SessionLeaseError{Path: path, Info: info}
+	}
+	// Respect an unexpired handoff reservation: while the previous holder has
+	// released the lease for a named target, only that target may acquire
+	// (via TryAcquireSessionLeaseWithHandoff). A plain acquire during the
+	// window would defeat the user-authorized transfer.
+	if info, err := LoadSessionLeaseInfo(path); err == nil && info != nil && info.HandoffTo != "" {
+		if info.HandoffExpiresAt.IsZero() || time.Now().UTC().Before(info.HandoffExpiresAt) {
+			sessionLeaseOwners.CompareAndDelete(path, ownerID)
+			return nil, &SessionLeaseError{Path: path, Info: info}
+		}
 	}
 	leaseLock, err := tryTakeSessionLeaseLock(path)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


Linearl:
All four review points are addressed in `5b33765b`:

1. **Target-identity invariant** — `TryAcquireSessionLeaseWithHandoff` now checks `SessionWriterID() == info.HandoffTo` (the acquiring runtime's own identity), so echoing `handoff_to` back from the lease info no longer bypasses the check. The caller-supplied `sourceWriterID` is re-interpreted as the **releasing** runtime and validated against `info.WriterID` — matching `/takeover-session`'s `from` field semantics.
2. **Tests** — `TestSessionLeaseHandoffTransfer` now simulates distinct runtimes by swapping the package-level `sessionWriterID` (with restore); a spoofed non-target acquire and a wrong-source acquire are both refused. `TestSessionLeaseHandoffExpiry` exercises the named-target path with the correct runtime identity.
3. **releaseSession** — takes `s.bindMu` (serialized with `/resume`, `/new`, `/fork`, `/takeover-session`, model switch), refuses handoff while `s.ctl().Running()`, and absolutizes `dir`/`SessionPath` before `Rel`/compare.
4. **releaseSession** — missing `to` is a 400 validated up front; after a successful `ReleaseForHandoff`, the local keeper drops the lease via `Rebind("")` so a later rebind to the same path performs a real acquire.

Verification: `go build ./...` + `go test ./internal/agent/ -run TestSessionLeaseHandoff` (3/3 pass) + `go test ./internal/serve/` (full pass).

Linearl:
Closing as superseded by upstream #9692 (`734dd7c12` — *feat(serve,desktop,cli): session takeover with cooperative handoff and reclaim*).

Verified against `upstream/main-v2` = `2330104af`:

- `SessionLease.ReleaseForHandoff(targetWriterID, **handoffID** string)` — `internal/agent/session_lease.go:353`. The upstream signature carries a **generation fence** this PR does not have.
- `TryAcquireSessionLeaseWithHandoff(path, sourceWriterID, handoffID)` — `internal/agent/session_lease.go:414`
- `internal/control/session_lease_keeper.go` (+250 lines, incl. bidirectional return handoff)
- `internal/cli/session_takeover_cli.go` (+995 lines)
- `desktop/session_takeover.go` (+1314 lines) plus its adopt/promote/readonly companions and the desktop frontend components
- `internal/eventwire/mirror_queue.go` (+277 lines) — the mirror-spectator transport
- Endpoints: upstream exposes `GET /ownership`, `POST /handoff`, `/adopt`, `/reclaim`, `/mirror-end`. This PR's `/release-session` and `/takeover-session` have **zero hits** anywhere upstream.

This PR touches 3 files (+327/−67) and offers nothing upstream lacks — the only differences are endpoint and field naming, and they are not compatible with upstream's set. Upstream also goes further on safety (generation fencing), completeness (CLI, desktop, mirror transport) and runtime semantics (drain vs hard refusal).

The 30s reservation window and its "only the named target may take the lease" semantics did carry over intact — good to see. Thanks for landing this.

PR #8773, opened 2026-08-13
fix(lsp): default Kotlin LSP to official kotlin-lsp / Kotlin 默认 LSP 切换为官方 kotlin-lsp

### LINKED ISSUE #7971: [Feature]: Update default Kotlin LSP command and InstallHint to official JetBrains project
### Version line

Currently, the default language server command for Kotlin in `internal/lsp/manager.go` is set to `kotlin-language-server`. This defaults to the `fwcd/kotlin-language-server` project, which has been officially deprecated. As stated in their upstream README (Jun 2, 2025):

> There is now an [official language server](https://github.com/Kotlin/kotlin-lsp), so this project can be considered deprecated.

What gets in the way today is that the current default values misguide users (especially new ones) into installing an abandoned project.

1. When the server is not installed, `lsp_diagnostics` throws the following error based on `internal/lsp/manager.go:94`:

https://github.com/esengine/DeepSeek-Reasonix/blob/a720a3afad6531b6723e4b21614af5e808115b48/internal/lsp/manager.go#L94

2. Following standard Windows package management (e.g., Scoop) to install this tool fails, as the manifests do not exist in the common buckets (`main`, `extras`, `jack-mil`):

```powershell
PS ...> scoop install kotlin-language-server
Couldn't find manifest for 'kotlin-language-server'.
PS ...> scoop install kotlin-lsp
Couldn't find manifest for 'kotlin-lsp'.
PS ...> scoop search kotlin
Results from local buckets...

Name                     Version Source Binaries
----                     ------- ------ --------
kotlin-native            2.4.10  main
kotlin                   2.4.10  main
kotlin-interactive-shell 0.5.2   extras
```

### Proposed solution

1. Update the default command for Kotlin from `kotlin-language-server` to the official JetBrains executable: `intellij-server`.
2. Update the `InstallHint` to point to the official installation methods:
	- macOS/Linux: `brew install JetBrains/utils/kotlin-lsp`
	- Windows: Download the standalone zip from `Kotlin/kotlin-lsp` Releases.

JetBrains' `Kotlin/kotlin-lsp` has published official releases (currently v262.9593.0). And the official executable is now `intellij-server`, the legacy `kotlin-lsp` script is explicitly being deprecated:

```powershell
PS ...> intellij-server --help
Usage: intellij-server [OPTIONS] [COMMAND]

Options:
  --version             Print version and exit
  --socket VALUE        A socket which will be used for a Kotlin LSP
  ...
```

(Executing the legacy script throws: `WARNING: kotlin-lsp.cmd is deprecated and will be removed in a future release. Use bin\intellij-server.exe instead.`)

**Additional**

This is not a blocking issue, but fixing the defaults will prevent out-of-the-box friction. Users can successfully bypass this by overriding the LSP configuration in their user config (`[lsp.servers.kotlin]`, ref `internal/config/config.go:711-718`).

https://github.com/esengine/DeepSeek-Reasonix/blob/a720a3afad6531b6723e4b21614af5e808115b48/internal/config/config.go#L707-L718

Tested and confirmed working with the `--stdio` argument:

```ini
[lsp.servers.kotlin]
command = "intellij-server"
args = ["--stdio"]
```

zcxads666:
## Summary

The Kotlin default `kotlin-language-server` points at the deprecated
`fwcd/kotlin-language-server` project (upstream README: "this project can be
considered deprecated"), which misguides new users into installing abandoned
tooling. JetBrains' official `Kotlin/kotlin-lsp` exposes `kotlin-lsp` on PATH
(the binary inside the archives is `bin/intellij-server`).

- Command: `kotlin-language-server` → `kotlin-lsp` (the official PATH name);
  `intellij-server` is kept as a generic `ServerSpec.Fallbacks` entry so the
  Windows zip `bin/` layout works without a reverse symlink.
- Args: `["--stdio"]` — the client talks stdio and the server defaults to
  socket mode; the official VS Code extension launches it with `--stdio`
  (`vscode-extension-core/src/lspClient.ts`), and the issue author confirmed it.
- InstallHint: platform-specific official steps for macOS, Linux, and Windows.

Implements #7971.

## Issues

Fixes #7971

## Verification

- `gofmt -l internal/lsp/` (clean)
- `go vet ./internal/lsp/`
- `go test ./internal/lsp/` — includes new `TestKotlinDefaultSpec` (pins
  `kotlin-lsp` + `--stdio`) and `TestResolveCommandFallback` (fallback used
  when `kotlin-lsp` is missing; primary wins when both are present)
- `go test ./...` (full suite) passes
- `go test -race ./internal/lsp ./internal/boot`
- `go run ./tools/repolint` and focused `golangci-lint` pass
- Local real-server check (not part of this PR): drove the new default
  (`kotlin-lsp --stdio`) against a real `main.kt` — hover, definition, and
  diagnostics all worked, ~6s including cold start.

## Documentation impact

Documentation-impact: none - no `docs/*.md` references the Kotlin server
default; the InstallHint carries the install guidance.

## Cache impact

Cache-impact: none - LSP server defaults are runtime-only; no system prompt,
memory prefix, skill index, or tool surface change.
Cache-guard: TestDefaultSpecsInvariants, TestKotlinDefaultSpec, and the LSPSpecs override tests.
System-prompt-review: Reviewed by @SivanCola - LSP executable selection is runtime-only; system prompt bytes, tool registration, schemas, and ordering are unchanged.

### DIFF
--- a/internal/boot/boot.go
+++ b/internal/boot/boot.go
@@ -2851,39 +2851,6 @@ func MCPStartupNotice(failures []plugin.Failure) (text, detail string, ok bool) 
 		len(failures), strings.Join(names, ", "), more, strings.Join(details, "\n")), true
 }
 
-// LSPSpecs returns the language → server map: the built-in defaults overlaid with
-// any user overrides. A user entry may set only the fields it wants to change;
-// empty fields keep the default for that language.
-func LSPSpecs(cfg config.LSPConfig) map[string]lsp.ServerSpec {
-	specs := lsp.DefaultSpecs()
-	for lang, s := range cfg.Servers {
-		spec := specs[lang]
-		if s.Command != "" {
-			spec.Command = s.Command
-		}
-		if s.Args != nil {
-			spec.Args = s.Args
-		}
-		if s.Env != nil {
-			spec.Env = s.Env
-		}
-		if s.LanguageID != "" {
-			spec.LanguageID = s.LanguageID
-		}
-		if s.Extensions != nil {
-			spec.Extensions = s.Extensions
-		}
-		if s.InstallHint != "" {
-			spec.InstallHint = s.InstallHint
-		}
-		if spec.LanguageID == "" {
-			spec.LanguageID = lang
-		}
-		specs[lang] = spec
-	}
-	return specs
-}
-
 func providerNames(cfg *config.Config) string {
 	names := make([]string, len(cfg.Providers))
 	for i, p := range cfg.Providers {
--- a/internal/boot/lsp_specs.go
+++ b/internal/boot/lsp_specs.go
@@ -1,0 +1,40 @@
+package boot
+
+import (
+	"reasonix/internal/config"
+	"reasonix/internal/lsp"
+)
+
+// LSPSpecs returns the language → server map: the built-in defaults overlaid
+// with user overrides. Empty fields keep the language default. An explicit
+// command replaces the built-in command family, including fallback names.
+func LSPSpecs(cfg config.LSPConfig) map[string]lsp.ServerSpec {
+	specs := lsp.DefaultSpecs()
+	for lang, s := range cfg.Servers {
+		spec := specs[lang]
+		if s.Command != "" {
+			spec.Command = s.Command
+			spec.Fallbacks = nil
+		}
+		if s.Args != nil {
+			spec.Args = s.Args
+		}
+		if s.Env != nil {
+			spec.Env = s.Env
+		}
+		if s.LanguageID != "" {
+			spec.LanguageID = s.LanguageID
+		}
+		if s.Extensions != nil {
+			spec.Extensions = s.Extensions
+		}
+		if s.InstallHint != "" {
+			spec.InstallHint = s.InstallHint
+		}
+		if spec.LanguageID == "" {
+			spec.LanguageID = lang
+		}
+		specs[lang] = spec
+	}
+	return specs
+}
--- a/internal/boot/lsp_specs_test.go
+++ b/internal/boot/lsp_specs_test.go
@@ -1,0 +1,36 @@
+package boot
+
+import (
+	"testing"
+
+	"reasonix/internal/config"
+)
+
+func TestLSPSpecsExplicitCommandDoesNotInheritDefaultFallbacks(t *testing.T) {
+	spec := LSPSpecs(config.LSPConfig{Servers: map[string]config.LSPServer{
+		"kotlin": {
+			Command: "company-kotlin-server",
+			Args:    []string{"--company-mode"},
+		},
+	}})["kotlin"]
+
+	if spec.Command != "company-kotlin-server" {
+		t.Fatalf("Command = %q, want explicit user command", spec.Command)
+	}
+	if len(spec.Args) != 1 || spec.Args[0] != "--company-mode" {
+		t.Fatalf("Args = %v, want explicit user arguments", spec.Args)
+	}
+	if len(spec.Fallbacks) != 0 {
+		t.Fatalf("Fallbacks = %v, want none for an explicit user command", spec.Fallbacks)
+	}
+}
+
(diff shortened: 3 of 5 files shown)


SivanCola (changes requested):
Thanks, this looks directionally right. Switching Kotlin off the deprecated `fwcd/kotlin-language-server` default, and passing `--stdio`, are both correct: Reasonix talks over stdio, and `intellij-server` defaults to a socket.

Before merging, please address the PATH name. `Manager.spawn` only `LookPath`s a single `Command`. This head sets that to `intellij-server`, but the official install methods put `kotlin-lsp` on PATH, not `intellij-server`:

- `brew install JetBrains/utils/kotlin-lsp` symlinks `libexec/bin/intellij-server` → `kotlin-lsp`
- `brew install --cask kotlin-lsp` exposes `kotlin-lsp.sh` as `kotlin-lsp`
- The Kotlin/kotlin-lsp README asks users to symlink the wrapper as `kotlin-lsp`
- Helix / the official neovim notes say to keep `kotlin-lsp` on PATH

The binary inside the archive is `bin/intellij-server`, and the `kotlin-lsp.sh` / `kotlin-lsp.cmd` wrappers are deprecated. That is real. It does not change the PATH name those official installers actually expose. The current `InstallHint` even asks brew users to reverse-symlink `kotlin-lsp` to `intellij-server`. That extra step is why a clean official install still fails with `intellij-server not found`.

A single command also cannot cover both official layouts:

- brew / README / cask → `kotlin-lsp` on PATH
- Windows zip with `bin/` on PATH → `intellij-server.exe` (the #7971 setup)

Please:

1. Keep `Args: []string{"--stdio"}`.
2. Make the default primary command the official PATH name `kotlin-lsp`.
3. `LookPath` `intellij-server` as a fallback so the Windows zip `bin/` layout works without a reverse symlink. A small `ServerSpec` alias/fallback list is fine; do not special-case only Kotlin in `spawn` if a generic field is just as small.
4. Rewrite `InstallHint` to the official brew / standalone zip steps only. Do not ask users to invent a second name.
5. Add a focused test that pins the Kotlin default to `--stdio` and that the fallback is used when `kotlin-lsp` is missing but `intellij-server` is on PATH.

`TestDefaultSpecsInvariants` only checks that a command exists, so this PATH mismatch can stay green.

zcxads666:
Thanks for the review! All points addressed:

- Primary command is now the official PATH name `kotlin-lsp`, with
  `intellij-server` as a generic `ServerSpec.Fallbacks` entry — no Kotlin
  special-case in `spawn`.
- InstallHint rewritten to the official brew / standalone zip steps only.
- Added `TestKotlinDefaultSpec` (pins `kotlin-lsp` + `--stdio`) and
  `TestResolveCommandFallback` (fallback used when `kotlin-lsp` is missing;
  primary wins when both are present).
- Verified locally: `gofmt` / `go vet` / `go test ./internal/lsp/` /
  `go test ./...` all pass. I also ran a quick smoke test against the real
  `kotlin-lsp --stdio` server — hover, definition, and diagnostics all
  worked (~6s cold start).

PR title and body updated accordingly.

SivanCola (approved):
Approved. This is a well-scoped and worthwhile fix. The PR now moves Kotlin users to the official JetBrains language server while preserving explicit user command ownership.

- The default command family handles the official kotlin-lsp PATH name and the Windows archive fallback without special-casing process startup.
- An explicit user command now replaces that built-in command family, so a missing custom binary cannot silently launch intellij-server with unrelated arguments or environment.
- The install hint now gives actionable macOS, Linux, and Windows paths.

The two review gaps are fixed in b6b488bba with focused coverage for platform guidance and LSPSpecs override semantics.

Verified:
- `go test ./internal/lsp ./internal/boot ./internal/config` passed
- `go test -race ./internal/lsp ./internal/boot` passed
- `go test ./...` passed on the latest-base merge tree
- `go run ./tools/repolint` and focused `golangci-lint` passed
- CI, CodeQL, cache-impact, and docs-impact are green

PR #8813, opened 2026-08-13
Fix phantom transcript bottom after interrupted turns / 修复中断回合的虚假底部按钮

SivanCola:
## Summary

- Prevent manual scroll intents from releasing tail-follow when the transcript has no physical native scroll range.
- Self-heal false at-bottom callbacks and render the jump-to-bottom control only when a native range exists.
- Add regression coverage for a short interrupted transcript and preserve the working long-transcript jump path.

## Root cause

The scroll controller treated wheel, touch, keyboard, middle-button, and nested-scroll input as manual intent even when scroll height did not exceed viewport height. That allowed logical scroll state to become manual while the native viewport had nowhere to move, and the jump control rendered from logical state alone.

## Verification

- pnpm test:transcript
- pnpm test:typecheck
- pnpm typecheck
- pnpm build
- Microsoft Edge real-browser check: a 1185/1185 no-range transcript remains in tail-follow with no jump control after wheel input
- Microsoft Edge real-browser check: a 5235/585 overflowing transcript still exposes the control and reaches bottom distance 0 after click
- go run ./tools/repolint
- git diff --check

## Compatibility and risk

- Frontend-only scroll-state handling; no persisted-data, Wails API, dependency, provider, or prompt-cache changes.
- Uses the existing 4 px at-bottom tolerance, so genuine overflow and explicit jump-to-bottom behavior remain unchanged.

Documentation-impact: none - This restores existing transcript scroll behavior and does not change documented user workflows or interfaces.

### DIFF
--- a/desktop/frontend/src/__tests__/transcript-virtualization.test.tsx
+++ b/desktop/frontend/src/__tests__/transcript-virtualization.test.tsx
@@ -277,5 +277,70 @@ function firstTextNode(root: Node): Text | null {
   }
 }
 
+// ── A short interrupted turn has no phantom "bottom" to jump to ─────────────
+// Once alignToBottom was removed, short content correctly stayed at the top.
+// An upward wheel intent still unpinned it even though the scroller had no
+// overflow, exposing a jump-bottom button whose click could not move anywhere.
+{
+  const harness = await createTranscriptHarness({ viewportHeight: 700, rowHeight: 60 });
+  try {
+    const interrupted: Item[] = [
+      { kind: "user", id: "u-interrupted", text: "inspect the four metrics" },
+      { kind: "assistant", id: "r1", text: "", reasoning: "checking the first source", streaming: false },
+      { kind: "tool", id: "t1", name: "read_file", args: "{}", output: "ok", status: "done", readOnly: true },
+      { kind: "assistant", id: "r2", text: "", reasoning: "checking the second source", streaming: false },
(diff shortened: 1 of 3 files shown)


SivanCola:
Review completed on 81ee7fa54. No blocking findings.

- Functional correctness: the fix checks the native scroll range before releasing tail-follow, self-heals false at-bottom reports for non-scrollable transcripts, and keeps the jump control behind the same physical-range invariant.
- Regression coverage: the interrupted short-transcript fixture remains in tail-follow with no dead control; a genuinely overflowing transcript still exposes the control and reaches the native bottom after click.
- Concurrency and performance: VERIFIED SAFE. This is frontend-local synchronous state handling with constant-time DOM metric reads. It adds no goroutines, locks, waits, I/O, network work, timers, or unbounded loops. The Linux desktop race leg passed.
- Security and cache: no dependencies, command execution, network or file access, auth or secret handling, persistence, provider serialization, prompt construction, tool schemas, or cache-visible contracts changed. CodeQL and the cache-impact guard passed.
- Integration: the PR merge tree is clean against main-v2 at 15abb0394, and there are no unresolved review threads. PR #5634 touches the nearby jump-control condition but is 2036 commits behind; if revived, it should rebase after this PR and preserve the physical-range guard.
- Verification: transcript suite, test and production typechecks, production build and bundle budgets, repository lint, real Microsoft Edge short/long transcript interactions, all required CI jobs, and platform desktop jobs passed.

Technical conclusion: ready to merge. Repository policy still requires maintainer review; GitHub does not allow the PR author to self-approve.

PR #7910, opened 2026-08-07
feat(composer): insert slash command/skill at any time

bfxh:
## Summary

前端 Composer 增强：**任意时刻可插入斜杠命令 / Skill**（不再局限于空输入时）。

- `+`（内容菜单）的 "Use command or Skill" 项去掉了 `disabled={text.trim().length > 0}` 限制——非空输入时点击会在光标处插入 `/` 并弹出 slash 菜单（Skill/子代理为结构化调用，可在消息中间插入）
- actions/management 类命令（如 /clear）在非空时保持灰显（语义正确：它们需要整条消息）
- 更新 en/zh/zh-TW 三语言文案（"Available when the input is empty" → "Insert a slash command or Skill at the cursor"）

## Verification
- composer 菜单相关测试 9/9 PASS、内容菜单测试 15/15 PASS（tsx）
- `tsc --noEmit` 零错误（仅环境缺失的 @xterm 既有错误，与本次无关）

## Notes
- 独立 PR：从主 PR #7791（OSWorld 闭环）拆出，前端零 Go 冲突

## CI Declarations

Documentation-impact: none - 纯前端交互改动（斜杠命令/Skill 可用性），产品文档无需更新

PR #8821, opened 2026-08-13
feat(composer): reference open windows, relay results, and expand as text

bfxh:
## Summary

Add a **"Reference window"** action to the composer's `+` content menu: a picker of all **open windows**, **including the current window itself**, which inserts a reference card. The referenced window's recent transcript rides along as context on submit (reuses the existing `PreviewSession` / `buildSessionContext` pipeline).

Window reference cards get two extra controls (not available on history-session cards):

- **Run-next relay** (⇲): when armed, the current window waits for its turn to finish, then pushes the finished result into the referenced window via `SubmitToTab`, **triggering that window to run automatically**. No result text → a fallback notice is relayed instead.
- **Expand as text**: pulls the referenced window's content and inserts it inline as plain text in the chat box (the "TXT into the chat" path).

Picker behavior mirrors the recent-session picker: search box, keyboard navigation (↑/↓/Enter/Escape), back row to close, menus stay mutually exclusive, and the current window is marked `(this window)`.

## Verification
- `composer-goal-toggle.test.tsx`: 237 passed / 3 pre-existing failures (durable-guidance suite fails identically on `main-v2` without these changes) — new coverage: 5 add-content actions incl. window action, picker lists this+another window with current marked, relay toggle arms run-next, submit attaches window content as context, relay calls `SubmitToTab` with the finished result, expand inserts content as text.
- `tsc --noEmit` clean, eslint clean, CSS syntax + z-index token checks pass.
- Branch is based on latest `main-v2` and contains only this feature (no overlap with #7910).

## Notes
- Relay waits for the current window's turn completion (`running` false) before submitting to the referenced tab; per-submit re-arm, no stale relays.
- Relies on existing `app.ListTabs()`, `app.PreviewSession()`, `app.SubmitToTab()` — no new backend API needed.

## CI Declarations

Documentation-impact: none - 前端交互改动（引用窗口/接力/展开文本），产品文档无需更新


bfxh:
Closed by author (bfxh) on request.

PR #8831, opened 2026-08-14
fix(desktop): recover answers after missed turn events / 修复漏事件后回答一直转圈

SivanCola:
## Summary

- arm stale-turn recovery from the optimistic submission timestamp even when the entire `agent:event` stream, including `turn_started`, is missed;
- keep a lightweight active-tab watchdog running while the backend still reports work in progress;
- hydrate the persisted answer once the backend becomes idle, so the result appears without switching conversations;
- add deterministic dropped-stream coverage plus adjacent controller regression coverage.

## Root cause

The existing watchdog required a received turn event before it considered a turn stale. If Wails dropped the whole stream, the optimistic frontend state stayed `running` with no activity timestamp, so recovery never started. In addition, a watchdog probe that found the backend still running did not schedule another probe, leaving a later missed answer and `turn_done` unreconciled.

## User impact

Long desktop conversations no longer remain on an empty spinning assistant placeholder when the latest event stream is lost. Once the backend has persisted the completed turn, the visible conversation converges automatically without a tab switch.

## Concurrency and performance

- only the active tab owns a watchdog timer;
- tab identity is checked before probing, after asynchronous reconciliation, and before re-arming;
- effect cleanup cancels the timer and prevents stale completions from scheduling work after a tab change;
- a still-running backend is probed again after 30 seconds;
- watchdog probes use `ListTabs` only and skip jobs, effort, and balance refreshes;
- history hydration runs only when reconciliation observes the backend transition to idle.

## Compatibility, security, and cache

- no Wails method or JSON contract, persistence format, provider, prompt, tool schema, dependency, credential, or network boundary changes;
- prompt-cache behavior is unaffected;
- the fix reuses existing tab status and history APIs.

## Related work

#4522 addresses replaying a pending approval after stale reconciliation. Current `main-v2` already preserves that behavior in runtime status dispatch. This PR covers a different failure mode: zero received turn events and repeated reconciliation while the backend remains active.

## Verification

- stale-turn watchdog regression: 9 passed;
- controller meta/reducer coverage: 146 passed;
- live-history race: 5 passed;
- running-tab history hydration: 9 passed;
- send fallback: 5 passed;
- cancel reconciliation: 18 passed;
- `pnpm typecheck`;
- `pnpm lint:hooks`;
- `pnpm build` (all bundle budgets pass; initial JavaScript gzip 411.7 KiB / 425.0 KiB);
- `go run ./tools/repolint` (`clean`, 1932 baselined findings);
- `git diff --check`.

`pnpm test` is not fully green in this checkout: it reaches `composer-goal-toggle.test.tsx` and reports three durable-guidance assertion failures. An isolated rerun reproduces those three failures (227 passed, 3 failed); that suite renders `Composer` directly and does not exercise the controller/watchdog path changed here.

Documentation-impact: none - This is automatic desktop recovery behavior with no new setting or user workflow.

### DIFF
--- a/desktop/frontend/src/__tests__/use-controller-meta.test.ts
+++ b/desktop/frontend/src/__tests__/use-controller-meta.test.ts
@@ -1,6 +1,7 @@
 // Run: tsx src/__tests__/use-controller-meta.test.ts
 
-import { currentTurnWaitMs, effortSwitchNoticeText, foregroundRunningFromRuntimeMeta, historyMessagesToItems, initialState, localizedBackendNoticeText, localizedNoticeText, metaFromTab, modelSwitchNoticeText, reducer, sameMeta, shouldReconcileStaleTurn, tokenModeSwitchNoticeText, type Item } from "../lib/useController";
+import { currentTurnWaitMs, effortSwitchNoticeText, foregroundRunningFromRuntimeMeta, historyMessagesToItems, initialState, localizedBackendNoticeText, localizedNoticeText, metaFromTab, modelSwitchNoticeText, reducer, sameMeta, tokenModeSwitchNoticeText, type Item } from "../lib/useController";
+import { shouldReconcileStaleTurn } from "../lib/useStaleTurnWatchdog";
 import { parseTodos } from "../lib/tools";
 import { resolveTodoPanelTodos } from "../lib/todoVisibility";
 import type { HistoryMessage, Meta, TabMeta, WireUsage } from "../lib/types";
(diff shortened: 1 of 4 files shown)


SivanCola (commented):
Self-review result for `6a7bd8116f55219a03d62f892f418ffb80cadfc7`: no blocking code findings.

### What this PR does and why it helps

It makes the active conversation recover when Wails drops the complete turn event stream. The optimistic submit timestamp now arms stale-turn reconciliation, a still-running backend causes another bounded probe, and an idle backend triggers the existing authoritative history hydration. This removes the need to switch conversations to reveal an already-persisted answer.

### Regression, concurrency, and performance audit

- the watchdog is active-tab scoped and owns one timer;
- cleanup and active-tab checks fence tab switches and stale asynchronous completions;
- a new user turn clears the previous turn's activity timestamp, preventing immediate false-positive probes;
- prompt replay and cancel reconciliation remain on the existing runtime-status path;
- still-running probes are spaced by 30 seconds and use `ListTabs` only, with jobs, effort, and balance refreshes disabled;
- history is fetched only after the backend is observed idle;
- overlapping open PRs that also touch `useController.ts` change event batching, compaction/resume UX, or hydration prompt state in separate semantic areas. #4522 covers prompt replay rather than zero-event recovery.

### Security, compatibility, and cache audit

No new input, file, command, credential, network, dependency, Wails schema, persistence, provider, prompt, tool, or cache boundary is introduced. Prompt-cache behavior is unchanged.

### Verification and merge readiness

The deterministic dropped-stream test covers a fully missed event stream, a first probe that still sees the backend running, re-arming, and final answer hydration without a tab switch. Focused adjacent controller tests, typecheck, Hooks lint, production build and bundle budgets, repolint, and diff checks pass locally. The live PR head is mergeable with no unresolved review threads. All completed CI checks are green; `desktop-windows` is still running, so merge should wait for that final required check to finish successfully.

Because this is my own PR, this is a COMMENTED self-review rather than an approval.

PR #8875, opened 2026-08-14
Add per-project cachecontext user attribution id

BuGlessRB:
# Per-project `cachecontext` user attribution id

## Summary

Adds a top-level, project-scoped `cachecontext` config field that is sent to
providers as the DeepSeek `user_id` — `metadata.user_id` on the
Anthropic-compatible endpoint, `user` on the OpenAI-compatible endpoint. DeepSeek
uses it for KV-cache isolation and usage attribution, so it should differ **per
project** rather than per provider.

Because it lives **outside** the provider config, one shared provider entry works
across all projects; each project sets its own value in its `.reasonix.toml`
without repeating the provider entry (which would otherwise be shadowed by the
user-global entry).

## What changed

- **Config**: new top-level `cachecontext` scalar on `Config`, plus render
  support (annotated and project-delta) and an example in
  `reasonix.example.toml`.
- **Load**: `resolveProviderCredentialsForRoot` stamps the workspace
  `cachecontext` onto every provider entry, so the project-local value reaches
  providers regardless of which file defined the shared entry. Exposed via
  `ProviderEntry.CacheContextValue()`.
- **Boot**: `boot.NewProviderWithProxy` threads it into the provider
  `Config.Extra` as `user_id`.
- **Anthropic provider**: emits `metadata: { user_id }` on the request body when
  set (omitted when empty).
- **OpenAI provider**: emits the `user` body field when set (omitted when empty).
- **Desktop settings**: a new **Project** section appears only when Settings is
  opened on a project tab (hidden in global settings). `SetCacheContext` is
  project-scoped by construction — it writes only to the active workspace's
  `.reasonix.toml` and refuses to run on a global tab, so the value can never
  leak into the user-global config. i18n added in `en` / `zh` / `zh-TW`.
- **Docs**: `docs/GUIDE.md`, `docs/CONFIG_PATHS.md` (+ `zh-CN` mirrors) document
  the field, its allowed pattern, and the desktop Project surface.

## Behavior

- `cachecontext` set in `<project>/.reasonix.toml` → sent as `user_id` for that
  project's provider requests.
- Value pattern `^[a-zA-Z0-9_-]+$`, ≤512 chars; anything else makes DeepSeek
  return HTTP 400.
- Unset → no `user` / `metadata.user_id` is sent (unchanged behavior).

## Tests

- `internal/config` — `TestLoadForRoot_ProjectCacheContextPropagatesToSharedProviders`
  (a provider defined once in the user config carries the project's cachecontext)
  and `TestLoadForRoot_NoCacheContextStaysEmpty`.
- `internal/provider/anthropic` — `TestStreamSendsMetadataUserID`,
  `TestStreamOmitsMetadataUserIDWhenUnset`.
- `internal/provider/openai` — `TestStreamSendsUserID`.
- `internal/boot` — `TestNewProviderPropagatesProjectCacheContextAsUserID`
  (effect test through a real loaded config).

## Impact

- Cache-impact: none - the provider-visible system-prompt prefix is
  byte-identical; `user_id`/`metadata.user_id` is a separate top-level wire
  field that does not touch the prompt prefix (DeepSeek's KV-cache isolation is
  keyed on the prompt content).
- Cache-guard: provider wire tests (`TestStreamSendsUserID`,
  `TestStreamSendsMetadataUserID`) plus the boot effect test
  (`TestNewProviderPropagatesProjectCacheContextAsUserID`).
- System-prompt-review: srb (author) - touches `internal/config/` and
  `internal/boot/` but only adds a per-request attribution field; the
  system-prompt prefix and tool surface are unchanged.
- Documentation-impact: updated - `docs/GUIDE.md`, `docs/CONFIG_PATHS.md`,
  and their `zh-CN` mirrors document the `cachecontext` field and the desktop
  Project settings surface.


BuGlessRB:
Going to create a revised one, this one is pulled back.

PR #8889, opened 2026-08-14
fix: harden provider, desktop recovery, and source update checks

Semidia:
# Reasonix 贡献合并请求草稿（main-v2 / v1.25.1-era）

状态：已重放到实时 `main-v2` 基线、推送到 `Semidia/DeepSeek-Reasonix` fork，并创建上游 Pull Request #8889；尚未合并。当前 GitHub 认证有效，上游仓库对 `Semidia` 只有只读权限。

## Summary

This contribution is prepared on an isolated worktree based on the local
live `main-v2` reference `a88f117a66fe72d5e3cea70fa978002242163042`.
The candidate is thirteen commits ahead of that live reference and contains ten behavior/build
changes plus three focused test-reliability commits.

The changes address OpenAI-compatible URL completion, configured-proxy key
validation, cold-start history loading, explicit-runtime WebView2 profile
isolation, bounded persisted tool output, transport error classification,
WebView2 smoke cleanup, first-launch blank history, and provider request
replay after a streaming POST has already been written. They also add a
read-only upstream source update check, a Windows scattered-source adapter,
and a portable CLI statusline test fixture. The desktop recovery test now
also verifies that an event-log-only update is not mistaken for a duplicate
trash copy.

## Candidate commits

- `09bbb13 fix(provider): complete OpenAI base request URLs`
- `52df4e0 fix(desktop): honor configured proxy for key validation`
- `f15d719 fix(desktop): tolerate missing startup history files`
- `33d97e6 fix(desktop): isolate WebView2 profiles for explicit runtimes`
- `c3dcb79 fix(agent): bound persisted tool output`
- `df51e20 fix(provider): classify transport failures safely`
- `fe63581 test(desktop): make WebView2 smoke cleanup reliable` (test-only)
- `1c6aac1 fix(desktop): treat first-launch blank history as empty`
- `cae31e7 fix(provider): stop replaying written streaming requests`
- `87c1414 test(desktop): preserve event-log divergence fixture` (test-only)
- `e81e133 feat(cli): add read-only source update check`
- `0ccb5cb feat(windows): add scattered source update adapter`
- `06b54cc test(cli): use a portable statusline command` (test-only)

## Issues

The original `D:\reasonix-src` worktree is dirty and contains local acceptance
data and unrelated experiments. It is not used as the contribution source.
The candidate was rebased onto the current `main-v2` line instead of copying
the older session catalog/recovery patch set, because the upstream line
already contains a different write/recovery architecture.

## Verification

- Provider retry safety focused tests passed five consecutive times, including
  the written-request no-replay assertion and OpenAI connection-reset cases.
- `TestTrashSessionMatchesLiveSeesEventLogDivergence` passed five consecutive
  times after the fixture began preserving the old checkpoint while the event
  log advanced.
- The reusable lifecycle lane at
  `D:\agent跑测试测试复用脚本区\杂务专用\Reasonix\Invoke-ReasonixLifecycleVerification.ps1`
  previously passed provider/control/i18n, recovery/catalog/inbox, agent
  durability, desktop lifecycle, event/TurnDone/autosave, tab close/session
  persistence, cold history, cancellation display and bridge reconciliation.
- Current `HEAD` CLI build passed, and
  `go test ./internal/cli ./internal/sourceupdate -count=1` passed after
  the Windows statusline fixture fix and live-base rebase. The current frontend passed the direct
  ESLint, TypeScript, Vite production build, CSS/z-index/WAAPI checks and
  bundle-budget checks. The normal `pnpm build` entry remains blocked because
  this machine has pnpm `11.21.0` while the project requires `>=10 <11`;
  no global package-manager change was made.
- Current Wails `v2.13.0` native build passed from the contribution worktree;
  the deployed desktop artifact is
  `D:\Reasonix\versions\v1.25.1-sandbox\reasonix-desktop.exe` with SHA-256
  `E6C4F1262B1177DB471DBBFE0905B4F0FCE1BE420F7EDFB0680B11F8353D0F6B`.
- The same source HEAD built the CLI and source-update CLI. Both report
  `v1.25.1-sandbox`, commit `06b54cc0e008`, and share SHA-256
  `74842FCF1C39E52B85CE00DFB0EA70F8F221C8221D6CBDBAA57AE5D4B68FA25B`.
- `gofmt` and `git diff --check` are required before push. Sensitive-data
  scanning must find no API key, real configuration, session database,
  browser profile, binary patch or acceptance artifact in the candidate range.

The following are not claimed as green yet: real GUI key re-entry, real cloud
provider send, real history switching/archive/restart recovery, live remote
forward comparison, full upstream CI, and maintainer review. Full root `go test ./...` reached the end but had
environment/baseline failures in `cmd/e2ebench` because the Docker/WSL
`ext4.vhdx` path is unavailable, plus a full-suite `internal/serve` concurrency
timeout; the same serve test passes in isolation. `-race` is unavailable in
this environment because no C compiler is installed. The local Go toolchain is
used with `GOTOOLCHAIN=local`; `go.mod` is not changed to bypass that gap.

## Documentation impact

Documentation-impact: updated - provider URL behavior, proxy validation and
isolated runtime paths are covered by the existing configuration and guide
documentation. The request replay rule is an internal transport-safety
contract and does not change user configuration syntax.

## Cache impact

Cache-impact: none - no system prompt, memory prefix, output style, skill
index, tool schema or provider-visible cache key changed.
Cache-guard: provider, desktop, history, lifecycle, frontend contract, CLI,
Wails and native WebView2 focused checks were rerun for the current `HEAD`.
System-prompt-review: Reviewed by Semidia: `internal/config/config.go` only changes the `RequestURL` field comment; no system prompt, memory prefix, output style, skill index, or provider-visible prompt construction changes.

## Review notes

- Target branch: `main-v2`.
- Source worktree: `D:\其他agent杂务工作区\agent杂务项目与最终成果\Reasonix\reasonix-contrib-main-v2-latest`.
- Source branch: `codex/contrib-prep-main-v2-latest`.
- Fork branch: `Semidia/DeepSeek-Reasonix:codex/contrib-prep-main-v2-latest`.
- This PR contains only source changes from the contribution worktree; local
  scattered deployment scripts and version manifests are not part of the
  candidate range.
- The candidate is rebased onto live `main-v2` at
  `a88f117a66fe72d5e3cea70fa978002242163042`; local `HEAD` is
  `06b54cc0e008290de4648ab7dd13a530d18727c8`. The branch has been pushed to
  `Semidia/DeepSeek-Reasonix`; Pull Request #8889 is open against upstream
  `main-v2`; maintainer review and merge remain pending.


SivanCola:
Thanks for the substantial work here. I reviewed live head `14ceead601c2`. There are several useful fixes in this bundle, but please address the following before merge:

1. **Preserve the no-replay guarantee end to end.** `SendWithRetry` correctly marks an error when the POST may already have reached the server, but the provider adapters pass that error through `StreamFailure`, which wraps it as `StreamInterruptedError`. `Agent.handleSamplingError` retries every stream interruption, so the same billable request can still be sent again by the Agent even though the provider-local retry loop stopped. Please make this ambiguity terminal/non-replayable (or add a distinct recovery classification) and add an Agent-level regression test proving the provider call count remains exactly one after `RequestMayHaveReachedServer`.

2. **Restore the current desktop test contract.** On the reviewed merge tree, `(cd desktop && go test ./...)` failed in two PR-related places:
   - `TestWindowsWebView2SmokeUsesExternalProductionBinaryContract` no longer finds the required production process-tree cleanup contract after the PowerShell refactor.
   - `TestNewSessionKeepsFreshRuntimeWhenTopicRepairFails` is broken because `isolateDesktopUserDirs` now points `REASONIX_HOME` at the directory that this existing failure fixture needs to replace with a file.

   Please update the implementation/tests without weakening the production cleanup guarantee, then rerun the full desktop suite.

3. **Redact project session failures in `doctor`.** `internal/doctor/sessions.go` concatenates the complete session directory into `Sessions.Error`, while `Collect` only redacts `Sessions.Dir`. Project session slugs can encode the workspace path, so a shared `doctor` report may expose machine-local paths. Please redact each collected error and add a regression test for the JSON and text reports.

4. **Tighten the PR scope and public metadata.** Please split or remove the personal-fork/scattered-build operational docs and reports from the general product fixes unless they are an intentional upstream product contract. Refresh the PR body to describe all 21 commits, remove machine-local absolute paths, update the head/base and verification claims, and classify the persisted `RawContent`/compaction change as at least low cache impact rather than none. Please also coordinate the overlapping provider URL/diagnostic work with #8870 and #8871.

5. **Rebase and resolve the new base conflicts first.** The current live base is `main-v2` at `97085d1ab22d`, and GitHub now reports this PR as `DIRTY/CONFLICTING` with merge-base `8330752a06f2`.

Verification completed before the latest base movement:

- focused Go suites for provider, agent, control, CLI, doctor, and source-update passed
- focused `-race` suites for provider, doctor, source-update, and the touched desktop paths passed
- provider endpoint frontend tests passed (41/41), plus production and test TypeScript typechecks
- `git diff --check` passed
- the full desktop suite failed on the two contracts above

Once these items are addressed, conflicts are resolved, and CI is green on the updated head, I will be happy to re-review.

Semidia:
Thanks for the detailed review. I will avoid repeating the five items already listed above. I would like to add two follow-up suggestions that may be useful for the upstream product and for source-built distributions:

1. **Keep the OpenAI-compatible endpoint contract shared across Chat and Responses.**

   #8870 and #8871 address the same misconfiguration class on two separate stream paths: a gateway may accept `POST /v1` with a non-SSE landing page, which otherwise surfaces as a bare EOF. After those fixes land, it may be worth extracting a shared endpoint-diagnostic contract and fixture set so both paths consistently explain `base_url` versus `request_url`, recommend `/v1/chat/completions` or `/v1/responses`, preserve the existing behavior for valid custom gateways, and remain diagnostic-only rather than rejecting gateways with missing or mislabeled `Content-Type` headers.

2. **Consider a first-class read-only source-update diagnostic for source-built or portable distributions.**

   Source-built forks need to track `main-v2` without silently installing official binaries or overwriting local fixes. A small CLI command could compare the local tracking baseline with the remote `main-v2` and return machine-readable states such as `up-to-date`, `upstream-update`, `baseline-missing`, and `check-failed`. The command should be explicitly read-only: no automatic fetch, merge, rebase, reset, checkout, build, or user-data replacement. A Windows wrapper could persist only a redacted result and keep application startup non-blocking.

   Suggested contract tests would cover a missing baseline, a dirty worktree, an unexpected remote, ordinary DNS/TLS failure, an upstream advance, and a refused local loopback proxy. This would make it easier for maintainers and contributors to monitor upstream progress while keeping source integration and release decisions explicit.

These are suggestions for a future upstream maintenance surface, not a request to merge personal scattered-build operational files into the product. I am happy to prepare a narrowly scoped follow-up if either direction is useful.

PR #8927, opened 2026-08-15
fix(anthropic): always replay the thinking block on DeepSeek tool-call turns

oll4com:
## Problem

DeepSeek's Anthropic-compatible endpoint (`https://api.deepseek.com/anthropic`) 400s assistant tool-call turns whose thinking block is absent from the replayed history:

- `The content[].thinking in the thinking mode must be passed back to the API.`
- and a thinking block that lacks the `thinking` field: `missing field `thinking``.

Both were reproduced live against the endpoint (2026-08-15). An empty value is accepted: `{"type":"thinking","thinking":""}` → 200.

The OpenAI-compatible wire already implements the verified policy for this 400 family (`internal/provider/openai/openai.go`): on DeepSeek tool-call turns always send the reasoning key, empty included, because reasoning can be lost upstream (gateway rename, filtering, session restore from older builds, the missing-reasoning retry fallback in `run_loop.go`). The Anthropic wire instead emitted the thinking block only when `ReasoningContent != ""`, so after any lost-reasoning turn every subsequent request failed with the user-facing "Malformed request (HTTP 400)" error until the history was rebuilt.

## Change

`internal/provider/anthropic/anthropic.go`:

- For `c.deepseek` assistant turns with tool calls, always emit the thinking block (`Thinking` becomes a `*string` so `"thinking": ""` is serialized instead of being stripped by `omitempty` — a block without the field 400s).
- Native Anthropic / adaptive behavior unchanged.

## Tests

- `TestBuildRequestDeepSeekReplaysOnlyToolCallReasoningFromHistory/missing_reasoning_on_a_tool-call_turn_still_emits_the_thinking_key` — asserts the empty thinking block is emitted and the wire JSON contains `"thinking":""`.
- `.../native_Anthropic_tool-call_turn_without_reasoning_stays_unchanged` — native path untouched.
- `go test ./internal/provider/...` green.

## Verification

- Live API probes against `api.deepseek.com/anthropic` (thinking enabled, effort high/max, deepseek-v4-flash/pro) confirmed the 400 shapes above and that empty thinking blocks are accepted.
- Multi-turn tool-call sessions through the built CLI (10 turns, web search + bash + resume) complete with 200s.

### DIFF
--- a/internal/provider/anthropic/anthropic.go
+++ b/internal/provider/anthropic/anthropic.go
@@ -374,12 +374,18 @@ func (c *client) buildRequest(_ context.Context, req provider.Request) anthReque
 			// Replay provider reasoning before the tool_use it led to. DeepSeek uses
 			// unsigned thinking blocks and requires the reasoning from a tool-call
 			// turn in every subsequent request, even if the current request no longer
-			// declares tools or has since disabled thinking. Anthropic proper requires
-			// a signature, so reasoning without one cannot be replayed on that endpoint.
-			if c.deepseek && len(m.ToolCalls) > 0 && m.ReasoningContent != "" {
-				blocks = append(blocks, contentBlock{Type: "thinking", Thinking: m.ReasoningContent})
+			// declares tools or has since disabled thinking. The endpoint 400s an
+			// assistant tool_calls turn whose thinking block is absent from the
+			// request JSON ("content[].thinking … must be passed back"), and accepts
+			// an empty thinking block — so, mirroring the OpenAI-compatible wire's
+			// policy for the same 400 family, always emit the block on DeepSeek
+			// tool-call turns (empty included) rather than fail the request when
+			// reasoning was lost upstream. Anthropic proper requires a signature, so
+			// reasoning without one cannot be replayed on that endpoint.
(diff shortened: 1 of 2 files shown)


oll4com:
Closing this as superseded — thanks for the thorough repro and fix!

The same bug family was fixed upstream in the meantime:

- **#8924** "Fix DeepSeek Anthropic reasoning replay" (merged 2026-08-15, released in v1.25.3): replays unsigned thinking before server-search/client-tool blocks, retries frozen requests when reasoning is missing, and fails closed before executing client tools.
- **#8952** "Fix provider reasoning replay recovery" (merged 2026-08-16, released in v1.25.4): repairs interrupted/legacy poisoned sessions and references this PR — the synthetic empty-thinking replay approach explored here was reviewed but not adopted there.

`main-v2` has since moved on and this branch now has a merge conflict in `internal/provider/anthropic/anthropic.go`. Since the fix shipped in v1.25.4, I'm closing this to keep the queue clean. Please reopen if you think any part of this approach is still worth landing (e.g. the always-emit-empty behavior on DeepSeek tool-call turns).

PR #8955, opened 2026-08-16
Fix provider reasoning replay recovery / 修复供应商推理重放恢复

SivanCola:
## Summary

- give strict reasoning-replay providers one exact retry per sampling boundary, independent of the persisted diagnostic incident state
- buffer only attempts whose replay policy can reject or replace the first response; keep fallback-compatible GLM responses streaming live
- keep a second malformed strict response and truncated preserved reasoning fail-closed before client tools execute
- preserve provider-issued GLM reasoning byte-for-byte and serialize the explicit empty `reasoning_content` field for valid tool turns where Coding Plan emitted none
- add deterministic production-adapter coverage and credential-gated live matrices for OpenCode Go, DeepSeek, LongCat, and Zhipu Coding Plan

## Problem

Strict Anthropic/Responses-compatible providers could omit reasoning on a tool-call turn. The Agent detected the unsafe turn, but its single retry was coupled to a persisted warning circuit breaker, so a previous incident could suppress recovery for later user retries. Providers that required replay but suppressed diagnostics could also emit speculative text/tool cards before validation, leaving the UI showing tool work that was never executed.

The same broad replay capability caused every thinking-enabled GLM response to be buffered until the complete SSE response arrived. Real Zhipu Coding Plan responses demonstrated that GLM tool turns may validly omit reasoning, so long text-only or valid empty-reasoning turns appeared stalled even though their chunks had already arrived.

## Official GLM protocol basis

- Zhipu's [Coding Plan integration guide](https://docs.bigmodel.cn/cn/guide/develop/others) documents the OpenAI-compatible Coding Plan endpoint used by this provider.
- Zhipu's [thinking-mode guide](https://docs.bigmodel.cn/cn/guide/capabilities/thinking-mode) requires interleaved/preserved reasoning to be returned complete and unmodified, and its tool-history example includes `reasoning_content` even when the value is empty.
- The [OpenCode Coding Plan guide](https://docs.bigmodel.cn/cn/coding-plan/tool/opencode) uses streaming requests for the supported workflow.

## Fix

- Recovery ownership now follows the provider's replay/fallback contract. Strict protocols always receive one exact frozen-request retry; the retry finalizer rejects a second malformed response without recursion or tool execution.
- Event buffering follows whether the first response can be rejected or replaced. A no-retry empty-fallback GLM turn now publishes SSE text immediately instead of waiting for `[DONE]`.
- GLM uses message-aware replay policy: provider-issued reasoning is retained byte-for-byte, while a valid tool turn with no emitted reasoning is serialized as `"reasoning_content":""` on continuation.
- Empty fallback applies only to genuinely missing reasoning. Reasoning truncated by the client byte limit is rejected before any tool executes, preserving the official complete-and-unmodified requirement.
- The terminal message now states that the automatic retry already ran and recommends a fresh attempt.

## Related work

- Refs #8927 by @oll4com. That PR explores serializing an empty Anthropic thinking block. I reviewed that approach but did not adopt it here: strict Anthropic/Responses turns remain fail-closed, and this change first retries the exact provider request before any client tool execution.

## Verification

- deterministic red/green regression: GLM text must reach the event sink before the held SSE response completes
- deterministic wire regression: GLM tool history contains an explicit empty `reasoning_content`
- deterministic safety regression: overflowed GLM reasoning stops before tool execution or continuation
- `go test ./...`
- `go test -race -count=1 ./internal/agent ./internal/provider/openai ./internal/provider/anthropic`
- `go vet ./...`
- `go run ./tools/repolint`
- `./scripts/cache-guard.sh` — all scenarios passed; tail averages remained 92%–96% against the 90% threshold
- `go test -tags=live ./internal/agent ./internal/provider/openai ./internal/provider/anthropic ./internal/provider/responses -run '^$' -count=1`

Earlier credential-gated live verification used disposable homes/workspaces and covered:

- OpenCode Go DeepSeek Flash: OpenAI tool replay, strict Anthropic tool/search replay, stateless Responses search replay, and 10 production Agent tool loops on each strict protocol
- DeepSeek official: Anthropic tool/search continuation, Responses search, interrupted save/load resume, transient missing-reasoning recovery, persistent omission fallback, and a 17-round tool loop
- LongCat: reasoning replay, five repeated Agent tool loops, deterministic error recovery, and a 17-round tool loop
- Zhipu Coding Plan: OpenAI tool replay, deterministic error recovery, a 17-round tool loop, and five clean Agent tool loops after the GLM policy correction

## Compatibility, cache, and security

- no persisted config, session, state, or Wails contract changed
- normal schemas, ordering, and stable prompt/tool prefixes are unchanged; GLM empty tool history gains only the deterministic documented field
- no dependency, privilege, sandbox, filesystem, or credential-handling changes
- live credentials were process-local, were not added to source/config/artifacts, and the final diff passed a secret and machine-local path scan

Cache-impact: low - GLM tool history with provider-omitted reasoning now adds a deterministic empty reasoning_content field; normal schemas and ordering stay unchanged, and malformed strict retries remain exact.
Cache-guard: `./scripts/cache-guard.sh` passed every scenario with 92%–96% tail averages against the 90% threshold.
Documentation-impact: none - This follows the existing official GLM Coding Plan protocol without adding Reasonix configuration or workflow.

### DIFF
--- a/internal/agent/errors.go
+++ b/internal/agent/errors.go
@@ -27,7 +27,7 @@ func (e *ReasoningReplayError) Error() string {
 	if e != nil && e.Kind == ReasoningReplayOverflow {
 		return "The provider reasoning exceeded the client safety limit, so Reasonix did not run the requested tools. Existing work was kept; retry to continue safely."
 	}
-	return "The provider omitted reasoning required to replay this tool turn, so Reasonix did not run the requested tools. Existing work was kept; retry to continue safely."
+	return "The provider omitted reasoning required to replay this tool turn after an automatic retry, so Reasonix did not run the requested tools. Existing work was kept; retry to make a fresh attempt."
 }
 
 // PauseClass names the guard that deliberately ended a run, so a host can
--- a/internal/agent/live_opencode_go_replay_test.go
+++ b/internal/agent/live_opencode_go_replay_test.go
@@ -1,0 +1,121 @@
+//go:build live
+
+package agent
+
+import (
+	"context"
+	"os"
+	"strings"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"reasonix/internal/event"
+	"reasonix/internal/provider"
+	"reasonix/internal/provider/anthropic"
+	"reasonix/internal/provider/openai"
+	"reasonix/internal/provider/responses"
+	"reasonix/internal/tool"
+)
+
+// TestLiveOpenCodeGoDeepSeekAgentToolLoops repeatedly exercises the production
+// Agent recovery boundary against both strict OpenCode Go Flash protocols.
+func TestLiveOpenCodeGoDeepSeekAgentToolLoops(t *testing.T) {
+	key := os.Getenv("OPENCODE_GO_API_KEY")
+	if key == "" {
+		t.Skip("OPENCODE_GO_API_KEY not set")
+	}
+	providers := []struct {
+		name string
+		new  func() (provider.Provider, error)
+	}{
+		{name: "anthropic", new: func() (provider.Provider, error) {
+			return anthropic.New(provider.Config{
+				Name: "opencode-go-deepseek-anthropic", BaseURL: "https://opencode.ai/zen/go", Model: "deepseek-v4-flash", APIKey: key,
+				Extra: map[string]any{
+					"api_key_env": "OPENCODE_GO_API_KEY", "reasoning_protocol": "deepseek",
+					"thinking": "adaptive", "effort": "high", "web_search": true,
+				},
+			})
+		}},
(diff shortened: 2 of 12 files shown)


SivanCola:
Superseded by #8952.

This PR's complete head through `798f138b7` is retained in #8952. The integration PR keeps the exact reasoning-replay recovery, live GLM streaming, documented empty `reasoning_content` continuation, and overflow fail-closed behavior from this branch, then centralizes the related DeepSeek Anthropic replay projection at the provider boundary.

Verification completed on this head with 24 successful checks, including race, coverage, CodeQL, cache impact, and all desktop/platform jobs. Closing this child PR so review and merge can continue on #8952.

PR #8950, opened 2026-08-16
fix(transcript): keep the viewport steady when auto-fold collapse shrinks the transcript / 思考区自动折叠收缩时不再追底跳动

XTLine:
## Summary

- Auto-fold collapse of the reasoning panel (completion shrink) no longer drags the transcript viewport with an imperative tail re-aim
- A height-drop detector (>=24px, excluding Virtuoso measurement jitter) distinguishes "fold collapse" from normal growth: tail
ownership is kept, but no `AUTOSCROLL_TO_BOTTOM` is emitted — the browser's native `scrollTop` clamp settles the viewport on
the new bottom

## Issues

Fixes #8914

## Verification

- New reducer contract tests (`shrink`, `shrinkOffBottom`, `isTranscriptContentShrink`): fold collapse keeps mode `tail-follow` with zero commands; growth after a shrink still re-engages tail follow
- Transcript suite 18/18 passing, including `history-load-failure-contract.test.ts` (footer resize contract, no regression)
- `tsc --noEmit` clean
- Recommend a manual check per the issue repro steps: long task → reasoning completion collapse → window no longer jumps up

## Documentation impact

Documentation-impact: none - internal scroll-coordination logic only; restores expected scroll behavior with no new config,
command, or permission; embedded documentation stays correct.

## Cache impact

Cache-impact: none - frontend scroll state machine only (`transcriptScrollController` / `useTranscriptVirtuosoScroll`); no cache or persistence path involved.
Cache-guard: no cache code involved; existing transcript suite covers shrink and growth paths.
System-prompt-review: N/A - no provider-visible system prompt, memory prefix, output style, or skill index changes.

PR #8960, opened 2026-08-16
fix(desktop): rebuild transcript scroll around a single-writer arbiter / 重构会话滚动为单一写者仲裁,修复超长会话无法滚到底

SivanCola:
## Summary

Rebuilds the desktop transcript scroll system around a single-writer arbiter and removes the keyed-remount chain that made very long sessions impossible to scroll to the bottom. Refs #8657 (the v1.25.3 residual reported on Windows/WebView2: sessions with dozens of turns could never reach the newest content).

## Problem

v1.25.3 (#8903/#8925) fixed medium-length sessions, but very long sessions still could not be scrolled to the bottom: approaching the tail, the view was pulled back and the newest content stayed unreachable. Twelve scroll-fix PRs in v1.25.0–v1.25.3 each patched one symptom because they all preserved the same structural premises.

## Root cause

A chain in which every link scales with session length:

1. **Patch storm**: opening a session resolves every lazy content ref (`history_items_patch` per resolved entry, revision bump each time). On long sessions this storm lasts minutes.
2. **Keyed remounts**: each revision bump scheduled a full Virtuoso remount at scroll idle, collapsing the measured size tree back to static height estimates.
3. **Estimate drift**: the estimator assumed ~88 chars/line; CJK rows are ~2x taller than estimated, and scrollToIndex landing error accumulates linearly with row count — several screens on long sessions.
4. **Unarbitrated recovery**: scroll writes lived in three modules, and a timed-out recovery could wait forever for an idle callback or be silently lost when the real Transcript event order transferred ownership to the user.
5. **Unsupported snapshot reuse**: `restoreStateFrom` was offered across changed `data`/`totalCount`, and the blank watchdog could restore the same measured tree it had just declared corrupt.

## Fix

Six layers, each covered by deterministic tests:

- **Phase 0 — observability**: `transcriptScrollProbe` records every imperative Virtuoso scroll write (`{owner, kind, top}`); `check-single-scroll-writer.mjs` fails the desktop build on any scroll writer outside the allowlist.
- **Phase 1 — trustworthy geometry**: East-Asian-Width-aware text estimation (full-width glyphs cost 2 columns) and a session-scoped measured-height cache (`rowKey → height`, per-kind medians) make fallback geometry converge on real measurements.
- **Phase 2 — no remounts on patches** (core fix): content patches flow through `data` updates only; Virtuoso re-measures mounted rows itself. The entire revision→rebuild chain is deleted; the blank-viewport watchdog (storm-proof cooldown dedup) is the only remaining rebuild trigger.
- **Phase 3 — single writer**: all writes route through `transcriptScrollArbiter` + `useTranscriptScrollArbiter`. Preemption priority is fixed: selection > user intent > programmatic > recovery > tail-follow. Every recovery reaches done/cancelled/expired; budget expiry schedules bounded quiet-window retries, while a genuine user takeover cancels the retry and makes the user's anchor authoritative.
- **Phase 4 — contract-safe remounts**: only a surface remount with the identical row-key sequence restores a captured Virtuoso snapshot. Append, prepend, rewind, session switch, and watchdog corruption discard it and fall back to the measured cache plus logical anchor recovery. This stays within Virtuoso's same-`data`/same-`totalCount` snapshot contract.
- **Phase 5 — steady auto-fold shrink**: a fold-sized height decrease keeps tail ownership but emits no tail re-aim and cancels a pending settle loop, allowing the browser's native `scrollTop` clamp to settle the new bottom. Later real growth can resume normal tail-follow.

Supporting fixes use `fileURLToPath` in the Windows scroll-writer checker, extract oversized bridge history fixtures so repolint stays clean, and unmount the navigation-race test root so current `main-v2` controller timers cannot keep the full runner alive.

## Related PR integration

- Adapted #8950's `CONTENT_SHRANK` mechanism onto the single-writer arbiter. Credit to @XTLine; the integration commit retains a public `Co-authored-by` trailer.

## Verification

- `pnpm test`: all 180 discovered desktop frontend suites pass (including pretest).
- `pnpm test:transcript`: pass; `transcript-recovery-race` has 57 deterministic assertions, including automatic no-input retry, real Transcript user-intent ordering, explicit cancellation, zero intermediate `scrollBy`, watchdog snapshot discard, and changed-data snapshot rejection.
- `transcript-scroll-release`: 21/21 assertions, including auto-fold shrink ownership and command suppression.
- `pnpm build`: pass (hooks lint, WAAPI, single-scroll-writer check, CSS/z-index checks, TypeScript, Vite, and bundle budgets).
- `go run ./tools/repolint`: clean (1289 baselined findings).
- `pnpm test:transcript-browser`: pass in real Chromium, including measurement churn and a live ref-resolution storm with scroll-idle pauses; both reach and hold the physical bottom with zero recovery-owned writes.
- **End-to-end before/after**: on the baseline, the continuing storm leaves the tail-following view 2073 px (~2.6 viewports) above the bottom. On this branch it reaches and holds 0 px through full resolution, with no persistent pull-back.

## Compatibility / cache impact

No persisted-format or Wails contract changes. `LiveTurnRegion`, tail-follow ownership, selection retention, native-scrollbar drag freeze, and creation-mode scrollbar remain behind the arbiter's interface-compatible API.

Documentation-impact: none - internal desktop scrolling behavior; contributor-only guidance changed in `CONTRIBUTING.md` and `desktop/AGENTS.md`, with no user documentation change.

Cache-impact: low - `desktop/AGENTS.md` adds stable directory-scoped guidance to the provider-visible prefix for desktop workspaces; runtime transcript rendering remains cache-neutral.

Cache-guard: `scripts/check-cache-impact.sh` - standing-instruction impact is declared and the static prefix was reviewed; runtime request serialization and tool schemas are unchanged.

System-prompt-review: reviewed - `desktop/AGENTS.md` contains static transcript engineering guidance only, with no dynamic, user, or session-local content.

## Boundaries

- The browser gate's persistent-pull-back assertion tolerates a transient one-frame remount flash during a genuine blank-watchdog rebuild (Virtuoso renders its initial frame at scrollTop 0 by design); only displacements persisting 5+ frames fail.
- A Windows/WebView2 recheck by the #8657 reporter is still the final confirmation for that environment. The reproducer here is Chromium, while the root mechanism (patch storm → remount → estimate landing) is platform-independent; the Windows path used by the writer checker is now normalized with `fileURLToPath` and remains covered by Windows CI.


SivanCola:
## Local Windows follow-up

Validated exact head `f52b00acb01df8fc5e534eb2a312f7dd1459d41a` in Windows 11:

- Production frontend build and transcript virtualization (43/43) pass; full desktop `go test ./...` passes with Git for Windows available on `PATH`.
- Microsoft Edge 151 / WebView2 Runtime 151 interactive gate: `bench:storm-40t` repeatedly reaches the physical bottom, holds at 0 px through full ref resolution, and records zero recovery-owned writes; rapid-scroll frame and long-task budgets also pass.
- Production `windows/amd64` Wails build with embedded WebView2 passes. The native startup gate twice observed a live window, renderer, document, and composer for five consecutive seconds and a graceful application exit; this VM then hit a post-success temp-directory cleanup race in the harness.

Boundary: the interactive long-session fixture runs in Microsoft Edge, while the embedded Wails/WebView2 startup path is validated separately. The fixture is not injected into the embedded native window; a #8657 reporter recheck against the original persisted session remains the final environment-specific confirmation.

lunky2021:
长会话现在可以滚动到底部，26版本仍然发生。

PR #8180, opened 2026-08-10
fix(agent): skip warm-cache prune to preserve DeepSeek prefix caching

Culeot:
## What

On a **warm** provider cache, the request-time prune (`applyToolResultMaintenanceView` → `tryToolMaintenance`) rewrites the prompt prefix before every request, invalidating DeepSeek's prefix cache. Cost difference between cache hit and miss is roughly **50–120×** (see #3968 and #8118). This makes a warm session silently pay full input price on every request after accumulated tool results cross the fold threshold.

## Fix

In `internal/agent/context_manager.go`, `tryToolMaintenance`: when the fold threshold is crossed, only elide stale tool results if the cache is **cold/unknown**, or the request is **manual/forced compaction / overflow**. On a warm cache the prune is skipped entirely, preserving the cache-first prefix. Canonical transcript is never touched (projection-only, as before).

## Why this is safe

- Manual compaction (`/compact`), forced compaction, and context-overflow still prune as before — the guard only affects the *automatic* request-time prune.
- Snip band (below fold threshold) is untouched.
- Cache state is already tracked (`SetCacheState`/`CacheState`, used by cold-resume prune); we just honor it at request time too.
- Matches the design intent of `cacheColdAfter` gating: only rewrite the prefix when the cache is already cold.

## Related

Fixes #8118

PR #9033, opened 2026-08-17
Fix tool-result cache prefix stability / 修复工具结果缓存前缀稳定性

SivanCola:
## Summary

- Keep each provider-visible tool result on the stable persisted `Content` representation, bounded to 32 KiB, while retaining the complete local value in `RawContent`.
- Add the fixed internal `session:tool_result` capability behind the existing `use_capability` schema for explicit, UTF-8-safe paged recovery.
- Normalize only exactly matching promoted-`RawContent` projection sidecars back to bounded prefix hashes; invalid projections are discarded without rewriting the canonical transcript.

Cache-impact: low - ordinary prefixes become bounded and append-only again; the top-level tool surface and system prompt remain byte-identical.
Cache-guard: `./scripts/cache-guard.sh` passes, including 64 KiB and 256 KiB tool-result cases capped at 32,768 provider-visible bytes.
System-prompt-review: reviewed by the author - no system-prompt text, provider-visible tool schema, description, or order changes; `task.go` only binds and filters the internal session capability.
Documentation-impact: updated - English and Chinese specifications, tool contracts, cache design notes, and release notes now document bounded `Content`, local `RawContent`, explicit paging, and sidecar compatibility.

## Root cause

Recent request preparation promoted complete tool `RawContent` into ordinary sampling, retries, summaries, projection replay, and covered-prefix hashing. Large tool results therefore increased the provider prompt in proportion to the complete local result and made the model-visible representation differ from the bounded `Content` persisted for compatibility.

This change restores one provider projection boundary: tool `Content` is always model-visible, and `RawContent` remains local unless the model explicitly requests a page.

## Behavior

- The existing 32 KiB, UTF-8-safe, tool-aware head/tail truncation policy remains fixed and non-configurable.
- New truncation markers include the tool name, call ID, byte counts, a stable `result_ref`, and an exact `use_capability` recovery example.
- `session:tool_result` uses byte offsets, defaults to 16 KiB pages, caps pages at 24 KiB, and returns `result_ref`, offsets, total bytes, full SHA-256, and completion state.
- Duplicate call IDs require an exact reference; legacy unique call IDs remain readable without one. Legacy truncated records without retained raw data fail explicitly and recommend rerunning the original tool.
- Capability readers bind to the current Agent session. `SetSession` follows the replacement session, capability clones do not inherit a parent reader, and allowed-tool profiles without `use_capability` are not widened.
- The top-level tool name, description, schema, order, system prompt, configuration surface, compaction ratios, and pressure thresholds are unchanged.

## Cache evidence

The deterministic cache guard now includes 64 KiB and 256 KiB tool results. Both keep the maximum provider-visible tool message at 32,768 bytes, preserve prior request messages byte-for-byte on later turns, and pass the existing 90% tail-average guard.

Sanitized DeepSeek live comparisons used the same large-result fixture and prompt shape:

| Build | Provider-visible large result | Second request prompt | Third request cache hit rate | Full-result sentinel on wire |
| --- | ---: | ---: | ---: | --- |
| Current broken baseline | 262,144 bytes | 132,476 tokens | 99.89% after the large cold miss | yes |
| This change | <=32,768 bytes | 9,568 tokens | 98.82% | no |
| v1.23.0 reference | 16,212 bytes | 9,472 tokens | 98.47% | no |

These observations validate bounded, stable request bytes; they do not promise a fixed provider-side hit rate.

## Compatibility and security

- No session JSON field or schema version is added. Existing `Content` / `RawContent` sessions remain bidirectionally readable; older builds may promote `RawContent` again but do not lose data.
- Promoted-`RawContent` sidecars migrate only when the historical hash matches exactly, including the corresponding receipt hash. Unprovable projection bodies are dropped while canonical messages and maintenance records remain intact.
- The session capability performs no file or network access and cannot select another Agent or session. Page contents are returned only after an explicit model call.
- `use_capability` provider serialization is asserted byte-identical before and after reader binding.

## Verification

Passed:

- Focused `internal/agent` tests for bounding, sampling/summary/projection isolation, persistence, paging, duplicate IDs, UTF-8 boundaries, sidecar migration, and Agent/child isolation.
- Focused `go test -race ./internal/agent` coverage for session binding and paged reads.
- `./scripts/cache-guard.sh`, including 64 KiB and 256 KiB tool results.
- `go test ./internal/boot ./tools/repolint`.
- Environment-gated DeepSeek live regression test and comparative runs against the broken baseline and v1.23.0.

Additional post-fix verification passed on commit `0ead5a732`:

- `go test ./internal/agent ./internal/boot`
- `go test ./...`
- `go vet ./...`
- `go run ./tools/repolint`
- `./scripts/cache-guard.sh`

The active-turn compression and overflow-progress regressions are now covered by direct tests, along with request-input immutability and pressure-prune RawContent isolation.

## Related open work

- #8180 targets an older warm-cache pruning path and does not overlap this provider projection boundary.
- #8256 targets Desktop resume/compaction presentation; documentation conflicts may need routine resolution.
- #7911 memoizes tool execution results. It is semantically independent, though both branches touch `internal/agent/agent.go`.

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -257,9 +257,9 @@ when the sole automatic threshold is crossed.
   `agent.compact_ratio` (default **0.80**; presets 0.70 / 0.80 / 0.85; range
   0.65–0.85).
   `triggerTokens = floor(context_window × compact_ratio)`.
-- **Below the trigger** the model receives complete tool `RawContent` from a
-  temporary request projection; no sidecar is written. Durable `Content` remains
-  bounded for older Reasonix readers.
+- **Below the trigger** ordinary requests remain append-only and no sidecar is
+  written. Every provider request uses the durable, bounded tool `Content`;
+  local `RawContent` is never promoted into sampling, retry, summary, or replay.
 - **At the trigger** one singleflight maintenance transaction first persistently
   prunes every tool result over 8192 Unicode code points to `4096 head +
   "[... tool result middle pruned ...]" + 1024 tail`. If this clears pressure,
@@ -296,10 +296,12 @@ when the sole automatic threshold is crossed.
     the physical remainder. A negative value force-omits optional wire limits;
     if the known auto budget no longer fits, Reasonix compacts instead of
     overriding that choice.
-- Canonical tool storage remains backward compatible: `Content` is a stable
-  ≤32KB form and `RawContent` holds the full original. New request projections
-  promote tool `RawContent` below pressure; prune projections never rewrite either
-  canonical field. Older supported readers therefore remain bounded.
+- Canonical tool storage remains backward compatible: `Content` is the stable
+  provider-visible ≤32KB form and `RawContent` holds the full local original.
+  Full results are returned to the model only after an explicit paged
+  `use_capability` call to `session:tool_result`; sampling, stream retry, summary,
+  and projection replay all use the same bounded `Content`. Prune projections
+  never rewrite either canonical field. Older supported readers remain bounded.
 - Automatic maintenance is planned once in `ContextManager.Prepare` from the
   current projection plus the append-only canonical tail. The canonical
   transcript is never rewritten. Subsequent thresholds merge
--- a/docs/SPEC.zh-CN.md
+++ b/docs/SPEC.zh-CN.md
@@ -148,8 +148,8 @@ transcript，仅在唯一自动阈值被跨越时安装 provider 可见的短 **
 - 每个 provider 声明 `context_window`（tokens）。唯一自动触发值是
   `agent.compact_ratio`（默认 **0.80**；预设 0.70 / 0.80 / 0.85；范围 0.65–0.85）。
   `triggerTokens = floor(context_window × compact_ratio)`。
-- **阈值以下**不写 projection；普通请求临时将 tool `RawContent` 提升为模型
-  可见 `Content`，因此新版本可看到完整工具结果，而旧版本仍安全读取有界内容。
+- **阈值以下**普通请求保持 append-only，不写 projection。所有 provider 请求只使用
+  持久化且有界的 tool `Content`；本地 `RawContent` 不会进入 sampling、重试、摘要或 replay。
 - **达到阈值**后，单飞维护事务先持久剪枝：所有超过 8192 个 Unicode code point
   的工具结果变为 `4096 头部 + "[... tool result middle pruned ...]" + 1024 尾部`。
   若已解除压力则不调摘要模型；否则将连续旧前缀摘要，并仅原样保留最近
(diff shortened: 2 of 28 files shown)


SivanCola (commented):
## Author self-review

Reviewed live head `389548c3b` against `main-v2` `5842ea8af`; the local merge-base matches the live base exactly.

### Findings

- Fixed during review: a new-format truncated record without `RawContent` now retains the marker's original `result_ref` and returns the stable unrecoverable response instead of a misleading not-found response.
- Fixed during review: helper extraction brought the touched files back under repository size policy without raising the lint baseline.
- Fixed during review: iteration uses the repository's current Go modernization conventions.
- No unresolved PR-attributable correctness, security/privacy, cache-prefix, compatibility, or concurrency defect was found.

### Evidence

- The provider-visible form of 64 KiB and 256 KiB tool results is capped at 32,768 bytes; subsequent requests preserve prior provider message bytes, and the retained `RawContent` sentinel is absent from sampling, retry, summary, and projection-replay requests.
- Explicit recovery paging reconstructs the original UTF-8 bytes and SHA-256; duplicate call IDs, legacy unique IDs, invalid boundaries, end-of-result, and missing raw data are covered.
- Session readers are scoped to the current Agent and Session; `SetSession`, clone isolation, restricted children, and the stable `use_capability` provider schema are covered by focused and race-enabled tests.
- Cache guard, focused agent/race tests, boot tests, repolint, golangci-lint, CodeQL, govulncheck, documentation/release checks, and all desktop jobs passed.
- Sanitized live validation reached 98.82% and 98.65% cache hit rates on warmed turns while keeping large tool output bounded; this is supporting evidence, not a provider-rate guarantee.

### Concurrency classification

- **CONFIRMED:** no race was found in the changed paths.
- **VERIFIED SAFE:** frontend binding is mutex-protected, session snapshots use the existing session lock, `SetSession` follows the new session, and clones do not inherit the parent reader.
- **UNVERIFIED:** the repository-wide race matrix cannot complete because its package test phase stops on the two pre-existing failures below; the focused race suite passes and CI contains no `WARNING: DATA RACE` output.

### Merge status: HOLD

The implementation review is clean, but required checks remain red. Linux, macOS, Windows, race, and coverage all stop on the same two order-dependent `internal/agent` tests:

- `TestCompressContextAfterExcludesActiveTurnAndAppendsToolResult`
- `TestToolLoopRetriesOnlyAfterOverflowMaintenanceProgress`

Both failures reproduce in the pristine base full-package suite, while isolated runs on the current base pass. They are not introduced by this PR, but the PR should remain blocked until the upstream baseline is repaired or maintainers explicitly apply their accepted baseline policy and rerun the matrix.

PR #9065, opened 2026-08-18
fix(desktop): land question navigation on the selected question / 问题导航真实落点修复

SivanCola:
## Summary

- Fixed the question navigator (and rewind) in long transcripts jumping to the
  transcript tail instead of the selected question.
- `scrollToDataIndex` now takes the data-relative index only; both call sites
  (question navigation, rewind signal) pass the row index straight through.
  #9054's stale-selection cleanup in `handleJumpToQuestion` is preserved
  unchanged. Recovery restores keep using `firstItemIndex + index` for
  `initialTopMostItemIndex`, where Virtuoso does expect the absolute location.
- Added real-landing regression tests (`transcript-question-jump.test.tsx`)
  and wired them into `pnpm test:transcript`.

## Root cause

React Virtuoso's `scrollToIndex` expects a data-relative index even when
`firstItemIndex` is set; its index normalizer clamps any index to
`totalCount - 1`. The previous code dispatched `firstItemIndex + dataIndex`
(1,000,000 base), so every jump clamped to the last row. 1.27.0 shipped
#9054 (selection cleanup) and #9013 (single activation), which fixed different
failure modes; the index defect itself was still live.

## Attribution

The index fix is authored by @wuhanwoaini521 in closed PR #9027. That PR was
based on older code and conflicted with #9054, so this PR re-applies the fix
on current `main-v2` (commit-level `Co-authored-by` trailer included) and
extends the landing tests (first/middle/tail, back-and-forth, stale selection,
prepend, rewind) with viewport-visibility assertions wired into the standard
script.

## Verification

- `tsx src/__tests__/transcript-question-jump.test.tsx`: 19 passed, 0 failed;
  re-run against the pre-fix code it fails on the first jump (timeout waiting
  for the target question to mount) — the test catches the original bug
- `pnpm run test:transcript`: all suites green
- `pnpm run test:typecheck`, `pnpm run lint:hooks`, `check:scroll-writer`: OK

Cache-impact: none - desktop-only transcript scroll indexing change; no provider-visible prompt, memory prefix, or tool registration is touched

Cache-guard: transcript-question-jump.test.tsx plus the full test:transcript suite pin the changed scroll-arbiter contract

Documentation-impact: none - internal scroll-index fix with no user-facing setting, command, or documented behavior change; existing docs remain accurate


SivanCola (commented):
(Superseded by the PR-level comment below — kept for review-history integrity.)

SivanCola:
Approved. This is a well-scoped and worthwhile fix. The PR correctly addresses
the long-transcript question navigator (and rewind) jumping to the transcript
tail by passing the data-relative index to Virtuoso's `scrollToIndex`, whose
index normalizer clamps any index to `totalCount - 1` — so the previous
`firstItemIndex + dataIndex` (1,000,000 base) always clamped to the last row.

- `scrollToDataIndex` now owns one unambiguous contract (data-relative index),
  matching `scrollToIndex`. The absolute-index path that legitimately needs
  `firstItemIndex` (`initialTopMostItemIndex` recovery restores in
  `transcriptAnchorInitialLocation`) is untouched.
- #9054's stale-selection cleanup in `handleJumpToQuestion` is preserved, so
  the WebView2 lost-pointerup fix keeps working alongside.

The main thing I wanted to tighten was real landing. The original coverage
proved a jump command was dispatched, but not where it landed. I added
`transcript-question-jump.test.tsx` (19 assertions) verifying the selected
question actually mounts inside the viewport for first/middle/tail positions,
back-and-forth jumps, stale-text-selection jumps, rewind, and older-history
prepend; reverted against the pre-fix code it fails on the first jump.

Verified:
- `tsx src/__tests__/transcript-question-jump.test.tsx` passed (19/0)
- `pnpm run test:transcript` full suite passed
- `pnpm run test:typecheck`, `lint:hooks`, `check:scroll-writer` passed
- All CI checks green on the rebased head (lint unblocked by #9067)

PR #9141, opened 2026-08-19
fix(desktop): boxed frame collapse

lightfront:
## Summary

Fix `\boxed` rendering so the frame is a proper rectangle instead of a collapsed vertical bar in WKWebView, and pin pnpm to v10 as required by the frontend `engines` field.

## Root cause

KaTeX draws the `\boxed` frame as an empty `.fbox` span whose width comes from `.katex-stretchy { width:100% }`. Under WKWebView that percentage resolves against the span's intrinsic width (**0**) inside the `inline-table`/`table-layout: fixed` vlist, so:

- top/bottom borders get zero length → invisible
- left/right borders merge → one tall vertical bar (one per box)

The override uses `-webkit-fill-available`, which fills the frame to its containing row while keeping the span's inline-block baseline alignment, so vertical position is preserved exactly.

## Changes

- `desktop/frontend/src/styles.css` — scope `.md .katex .fbox { width: -webkit-fill-available }`.
- `desktop/frontend/package.json` — add `"packageManager": "pnpm@10.34.5"` so Corepack/Wails pick pnpm 10 (engines require `>=10 <11`), instead of defaulting to pnpm 11.
- `internal/cli/latex.go` — TUI math renderer: handle `\boxed` (render inner content), `\left.`/`\right.` (drop the invisible dot delimiter), and add `\vert`/`\lvert`/`\rvert` → `∣`, `\Vert`/`\lVert`/`\rVert` → `‖`.
- `internal/cli/latex_test.go` — regression cases for the above.

## Verification

- `go test ./internal/cli` — pass
- `gofmt -l internal/cli/latex.go internal/cli/latex_test.go` — clean
- `go vet ./internal/cli` — clean
- `pnpm build` (lint:hooks + tsc + CSS checks + vite build + bundle budget) — pass
- Manually verified in the desktop app (WKWebView): boxes now fully enclose the formulas.

Cache-impact: none
Cache-guard: none
System-prompt-review: not needed
Documentation-impact: none - rendering-only fix and dependency pin; no docs/*.md updated because existing documentation remains correct

Co-authored-by: Xingbo Zhao <xbzhao@impcas.ac.cn>


SivanCola (approved):
Reviewed updated head 9fb0a2858. The boxed KaTeX width fallback, frontend contract test, TUI regression coverage, and latest main-v2 integration are verified locally. Approving the code-owner review; CI must still complete successfully before merge.

SivanCola (approved):
Re-approved updated head e42b63ebb after the narrow bundle-budget ratchet. The CSS fallback, contract test, and build budget change are locally verified.

PR #9155, opened 2026-08-20
fix(agent): rebuild zero-filled corrupt branch meta instead of failing saves / 修复：全零损坏的会话 meta 自动重建，不再阻断保存/关闭

Linearl:
## Summary

A branch-meta sidecar (`.jsonl.meta`) can be torn into **all-NUL (or all-whitespace)** bytes when a Windows non-atomic `ReplaceFile` in-place copy is interrupted by a forced reboot (`DPC_WATCHDOG_VIOLATION 0x133` / power loss — see #6325). After that, every save/close fails with:

```
在切换标签页前保存会话失败: decode branch meta <path>.jsonl.meta: invalid character '\x00' looking for beginning of value
```

…and the tab cannot be closed, because `LoadBranchMeta` surfaces the decode error and callers (`save.go:807` `persistBranchMeta`, `branch.go` `ensureBranchMetaUnlocked`) have no rebuild fallback for an already-corrupt file.

`internal/agent/branch.go`:
- `LoadBranchMeta` now treats an **all-zero / all-whitespace undecodable** meta file as **absent** (returns `ok=false, err=nil`), so callers walk the normal "meta missing → rebuild" path on the next save.
- Genuine **partial JSON (non-zero content)** still returns an error, so real corruption is not silently swallowed.

New `internal/agent/branch_meta_corrupt_test.go` covers: zero-filled → absent, whitespace-only → absent, partial JSON → error, `EnsureBranchMeta` rebuilds a corrupt sidecar.

- Refs: #6325

## Motivation

Observed on Windows: a forced reboot corrupted a `jsonl.meta` to 1043 all-NUL bytes; the session's transcript/checkpoints were intact, but the app could not save or close the tab. This change makes the torn meta self-heal on next save instead of blocking the conversation.

## Test plan

- `go build ./internal/agent/` — pass
- `go test ./internal/agent/ -count=1` — pass
- `go test ./internal/agent/ -run 'TestLoadBranchMetaZeroFilledIsAbsent|TestLoadBranchMetaAllWhitespaceIsAbsent|TestLoadBranchMetaGenuineCorruptionStillErrors|TestEnsureBranchMetaRebuildsCorruptMeta|TestBranchMeta'` — pass

Cache-impact: none- branch-meta read path only; no cache behavior change
Cache-guard: updated- TestLoadBranchMetaZeroFilledIsAbsent, TestLoadBranchMetaAllWhitespaceIsAbsent, TestLoadBranchMetaGenuineCorruptionStillErrors, TestEnsureBranchMetaRebuildsCorruptMeta, TestBranchMeta*
Documentation-impact: none- internal behavior only; no user-facing docs
System-prompt-review: none- no agent/boot/config system-prompt changes


SivanCola (approved):
Reviewed live head 505d752198cb4e9d25fb6fb3baa1cb52aad092fc against the latest main-v2 three-dot diff.

Approval rationale:
- LoadBranchMeta now treats only empty/all-NUL/all-JSON-whitespace undecodable sidecars as absent, so the existing EnsureBranchMeta path rebuilds them. Genuine partial JSON still returns an error.
- The change is scoped to branch-meta recovery, with no cache, security, dependency, or new concurrency surface.
- Verified locally: go test ./internal/agent/ -count=1; focused go test -race covering the four new cases.

Non-blocking follow-up: the helper comment saying JSON cannot otherwise “begin that way” is imprecise because valid JSON may begin with whitespace; please reword it to say a valid JSON value cannot consist solely of those bytes.

SivanCola (approved):
Re-approved at current head b515d78dbc713f167acec179ebc60dec7f285d66.

The helper comment has been corrected to state that a valid JSON value cannot consist solely of NUL/JSON-whitespace bytes. The branch now includes the latest main-v2 base without rewriting the contributor history.

Verified after the update:
- go test ./internal/agent/ -count=1
- focused go test -race covering the four branch-meta corruption cases
- diff is limited to internal/agent/branch.go and internal/agent/branch_meta_corrupt_test.go.

PR #9205, opened 2026-08-20
fix(desktop): rebuild session sidebar from authoritative sessions / 修复会话侧栏索引丢失

SivanCola:
## Problem

After an upgrade, the desktop session sidebar could appear to lose conversations even when the authoritative JSONL and sidecar files were still present. A `ready` catalog could have the same row count as disk while pointing at the wrong paths or topic projection, and recovery-only branches could be hidden from the ordinary tree without a visible recovery entry.

## Root cause

The disposable projection trusted a v4 cache lineage and a count-only readiness check. Target discovery did not consistently include restored or currently selected project directories, and recovery visibility treated all recovery-only topics as hidden. The source files were not the cause of the loss.

## Fix

- Move the disposable session index to `v5.sqlite` and rebuild it from authoritative files on the first startup of the repaired generation.
- Validate exact path, scope, topic, lineage, fingerprints, and derived projection state before skipping a directory scan.
- Cover global, global-workspace, registered-project, and restored-tab directories without scanning outside explicit targets.
- Keep rebuild backups and never modify or delete JSONL, metadata, event, or recovery files.
- Show non-empty recovery-only sessions as one logical sidebar row with a recovered/recoverable state; keep covered physical copies in recovery history.
- Expose repair status, diagnostics, `sessions reindex --dir PATH`, and a desktop rebuild action.

## Compatibility and overlap

This is intended for the next patch release after 1.31.0 (1.31.1). The `v4` cache remains untouched; rebuilds retain the replaced projection for rollback. The implementation is independent of the user's source version and does not assume a particular upgrade path.

The catalog/recovery visibility portions of [#9203](https://github.com/esengine/DeepSeek-Reasonix/pull/9203) and [#9204](https://github.com/esengine/DeepSeek-Reasonix/pull/9204) were reviewed and adapted here; this PR does not mechanically cherry-pick either branch. Project-root registration in [#8629](https://github.com/esengine/DeepSeek-Reasonix/pull/8629) and deferred startup behavior in [#8650](https://github.com/esengine/DeepSeek-Reasonix/pull/8650) were also reviewed, while preserving background/non-blocking reconciliation.

## Verification

- `go vet ./...`
- `go test ./...`
- `go test -race ./internal/sessioncatalog ./internal/projectiondb`
- `cd desktop/frontend && pnpm typecheck && pnpm test:typecheck && pnpm lint:hooks && pnpm check:css`
- `go run ./tools/repolint`

Documentation-impact: updated - session catalog v5, strict integrity checks, recovery visibility, diagnostics, and rebuild behavior are documented in `docs/SESSION_CATALOG*` and `docs/CONFIG_PATHS*`.

Cache-impact: none

Cache-guard: N/A — no prompt/provider/tool schema/memory/compaction path touched.

System-prompt-review: N/A

Privacy: no session contents, private paths, usernames, or environment values are included in logs or this PR.


SivanCola (commented):
Self-review complete.

Review matrix: functional behavior, authoritative-source safety, strict index integrity, recovery-only visibility, target coverage, lifecycle/concurrency, CLI diagnostics, frontend status flow, Windows/macOS/Linux compatibility, and regression risk. No blocking findings remain.

All required CI checks are green, including desktop, desktop-windows, desktop-macos, WebKitGTK 4.1, race, CodeQL, govulncheck, and frontend bundle/type checks. GitHub does not allow the PR author to self-approve; this is technically ready for maintainer approval.

PR #9256, opened 2026-08-21
fix(desktop): label the compaction-threshold pin so it is not read as context usage

bfxh:
## Summary

The context-window capacity meter shows two pins on the same scale: the used-context percentage and the auto-compaction threshold. The compact pin rendered as a bare percentage (e.g. `80%`), which can be misread as current context usage — especially confusing right after a session crosses the threshold and the usage pin reads `101%`.

This PR labels the compact pin with the localized "Compression threshold" text so the two pins are visually distinguishable, adds hover tooltips to both pins for assistive clarification, and widens the pin cap so the longer label is not truncated.

## Verification

- `desktop/frontend/src/__tests__/context-panel-capacity.test.tsx`: 10 passed with en-US locale. New assertions cover:
  - compact pin shows a label, not a bare ratio
  - compact pin tooltip includes the threshold ratio
  - used pin exposes a tooltip for assistive clarification
- `tsc --noEmit`: clean
- eslint: clean
- CSS syntax: valid

Note: 2 pre-existing failures occur only under zh-CN system locale because the test baseline renders Chinese; with en-US injected the suite is fully green. These are unrelated to this change.

## Documentation impact

Documentation-impact: none - the change is a self-explanatory UI label improvement; no docs reference the compact pin rendering.

## Cache impact

Cache-impact: none - frontend rendering only; no provider-visible system prompt, memory prefix, output style, skill index, tool schema, or request serialization touched.
Cache-guard: N/A - no cache-sensitive path modified.
System-prompt-review: N/A

### DIFF
--- a/desktop/frontend/src/__tests__/context-panel-capacity.test.tsx
+++ b/desktop/frontend/src/__tests__/context-panel-capacity.test.tsx
@@ -98,6 +98,7 @@ const capacity = document.querySelector(".context-panel__capacity-card");
 const meter = capacity?.querySelector(".context-panel__capacity-meter");
 const fill = meter?.querySelector(".context-panel__progress-fill") as HTMLElement | null;
 const usedPin = meter?.querySelector(".context-panel__capacity-pin--used");
+const compactPin = meter?.querySelector(".context-panel__capacity-pin--compact");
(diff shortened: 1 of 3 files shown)


bfxh:
Closed by author (bfxh) on request.

PR #9304, opened 2026-08-22
fix(billing): bill whole Beijing weekends off-peak for DeepSeek V4

xyzs996:
## What's wrong

`DeepSeekRateBand` decides the band from the UTC hour alone. DeepSeek's pricing
page now says:

> Effective 00:00 (Beijing Time) on Sunday, August 23, 2026, we will adjust our
> peak/off-peak billing rules … with off-peak rates applying throughout the day
> on weekends (Saturdays and Sundays, Beijing Time)

That instant is **2026-08-22T16:00Z — already past.** So on a Beijing Saturday or
Sunday, `01:00–04:00` and `06:00–10:00` UTC still resolve to `peak`, while
DeepSeek bills off-peak. Off-peak is exactly half of peak, so every quote landing
in those windows comes out **2×**.

The first window this gets wrong is **2026-08-23 01:00–04:00 UTC** (Beijing
Sunday 09:00–12:00).

## The part that is easy to get subtly wrong

The weekend is bounded in **Beijing** time, not UTC: it runs from `16:00Z Friday`
to `16:00Z Sunday`. Writing the check as `at.UTC().Weekday()` selects a different
48 hours.

I want to be straight about the size of that second point: with today's peak
windows it changes nothing, because `01:00–04:00` and `06:00–10:00` UTC both sit
outside the 16 hours where the two spellings disagree. It is not a live bug — it
is a trap that arms itself the day the windows move. I took the weekday in a
fixed `UTC+08:00` zone anyway and left a comment saying exactly that, so the next
reader knows it was a decision rather than an accident.

`time.FixedZone` rather than `time.LoadLocation("Asia/Shanghai")`: China has run
a fixed UTC+08:00 with no daylight saving since 1991, so the fixed zone is exact
here, and it keeps the billing path from depending on host tzdata being installed.

## Changes

- `internal/billing/catalog.go` — `deepSeekWeekendOffPeak` predicate, gated on
  `deepSeekWeekendOffPeakFrom` (00:00 Beijing on 2026-08-23), consulted first by
  `DeepSeekRateBand`.
- `internal/billing/catalog_test.go` — six cases through `DeepSeekRateBand`
  (including a Beijing Saturday *before* the rule takes effect, which must still
  be `peak`), plus four that pin the 16:00Z weekend edges directly on the
  predicate. Those four are all off-peak by the hour, so they cannot be asserted
  through `DeepSeekRateBand` today — that is precisely why they are worth pinning.
- `docs/BILLING.md`, `docs/BILLING.zh-CN.md` — the rule and its UTC edges.

`TestResolveDeepSeekScheduledRateBoundaries` and
`TestDeepSeekScheduledRatesAllModelsCurrenciesAndTokenClasses` are unchanged and
still pass: their timestamps are Beijing weekdays, and the `2026-08-16T16:00Z`
cutover case predates the weekend rule.

## Verification

```
gofmt -l internal/billing/   # clean
go vet ./internal/billing/   # clean
go test ./internal/billing/ ./internal/event/ ./internal/eventwire/   # all ok
```

`go test ./internal/cli/` fails on this host both with and without the patch
(`TestMiddleClickUsesPrimarySelectionOutsideTmux`, no X11 primary selection in a
headless container) — pre-existing and unrelated.

Cache-impact: none — billing resolution only, no provider-visible prompt, tool
schema, or request serialization touched.
Cache-guard: not applicable; no cache-sensitive path in the diff.

### DIFF
--- a/docs/BILLING.md
+++ b/docs/BILLING.md
@@ -57,7 +57,11 @@ For official DeepSeek OpenAI, Responses, and Anthropic endpoints, V4 Flash,
 `deepseek-v4-flash-vision-exp` (same list price as Flash), and V4 Pro use
 occurrence-time pricing from 2026-08-17 00:00 Beijing time. Peak windows are
 09:00–12:00 and 14:00–18:00 Beijing time; boundaries are left-closed/right-open
-and all other times are off-peak. The request-completion timestamp is used
+and all other times are off-peak. From 2026-08-23 00:00 Beijing time, Saturdays
+and Sundays bill off-peak for the whole day, peak hours included. That weekend
+is bounded in Beijing time, so it runs 16:00Z Friday to 16:00Z Sunday — the same
+window written against a UTC weekday would be a different 48 hours. The
+request-completion timestamp is used
 because the provider does not report per-token billing time. Images sent to the
 vision SKU are billed as input tokens from provider usage.
 
--- a/docs/BILLING.zh-CN.md
+++ b/docs/BILLING.zh-CN.md
@@ -51,7 +51,9 @@ billing_mode = "payg"       # payg | subscription_equivalent
 从北京时间 2026-08-17 00:00 起，DeepSeek 官方 OpenAI、Responses 与
 Anthropic 端点上的 V4 Flash、`deepseek-v4-flash-vision-exp`（价卡与 Flash 相同）
 和 V4 Pro 按请求发生时刻计价。北京时间 09:00–12:00、14:00–18:00 为高峰，区间左闭右开，
-其余为低峰。由于供应商不提供逐 token 计费时刻，Reasonix 使用取得 usage 的请求完成时刻，
+其余为低峰。从北京时间 2026-08-23 00:00 起，周六和周日全天按低峰计价，高峰时段也在内；
+这个周末以北京时间划界，落在 UTC 上是周五 16:00Z 到周日 16:00Z——按 UTC 星期几判定，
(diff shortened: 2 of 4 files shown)


xyzs996:
Citation update, since the sentence I quoted above no longer exists on the page.

The rule went into force today, and the announcement banner has been replaced by
the steady-state footnote. Verbatim, both language versions, read 2026-08-23:

> **EN** — Off-peak rates are half of the peak rates. Peak hours are 01:00 - 04:00
> and 06:00 - 10:00 UTC, Monday through Friday (all other hours are off-peak).

> **ZH** — 空闲时段价格为高峰时段价格的一半。高峰时段为北京时间周一至周五 9:00 - 12:00、14:00 - 18:00（其余为空闲时段）。

Same rule, so nothing in the patch changes: a weekend is off-peak because it is
not Monday–Friday.

Worth knowing while you read it, though: those two sentences do not agree on the
calendar. The Chinese one puts the weekday in Beijing time; the English one
attaches `UTC` to the hours and leaves "Monday through Friday" unqualified, which
reads as UTC weekdays. They differ over 16:00–24:00 UTC on Friday and on Sunday.
With the current windows both readings price all 168 hours identically, so no
test taken from the published schedule can tell them apart — which is why the
patch fixes the zone explicitly rather than relying on a vector to catch it.

<https://api-docs.deepseek.com/quick_start/pricing/> · <https://api-docs.deepseek.com/zh-cn/quick_start/pricing/>

PR #9381, opened 2026-08-24
fix(mcp): dispatch host-session MCP servers supplied over ACP

hdkshingala:
## Problem

An MCP server supplied through ACP `session/new.mcpServers` connects, completes `initialize` + `tools/list`, and its tools are advertised in the capability catalog as `status: "ready"` — but **every call to them is refused**:

```
MCP server "acppassed" is disabled in this session
```

and the server itself is not addressable at all:

```
use_capability action=inspect capability_id=mcp-server:acppassed
  -> unknown capability_id "mcp-server:acppassed"
```

The server was never disabled by anyone. It is simply absent from the registry that guards dispatch, and the catalog and the gate disagree.

This is invisible to a happy-path turn: discovery looks correct, and the failure only appears once the model actually calls the tool mid-turn.

## Root cause

There are two registries, and host-session specs only reach one of them.

`internal/cli/acp.go` passes the ACP servers in as `ExtraPlugins`, and inside `build()`:

- `eagerSpecs = append(eagerSpecs, extraSpecs...)` → **`pluginHost`**. This is why the server connects and its tools list.
- `capSpecs := PluginSpecsForRootWithOptions(cfg.Plugins, ...)` → **config only**.
- `enabledMCPNames` is built from `autoStartEntries` → **config only**.

Both of the latter feed the one call that seeds the dispatch registry:

```go
capRuntime.ConfigureServers(cfg.Plugins, capSpecs, enabledMCPNames)
```

The gate is `present && enabled`:

```go
func (r *MCPCapabilityRuntime) serverEnabled(server string) bool {
	...
	configured, ok := r.servers[strings.TrimSpace(server)]
	...
	return ok && configured.enabled
}
```

A host-session server is absent from `r.servers`, so `ok` is false and `resolveMCPTool` reports it as "disabled". `mcp-server:<name>` is unknown for the same reason — `catalogStateLocked()` iterates `r.servers`.

## Fix

Register host-session specs with the capability runtime and mark them enabled. This follows the intent already stated where they are built — they arrive through an explicit host action, so they are authoritative exactly like an auto-started config plugin (the code already sets `Authorized = true` for them on that basis).

Both changes are required and neither is sufficient alone:

- The `capSpecs` change alone registers the server with `enabled == false`, because `ConfigureServers` reads an absent `enabled[name]` as false.
- The `enabledMCPNames` change alone marks a name that is not in `r.servers`, so `serverEnabled` still fails its `ok` check.

Also separates **"not registered"** from **"disabled"** in the refusal message. They have different causes and different fixes, and reporting the former as the latter sends users looking for a toggle that does not exist. `lockAuthorizedRuntimeServer` already made this distinction; the three `use_capability` resolve paths did not.

### Precedence on a name collision

Host-session servers take precedence over a same-named config entry, **including a disabled one**.
If a user has `[[plugins]] name = "github"` with `auto_start = false` and a client sends
`session/new.mcpServers = [{"name":"github", …}]`, the client's endpoint is what the session
dispatches to.

This is not new intent — the registration tier already makes exactly that choice, filtering colliding
config specs out of `configSpecs` before `registerEnabledMCP` (*"eagerSpecs already includes
extraSpecs; avoid double registration of host-session servers that connected above"*). What is new is
that `mergeHostSessionCapabilitySpecs` now **states** it where the inventory is built, instead of
letting it fall out of slice order plus `ConfigureServers`' `next[name] = …` map overwrite. Reorder
that append, or make `ConfigureServers` first-wins, and the old arrangement silently inverts into the
worst case: a user-disabled config spec marked enabled, with an identity that no longer matches the
tools `pluginHost` registered.

The merge drops the shadowed config **entry** as well. `ConfigureServers` pairs an entry with a spec
by name, so a surviving entry paired `auto_start = false` with the live host-session spec, and `/mcp`
showed one server as simultaneously not-auto-start and not-disabled.

## Reproduction (byte-level, against v1.31.3)

Probed with a local Streamable-HTTP MCP server and a config containing a provider and **zero** `[[plugins]]` stanzas, so this is not the config path in disguise.

| Probe | v1.31.3 | with this change |
| --- | --- | --- |
| `use_capability list` → `mcp-server:acppassed` | *absent* | `ready` |
| `use_capability list` → `mcp-tool:acppassed/traycer_probe_ping` | `ready` | `ready` |
| `inspect mcp-server:acppassed` | `unknown capability_id "mcp-server:acppassed"` | completed, `auto_start: true` |
| `call mcp-tool:acppassed/traycer_probe_ping` | `MCP server "acppassed" is disabled in this session` | returns the tool result |
| MCP server received `tools/call` | never | yes |

The frame that motivated this, from the unpatched build — note the tool is advertised ready and refused in the same session:

```json
{"sessionUpdate":"tool_call_update","toolCallId":"c2","status":"failed",
 "content":[{"type":"content","content":{"type":"text",
   "text":"MCP server \"traycerprobe\" is disabled in this session"}}]}
```

Verified transport-independent: a `stdio` host-session server fails identically after the same successful `initialize` + `tools/list`.

## Tests

Six tests now. The three behavioural ones were each mutation-checked — the guard is not that they pass, but that they fail when the thing they pin is removed:

| Test | Mutation that must break it |
| --- | --- |
| `TestBuildEnablesHostSessionMCPForCapabilityDispatch` | revert the host-session inventory enablement/registration |
| `TestMergeHostSessionCapabilitySpecsPrefersHostSessionOnNameCollision` | replace the merge with a naive append |
| `TestBuildLeavesDisabledConfigMCPUndispatchableAlongsideHostSession` | widen the enablement loop from "every host-session name" to "every name" |

All three confirmed failing under their mutation and passing without it.
`TestBuildLeavesUnknownMCPServerUndispatchable` passes before and after by design — it exists so the fix cannot be satisfied by defaulting unknown servers to enabled.

### Why this went unnoticed

Not for lack of a test exercising host-session MCP — there is a green one.
`TestBuildDeepSeekTextParentCanUseImageReturningMCPAndVisionSubagent`
(`boot_test.go:607`) builds with `ExtraPlugins: [{Name: "vision-reader", …}]`, runs a real turn, and
asserts `mcpCalls == 1`. A host-session MCP server is successfully called on `main-v2` today.

It passes because there are **two dispatch routes and only one carries the enablement gate**:

| Route | Gate | Host-session server |
| --- | --- | --- |
| Registry-resident `mcp__server__tool` | `mcpServerAuthorized` only (`execute_one.go:501-507`) | works today |
| `use_capability` → `mcp-tool:<server>/<tool>` | `serverEnabled` | refused |

That test emits the raw registry name `mcp__vision-reader__inspect` (`boot_test.go:1032`) and takes
the ungated route. **In an ACP session that route does not exist**:
`applyUnifiedProviderToolSurface` narrows the provider-visible set to the fixed
`UnifiedProviderToolNames()` allowlist, so the model never sees an `mcp__*` name and `use_capability`
is the only way to reach an MCP tool.

So the accurate gap is narrower than "no test covers enablement": **no test covers a host-session
server reached through `use_capability`** — which happens to be the only route an ACP client has.
Post-patch the two routes agree for host-session servers; before it they disagreed, and the green
end-to-end test sat on the agreeing side.

## Checks run locally (go1.26.6, darwin/arm64)

```
gofmt -l          clean
go vet ./internal/boot/ ./internal/agent/ ./internal/control/   clean
go build ./...    ok
go test -p 4 ./internal/boot/... ./internal/agent/... ./internal/acp/...   ok
go test -p 1 ./internal/control/...                                        ok
go test -race ./internal/boot/ ./internal/agent/ ./internal/control/       ok
```

## Notes

- No protocol, schema, or capability surface change. `initialize` output is byte-identical. The only observable difference is that a previously-refused call now succeeds.
- The host-session regression now drives the real `use_capability` proxy through `tools/call`, so the test-only `ServerEnabled` / `CapabilityServerEnabled` accessors were removed after the ownership refactor. `go run ./tools/repolint` is clean without any baseline change.

## Impact declarations

Documentation-impact: none - no `docs/*.md` changed. This restores the already-documented behaviour of `session/new.mcpServers` (an MCP server the client asks the agent to connect) rather than introducing or altering a documented surface; no protocol, schema, config key, or CLI flag is added.

Cache-impact: none - structurally, not just observationally. (1) The system prompt cannot be affected: the last write to `sysPrompt` is `boot.go:673` (`config.ApplyOfficialDeepSeekV4ProPersona`, just past the `skill.ApplyIndex` call at `:670`), while the capability inventory does not exist until `boot.go:1564` — there is no write-back. (2) Provider-visible tools cannot be affected: `applyUnifiedProviderToolSurface` sets the visible set from the fixed `UnifiedProviderToolNames()` allowlist, so registering an MCP server cannot add a tool to it — `SetProviderVisibleTools` is an allowlist, not a diff. (3) The skill index cannot be affected, the one non-obvious path since the widened inventory does flow into `skillMCPBindings`: ACP specs are built with no `Package` (`service.go:3035-3046`), `skillMCPBindings` matches on `binding.Package == sk.Plugin`, and `skill.Store.Prepare` short-circuits on empty `sk.Plugin` — so a host-session server can never contribute a skill binding. Removing one cannot either, and that motion is new in this round: `mergeHostSessionCapabilitySpecs` drops the shadowed config spec from the slice passed to `skillMCPBindings` (`boot.go:1571`). That is normally a no-op — `skillMCPBindings` already skips any spec whose name is in `liveServers`, and the host-session server registered its tools under that same `mcp__<name>__` prefix, so the shadowed spec was being skipped anyway. In the one residual case (the host-session server fails to connect and registers nothing) it still cannot move the prompt: bindings are consumed lazily by `skill.Store.Prepare` at invocation time, and `skill.ApplyIndex` has already run regardless. Confirmed empirically as well: the same ACP session driven against the released `v1.31.3` binary and against this branch produced byte-identical provider requests — `tools=13 sha256=6ea9ba724c1851f6e50c2bba699406cc`, `system_msgs=1 sha256=02c53689292e07c4a443bf1d85627eef`.

Cache-guard: three focused guard tests near the changed surface, each mutation-checked — `TestBuildEnablesHostSessionMCPForCapabilityDispatch` (fails if the host-session inventory enablement/registration is reverted), `TestMergeHostSessionCapabilitySpecsPrefersHostSessionOnNameCollision` (fails if the merge becomes a naive append), and `TestBuildLeavesDisabledConfigMCPUndispatchableAlongsideHostSession` (fails if the enablement loop widens from host-session names to every name). `TestBuildLeavesUnknownMCPServerUndispatchable` and `TestMergeHostSessionCapabilitySpecsIsIdentityWithoutHostSession` pass either way by design: they pin that the fix is not default-allow and that the ordinary config path is handed through untouched. For the prompt/tool-prefix surface specifically, the guard is the released-vs-branch request comparison described above, reproducible without credentials: drive one `session/new` + `session/prompt` against a local OpenAI-compatible endpoint that logs the request body, once per binary, and diff the `tools` array and system message.

System-prompt-review: no maintainer sign-off has been obtained — I am an outside contributor and cannot self-approve one, so please treat this as a request for review rather than a record of it. What I can attest concretely: `internal/boot/boot.go` is flagged system-prompt-sensitive because it assembles the agent, but this change adds only MCP registration/enablement inside `build()` and touches no prompt, tool description, or instruction text. The hash comparison under Cache-impact is the evidence, and it covers the assembled system message end to end.

### DIFF
--- a/internal/agent/mcp_availability.go
+++ b/internal/agent/mcp_availability.go
@@ -1,0 +1,50 @@
+package agent
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+func (r *MCPCapabilityRuntime) serverRegistered(server string) bool {
+	if r == nil {
+		return false
+	}
+	r.mu.RLock()
+	_, ok := r.servers[strings.TrimSpace(server)]
+	r.mu.RUnlock()
+	return ok
+}
+
+func mcpServerUnregisteredMessage(server string) string {
+	return fmt.Sprintf("MCP server %q is not registered in this session", server)
+}
+
+func mcpServerDisabledMessage(server string) string {
+	return fmt.Sprintf("MCP server %q is disabled in this session", server)
+}
+
+func mcpServerUnregisteredError(server string) error {
+	return errors.New(mcpServerUnregisteredMessage(server))
+}
+
+func mcpServerDisabledError(server string) error {
+	return errors.New(mcpServerDisabledMessage(server))
+}
+
+func (t *UseCapabilityTool) serverRegistered(server string) bool {
+	if t.runtime != nil {
+		return t.runtime.serverRegistered(server)
+	}
+	// Standalone proxies have no authoritative runtime inventory to miss from.
+	return true
+}
+
+// serverUnavailableReason distinguishes a disabled server from one this
+// session never registered so the refusal points to the correct remedy.
+func (t *UseCapabilityTool) serverUnavailableReason(server string) string {
+	if !t.serverRegistered(server) {
+		return mcpServerUnregisteredMessage(server)
+	}
+	return mcpServerDisabledMessage(server)
+}
--- a/internal/agent/mcp_server_unavailable_reason_test.go
+++ b/internal/agent/mcp_server_unavailable_reason_test.go
@@ -1,0 +1,35 @@
+package agent
+
+import (
+	"strings"
+	"testing"
+
+	"reasonix/internal/config"
+)
+
+// "disabled" and "never registered" have different causes and different fixes.
+// Reporting the second as the first sends the user looking for a toggle that
+// does not exist — which is exactly what a host-session MCP server used to hit,
+// since it connects and lists its tools while the capability runtime has no
+// record of it at all.
+func TestServerUnavailableReasonSeparatesUnregisteredFromDisabled(t *testing.T) {
+	r := runtimeWithServers(config.PluginEntry{Name: "enabled-server"})
+	r.servers["disabled-server"] = mcpRuntimeServer{
+		entry:   config.PluginEntry{Name: "disabled-server"},
+		enabled: false,
+	}
+	frontend := r.NewFrontend(nil, nil)
+
+	if got := frontend.serverUnavailableReason("disabled-server"); !strings.Contains(got, "is disabled in this session") {
+		t.Errorf("registered-but-disabled server: reason = %q, want it to say disabled", got)
+	}
+	if got := frontend.serverUnavailableReason("ghost-server"); !strings.Contains(got, "is not registered in this session") {
+		t.Errorf("unregistered server: reason = %q, want it to say not registered", got)
+	}
+	if !frontend.serverEnabled("enabled-server") {
+		t.Error("enabled-server should dispatch")
+	}
+	if frontend.serverEnabled("disabled-server") || frontend.serverEnabled("ghost-server") {
+		t.Error("neither a disabled nor an unregistered server may dispatch")
+	}
+}
--- a/internal/agent/usecapability.go
+++ b/internal/agent/usecapability.go
@@ -3,7 +3,6 @@ package agent
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"maps"
 	"sort"
@@ -291,26 +290,6 @@ func (r *MCPCapabilityRuntime) enabledSpec(server string) (plugin.Spec, bool) {
 		return plugin.Spec{}, false
 	}
 	return cloneMCPSpec(configured.spec), true
-}
-
-// ServerEnabled is the exported form of serverEnabled. Hosts and regression
-// tests need to assert against the real dispatch gate rather than a proxy for
-// it, because a server can connect and list tools while still being refused.
-func (r *MCPCapabilityRuntime) ServerEnabled(server string) bool {
-	return r.serverEnabled(server)
-}
-
-// serverRegistered reports whether this runtime knows the server at all,
-// independent of whether it is enabled. The two states have different causes and
-// different fixes, so they must not share one message.
-func (r *MCPCapabilityRuntime) serverRegistered(server string) bool {
-	if r == nil {
-		return false
-	}
-	r.mu.RLock()
-	_, ok := r.servers[strings.TrimSpace(server)]
-	r.mu.RUnlock()
-	return ok
 }
 
 func (r *MCPCapabilityRuntime) serverEnabled(server string) bool {
@@ -1404,7 +1383,7 @@ func (t *UseCapabilityTool) lockAuthorizedRuntimeServer(ctx context.Context, ser
 	if t.runtime == nil {
 		spec, ok := t.specFor(server)
 		if !ok {
-			return plugin.Spec{}, func() {}, errors.New(mcpServerUnregisteredMessage(server))
+			return plugin.Spec{}, func() {}, mcpServerUnregisteredError(server)
 		}
 		spec = plugin.ResolveStoredAuthorization(ctx, spec)
 		if !spec.ServerAuthorized() {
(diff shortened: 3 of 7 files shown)


SivanCola (approved):
Approved. This is a well-scoped and worthwhile fix. The PR corrects ACP host-session MCP servers that connected and advertised tools but were absent from the capability runtime inventory, which caused every `use_capability` dispatch to be refused.

- Host-session specs remain session-scoped and now participate in the same authoritative runtime inventory and dispatch gates as configured servers.
- A same-name host-session endpoint explicitly overrides the shadowed configured entry without default-enabling unrelated or unknown servers.
- Provider-visible schemas and the system prompt remain unchanged; dynamic MCP tools stay behind the fixed `use_capability` schema.

I added `0a98eca6c` to resolve the repository-ownership/capacity gate without widening the repolint baseline. The follow-up moves availability and host-session inventory policy into focused owner files, removes the test-only Controller accessor, and upgrades the regression to cross the real `use_capability -> tools/call` boundary while verifying disabled/unknown refusals and schema containment.

Concurrency and security review:

- Inventory assembly happens during boot, before runtime publication; dispatch continues to use the existing runtime locks.
- No dependency, process-launch, persistence, secret-handling, authorization, or production network boundary was added.

Verified on the final head:

- `go run ./tools/repolint` passed with no baseline change
- `go test ./...` passed locally
- focused `go test -race` for `internal/boot`, `internal/agent`, and `internal/control` passed
- `go vet ./internal/boot ./internal/agent ./internal/control` passed
- `gofmt` and `git diff --check` passed
- GitHub Actions and CodeQL passed on the final head

PR #8740, opened 2026-08-13
feat: entry points for turn detach - desktop button, CLI /detach

Linearl:
## Summary

#8170 入口部分（核心在 PR #8735）：**用户可触发的转后台入口**。

1. **Desktop**：
   - 新 IPC `DetachRunningTurn(tabID)`（desktop/app.go → `controller.RequestDetach()`，只读 tab 拒绝）
   - Composer 停止按钮旁新增**"转后台"按钮**（`SendToBack` 图标，turn 运行中显示；`t("composer.detach")` en/zh/zh-TW）
   - **输入排队已天然存在**：Composer 在 turn 运行中提交消息时走 durable inbox（steer follow-up），turn 结束后自动 dispatch——detach 生效（主 turn 结束）后排队输入立即执行，与设计共识一致，**零新增**
2. **CLI**：`/detach` 命令（`runSlashCommand` case → `ctrl.RequestDetach()`；`builtinSlashSpecs` 注册进斜杠菜单；i18n `CmdDetach`/`SlashDetachQueued` en/zh/zh-TW）——CLI 输入排队（inbox）已有，detach 后排队输入自动接续
3. `TurnControl` 接口（internal/control/port.go）新增 `RequestDetach() error`

行为（grill 共识）：点击/命令后 turn 在当前工具轮结束后转后台 job（快照 + subagent runner，核心 PR）；输入框立即可用；停止按钮语义不变（取消即杀死）。

English: entry points for #8170 (core in #8735) — desktop "move to background" button next to the stop button in the Composer (new `DetachRunningTurn` IPC), and a `/detach` slash command in the CLI. Desktop input queueing already exists (durable inbox steer follow-ups auto-dispatch after TurnDone), so queued input runs right after detach takes effect. No behavior change to the stop button.

## Issues

Refs #8170 (part 2: entry points; core in #8735)

## Verification

- `go build ./...` (control/i18n/cli/desktop) — pass
- `go test ./internal/i18n/...` — pass (catalog parity)
- `go test ./internal/cli/...` — pass
- `pnpm typecheck` / `pnpm lint:hooks` / `pnpm check:css` — pass
- `pnpm test:workspace` — 7 passed, 0 failed

## Documentation impact

Documentation-impact: none- new IPC + slash command only; no config, provider, permission, or tool surface changed.

## Cache impact

Cache-impact: none- no prompt, tool schema, or provider request changes.
Cache-guard: none- not required - changed paths are outside provider-visible cache construction.
[...]

PR #8735, opened 2026-08-13
feat(control): detach a running turn to a background job

Linearl:
## Summary

#8170 核心部分：**运行中的前台 turn 可在工具轮边界转为后台 job**（用户点击转后台 → 当前工具调用完成后，会话快照 + 剩余任务以 subagent 式 job 后台继续；主 turn 正常结束，排队输入立即接续执行）。

设计经 14 项评审共识（grill）定稿，要点：黑盒隔离（job 看不到主会话后续）、文件快照（job 完成后删除）、subagent 式 runner（复用主注册表/模型）、尽力而为（turn 尾声自然结束则 detach 失效）、取消即杀死。

**改动（纯 Go 核心，前端/CLI 入口为后续 PR）**：

1. **`internal/agent`**：
   - `turn_detach.go`（新）：`ErrTurnDetached` sentinel + `WithTurnDetachSignal` 上下文谓词
   - `run_loop.go`：每轮工具调用提交后检查 detach 信号 → 返回 `ErrTurnDetached`（最终响应分支不检查 → 自然结束 = 尽力而为）
   - `subagent_store.go`：`PrepareDetached(snapshotPath, spec)`（复制快照进 transcript store + 标记 `kind=detach`）+ `DeleteTranscript(ref)`（完成后清理）+ `copyFile`
2. **`internal/control`**：
   - `detach.go`（新）：`RequestDetach()`（仅 running 时置位，自清除）+ `DetachSpec`/`DetachRunner` 类型 + `performDetach`（持久化 → 调用 runner → notice）
   - `controller.go`：`detachRequested` 字段 + `Options.DetachRunner` 注入
   - `turn_orchestrator.go`：注入 detach 信号（defer 自清）+ `ErrTurnDetached` 分支 → `performDetach` → 正常结束（不 strip、不视为失败）
3. **`internal/boot`**：`detachRunner` 闭包——`PrepareDetached` + `jobs.StartForSession`（异步）+ `RunSubAgentWithSession`（主注册表/模型）+ 完成时 `DeleteTranscript`；注册进 `Options.DetachRunner`

**前缀链**：主会话链与 job 独立链物理隔离（独立文件），各自严格顺序；结果经 jobs manager（`<background-jobs>` + notice）回主会话。

[...]

PR #9409, opened 2026-08-25
feat(runtime): make turns durable, ordered, and interrupt-safe

SivanCola:
## Summary

This PR makes Reasonix turns durable, ordered, interrupt-safe, and bounded in long-running sessions.

It addresses two user-visible failure classes:

- conversations appearing to roll back after Stop, Ask, tab switches, replay repair, or a late history load;
- turns ending early or remaining permanently running when provider streams, tools, prompts, storage, and cancellation race.

The runtime now has one explicit session -> turn -> item lifecycle. The local turn ledger is the source of runtime truth; the frontend is an ordered projection.

## Root cause

Runtime truth was previously split across provider transcript reloads, controller-local state, active-tab routing, and frontend snapshots. Cancellation could replace newer UI state with an older transcript, prompt answers could target a replacement controller, and side effects could begin even when their lifecycle record was not known to be durable.

The first implementation also wrote provider stream chunks too frequently and retained the event sidecar indefinitely, creating avoidable file-open, write, replay, and disk-growth costs in long conversations.

### Durable turn protocol

- Adds stable sessionId, turnId, submissionId, runtimeEpoch, itemId, attemptId, monotonic seq, status, and transcript revision/digest envelopes.
- Enforces exactly one terminal event per Turn.
- Adds the stable host-consumed finish tool with completed, partial, and blocked outcomes, one protocol-repair round, then protocol_failed rather than false completion.
- Keeps providers without tool support on the pure-text compatibility path.
- Keeps interrupted partial answers display-only and excludes incomplete assistant/tool tails from provider continuation.

### Checked side-effect barriers

- Adds an internal checked sink capability while preserving the existing Sink interface for compatibility.
- Commits Turn admission, accepted tool dispatches, RunShell/skill dispatches, Ask and approval requests, prompt answers, and terminal transitions before releasing the corresponding side effect.
- Makes Ask and ordinary approval resolution two-phase and retryable. Concurrent duplicate answers wait for the same persistence transaction and receive the same result.
- If ledger writes fail, the Turn is cancelled, unpublished chunks remain unpublished, provider/prompt/process work is stopped, and the session rejects new Turns instead of reporting completed.
- Cancellation still proceeds even when storage is unavailable.

### Bounded write path and schema-v2 checkpoints

- Moves the single 16 ms / 16 KiB Text and Reasoning coalescer before persistence. The first delta remains immediate; ToolDispatch, ToolResult, Ask, approval, status, and terminal events flush the stream prefix.
- Removes the downstream duplicate coalescer so the ledger and frontend observe identical event boundaries.
- Reuses one append handle for the active Turn and performs one terminal fsync/close. A Desktop projection acknowledgement may use one additional short open cycle.
- Adds schema-v2 event, projection_ack, and checkpoint records while retaining v1 readers and active-v1 continuation.
- Checkpoints keep the latest transcript fingerprint and 16 content-free terminal summaries, including bounded submission receipts.
- Default compaction thresholds are 8 MiB, 4096 retained events, or 256 KiB on controller close. Active Turns and unacknowledged display projections are never discarded.
- Compaction uses strict atomic replacement. A failed compaction leaves the valid sidecar intact and does not turn an already durable projection into a user-visible failure.
- Torn recognized tails are isolated; unknown future schema versions fail without modifying the source file.
- No gzip, segmentation, ToolProgress semantic merging, new dependency, or user setting is added.

### Display projection and crash recovery

- Adds optional turnId to the display-only sidecar and performs idempotent upserts.
- Desktop order is terminal WAL commit, display persistence, projection acknowledgement, optional checkpoint, then frontend turn_done.
- Failed display writes retain the complete WAL and retry asynchronously; successful retry acknowledges the exact captured controller rather than whichever tab is active later.
- Startup rebuilds unacknowledged interrupted output from retained Turn events. Write tools are never replayed.
- CLI/Serve automatically acknowledge terminal delivery because they have no separate display sidecar.

### Replay reset and frontend rebase

- TurnEventsForTab now returns a paged replay view: events, floorSeq, latestSeq, nextAfterSeq, hasMore, resetRequired, transcript fingerprint, and runtime epoch.
- Pages are bounded to 512 events and a 2 MiB soft limit, with at least one oversized event returned.
- Each tab owns its replay promise, live queue, cursor, runtime epoch, and generation fence.
- A compacted cursor triggers explicit transcript rehydration and history_rebase. Only the persisted history prefix is replaced.
- The active Turn, optimistic user message, and live assistant tail are retained and deduplicated by stable identity/signature.
- Older transcript revisions and stale controller epochs are rejected. A stale seq=1 event cannot reset a monotonic ledger.
- Normal Stop still patches only the current Turn; it does not perform a full history replacement.
- Existing scroll-arbiter ownership remains unchanged, and completed measured row heights are not replaced by estimates.

### Metrics and privacy

Content-free aggregate metrics cover append/replay/compaction latency buckets, file-size buckets, stream compression ratio, replay/reset volume, compaction bytes, write/torn-tail failures, projection retries, and pending projection count.

Metrics do not contain session IDs, prompts, answers, tool arguments/output, local paths, credentials, or diagnostic text. No per-event metrics file is created.

## Performance and capacity

Single-run benchmarks on Apple M1 Max:

- 10,000 Text deltas through coalesce -> ledger -> frontend: 34.20 ms.
- 10,000 unmerged ToolProgress records through the persistent handle: 67.03 ms.
- Replay the final 100 events after a synthetic one-million-event checkpoint: 0.124 ms.
- Normal WAL lifecycle assertion: one open, one terminal fsync, one close; Desktop acknowledgement adds at most one short open/close.
- Fully acknowledged idle checkpoint target: below 64 KiB.

Frontend production bundle after merging the latest navigation transaction:

- initial JavaScript gzip: 437.4 KiB / 437.7 KiB;
- initial raw JavaScript and CSS: 2379.2 KiB / 2379.8 KiB;
- combined turn-runtime and navigation cost over the #9422 base: 2.56 KiB gzip (0.59%) and 10.12 KiB raw (0.43%), with 0.34/0.58 KiB deterministic headroom.

## Compatibility

- Provider transcripts and old sessions are not destructively rewritten.
- v1 sidecars remain readable; an active v1 Turn finishes in v1 before checkpoint upgrade.
- Sequence numbers never reset after compaction.
- Unknown JSON fields and empty Wails arrays remain compatible.
- This performance follow-up does not change provider-visible prompts, system instructions, tool schemas, or the finish protocol.
- Mechanical lint-baseline commits refresh exact legacy-owner and repository totals after the two functional commits; it changes no runtime behavior. The one intentional provider-prefix change remains the fixed finish tool introduced by the first four commits.

## Validation

Final local head dcce75411:

- Final-head GitHub CI run 32900548853 passed every executed job; the site job was skipped by its path filter.

- go test ./...
- go test -race ./internal/agent ./internal/control ./internal/turnevent ./internal/event
- go run ./tools/repolint
- GOOS=windows GOARCH=amd64 go test -c ./internal/control
- cd desktop && go test ./...
- frontend test:typecheck
- frontend test:transcript
- replay-reset and cancel-reconcile focused suites
- frontend production build, hook/scroll-writer checks, and bundle budgets
- deterministic coalescing, persistent-FD, v1/v2, checkpoint, atomic-failure, replay, projection-ack, duplicate-answer, tool fail-closed, and cross-session tests
- provider-visible prompt/tool schema and cache fixtures unchanged by the performance, replay, lint, and Windows-storage follow-ups
- Windows CI storage-admission failure reproduced and fixed: open/append failures retain ErrTurnLedgerUnavailable and cannot be misclassified as provider disconnects
- latest main-v2 at ce96692e4 merged; five frontend conflicts were resolved by retaining both navigation paint commits and durable history rebasing
- repolint baseline recomputed on the exact merge tree and remains clean with 1273 findings
- golangci-lint v2.12.2: 0 issues after preserving wrapped I/O errors, modernizing integer/reverse ranges, and deleting superseded approval helpers

### Windows 11 ARM64 / Parallels native smoke

Core runtime scenarios passed on the pre-merge implementation snapshot; the final head also passed the complete Windows CI suite and the native scroll gate below:

- Wails production build with embedded WebView2;
- native window, renderer, document, and composer startup health;
- Stop during a long shell command, durable interrupted terminal, retained partial output, and immediate resend;
- real Ask and approval flows with duplicate-submit protection;
- two concurrent tabs, with stopping A not affecting B;
- 5000-line output, checkpoint compaction to a 1772-byte sidecar, restart, and transcript rebuild.

Final-head native scroll gate passed on dcce75411:

- The final production frontend was served to a real WebView2 window in Parallels Windows 11 ARM64, with the amd64 native host running through Windows emulation.
- Native SendInput wheel traversal covered 123 virtual rows and 4674 painted frames while 64 footer-growth ticks changed the physical extent.
- scrollTop advanced from 3759 to 10869; maximum reverse movement was 0 CSS px; blank frames were 0; every sampled viewport remained occupied.
- Final bottom distance was 0 px and the owner returned to tail-follow. The measured footer-bottom residual was -0.34375 CSS px, within the 1 CSS px acceptance bound.
- The original #9416 harness expected the pre-#9422 unbounded tools fixture (at least 400 rows). The final-head run used the dedicated geometry fixture because #9422 now pages that tools history; result thresholds for reverse movement, blank frames, occupancy, and final tail convergence were unchanged.

## Coordination

- #9421 is now the catalog baseline. Its atomic directory projection and this PR's display projection acknowledgement are independent layers; both full suites pass together.
- #9422 is now the navigation baseline. Conflict resolution preserves its data/paint commit and historyMutation contract alongside this PR's historyPrefixCount, history_rebase, epoch fence, and live-tail retention.
- #9416 remains open and conflicts with the current tree. It directly overlaps six #9409 scroll/diagnostic files and a simulated merge reports eight conflicts because it also predates #9422. It must rebase after #9422/#9409 and preserve the single-writer, navigation permit, stable rebase anchor, and pending-only measurement contracts rather than being mechanically merged.
- Related: #9366. The focused pending-only row-height fix is incorporated with source attribution and a co-author trailer for @Linearl.
- This PR also touches runtime/transcript owner files changed by #9385, #9248, #9187, #8735, and #8740; those branches should retain ordered ledger projection and exact-turn routing when rebased.

## Impact declarations

Cache-impact: medium - the upgrade intentionally adds one stable provider-visible finish tool, causing one prefix cache miss; subsequent tool and system-prompt bytes are deterministic.

Cache-guard: golden prefix_shape, provider_request, and tool_schemas fixtures plus focused internal/boot and internal/agent tests verify byte stability after the upgrade.

System-prompt-review: self-reviewed - no turn IDs, runtime state, diagnostics, or ledger data are injected into the system prompt; the two performance/replay commits do not alter provider-visible bytes.

Documentation-impact: updated - the English and Chinese tool-contract documents in the first four commits define finish outcomes, Ask separation, one repair round, and protocol_failed behavior.


SivanCola (commented):
Review outcome for final head `2d2180846`:

I found no remaining actionable code-correctness issue in the turn-ledger, checked-barrier, projection-ack, replay-reset, or frontend rebase paths after the follow-up fixes. The review also caught and fixed two CI-only seams: Windows ledger-admission failures now preserve `ErrTurnLedgerUnavailable` instead of being misclassified as provider disconnects, and the new ledger code is clean under golangci-lint v2.12.2, including cross-platform build tags.

The final head is merged with current `main-v2`. All reported checks are terminal green: root tests on Linux/macOS/Windows, targeted Windows control, race, coverage, lint, govulncheck, SDK matrix, Desktop Linux/macOS/Windows, WebKitGTK 4.1, native WebView2 startup, replay/browser tests, vendored WebView2, and Windows packaging. The local benchmark/capacity numbers and compatibility/privacy claims are documented in the PR description.

One release gate remains unresolved, so I am leaving this as a review comment rather than approval: the manual Windows WebView2 scroll acceptance has not demonstrated actual wheel/keyboard/thumb movement and <=1 CSS px anchor preservation on the final head. CI proves the native startup and state-machine contracts, but it does not prove real user-input viewport motion. Please reproduce that native-scroll gate successfully on final head, or identify and fix the Windows defect, before approval/merge.

SivanCola (commented):
Final self-review on dcce75411: no blocking findings.

The latest main-v2 merge preserves both #9422 navigation data/paint commits and this PR’s replay/history-rebase fencing. #9421’s catalog projection remains independent. #9416 still requires a deliberate rebase after these changes; a mechanical merge would conflict with the single-scroll-writer and pending-only measurement contracts.

All executed CI jobs in run 32900548853 passed (site was path-filtered). The final-head Parallels Windows 11/WebView2 native scroll run also passed with 0 CSS px reverse movement, 0 blank frames, and a -0.34375 CSS px footer residual. Local Go, race, Desktop, frontend transcript/typecheck/build, vet, repolint, cache-golden, and bundle gates are green.

This head is approve-ready from the code, compatibility, performance, and native UX perspectives. A separate maintainer review is still required for formal GitHub approval because the PR author cannot approve their own PR.

PR #9434, opened 2026-08-26
fix(mcp): expose structured tool results to models

ChaonanShen:
## Summary

- preserve existing MCP `content` text and image handling
- append compact `structuredContent` JSON when no equivalent JSON TextContent exists
- avoid duplicating dual-channel results while exposing structured-only and generic-text results to models
- cover content-only, duplicate JSON, structured-only, scalar, error, image, null, and malformed result cases

## Issues

MCP servers may place the semantic tool result in `structuredContent` while returning only a generic success message in `content`. Reasonix currently drops that structured value before it reaches the model.

## Verification

- `go test ./internal/plugin`
- `go test ./...`
- `go vet ./...`
- `gofmt -l internal/plugin/plugin.go internal/plugin/toolresult_structured_test.go`

## Documentation impact

Documentation-impact: none - this is MCP result compatibility with no new user-facing configuration or command surface.

## Cache impact

Cache-impact: low - tool schemas and the stable prompt prefix are unchanged; only per-call MCP result text gains missing structured data.
Cache-guard: focused `internal/plugin` structured-result tests plus `go test ./...`.
System-prompt-review: N/A

### DIFF
--- a/internal/plugin/plugin.go
+++ b/internal/plugin/plugin.go
@@ -6,6 +6,7 @@
 package plugin
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


Luo-root:
👍

wei-xiao123:
👍🏾

PR #9488, opened 2026-08-27
feat(cli): skip /clear confirmation in YOLO mode / YOLO 模式下跳过 /clear 确认

BuGlessRB:
## Summary

Make `/clear` non-interactive when YOLO is already active: it clears the session immediately instead of opening the confirmation overlay. Non-YOLO modes (`ask`/`auto`/`plan`) keep the existing confirmation unchanged.

Rationale: you rarely type `/clear` by accident, and even a mistaken `/clear` is at worst inconvenient — never catastrophic. YOLO is already an explicit commitment to skip confirmations, so asking again there is pure friction.

Implementation: the clear action moves into a shared `clearContext()` helper used by both the confirm handler and the new YOLO fast path.

## Documentation

- `docs/GUIDE.md` + `docs/GUIDE.zh-CN.md`: the `/clear` confirmation key table and the slash-command prose now note that YOLO mode clears immediately without asking.
- `docs/TOOL_APPROVAL_MODES.md` + `docs/TOOL_APPROVAL_MODES.zh-CN.md`: the Yolo section states the CLI `/clear` fast path.

## Tests

- `TestClearCommandRequiresConfirmationAndDiscardsSession` (existing) — non-YOLO still confirms, cancel keeps the session, confirm clears.
- `TestClearCommandInYOLOModeSkipsConfirmation` (new) — YOLO `/clear` clears immediately, no overlay, session rotated, only the system prompt remains.
- `go vet`, `go run ./tools/repolint` clean; focused `TestClearCommand*` suite passes.

## PR metadata

Documentation-impact: updated - GUIDE.md/GUIDE.zh-CN.md and TOOL_APPROVAL_MODES.md/TOOL_APPROVAL_MODES.zh-CN.md document the YOLO /clear fast path

Cache-impact: none - CLI-only /clear dispatch; provider-visible prefix unchanged

Cache-guard: existing internal/cli test suite

System-prompt-review: n/a (no config/memory/outputstyle/skill/boot change)


SivanCola (changes requested):
Thanks, this is a worthwhile and well-scoped CLI UX improvement. Before merging, please address the following correctness issues:

1. Preserve the clear-screen success/failure contract. `clearContext()` currently discards the result of `Controller.ClearSession()`, so `confirmClearContext()` always returns `tea.ClearScreen` even when the clear failed. The previous implementation returned no clear-screen command on an error. An extension veto, rotation error, or filesystem failure can therefore erase the visible screen while leaving the original session intact; in native-scrollback mode this can make the operation look successful. Please make the helper return a success result or command and emit `tea.ClearScreen` only after a successful clear. Add focused failure-path coverage.

2. Keep the YOLO fast path equivalent to a successful confirmed clear. The new `/clear` branch calls `clearContext()` but returns no command, whereas the existing success path emits `tea.ClearScreen`. This matters for inline/native-scrollback terminals, where committed transcript lines are outside the model's redraw surface. A successful YOLO clear should return the same clear-screen command. Please update `TestClearCommandInYOLOModeSkipsConfirmation`, which currently requires a nil command; also snapshot the session before asserting that its old transcript file was removed so that assertion is meaningful.

Please also rebase onto the latest `main-v2` and regenerate `tools/repolint/baseline.json`. The current head is conflicting there. The current `govulncheck` failure is from the old base's `golang.org/x/crypto v0.54.0`; current `main-v2` has `v0.55.0`, but the rebased head still needs a fresh full CI run. Please update the title to the repository's bilingual format as well.

Verified on the current head and a latest-base synthetic merge:

- `go test ./internal/cli -run 'TestClearCommand' -count=1` passed
- `go test -race ./internal/cli -run 'TestClearCommand' -count=1` passed
- `go test ./internal/cli -count=1` passed on the latest-base synthetic merge
- `go vet ./internal/cli` passed
- `go test ./internal/control -run 'TestClearSession' -count=1` passed
- `go run ./tools/repolint` passed on the current head
- `git diff --check origin/main-v2...origin/pr/9488` passed

Once these issues are addressed and CI is green on the updated head, I will consider merging this into `main-v2`.

SivanCola (approved):
Approved. The updated head preserves the /clear success/failure contract: both confirmed and YOLO clears return tea.ClearScreen only after ClearSession succeeds, while failures retain the visible transcript. The YOLO test now persists the old session before asserting deletion, and the active-turn regression covers failure atomicity.

The branch includes current main-v2, the repolint baseline is regenerated, and the title is bilingual.

Verified:
- all CI and CodeQL checks passed on 56e430739
- go test ./internal/cli -count=1
- go test -race ./internal/cli -run TestClearCommand -count=1
- go vet ./internal/cli
- go test ./internal/control -run TestClearSession -count=1
- go run ./tools/repolint

PR #9519, opened 2026-08-28
feat(desktop): add safe cross-platform shell support / 增加安全的跨平台 Shell 支持

mchenziyi:
## Summary

- Add a process-wide, cached shell inventory with deterministic Git for Windows discovery from explicit configuration, `PATH`, registry entries, and standard install locations.
- Resolve macOS shells as Bash -> native zsh -> native POSIX sh. Linux remains Bash-only for execution while still reporting zsh/sh capabilities.
- Model Git as an independent, lazily probed host capability. Git availability never implies Bash availability and never participates in shell selection.
- Make Windows repair manual-only: Settings links to the official Git for Windows download page and never launches `winget`, App Installer, MSI, or another package manager.
- Add copy-only Git repair guidance for macOS (`brew install git`) and allowlisted Linux distribution families. Missing Linux Bash retains separate Bash guidance. Reasonix never executes these commands and never adds `sudo`.
- Keep the configured preference, configured path, and currently bound controller unchanged during repair. Settings distinguishes the shell used by the current session from the shell that an explicit reload would resolve.
- Validate retained configured paths at runtime: known cross-interpreter executables, WSL launchers, and unresolved `git-bash.exe` MinTTY launchers are ignored without rewriting user configuration; intentional custom wrappers remain supported.
- Remove the previous cross-platform package-manager installers, implicit `prefer = "bash"` writes, and global `MSYS_NO_PATHCONV` injection.

### Windows

- Explicit compatible configured paths have highest priority; `git-bash.exe` is normalized to headless `bin/bash.exe` instead of starting MinTTY.
- WSL launchers under `%SystemRoot%` are excluded from both discovered candidates and explicit Bash paths.
- A retained path known to belong to another interpreter is ignored at runtime, so Bash and PowerShell argv contracts cannot be crossed when the preference changes.
- Git Bash is also discovered from Git executables on `PATH`, Git for Windows registry entries, and fixed standard install locations.
- A supported repair request always returns `manual_required` with `https://git-scm.com/download/win`. The frontend opens only that exact HTTPS host and path, falling back to the built-in URL when a bridge payload differs. `winget` presence is intentionally irrelevant: Reasonix does not probe or invoke it, so systems without `winget` receive the same safe official-download flow.
- No repair path requests administrator elevation, changes shell preference, or changes the live controller.
- Git is reported separately from the Git Bash interpreter.

### macOS

- Bash remains preferred when available.
- If Bash is unavailable, auto-detection selects zsh and then POSIX sh. No Bash installation is requested while a native fallback is usable.
- Git is probed independently. Only missing or unusable Git shows the copyable `brew install git` guidance.
- Reasonix never invokes Homebrew.

### Linux

- Bash remains the execution shell; zsh and POSIX sh are reported as capabilities but are not silent replacements.
- Missing Bash and missing Git have separate copy-only repair guidance selected from fixed `apt`, `dnf`, `pacman`, `zypper`, or `apk` command tables.
- Unknown distributions receive generic manual guidance without an untrusted command assembled from `/etc/os-release`.
- Reasonix never invokes a package manager or `sudo`.

## Compatibility

- The persisted configuration format and accepted shell preference values remain unchanged: `auto|bash|powershell|pwsh`.
- `ShellZsh` and `ShellSh` are appended internal runtime kinds; existing enum values are preserved.
- Shell/Git capability and repair-guidance fields are additive Wails fields. Older frontends ignore them, and newer frontends handle older backends omitting them.
- `InstallShellSupport` remains available for binding compatibility but returns a structured manual result; `CancelShellInstall` is an idempotent no-op because no installer can run.
- Prompt text and tool schemas are unchanged. Local shell execution metadata now accurately distinguishes zsh and sh.
- Repair never rewrites configuration, changes the current shell, or rebuilds the controller without the user's explicit reload.
- Changing preferences continues to preserve the configured path byte-for-byte. A known incompatible retained path is ignored only by runtime resolution, while unknown wrapper names remain valid.

## Verification

- Validated a clean merge tree against current `main-v2` (`4e6c0a068`).
- `go test ./... && go vet ./...` in a clean tracked checkout.
- `cd desktop && go test ./... && go vet ./...`.
- `cd desktop && golangci-lint run --timeout=5m ./...` with v2.12.2: 0 issues.
- `go test -race ./internal/boot ./internal/sandbox` and focused shell/terminal regressions repeated with `-count=20`.
- Windows test-binary cross-compilation for Desktop and sandbox packages on amd64 and arm64.
- Real Windows 11 ARM64 focused native tests for configured-path priority, matching capability projection, cross-interpreter rejection, explicit WSL exclusion, custom-wrapper compatibility, unresolved MinTTY launcher rejection, and WSL classification.
- Real Windows 11 ARM64 smoke test of the manual-only flow: no `winget` invocation, installer process, or UAC prompt.
- `go run ./tools/repolint`: clean with 1267 baselined findings. Aggregate limits are below current `main-v2` (`file-size` 107967 < 108531; `test-file-size` 69472 < 69481). Only the additive generated Wails `bridge.ts` and `types.ts` bindings retain necessary +29/+30 per-file growth.
- `node scripts/run-tests.mjs`: all 232 discovered frontend suites passed.
- `pnpm exec tsx src/__tests__/shell-support-install.test.tsx`: 26/26 passed.
- Remote tail suites and `pnpm test:performance` passed. The aggregate `pnpm test:remote` command still encounters one inherited static source-contract assertion in `remote-project-tree.test.tsx`; the tested files are byte-identical to current `main-v2`, and the implementation it expects now lives in `remote_tab_pending_selection.go`.
- `pnpm build`, including ESLint, TypeScript, Vite, and bundle budgets (initial gzip 449.8/449.9 KiB; initial raw 2422.4/2422.5 KiB).
- `git diff --check` and gofmt verification.

## Documentation impact

Documentation-impact: none - the Settings UI explains the capability and manual-repair behavior in product, while the persisted shell configuration contract remains unchanged.

## Cache impact

Cache-impact: low - prompt and tool schema bytes are unchanged; only the stable environment shell label can change when macOS selects zsh/sh or when an incompatible stale configured path is corrected to the interpreter actually bound.
Cache-guard: go test ./internal/boot ./internal/tool/builtin ./internal/sandbox
System-prompt-review: reviewed the provider-visible path; prompt and schema golden tests remain unchanged, and the edge-case environment label now reflects the interpreter actually bound.


SivanCola:
这个思路很好

PR #9672, opened 2026-08-31
fix(openai): accept provider-prefixed model refs in the DeepSeek vision SKU check / DeepSeek 视觉 SKU 判定容忍 provider 前缀（#9671）

Linearl:
**TL;DR**: 修复 DeepSeek 视觉 SKU 判定拒绝带 provider 前缀 model ref 的 bug（`deepseek/deepseek-v4-flash-vision-exp` → vision 被静默关闭、图片永远读不到）。#9671 根因部分。

## Summary

`IsOfficialDeepSeekVisionModel` matched `cfg.Model` as a **whole string**, so a multi-provider model ref like `deepseek/deepseek-v4-flash-vision-exp` failed the pinned-SKU check → `vision=false` → `image_url` was never embedded on official DeepSeek endpoints. The vision model could not see images **even for an explicit `@` image reference** — the gate effectively disabled the only official DeepSeek vision SKU (MiniMax-M3 was unaffected because its vision gate is not an exact model-name match; this contrast is what narrowed the root cause, see #9671).

Fix: match only the segment after the last `/` (the actual model id). Flash, Pro, prefixed non-SKU names, and future "vision"-containing names still fail the check, so the original intent — pinning image input to the one official vision SKU — is unchanged.

The same helper gates vision in 6 places (`config/deepseek_vision.go`, `config/vision.go`, `anthropic`, `responses`, `openai` ×2), so one change fixes all DeepSeek vision paths.

## Scope note

#9671 also carries a feature request (attached images without an `@` reference only feed the OCR summary). This PR addresses the root-cause bug only; the no-`@` injection design is better discussed on the issue.

## Issues fixes

Partially addresses #9671 (the DeepSeek vision gate root cause; the no-`@` injection feature remains open on the issue).

## Verification

- `go build ./...` — pass
- `go test ./internal/provider/openai/ ./internal/config/ -run 'TestIsOfficialDeepSeekVisionModel|...'` — pass (new table cases: prefixed SKU match, case/space combos, prefixed non-SKU rejection, deeper-path rejection)
- `go test ./internal/provider/...` — pass (covers the anthropic/responses call sites)

Cache-impact: none- provider-layer gate only; no cache format or version changes

Cache-guard: updated- go build ./... + internal/provider/... test suites

Documentation-impact: none- provider behavior fix; no user-facing docs

System-prompt-review: none- no prompt text changes

### DIFF
--- a/internal/provider/openai/host.go
+++ b/internal/provider/openai/host.go
@@ -43,9 +43,17 @@ func IsDeepSeek(baseURL string) bool {
 const OfficialDeepSeekVisionModel = "deepseek-v4-flash-vision-exp"
 
 // IsOfficialDeepSeekVisionModel reports whether model is the pinned official
-// DeepSeek vision SKU. Matching is case-insensitive and trims surrounding space.
+// DeepSeek vision SKU. Matching is case-insensitive and trims surrounding
+// space. Model refs may carry a provider prefix ("deepseek/deepseek-v4-flash-
+// vision-exp" from a multi-provider config), so only the segment after the
+// last "/" — the actual model id — participates in the SKU check (#9671).
+// Flash, Pro, and future names that merely contain "vision" still fail.
 func IsOfficialDeepSeekVisionModel(model string) bool {
-	return strings.EqualFold(strings.TrimSpace(model), OfficialDeepSeekVisionModel)
+	model = strings.TrimSpace(model)
+	if i := strings.LastIndex(model, "/"); i >= 0 {
+		model = model[i+1:]
+	}
+	return strings.EqualFold(model, OfficialDeepSeekVisionModel)
 }
 
 // OfficialDeepSeekAllowsVision reports whether this official DeepSeek endpoint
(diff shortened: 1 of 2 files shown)


qihanqihan-gif:
Field reproduction from a real Windows project confirms this is user-visible and high impact.

Environment:
- Reasonix Desktop v1.31.4 with an unrelated local action-preamble guard patch
- selected model ref: `deepseek/deepseek-v4-flash-vision-exp`
- provider catalog includes `deepseek-v4-flash-vision-exp`
- `vision_models = ["deepseek-v4-flash-vision-exp"]`

Observed:
1. A user image is attached/referenced in a session using the official vision SKU.
2. The UI/session identifies the executor as `deepseek/deepseek-v4-flash-vision-exp`.
3. The model reports that it cannot see the image. The same image can be analyzed through an external DeepSeek Vision script, so the file itself is valid.
4. Subagent `.meta.json` files also record the prefixed model ref, which makes this prefix form a normal runtime path rather than a hand-written edge case.

This is consistent with `IsOfficialDeepSeekVisionModel` comparing the full provider-qualified ref and silently disabling the image payload. I have not yet applied this PR locally, so this comment is a reproduction confirmation rather than a claim that the patch was field-tested.

Please consider merging/backporting this into the next Desktop release: selecting the official vision SKU through normal provider/model resolution can otherwise produce a model that is labelled as vision-capable but receives no pixels.

PR #9681, opened 2026-09-01
fix(studio): keep model search in settings scroll flow

### LINKED ISSUE #9648: [Bug]: 下滑时搜索框异常
### Version line

### Exact version

2.10.0

### What happened?

Image

### Steps to reproduce

往下滑就会出现

### OS / platform

Windows 11

### Relevant logs or output

```shell

```

jackie-cqz:
## Summary

- scope sticky model-search styles to popup menus
- use a dedicated non-sticky class for the settings model search
- add a regression test that prevents the settings page from reusing popup-menu search styling

## Issues

Fixes #9648

## Verification

- `pnpm test` — 33 files, 235 tests passed
- `pnpm typecheck`
- `pnpm build`
- `go vet ./...`
- repository lint passed

### DIFF
--- a/desktop/frontend-next/src/styles/app.css
+++ b/desktop/frontend-next/src/styles/app.css
@@ -339,14 +339,14 @@
 
   /* Sticky because the rows are the scrolling part: a field that scrolls away
      takes the query out of sight while its result is still on screen. */
-  .mfind {
+  .menu .mfind {
     position: sticky; top: -5px; z-index: 1; width: calc(100% + 10px);
     margin: -5px -5px 5px; padding: 9px 11px;
     border: 0; border-bottom: 1px solid var(--hair); border-radius: var(--r-md) var(--r-md) 0 0;
     background: var(--float); color: var(--text); font: var(--w-reg) 12.5px var(--ui);
   }
-  .mfind:focus { outline: none }
-  .mfind::placeholder { color: var(--ghost) }
+  .menu .mfind:focus { outline: none }
+  .menu .mfind::placeholder { color: var(--ghost) }
   /* What Enter takes while the field still holds focus. Deliberately the same
      weight as hover: it is a standing offer, not a selection. */
   .mi[data-lead] { background: var(--overlay) }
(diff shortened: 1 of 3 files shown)


esengine:
Thanks for tracking this down — the diagnosis is right. `.mfind` is doing double duty, and that is the cause of #9648.

We've landed a fix on `studio` that takes a wider cut, so I'm closing this PR. The reasoning, in case it's useful:

**1. There are three call sites, not two.** `ModelChoice.tsx:81` carries `.mfind` on the same wrapper `<div>` that `Models.tsx` does. This PR renames only the latter, so the collision survives in `AddProvider` / `EditConn` / `Completion`.

**2. `.mlist .mfind` becomes a dead rule.** That reset (`position: static; width: auto; margin: 0 0 8px; padding: 0; border: 0; background: none`) exists for exactly one reason: to undo the leaked sticky. Once the sticky rule is scoped to `.menu`, it has nothing left to undo. It is also the local patch written the last time this same collision surfaced — applied to `ModelChoice` and never to `Models`, which is precisely why only one of the two lists was ever visibly broken. Leaving it in place preserves the shape of the bug.

**3. The regression test asserts the wrong thing.** The defect lives in CSS; the test reads the class name out of rendered markup:

```js
expect(html).not.toContain('class="mfind"');
```

Change line 364 back to an unscoped `.mfind` and this test still passes while #9648 reproduces. It pins the edit, not the invariant.

**What landed instead:** the menu keeps `.mfind` (it is a bare `<input>`); both wrappers become `.msearch`; the sticky group is scoped to `.menu` as well, and the now-redundant `.mlist .mfind` reset is removed. The leak ran in both directions, which the scoping alone doesn't fix — the wrapper group's `margin-bottom: 12px` was also overriding the `5px` the menu's own rule declares, and `display: flex` was landing on the menu input.

Measured across all three call sites afterwards: the menu input keeps `position: sticky`, its `9px 11px` padding and `--float` fill, and regains its declared `5px` bottom margin; both wrappers resolve to `position: static`, no fill, no border — the flex row they were always meant to be.

Appreciate the report and the fix attempt; please do keep them coming.

PR #9746, opened 2026-09-03
fix(desktop): guard late transcript range commits / 防止延迟范围提交导致会话回跳

SivanCola:
## Summary

- Preserve the last logical transcript anchor through late WebView2/Virtuoso range commits that arrive after a reader transaction times out.
- Route delayed deliveries through the existing anchor-compensation writer and cancel the passive lease when a higher-priority owner takes over.
- Add a deterministic regression for the delayed range-replacement race.

## Issues

Field diagnostics from v1.36.0 on Windows showed repeated downward-reading jumps back to the same stale transcript offset after the reader transaction timed out. The viewport only appeared to recover after the user reached the tail. The trace contained no question-jump or click navigation; it matched a delayed virtual-range/measurement commit being accepted as the new resting anchor.

This is a narrow follow-up on current `main-v2`. Related open transcript PRs #9416 and #9673 cover broader or older scroll changes; this PR intentionally changes only the late-anchor ownership path and its regression coverage.

## Verification

- `pnpm --dir desktop/frontend test:transcript`
- `PLAYWRIGHT_BROWSERS_PATH=.pw-browsers REASONIX_TRANSCRIPT_READER_ITERATIONS=1 pnpm --dir desktop/frontend test:transcript-reader-browser` (Chromium and WebKit)
- `PLAYWRIGHT_BROWSERS_PATH=.pw-browsers pnpm --dir desktop/frontend test:transcript-browser`
- `git diff --check`
- PR CI: lint, desktop/macOS/Windows, Linux WebKit, macOS WKWebView, CodeQL, and metadata guards

The raw initial bundle measured `2469.347 KiB` after the passive lease (+`0.222 KiB`, +`0.009%`) and is guarded by a narrow `2469.4 KiB` ceiling; gzip and largest-chunk budgets are unchanged.

## Documentation impact

Documentation-impact: none - this is an internal desktop transcript-scroll reliability fix; it adds no user-facing setting, command, API, or workflow.

## Cache impact

Cache-impact: none - no provider-visible prompts, tool schemas, request serialization, memory, or skill metadata changed.
Cache-guard: not applicable; the patch is limited to frontend scroll ownership and deterministic UI tests.
System-prompt-review: N/A

### DIFF
--- a/desktop/frontend/scripts/check-bundle-budget.mjs
+++ b/desktop/frontend/scripts/check-bundle-budget.mjs
@@ -325,6 +325,9 @@ const rawInitialBytes = [...initialJS, ...initialCSS, ...appShellCSS]
 // Session takeover banners and #9703/#9711's provisional-selection handoff
 // combine with Sticky Context's pinned-file state at 2469.125 KiB raw on the
 // merged stable path. Retain only the next one-decimal ceiling.
-const rawInitialBudgetKiB = 2_469.2;
+// The passive reader-anchor lease for delayed WebView2 range commits measures
+// 2469.347 KiB raw (+0.222 KiB, +0.009%). Retain only the next one-decimal
+// ceiling; gzip and largest-chunk budgets remain unchanged.
+const rawInitialBudgetKiB = 2_469.4;
 assertBudget("initial raw JavaScript and CSS", rawInitialBytes, rawInitialBudgetKiB * 1024);
 assertBudget("largest initial JavaScript chunk raw", largestInitialJSRaw, 1_000 * 1024);
--- a/desktop/frontend/src/__tests__/transcript-anchor-compensation-race.test.tsx
+++ b/desktop/frontend/src/__tests__/transcript-anchor-compensation-race.test.tsx
@@ -305,6 +305,43 @@ check(
   "user scroll intent cancels a pending anchor compensation",
 );
 
+// WebView2 may commit a delayed Virtuoso range immediately after the active
+// reader transaction ends. Keep the reader's last logical anchor passive so
+// that commit cannot become a new (already-jumped) resting position.
+await act(async () => arbiter?.reset());
+scrollExtent = 20_000;
+scrollElement.scrollTop = 8_000;
+rowElement.getBoundingClientRect = () => rectAt(20 + (Number.parseFloat(
+  scrollElement.style.getPropertyValue("--transcript-reader-visual-offset"),
+) || 0));
+await act(async () => arbiter?.deliverScroll());
+await act(async () => arbiter?.setMode("manual", "test-passive-reader-anchor"));
+await act(async () => arbiter?.onWheelIntent({
+  ctrlKey: false,
+  deltaX: 0,
+  deltaY: 120,
+  target: scrollElement,
+} as React.WheelEvent<HTMLElement>));
+scrollElement.scrollTop = 8_120;
+await act(async () => arbiter?.deliverScroll());
+clockNow += 1_201;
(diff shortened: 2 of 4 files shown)


SivanCola (commented):
Review conclusion: no blocking findings; recommend merge after the required maintainer approval.

## 1. What this PR does

The PR fixes a Windows WebView2/Virtuoso ordering gap in the existing transcript scroll owner. When a reader transaction times out, a delayed virtual-range or measurement commit could arrive afterward and replace the last trustworthy logical anchor with the already-displaced viewport. The fix retains that anchor as a bounded passive lease, routes late scroll deliveries through the existing anchor-compensation writer, and cancels the lease on reset, new reader intent, selection, resize, navigation, recovery, or another offset owner. The added race coverage models a delayed range replacement and verifies that the original reader offset is restored.

The bundle-budget edit is a narrow ratchet from 2469.2 to 2469.4 KiB for the measured 2469.347 KiB raw payload; gzip and largest-chunk limits are unchanged. No provider, persistence, Wails, or public API contract changes are included.

## 2. Merge benefit

- Prevents the reported repeatable mid-transcript jump after a downward wheel gesture; the viewport no longer requires reaching the tail to recover.
- Preserves the user's logical reading anchor across late virtualized range commits while leaving explicit user/navigation owners authoritative.
- Makes the delayed-commit contract executable with deterministic coverage and keeps the test file within the repository's 800-line limit.

## 3. Regression and compatibility review

No introduced functional, security, persistence, or cross-session isolation issue was found. The passive lease is cleared by every higher-priority owner transition, so a later wheel, scrollbar drag, selection, resize, question jump, recovery, or surface reset cannot be overridden by the stale anchor. The exact customer's production transcript data and OS build were not available for replay; the bounded fixture and native WebView2 smoke provide the closest reproducible coverage and leave no merge-blocking gap.

## 4. Concurrency, performance, and security

- `VERIFIED SAFE` — late `SCROLL_DELIVERED` events compare against the retained anchor instead of sampling displaced geometry; owner transitions clear the lease before their writes; all actual writes still pass through the existing generation-aware writer gateway. The deterministic transcript race suite covers delayed commit ordering and one bounded correction.
- `VERIFIED SAFE` — the passive path performs one mounted-row lookup and schedules the existing two-frame compensation only for at least 96px drift. It adds no timers, network I/O, file I/O, unbounded loop, dependency, or provider-visible work.
- `VERIFIED SAFE` — the diff adds no command execution, auth/token handling, persistence, telemetry payload, or external data path.

The exact native input sequence from the reporter's machine remains `UNVERIFIED` because the supplied recording is evidence rather than a runnable host; the Windows WebView2, macOS WKWebView, Linux WebKitGTK, browser, and race gates all passed on the PR head.

## 5. Merge readiness and integration

- Current base/head: `d9cd71313839a00f25d0dbf46f94c6cf3bc7a2d2` → `ef1e3841486abdae00bb33f2ff2999510e03e465`; merge-base equals current `main-v2`.
- `MERGEABLE/CLEAN`; no unresolved review threads or prior reviews.
- Required checks passed: lint, desktop, desktop-macos, desktop-windows, desktop-linux-webkit41, macos-wkwebview, CodeQL (Go/JS/Actions), cache/docs/label/changes, SDK and platform tests. Race/coverage/site/govulncheck are explicitly skipped by this PR workflow.
- Open #9416 and #9673 are broader/older conflicting branches; this two-file runtime fix plus its budget/test adjustment is based directly on current `main-v2` and has no dependency on them.

Cache review: `N/A` beyond the passing guard; no provider-visible prompt, tool schema, serialization, memory, skill, or compaction bytes changed.

PR #9762, opened 2026-09-03
fix(provider,agent): replay DeepSeek thinking on all turns and self-heal stale-thinking 400 / DeepSeek thinking 全轮回传与 400 自愈修复

SivanCola:
## Summary

- Replay DeepSeek thinking blocks for **every** assistant turn that carries stored reasoning — DeepSeek's thinking mode requires `reasoning_content` back on all historical turns whenever the request declares tools, not only tool-call turns ([official doc](https://api-docs.deepseek.com/guides/thinking_mode)).
- Keep `thinking` first when merged assistant messages combine a projection-degraded plain-text turn with a following healthy turn (`[thinking, text, …]`, never `[text, thinking, …]`).
- Add a one-shot catch-and-repair for the `content[].thinking … must be passed back` HTTP 400: strip all assistant reasoning from the provider-visible projection (vendor-documented self-heal), retry exactly once, and pin the conversation to the strong projection so later rounds do not pay another 400. Canonical session history is never modified; unrelated 400s stay terminal.

Root cause of the report: histories containing plain question-answer assistant turns (reasoning stored, no tool calls) were replayed without their thinking blocks, so the next tool-carrying request was rejected, and the turn stopped with no recovery.

## Issues

Refs #9750

## Verification

- `go build ./...`
- `go test -count=1 ./internal/provider/... ./internal/agent/...`
- `go test -race -count=1 ./internal/agent/ -run 'Replay|Reasoning|Retry'`
- New deterministic coverage:
  - `internal/provider/reasoning_replay_error_test.go` — matcher accepts only the exact 400 thinking-replay shape (10 cases: non-matching 400/401/422/empty body all rejected).
  - `internal/provider/anthropic/replay_projection_test.go` — merged assistant turns keep thinking first after projection.
  - `internal/agent/reasoning_replay_recovery_test.go` — repair retry succeeds with only `Messages` changed in the frozen request; a second 400 terminates as before; retry budget is exactly one; unrelated 400 untouched.
  - `internal/agent/retry_e2e_test.go` — end-to-end against a stub Anthropic endpoint: first request 400 → repaired retry succeeds → later rounds keep the strong projection.
- Two pre-existing assertions were updated because they encoded the old "tool-call turns only" replay decision that conflicts with DeepSeek's documented contract (`anthropic_test.go`): reasoning on plain-text turns is now replayed, matching the long-standing "thinking disabled after tool call" behavior.

## Documentation impact

Documentation-impact: updated - `docs/REASONING_PROVIDERS.md` now describes replaying reasoning for every assistant turn carrying it.

## Cache impact

Cache-impact: medium - provider-visible history bytes change once for in-flight sessions (stored thinking blocks are now replayed on plain-text assistant turns; required by the DeepSeek API contract). Steady-state behavior stays append-only; healthy histories keep their backing slices (zero-copy projection unchanged). The 400-repair strong projection only activates on conversations that already hit the 400, i.e. ones that were failing anyway.
Cache-guard: `go test -count=1 ./internal/provider/ -run 'ReplaySafe|ReasoningStripped'` plus `internal/agent/retry_e2e_test.go` byte-equality assertion on the frozen request (only `Messages` may change).
System-prompt-review: N/A - system prompt, memory prefix, output style, and skill index are untouched; only assistant-history reasoning replay on the DeepSeek Anthropic endpoint changes.

### DIFF
--- a/docs/REASONING_PROVIDERS.md
+++ b/docs/REASONING_PROVIDERS.md
@@ -66,7 +66,9 @@ New official entries use this native Messages API path and enable provider-side
 `web_search`; existing explicit providers, including legacy
 `deepseek-anthropic` entries, keep their configured protocol. Reasonix emits
 `thinking.type=enabled|disabled` with `output_config.effort`, replays unsigned
-DeepSeek thinking blocks from historical tool-call turns, omits unsupported
+DeepSeek thinking blocks from every historical assistant turn that carries
+reasoning (tool-call turn or not, as DeepSeek's thinking mode requires when the
+request declares tools), omits unsupported
 images, and relies on DeepSeek's automatic prefix cache instead of ignored
 `cache_control` markers.
 
--- a/internal/agent/reasoning_replay.go
+++ b/internal/agent/reasoning_replay.go
@@ -102,6 +102,33 @@ func (a *Agent) finishReasoningReplayRetry(retry streamedTurn, sink *deferredStr
 	return retry
 }
 
+// finishReasoningReplayOverflow terminates an attempt whose required reasoning
+// was truncated by the client limit: audit, finalize usage, and hand the turn
+// to the unreplayable-reasoning policy.
+func (a *Agent) finishReasoningReplayOverflow(result streamedTurn, sink *deferredStreamSink, issue ReasoningReplayFailure, billable *provider.Usage, attemptID string, attempt int) streamedTurn {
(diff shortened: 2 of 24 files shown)


SivanCola:
Superseded by #9776, which has been merged into main-v2 and contains the complete reasoning replay, OpenAI-compatible support, stale-history 400 self-heal, cutoff/anchor protection, raw reasoning preservation, and synchronized documentation. Closing this older overlapping PR.

PR #9863, opened 2026-09-06
Make tool argument failures recoverable / 让工具参数错误可恢复

SivanCola:
## Problem

Malformed tool arguments were reported as ordinary schema failures at first, then escalated by a schema-specific loop guard into a blocked, permission-style state. This made recoverable calls appear locked, especially when a provider emitted an extra top-level `arguments` wrapper.

## Root cause

Argument validation owned duplicate counters, automatic inspect attachment, and a third-failure block path in addition to the shared storm breaker.

## Changes

- Keep strict pre-dispatch validation and return actionable unexecuted errors with stable diagnostic signatures.
- Add conservative, value-free detection of one redundant `arguments` wrapper without changing supplied arguments.
- Diagnose repairable `use_capability` envelope errors without masking unavailable or authorization failures.
- Remove schema-specific loop state, inspect unlock behavior, and retired metric writes.
- Let the existing storm breaker provide parameter-correction soft convergence while preserving permission and hook blockers.
- Add regression, concurrency, cache-prefix, extension, proxy, MCP, skill, and budget coverage.
- Document the contract in English and Chinese.

## Verification

- `go test ./...`
- `go test -race ./internal/agent ./internal/tool ./internal/capability`
- `go vet ./...`
- Focused boot/provider/cache-prefix tests

No provider-visible tool schema, system prompt, provider protocol, persisted format, or AMD adapter was changed. AMD live-model validation was not run because no live credential test was used.

Documentation-impact: updated - added bilingual tool-contract guidance for validation feedback, capability recovery, soft convergence, compatibility, and cache impact.


SivanCola (commented):
## Review: hold pending required checks

The implementation is technically sound and addresses the shared owner of the failure: ordinary argument validation is now an unexecuted, recoverable tool error, while the existing storm breaker remains the only repeated-failure convergence mechanism. The diff also preserves the permission boundary, extension re-validation, MCP skip behavior, legacy null compatibility, and provider-visible schema/prefix bytes.

Verified locally:

- `go test ./...`
- `go test -race ./internal/agent ./internal/tool ./internal/capability`
- `go vet ./...`
- focused boot/provider/cache-prefix and recovery tests

Security and concurrency review: no privilege expansion, argument rewriting, secret/value echo, extra model request, network access, or new shared state was introduced. The deleted schema counters remove duplicate mutable state; the deterministic concurrent recovery test and race suite pass.

Cache review: system prompt, tool schemas/order, provider protocol, and persisted formats are unchanged. The new failure feedback is appended to tool results and consumes some context tokens by design; the provider-prefix regression passes.

Related open work exists in #9581 (truncated streamed arguments) and #8553 (skill capability argument normalization). They address adjacent producer/skill paths; this PR owns host-side validation and recovery and should be evaluated with those changes for integration ordering.

I am holding approval until the current head `29dabcf0813f5d30322342276a57b1a8280463e8` finishes the required CI jobs. The earlier docs-impact failure was metadata-only and has been corrected with the required `Documentation-impact:` declaration. No unresolved review threads are present.

PR #9943, opened 2026-09-08
ci: shorten desktop checks and reuse signed release artifacts / 缩短桌面检查并复用已签名发版产物

SivanCola:
## Summary

Desktop PR checks currently repeat frontend suites in a serial job, and stable releases build and sign the entire desktop matrix again after the same matrix passes preflight.

- Build Linux frontend bindings/assets once and share them with independent frontend, browser and Go/race jobs. Preserve the existing required `desktop` aggregate check, including failure, cancellation and path-skip handling.
- Expand the existing frontend test entrypoints into one coverage-preserving plan, deduplicate 117 repeated TypeScript invocations and run at most two test processes concurrently. Keep the performance benchmark serial and retain all browser/native platform checks.
- Publish the same signed four-platform artifacts produced by successful stable preflight. Verify candidate/control SHAs, version, tag, channel, signing fingerprint, run/attempt identity and file digests before publication. Preserve signing, native smoke, cache/docs and release authorization gates; cache pnpm downloads.
- Keep all three independent memory processes and all 896 round trips per process. Remove only the explicit soak fixture's artificial 1.5-second hydration latency, retain asynchronous hydration and default delayed-hydration coverage, stabilize the resting baseline, and record phase timings. Version the screening protocol so old reports cannot be combined.

## Verification

- 378 frontend test commands passed before adding the focused hydration regression; the added hydration regression passed separately (final plan: 379 commands).
- 46 release transport, workflow gate, test coverage plan and memory protocol/timing tests passed, including after integration with current main-v2.
- Release workflow contract tests, actionlint (excluding the repository's existing custom ARM runner label), repository lint, production frontend build and bundle budgets passed.
- Real Chromium app lifecycle tests passed with default delayed hydration preserved.
- Three independent local Chromium processes each completed all 896 memory round trips in approximately 14m28s; every shard and the aggregate passed. Offline heap-retainer attribution remains pending. This local macOS measurement is not a direct Linux Actions before/after comparison.

GitHub PR checks will provide the current-head Linux/native integration evidence. A production signed release has not been executed. Removing the duplicate desktop matrix does not imply that the complete stable release duration halves: npm publication can remain the critical path.

## Documentation impact

Documentation-impact: none - No user-facing product contract changes. Bilingual contributor documentation in `.github/CI_PERFORMANCE*.md` explains the execution, artifact and memory protocol contracts without changing the embedded product documentation corpus.

## Cache impact

Cache-impact: none - No system prompt, tool schema/order, provider serialization, memory prefix or skill-index changes. The frontend bridge change is restricted to explicit mock soak URLs.
Cache-guard: Existing release cache guards remain mandatory; focused hydration tests verify that default fixture behavior is unchanged.
System-prompt-review: N/A


SivanCola:
Fixed the signing contract regression in e08d6dad273b89db42ce18a65a9fddc65cc1581d.

The existing desktop Go test still required the former build-only publish expression. It now checks successful-build or explicitly verified orchestrator reuse, plus mandatory signing/cache success and smoke/preflight publication exclusions. Signing-order assertions remain intact; the executable workflow tests cover successful publication and rejected failure/cancellation/unverified reuse paths.

All tests in `desktop/windows_signing_packaging_test.go` and the 22 workflow/artifact contract tests passed locally. Current-head CI is running; merge remains gated on its results. No release has been published by this PR.

Follow-up local validation: `go test ./...` in the desktop module passed. The root-module sweep had one `TestGoalTurnRunsPastTheOldRoundCeiling` timeout; that unchanged test passed three focused repetitions, and the complete `internal/control` package passed on rerun. The initial root sweep is therefore recorded as a failure followed by successful focused/package revalidation, not as an uninterrupted full-suite pass.

PR #9981, opened 2026-09-09
fix(agent): bill the turn phases that capability_phases reports / 修复 capability_phases 六个耗时桶恒为零 (#5178)

boscocp:
**简述**：`run --metrics` 导出的 `capability_phases` 有六个耗时桶，实际运行中全部为零，其中五个永远不可能非零——`checking`/`verifying` 无人发出，`subagent`/`user`/`compact` 无人映射，且 `beginRunTurn` 会整体重置 `a.turn` 从而丢弃未结束的阶段。本 PR 补齐阶段发出与计费，使指标真实反映耗时。Refs #5178。

## Summary

`run --metrics` exports a `capability_phases` object with six duration buckets.
On any real run **all six are zero**, and five of them cannot be anything else.

`emitTurnPhase` (`internal/agent/turn_phase.go`) bills the duration of the
*previous* phase and then opens the new one, so a span is only ever accounted
for when something closes it. Today there are exactly two emitters:

- `run_loop.go` — `working`, once per turn, in `beginRunTurn`
- `capability_gate.go` — `reviewing`, only when the delivery review gate fires

From that:

- `phaseAuditName` maps `checking`/`verifying` to the `tool` bucket, but neither
  phase is emitted anywhere, so `ToolExecMs` can never be non-zero.
- Nothing maps to `subagent`/`user`/`compact`, and `RecordPhaseMs` is their only
  writer, so those three buckets are dead code.
- `beginRunTurn` resets `a.turn` wholesale, discarding the open phase. On an
  ordinary turn nothing closes the last span, so **nothing is recorded at all**.
- When the review gate does fire, `reviewing` closes `working` and credits the
  entire turn up to the gate as `provider` — a misleading number rather than
  provider wait. `reviewing` itself is never closed, so `ReviewMs` stays zero.

### This is a regression, and the culprit is identifiable

`5e03e03ea` emitted `TurnPhaseVerifying` from `appendClosedLoopReadiness` in
`internal/agent/final_readiness.go`. `314beed8f` ("replace task-mode routing
with fact-driven execution", #8930) deleted that whole function during a
refactor and took the emission with it, without replacing it. The constant in
`event.go`, the labels in three locales, the TUI branch in `chat_tui.go` and the
`phaseAuditName` mapping were all left orphaned.

## Changes

**`internal/agent/turn_phase.go`** — the billing body moves into
`recordOpenPhase(now)` so `emitTurnPhase` keeps its exact semantics, plus
`closeTurnPhase()`, which bills the open phase without opening another one. That
is the load-bearing piece: without it a turn's tail span is lost when
`beginRunTurn` resets the turn state.

**`internal/agent/run_loop.go`** — three insertions:

- `handleToolRound`: `checking` immediately before `executeBatch`, `working`
  immediately after it returns. The pair is what makes the two buckets mean
  their names — the round's wait is billed to the provider, the batch to tools.
  The second emission sits *before* the `batch.err` check on purpose, so a
  failed batch still bills its tool span.
- `handleFinalResponse`: `verifying` before `finalReadinessCheckFor()`, which
  restores the emission #8930 dropped, then `working` immediately after it. It
  goes at the caller rather than inside the function because `ReadinessResult`
  runs the same check for the host, outside any turn — emitting inside would
  publish a phase with no turn. The reopening is load-bearing: four of that
  function's paths `return true` to continue the loop, which starts a fresh
  provider round, and leaving the tool-billed `verifying` phase open across it
  would bill model wait as tool time.
- `closeTurnPhase()` on both clean endings — the text final answer in
  `handleFinalResponse`, and a successful terminal tool (`submit_plan`) in
  `handleToolRound`.

### Known limitation

A turn that ends through an error or a pause still drops its last open span.
Covering those needs a defer, and the only two places to hang one are
`runToolLoop`, which has no line budget left, and `Agent.Run`, which is
system-prompt-sensitive. The buckets are a lower bound on the turn rather than a
full partition of it, and `docs/CLI.md` says so rather than overclaiming.

Nothing is added to `runToolLoop`: it is 122 lines against repolint's 120-line
ceiling and the file's recorded `function-size: 2` is entirely its own, so a
single added line there would have to widen the baseline. `handleToolRound`
(93 -> 97) and `handleFinalResponse` (113 -> 118) both had headroom, and
`repolint` is clean with the baseline untouched.

**Out of scope, deliberately:** `SubagentWaitMs`, `UserWaitMs` and `CompactMs`
still have no emitter. Wiring them means deciding where those phases begin and
end, which is new behavior rather than this regression. Happy to follow up.

### Observable effect

The CLI spinner and the desktop composer now show `checking` while a tool batch
runs and `verifying` during the readiness check, instead of staying on
`working`. Desktop result cards are driven only by actual running verification tools;
`checking` and `verifying` phase events update execution status without creating
verification evidence. The frontend reducer preserves a real check card across
phase changes and removes its running state when the check finishes.

A regression test in `completion-summary-presentation.test.ts` covers all four
phase names, stable result identity, and the actual dispatch/progress/result
sequence of a verification tool. The test failed before the fix. Browser DOM
verification also confirmed that generic phases create no result card while
real verification retains its card until completion. Frontend build and bundle
budgets, completion UI tests, Go vet, lint, and builtin/boot tests passed locally.

## Issues

Refs #8930 — the refactor that dropped the `verifying` emission.

Refs #5178 — that P0 asks "where did those 87 seconds go, and which tool call is
the bottleneck (network round-trip vs local computation)", and notes the data
source is per-tool stats `internal/agent` already has. The provider-wait vs
tool-exec split is exactly that attribution, and one half of it was structurally
zero. This is a prerequisite, not the panel that issue asks for, so Refs rather
than Fixes.

## Verification

- New `TestToolRoundAlternatesProviderAndToolPhases` asserts the full sequence
  `[working, checking, working, verifying, working]` for a turn with one tool
  call. On the unpatched tree it reports `phases = [working]`. It also asserts
  the accounting invariant that no tool-billed phase is ever left open at the
  end of the sequence, which is what stops a provider round from being billed
  as tool time.
- New `TestToolRoundBillsPhaseDurations` runs a `fakeTool` with a 5ms `delay`
  (the field already exists) and asserts `Phases.ToolExecMs > 0`. `RecordPhaseMs`
  drops anything under a millisecond, hence the deliberate delay. On the
  unpatched tree it reports `ToolExecMs = 0`.
- New `internal/boot/effect_turn_phase_test.go` is the effect test REASONIX.md
  asks for: a scripted provider that sleeps 10ms, the real `boot.Build`
  assembly, and the audit read back through the same path the CLI uses
  (`ctrl.Executor().CapabilityAudit().Snapshot().Phases`). Asserts
  `ProviderWaitMs > 0`; on the unpatched tree it prints all six buckets at zero.
  It deliberately does not assert `ToolExecMs`, which would need a real slow
  tool, and `internal/boot` runs in the scoped Windows suites.
- `gofmt -l .` clean, `go vet ./...` clean, `go build ./...` clean.
- `go run ./tools/repolint`: clean, baseline unchanged.
- `golangci-lint run ./...` at the pinned `v2.12.2`: 0 issues. Needed
  `GOTOOLCHAIN=go1.26.6` locally, since on Go 1.27 the bundled staticcheck
  panics in `buildir` on the stdlib `internal/poll` — unrelated to this diff.
- `go test ./internal/agent/ ./internal/capability/ ./internal/cli/ ./internal/boot/`: ok.
- `go test -race -run 'TurnPhase|Phase' ./internal/agent/`: ok.
- `REASONIX_RELEASE_CACHE_GUARD=1 go test -run 'TestCacheHitPrefixStable|TestCacheHitClimbsWithoutCompaction' ./internal/agent/`: ok.
- `go test ./internal/tool/builtin/ ./internal/boot/`: ok.
- `make lint` also fails on `scripts/check-wails-pin.test.sh` here because this
  machine has no `node`; the identical failure reproduces on unmodified
  `main-v2`, so it is an environment gap, not this diff.

## Documentation impact

Documentation-impact: updated - docs/CLI.md and docs/CLI.zh-CN.md gain a "Turn
phases" section next to the metrics/trajectory docs, listing each phase, when it
is emitted and which `capability_phases` bucket it bills. Nothing documented the
turn_phase contract or these buckets before. The Chinese mirror is a best-effort
translation and welcome to correction.

## Cache impact

Cache-impact: none - the provider-visible prefix is byte-identical. This only
adds frontend-sink events and host-side timing; no prompt text, tool schema,
memory or config surface changes.
Cache-guard: REASONIX_RELEASE_CACHE_GUARD=1 go test -run 'TestCacheHitPrefixStable|TestCacheHitClimbsWithoutCompaction' ./internal/agent/
System-prompt-review: requested from @SivanCola and @esengine (CODEOWNERS), who
are auto-assigned on this PR. No system-prompt surface is touched: the only
internal/boot change is the new effect test file, and the guard fires solely
because check-cache-impact.sh does not exempt *_test.go the way
check-docs-impact.sh does. The effect test has to live in internal/boot because
REASONIX.md requires asserting through the real boot.Build assembly, and
importing boot from a lower package would break the layering rule.

### DIFF
--- a/desktop/frontend/src/__tests__/completion-summary-presentation.test.ts
+++ b/desktop/frontend/src/__tests__/completion-summary-presentation.test.ts
@@ -46,7 +46,7 @@ const answer: Item = { kind: "assistant", id: "a0", text: "done", reasoning: "",
 const outside = partitionTurnItems([history!, answer]).flatMap(p => p.outsideItems);
 assert.deepEqual(outside.map(i => i.id), ["a0", "history"], "sidecar placement preserves a result footer");
 for (const phase of ["checking", "verifying", "working", "reviewing"]) {
-  const plain = { ...initialState, items: [user], seq: 1 };
+  const plain: State = { ...initialState, items: [user], seq: 1 };
   const next = reducer(plain, { type: "event", e: { kind: "turn_phase", phase } });
   assert.equal(next.turnPhase, phase);
   assert.equal(next.completionSummary?.checking ?? false, false, `${phase} is not verification evidence`);
@@ -58,11 +58,11 @@ for (const phase of ["checking", "verifying", "working", "reviewing"]) {
   assert.equal(active.completionSummary?.checking, true, `${phase} preserves real running checks`);
   assert.equal(active.items.find(i => i.kind === "notice")!.id, resultId);
 }
-let live = reducer({ ...initialState, items: [user] }, { type: "event", e: { kind: "tool_dispatch", tool: { id: "live", name: "bash", args: '{"command":"go test ./..."}' } } });
-live = reducer(live, { type: "event", e: { kind: "tool_progress", tool: { id: "live", name: "bash", verifying: true, output: "running" } } });
+let live = reducer({ ...initialState, items: [user] }, { type: "event", e: { kind: "tool_dispatch", tool: { id: "live", name: "bash", readOnly: false, args: '{"command":"go test ./..."}' } } });
+live = reducer(live, { type: "event", e: { kind: "tool_progress", tool: { id: "live", name: "bash", readOnly: false, verifying: true, output: "running" } } });
 assert.equal(live.completionSummary?.checking, true);
 live = reducer(live, { type: "event", e: { kind: "turn_phase", phase: "working" } });
 assert.equal(live.completionSummary?.checking, true);
-live = reducer(live, { type: "event", e: { kind: "tool_result", tool: { id: "live", name: "bash", output: "PASS" } } });
+live = reducer(live, { type: "event", e: { kind: "tool_result", tool: { id: "live", name: "bash", readOnly: false, output: "PASS" } } });
 assert.equal(live.completionSummary?.checking, false);
 console.log("turn result truth, stable identity, concurrent checks, phases and history passed");
--- a/desktop/frontend/src/__tests__/completion-summary-presentation.test.ts
+++ b/desktop/frontend/src/__tests__/completion-summary-presentation.test.ts
@@ -3,7 +3,7 @@ import { completionSummaryPresentation, normalizeCompletionSummary } from "../li
 import { mergeTurnResult, normalizeTurnChanges, turnChangeText, turnCheckState } from "../lib/turnResult";
 import { historicalResultNotice, withTurnResult, withRunningChecks } from "../lib/completionResultState";
 import { partitionTurnItems } from "../lib/transcriptRows";
-import { initialState, type State, type Item } from "../lib/useController";
+import { initialState, reducer, type State, type Item } from "../lib/useController";
 import { t } from "../lib/i18n";
 import type { TurnChanges } from "../lib/types";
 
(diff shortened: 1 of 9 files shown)


SivanCola:
Rebased this onto `main-v2` (`dd92f1664`) and pushed the result. The branch was 204 commits behind, and after the merge `handleFinalResponse` landed at 125 lines against repolint's 120-line ceiling, so it could not go green as pushed.

- `internal/agent/readiness_gate.go`: the whole `readiness.reason != ""` decision moved into `Agent.handleReadinessGap`, next to `readinessPauseActive`. `handleFinalResponse` is 104 lines now and no baseline entry was widened. Behavior is unchanged — empty reason falls through, a policy that ends normally still records the allowed-readiness audit and keeps going, a pausing policy still returns the same `FinalReadinessError`.
- `docs/CLI.md` / `docs/CLI.zh-CN.md`: the `reviewing` row no longer claims `ReviewMs`. `reviewing` is published at run exit, after the turn's phase clock has closed, so the bucket stays zero until an in-turn emitter exists; the row now sits with the other emitter-less buckets. The section also records that `working` is emitted after the final-readiness check as well, and that an approval prompt inside a tool batch bills to `ToolExecMs`.

Pushed as `61fea90cb` with `--force-with-lease`: your commit is preserved with authorship intact, rebased onto current `main-v2`. Verification on the pushed tree: repolint clean (1215 baselined findings, none new), `gofmt`/`go vet` clean, `go build ./...` clean, `go test ./internal/agent/` and `./internal/boot/` green, focused phase tests green under `-race`. CI, CodeQL and App memory screening are approved and running now.

PR #10007, opened 2026-09-09
chore: remove unreachable UI modules, dead locale keys and migration leftovers / 清理无引用前端模块、死语言键与迁移残留

SivanCola:
## Problem

`main-v2` carries code that nothing reaches. A dead-code audit across the Go kernel, the desktop host contract and the frontend found a small set of files whose only remaining references are a comment, a CSS class name, or a test that pins the very thing being removed.

## Fix

Frontend, proven unreachable by an import graph over all tracked sources:

- `components/InlineDiff.tsx`, `components/OnboardingOverlay.tsx`, `components/SearchFootnotes.tsx`, `components/ScrollDiagnosticPanel.tsx` (its `.css` stays — `FrontendDiagnosticsPanel.tsx` imports it), `lib/displayMode.ts`, `lib/providerPresetDescriptions.ts`, `assets/logo.svg`.
- The 330 CSS lines only those components used (`.onboarding__*`, `.inline-diff*`) and the two test assertions that pinned them (`diff-rendering.test.ts`, `theme-pack.test.ts`). The live `.diff__*` rules and the `--diff-row-bg` / `--tp-code-*` contracts are untouched.
- 537 locale keys with zero occurrences in any tracked file outside the three dictionaries. 54 dynamic-prefix call sites (`` t(`remote.status.${state}`) ``, `"prefix." + id`) were extracted and excluded, so keys reached only at runtime are kept. `settings.tab.providers` / `settings.tab.capabilities` are the two keys that survived that filter.

Repository leftovers:

- `docs/production_checklist.md` and its `.zh-CN` twin — frozen at v5.9.9 / PR #5217, references `system.StableSystemContract()` and `internal/system`, neither of which exists, and `docs/embed.go` ships every top-level `docs/*.md` into the binary as the current product corpus.
- `tools/write_heartbeat/` — a one-shot script that appends two hard-coded demo tasks; no build, CI or script path invokes it.
- `prototypes/electron-browser/` — 30 files whose behaviour was promoted into `desktop/electron/src/main/browser/`; its README still says "Production Desktop remains Wails".

## Verification

- `tsc --noEmit` and `tsc --noEmit -p tsconfig.test.json`: clean. `DictKey = keyof typeof en` plus `Record<DictKey, string>` on the zh/zh-TW dictionaries means any surviving static reference to a deleted key is a compile error.
- `pnpm test`: green.
- Focused suites: `diff-rendering` (23), `theme-pack` (305), `search-sources` (31), `frontend-diagnostics`, `transcript-scroll-diagnostics`.
- `pnpm check:css` (syntax + z-index tokens + theme-token contract over 21 CSS files / 317 tokens), `pnpm check:app-layers`, `check-desktop-host-boundary`: pass.
- `go build ./...`, `go vet ./...`, `go test ./internal/productdocs/ ./internal/boot/ ./internal/tool/builtin/`: pass.
- `make lint`: golangci-lint `0 issues` + `repolint: clean`. Note: running it in a checkout that still holds untracked `wt*/` worktree directories fails, because repolint walks the filesystem and counts those copies — the same change is clean in a fresh worktree.
- `scripts/check-cache-impact.sh` reports no cache-sensitive prompt/tool files changed.

## Not in this PR

Follow-up batches, each needing its own verification: the Go-side dead symbols and test-only helpers, the 113 host commands with no frontend call site, the remaining heuristic dead CSS, and the stale Wails wording across `docs/*.md`.

Documentation-impact: updated - deleted `docs/production_checklist{,.zh-CN}.md` (stale since v5.9.9, embedded into the binary) and the promoted `prototypes/electron-browser/` docs; no remaining doc references either, and the surviving `docs/*.md` links still resolve.


SivanCola:
`test (windows-latest)` fails on `TestBuildRunsCleanupPendingDespiteSafeModeEnv`:

```
internal/boot/config_home_test.go:30: close shared history catalog: context deadline exceeded
```

This is a pre-existing Windows flake in the `internal/boot` history-catalog helper (`closeBootTestHistoryCatalog` allows 5s), not a regression from this change:

- this PR contains no Go files; `internal/boot` is untouched
- `test (ubuntu-latest)` and `test (macos-latest)` pass on the same commit
- main-v2 CI is red on the same job already: run 34333648350 (merge of #9988, before this PR) shows the identical `close shared history catalog: context deadline exceeded` signature plus 8 other `--- FAIL` tests across boot/agent/bot/config/control, and run 34324106049 fails the same job. 3 of the last 4 completed main-v2 runs failed on it.

Not re-running the job: the signature reproduces on main-v2, so a green rerun would hide it rather than fix it.

PR #10022, opened 2026-09-09
test(desktop): move test-only helpers out of the shipped binary / 把仅测试使用的 helper 移出生产代码

SivanCola:
## Problem

Five production files in the desktop module carry helpers that only the test suite calls, so they compile into the app binary and read as production API:

- `sessions.go`: `setSessionTitle`, `deleteSessionFile`
- `heartbeat_store.go`: `HeartbeatEngine.loadTasks`, `.saveTasks`
- `app.go`: `historyMessages`, `skillRootsView`, `mcpFailed`, `saveExclusiveExportFiles`
- `delivery_worktree.go`: `pathWithinWorktree`
- `scroll_diagnostics_export.go`: `buildScrollDiagnosticsZip`

## Fix

Move all ten into a new `desktop/testhelpers_test.go`. No call site changes: the tests keep the same names, and no production path referenced any of them.

They did not go into the existing test files because `sessions_test.go` (1586 lines) and `app_test.go` already sit above repolint's 800-line ceiling — growing them would create a new `test-file-size` violation rather than remove one.

## Verification

- `go build ./...` and `go vet ./...` clean in the desktop module.
- `go test ./...` in the desktop module.
- `go run ./tools/repolint`: no finding for any touched file or the new one.
- Focused: `TestSession*`, `TestHeartbeat*`, `TestHistory*`, `TestSkill*`, `TestMCP*`, `TestExport*`, `TestDeliveryWorktree*`, `TestScrollDiagnostics*`.

## Not moved

`icon_repair_darwin.go`'s `writeMacAlias`, `resolveMacAlias` and `macURLReferencesReasonixBundle` stay in their `darwin && cgo` production file: they call cgo, and Go does not allow `import "C"` in a `_test.go` file.

Documentation-impact: none - no user-visible behavior changes; the helpers are test scaffolding that never ran in production.

### DIFF
--- a/desktop/app.go
+++ b/desktop/app.go
@@ -5376,10 +5376,6 @@ func historyCheckpointTurns(msgs []provider.Message, resolveUserContent func(str
 	return out
 }
 
-func historyMessages(msgs []provider.Message, resolveUserContent func(string) string) []HistoryMessage {
-	return historyMessagesWithPlannerDisplays(msgs, resolveUserContent, nil, nil)
-}
-
 func historyMessagesWithPlannerDisplays(msgs []provider.Message, resolveUserContent func(string) string, plannerTurns []plannerDisplayTurn, checkpointTurns map[int]int) []HistoryMessage {
 	replayedTodoArgs := historyTodoArgsWithCompleteSteps(msgs)
 	toolResults := historyToolResultsByID(msgs)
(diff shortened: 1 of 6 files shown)


SivanCola:
`desktop-frontend` fails on `src/__tests__/settings-refresh-snapshot.test.tsx`:

```
Error: timed out waiting for persisted display zoom sync
```

This is pre-existing on main-v2, not caused by this change:

- this PR only moves Go test helpers inside the desktop module; it contains no frontend change
- the failure reproduces on a re-run of the job: same test, same line 508, same message
- main-v2 already fails here: PR #10014 (merged 2026-09-09T11:43:51Z, and this branch base `88170d995`) shows the identical `timed out waiting for persisted display zoom sync` at the same assertion, and its check list also carries the `desktop` cascade failure

`desktop` is a gate job that asserts `FRONTEND_RESULT = success`, so it reports failure as a cascade of that same run rather than an independent failure.

The assertion waits for `GetDesktopZoomFactor` to reach the slider value, and `waitFor` in `test-support/settingsTestFixtures.ts:23` only flushes microtasks 20 times, so the zoom persistence path added by #10014 is the likely owner. Fixing it is a separate change.

PR #10027, opened 2026-09-09
fix(desktop): keep remote sends working against serves predating model settings / 修复对旧版远程 Serve 完全无法发送消息 (#9996)

### LINKED ISSUE #9996: [Bug]: Regarding the remote session, in the latest version, it completely fails to send messages and old messages are in an infinite connection state.!!
### Version line

### Exact version

v1.38.3

### What happened?

Regarding the remote session, in the latest version, it completely fails to send messages and old messages are in an infinite connection state.

Image

After opening the remote connection, your conversation will be unable to send messages normally. An error message "Send failed: model settings were saved but the remote session could not apply them: invalid character '<' looking for the beginning of the value" will appear.

Image

### Steps to reproduce

1. Create a remote session
2. Start chatting

### OS / platform

Windows 11

### Relevant logs or output

```shell

```

XTLine:
## Summary

- v1.38.3 gates every remote submit, compact, summarize and goal-resume behind the new model-settings protocol. Against a reused Serve only a few versions older — one that predates `/model-settings` — every send failed with `model settings were saved but the remote session could not apply them` ending in a JSON parse error about an invalid `<` character, making remote sessions completely unusable.
- The status probe now classifies a 200 whose body is an HTML document as the same "newer remote Serve" capability rejection as 404/405, instead of letting the HTML index reach the JSON decoder.
- Turn admission records that protocol-absent verdict for the current tab generation and admits the run without a revision — the behavior credential mode already supported before the gate — while transient refusals (busy turns, ownership conflicts) keep failing sends.

## Root cause

A Serve without the `/model-settings` route answers unknown GET paths through its catch-all `GET /` route with status 200 and the HTML index. The probe only anticipated 404/405 as "Serve too old", so the 200 HTML body went straight into JSON decoding (`invalid character '<' looking for beginning of value`). Because the probe runs before every admitted run, the error aborted every send. The reuse path kept such Serves eligible: its capability probe (`--port-file`) predates the model-settings protocol by six weeks, so the serve is reused, not upgraded.

## Issues

Fixes #9996

## Verification

- [x] `cd desktop && go build ./...`
- [x] `cd desktop && gofmt -l` clean and `go vet ./...` pass
- [x] New `TestRemoteModelSettingsLegacyServeHTMLIndexIsCapabilityRejection` covers the HTML index answer for both the status GET and the install POST
- [x] New `TestRemoteModelSettingsNonDocumentDecodeFailureStaysError` keeps corrupt non-document payloads as ordinary decode errors
- [x] New `TestEnsureRemoteModelSettingsAdmitsLegacyServeTurns` (fake kernel) covers turn admission, per-generation memory without re-probing, and re-probing after reconnect
- [x] `cd desktop && go test ./... -count=1` (full suite passes)
- [x] Root module `go build ./...`

## Documentation impact

Documentation-impact: updated - refreshed docs/desktop-migration/INVENTORY.md and inventory.json to match the current main-v2 desktop surface; no product behavior documentation changed.

## Cache impact

Cache-impact: none - no provider-visible prompt, transcript, or session-cache surface is touched; only the remote model-settings probe classification and admission for protocol-absent serves.
Cache-guard: `cd desktop && go test -run "TestRemoteModelSettings|TestEnsureRemoteModelSettings" -count=1 .`
System-prompt-review: N/A - no system prompt, memory prefix, output style, or tool schema behavior changes.

## Concurrency, security, and compatibility

- The per-generation verdict is written under `remoteTabMu` with the same still-current tab/generation/path fence the existing failure path uses; no new lock ordering.
- No network destination, credential, or secret handling changes: the classification only inspects an already-fetched bounded body, and admitted turns carry no revision header, which older Serves ignore.
- Capable Serves are unaffected: probe, install, and fencing paths are identical when the protocol answers.


XTLine:
Both review findings are addressed — thanks @chatgpt-codex-connector!

**P1 (recheck the generation before admitting)** — fixed in da871517d. The unsupported-protocol verdict now admits a turn without a revision **only when it was recorded on the still-current generation**. When the tab/generation/path fence no longer matches (reconnect replaced the probe target mid-flight), admission loops and re-probes the replacement target, which may speak the protocol, instead of releasing one unrevised send against it. Covered by `TestEnsureRemoteModelSettingsReprobesReplacedConnection`, which bumps the tab generation during the first probe and asserts the replacement generation is probed before admission.

**P2 (report legacy generations instead of leaving them pending)** — fixed in 0dd63d197. `appendRemoteModelSettingsStatus` now reports a target whose cached `unsupportedGen` matches the current generation as `not_required` and leaves the overall application state untouched, so the Settings receipt stops claiming those sessions will apply settings and stops polling once every other target settles. Covered by `TestAppendRemoteModelSettingsReportsLegacyTargetNotRequired`.

Verification for both: `cd desktop && go build ./...`, `gofmt -l` clean, `go vet ./...`, focused model-settings tests, and the full `cd desktop && go test ./... -count=1` suite pass.

XTLine:
@codex review

XTLine:
@codex review both findings fixed:

**P1 (fence cached unsupported admissions to the send target)** — 5b371e4dd: `ensureRemoteModelSettings` now returns the tab generation its decision was made under; `SubmitRemoteTab` and the gated `remoteTabPost` paths re-validate that generation against the resolved target and loop back through admission when a reconnect replaced it. Ungated sends keep generation 0 and skip the fence. Covered end-to-end by `TestSubmitRemoteTabFencesLegacyAdmissionToItsTarget`.

**P2 (require a recorded nonzero unsupported generation)** — 8bcc041c7: generation 0 is now treated as an unrecorded verdict at both consumers, so fresh/restored tabs neither skip the admission gate nor report `not_required`; they stay pending until a probe runs. Covered by `TestAppendRemoteModelSettingsFreshTabStaysPending`.

Full `cd desktop && go test ./... -count=1` passes.

PR #10029, opened 2026-09-09
fix(frontend): unbreak the zoom snapshot test after the Electron app-zoom move / 修复 Electron 缩放迁移后失败的 zoom 快照测试

SivanCola:
## Problem

`desktop-frontend` is red on `main-v2`:

```
FAIL src/__tests__/settings-refresh-snapshot.test.tsx
Error: timed out waiting for persisted display zoom sync
```

It fails on every branch based on `main-v2`, including PRs that touch nothing in the frontend, so the required check blocks unrelated work. The failure reproduces on a job re-run.

## Root cause

#10014 moved app zoom into the Electron shell. `SettingsPanel` now reads `host.native.getAppZoom()` when `desktopHost().kind === "electron"` and only falls back to the `App.GetDesktopZoomFactor` command otherwise:

```ts
const persisted = host.kind === "electron"
  ? await host.native.getAppZoom()
  : await app.GetDesktopZoomFactor();
```

The shared `installDesktopHostStub` always reports `kind: "electron"` and hard-coded `getAppZoom: async () => 1`, so the `0.5` this test injected through the retired command never reached the zoom slider and `waitFor` timed out after its 20 microtask flushes.

## Fix

Give the stub `appZoom` / `appZoomWrites` options, matching the existing `clipboardWrites` / `externalOpens` pattern, and drive the native path from the test: `persistedZoom` feeds `getAppZoom` and `setAppZoom` records the writes. The two legacy host-command stubs are removed because the panel no longer calls them.

No production code changes.

## Verification

- The suite passes: `106 passed, 0 failed` (before: `timed out waiting for persisted display zoom sync` at line 508).
- `pnpm test` — all discovered suites.
- `tsc --noEmit` and `tsc --noEmit -p tsconfig.test.json`.
- `eslint` on both changed files.

Documentation-impact: none - test-only change; the documented zoom behavior and the panel's production path are unchanged.

### DIFF
--- a/desktop/frontend/src/__tests__/desktopHostStub.ts
+++ b/desktop/frontend/src/__tests__/desktopHostStub.ts
@@ -17,6 +17,10 @@ export interface DesktopHostStubOptions {
   clipboardReadText?: string;
   /** Records native openExternal calls. */
   externalOpens?: string[];
+  /** Value returned by native app-zoom reads. */
+  appZoom?: () => number;
+  /** Records native app-zoom writes. */
+  appZoomWrites?: number[];
 }
 
 export interface DesktopHostStub {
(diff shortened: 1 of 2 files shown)


SivanCola:
Superseded by #10032, which is already merged and fixes the same failure by routing the stub's native zoom through the command table. That keeps the existing test unchanged, so this PR is redundant and would conflict on `desktopHostStub.ts`. Closing.

PR #10163, opened 2026-09-12
feat(e2ebench): honor -permission in suite mode / suite 模式不再忽略 -permission

boscocp:
## Summary

**简述**：`-permission` 此前只在 swebench 模式生效。suite 模式把姿态硬编码在
`buildRunTaskArgs` 里，传进来的值被静默忽略、非法值也不报错，而 `benchmarks/e2e/tasks/`
那 73 个任务的语料正是跑在 suite 模式下——于是"权限姿态"这个变量在真正做比较的那个
模式里无法表达。本 PR 把同一个 flag 接到 suite 模式，进入前用已有的 `permissionFlag`
校验（未知值以状态码 2 退出），并像其它实验臂一样把姿态记进结果与报告标题。默认姿态
不变，仍是 `workspace-write`，已记录的 suite 数据仍然可比。

> **Rebased onto `main-v2` after #10209.** That PR introduced the public
> `--permission-mode` preset contract and its own `permissionFlag`, so this
> branch now speaks presets (`read-only | workspace-write | danger-full-access`)
> instead of the `auto | yolo` spelling it opened with. The A/B below was
> measured on the pre-rebase binary; the two arms are the same two modes under
> their new names (`--auto` is a deprecated alias that resolves to
> `workspace-write`, and `bypassPermissions` is now `danger-full-access`).

`permissionFlag`'s own comment already records why the alternative posture exists:

> The alternative posture exists because comparable harnesses run without the
> dynamic-shell gate, so measuring against them under the gate measures our
> permission policy rather than the agent.

That argument is about cross-harness comparison, and suite mode is where the
corpus such a comparison runs on lives. The flag was simply never wired there.
Its help string scoped it to swebench mode, so this is an extension rather than
a repair — but until now `-mode suite -permission <anything>` was still accepted
on the command line and then ignored, bogus values included.

### What it makes measurable

The point is not that one posture is better. It is that the posture was not a
variable a suite run could set. Running the suite corpus under both arms — 31
tasks, 2 reps per arm, one local open-weight model, a binary built from this
branch — gives:

| | `workspace-write` (control) | `danger-full-access` (treat) |
| --- | ---: | ---: |
| inline-interpreter denials / 100 tool calls | 5.27 (reps 6.33 / 4.40) | 0.33 (0.00 / 0.61) |
| tasks with at least one denial | 12 | 1 |
| `blocked by permission policy` | 44 | 0 |
| tool-call failures | 210 / 664 (31.6%) | 139 / 611 (22.7%) |
| Pass@1 | 90% (85% / 95%) | 82.5% (80% / 85%) |

The Pass@1 ranges overlap at 85%, so the accuracy effect is **null**: dropping
the approval gate buys cost and latency, not correctness. Caveat stated plainly
— one small open-weight model on consumer hardware, n=2 per arm, not
reproducible by a maintainer, and not offered as evidence about agent quality.
It is offered as evidence that the flag reaches the agent and moves the failure
distribution it is supposed to move.

### Why this is not tuning the instrument toward the score

#7779 pinned the rule this PR has to satisfy: *"The benchmark is an instrument,
not a target… nothing that helps the agent in a way a user's own machine would
not."* Three things keep this inside that rule.

- `bypassPermissions` is not a benchmark-only affordance. It is a posture a user
  selects on their own machine (`reasonix run --permission-mode=bypassPermissions`).
  The harness gains no capability the product does not already ship.
- The default mode is unchanged, so every existing run — including the `/e2e`
  bot, which invokes suite mode with no `-permission` — stays on
  `workspace-write`.
- The measurement above is the strongest evidence of all: if this were tuning
  for the score, it failed. The posture does not move Pass@1. It moves the
  failure mix, which is what an instrument is for.

#7303, which introduced the hard-coded posture, asked that "explicit ask/deny
rules and the sandbox boundary" stay intact. They do, and not by this PR's
choice — `Controller.ApplyHeadlessApprovalMode` already defines the posture as
*"skip ordinary approval-gated decisions (nil approver); deny rules and fresh
decisions still fail closed"*, and *"deny rules and fresh-human tools (memory,
plan, sandbox, config) stay enforced by the gate for every mode."*

### The posture is an arm, so it is recorded like one

Every other arm axis is on the result — `plan_forced`, `anchor`, `cache_arm`,
`effort`, `segments` — several with a comment saying arms must not be mixed.
The posture was not, and the README line I am adding says "never mix postures in
one report". So each result now carries `permission`, and a non-default posture
is named in the report header beside the cache arm.

This is not hypothetical. In the A/B above the control and treat markdown
reports both opened with the byte-identical
``## 🤖 Reasonix e2e benchmark (arm `full`)`` and neither JSON held anything to
tell them apart — only the file names did. The added field is additive, so the
`/e2e` bot's `report.json` consumer, which reads named fields, is unaffected.

### Deliberately not done

- **No new posture.** The three presets are exactly the ones `permissionFlag`
  already mapped for SWE-bench mode; nothing new is reachable.
- **The default *mode* is untouched, the spelling is not.** `buildRunTaskArgs`
  now emits `--permission-mode=workspace-write` where it hard-coded `--auto`.
  Those resolve to the same mode (`internal/cli/cli.go:488` marks `--auto` a
  hidden deprecated alias that "uses workspace-write"), so recorded suite
  numbers stay comparable — but the argv is no longer byte-identical, and
  `profile_test.go`'s three whole-slice assertions are updated to the new
  spelling rather than left as a guard for the old one. Suite mode was the last
  caller of the deprecated alias; leaving it there to preserve a byte would have
  meant re-introducing it against #10209's direction.
- **`-mode diff` is left alone.** It has its own arg builder and no comparison
  story that needs the axis.
- **No mixing guard in `-mode compare`.** `compare.go` does not guard `anchor`
  or `cache_arm` either; adding a posture-only guard would be inconsistent.

## Verification

- New `cmd/e2ebench/suite_permission_test.go`, three guards:
  - `TestBuildRunTaskArgsCarriesPosture` — each of `""`, `workspace-write`,
    `read-only` and `danger-full-access` reaches the produced arg slice as the
    right flag, and **exactly one** posture flag appears (the CLI rejects two:
    `--auto/-y cannot be combined with --permission-mode`).
  - `TestPermissionFlagRejectsAnUnknownPreset` — the retired spellings
    (`auto`, `yolo`, `bypassPermissions`) and a nonsense value are all refused,
    so a stale command line fails loudly instead of running under a posture
    nobody chose.
  - `TestReportHeaderNamesANonDefaultPosture` — `workspace-write` renders the
    historical header unchanged, `danger-full-access` appends
    ` · danger-full-access-permission`.
- `profile_test.go`'s three whole-slice assertions
  (`TestBuildRunTaskArgsEnablesUnattendedWorkspaceWrites`,
  `…PassesTheAblationArmThrough`, `…PassesEffortThroughWithoutLegacyMode`) are
  updated to the new spelling and remain what pins the default argv: restoring
  a hard-coded posture in `buildRunTaskArgs` breaks
  `TestBuildRunTaskArgsCarriesPosture` on every non-default preset.
- Early validation: `go run ./cmd/e2ebench -mode suite -permission lixo` prints
  `unknown permission preset "lixo" (want read-only, workspace-write, or danger-full-access)`
  and exits 2 before loading the suite. On `main-v2` that validation sits inside
  the `if *mode == "swebench"` branch, so suite mode accepts the bogus value and
  runs under the hard-coded posture without a word.
- `go test ./cmd/e2ebench/` — pass. `gofmt -l .` and `go vet ./...` — clean.
  `make lint` — `golangci-lint` 0 issues at the pinned version, and
  `repolint: clean (1197 baselined findings)` with the baseline **untouched**.
  Note for local reproduction: `make lint` needs `GOTOOLCHAIN=go1.26.6` (the
  version `go.mod` pins and CI resolves via `go-version-file`) — the pinned
  staticcheck panics building IR for the Go 1.27 stdlib.
- End-to-end: the two-arm run in the table above, which is only possible with
  this change — the control arm logged 44 `blocked by permission policy`
  failures and the treat arm 0.

## Documentation impact

Documentation-impact: none - `cmd/e2ebench` is an internal benchmark tool with no `docs/*.md` page, the same reading #7733 used when it introduced this flag. The harness's own reference, `benchmarks/README.md`, is updated here: the flag table gains `-permission`, and three passages that presented the hard-coded posture as the only one are corrected, including the invocation line and "the `--auto` flag is deliberate so unattended fixture writes are allowed".

## Cache impact

Cache-impact: none - no provider-visible prompt or tool surface is touched; the diff is confined to `cmd/e2ebench/` and `benchmarks/README.md`.
Cache-guard: Not required - `scripts/check-cache-impact.sh` reports no cache-sensitive prompt/tool file in this diff, and `.github/workflows/cache-impact.yml` puts `benchmarks/**` and `*.md` in `paths-ignore`.
System-prompt-review: Not required - no system-prompt-sensitive path in this diff.


SivanCola:
Thanks — the suite-mode wiring is directionally right, but I found two benchmark-integrity gaps that should be addressed before approval.

1. **The warm-cache primer still ignores `-permission`.**

   `runTask` passes the requested posture to the graded invocation, but `warmPrefix` still hard-codes `reasonix run --auto`. Therefore:

   ```sh
   e2ebench -cache warm -permission read-only
   ```

   runs the ungraded primer as `workspace-write` and only the graded task as `read-only`. The primer runs in the same task work directory, so it can mutate fixtures under a posture the caller explicitly did not select; it also means a warm arm is not consistently using the recorded permission axis. A focused fake-binary test captured the primer argv as `run --auto --max-steps 1` for a `read-only` config.

   Please route `warmPrefix` through the same `permissionFlag(cfg.permission)` mapping (or a shared argument builder), and add a regression test proving that each supported preset reaches both the primer and graded invocation exactly once.

2. **Skipped results can erase the permission label from the report.**

   `Permission` is populated only by `runTask`; budget- and anchor-skipped results do not carry it. `render` then derives the header exclusively from `results[0]`. If the first task is skipped because an anchor seed is missing and a later task runs with `danger-full-access`, the report still renders the historical default header. A focused test reproduces this with a leading skipped result followed by a `danger-full-access` result.

   Please stamp the experiment axes, including `Permission`, onto every result shape, or derive and validate the report-level posture independently of the first row. Add coverage for a leading skipped result under a non-default permission mode.

The existing focused suite and lint are otherwise clean on the latest `main-v2` merge tree:

- `GOTOOLCHAIN=go1.26.6 go test ./cmd/e2ebench/` passed
- `GOTOOLCHAIN=go1.26.6 make lint` passed
- `git diff --check` passed

Once these two cases are fixed and the updated head has full CI, I can reassess approval.

boscocp:
Thanks — both were real, and both are fixed in 5d4aee6.

**1. The warm-cache primer now carries the selected posture.**

You were right that `warmPrefix` was the hole: the primer runs in the same task work directory the graded run is scored in, so a hard-coded `--auto` let it mutate fixtures outside the recorded permission axis.

Rather than only mapping the posture in the second place, I pulled out the two things the primer and the graded invocation must agree on:

- `mustPermissionFlag(cfg.permission)` — the posture, resolved through the same `permissionFlag` mapping, after `runSuiteMode` has already validated the preset;
- `appendPrefixShapingArgs` — model, effort, and ablation, the flags that rewrite the provider-visible prefix.

`buildWarmPrefixArgs` and `buildRunTaskArgs` are now both built from those, so the primer can't drift from the graded run again. The primer stays ungraded: no `--metrics`, no `--trajectory`, one step. Graded argv ordering is unchanged, so the existing `profile_test.go` expectations still hold byte-for-byte.

Regressions added in `suite_permission_test.go`:

- `TestWarmPrefixPrimerCarriesTheSelectedPosture` — for each of the four supported presets, the posture reaches the primer *and* the graded invocation exactly once, with the expected value.
- `TestWarmPrefixPrimerMatchesTheGradedPrefixShape` — every prefix-shaping flag appears in both argv, and the primer stays unmetered at one step.

**2. Skipped rows no longer erase the axes.**

Fixed at the boundary rather than in the renderer: `newResult(cfg, t)` stamps the experiment axes — arm, anchor, cache arm, effort, permission — wherever a result is created. Both skip paths (`runSuite`'s budget skip and `anchorSkip`) go through it, so a skipped row now belongs to the arm it was skipped under. This also closes the same gap for `Arm` and `CacheArm`, which were dropped from skipped rows too.

I also decoupled the header from row zero, as you suggested: `reportAxis` reads each axis from the first row that actually carries a value. A live run stamps every row, so this only matters for report JSON written before this change — but it means the header is no longer hostage to whichever row happened to sort first.

Your exact repro is now a test, `TestReportHeaderNamesThePostureBehindALeadingSkippedRow`: a leading anchor-skipped task followed by a `danger-full-access` result. Reverting `anchor.go` alone makes it fail with `skipped result dropped the permission axis`. `TestReportHeaderIgnoresALeadingRowWithoutAxes` covers the older-JSON shape.

Verification on the pushed head (5d4aee6, base `36852da4`):

- `GOTOOLCHAIN=go1.26.6 go test ./cmd/e2ebench/` passed
- `GOTOOLCHAIN=go1.26.6 make lint` passed (golangci-lint 0 issues, repolint clean)
- `gofmt -l cmd/e2ebench/` and `git diff --check` clean
- both new regressions confirmed failing against the pre-fix code

I left the branch on its current base and pushed fast-forward, so your earlier review context still lines up. Happy to rebase onto current `main-v2` if you'd rather re-verify against the latest tree.

PR #10177, opened 2026-09-12
fix(transcript): move commit-phase geometry dispatches off interactive lanes / 修复思考滚动时的 React #185 崩溃

SivanCola:
## Problem

Two production crash reports on build `036c7c50c5c1` (v1.38.7) show React error #185 — `Maximum update depth exceeded` — while the backend streams and the reader scrolls reasoning content. The two stacks throw from different dispatch sites (`onWheelCapture` and a bare setState), which is consistent with a root-level commit cascade where whichever update lands at the limit throws.

## Root cause

React 19's nested-update fuse counts consecutive commits that leave `Sync`/`InputContinuous`/`Default` lane work pending (`remainingLanes & 42`). The window adapter's measurement protocol legitimately closes every publication batch with a layout-effect state update — `setMeasurementRevision`, `virtualizer.resizeItem → rerender`, `setResidentStartKey`, `geometryChanged`, restore-transaction `refresh()` — so a sustained materialization cascade during stream + scroll produces 50+ consecutive spawned commits and trips the fuse even though every update is real, bounded progress.

## Fix

Dispatch the transcript geometry path's commit-phase updates inside `startTransition`, so spawned renders land on transition lanes and the fuse resets on every commit:

- `TranscriptWindow`: resident boundary reset/transfer, `presentationChanged` (the `geometryChanged` sink used by `MarkdownHistory`), `onGeometryWillChange` (restore-transaction refresh), the TanStack keyed-cache `resizeItem` batch + its batch close, the geometry acknowledgement, and pinned-jump acknowledgment.
- `Transcript`: both `setFolds` reconciliation effects.

All synchronous protocol work — staging, ledger commit, keyed-cache transfer, transaction creation, DOM measurement — still runs inside the same layout-effect task. The "layout-effect state update" and "same browser task" contracts are preserved verbatim; only the spawned render's lane changes.

## Verification

- New deterministic regression test `transcript-nested-update-fuse.test.tsx` fires `geometryChanged()` from a no-deps layout effect for 60 consecutive commits through the real `TranscriptWindow`: with plain dispatches React throws `Maximum update depth exceeded` at commit 53 (the production signature); with transition lanes all 60 settle and the window keeps mounting.
- `pnpm test:transcript` — full suite green (kernel, races, measurement ledger, materialization, viewport, selection, store).
- `tsc --noEmit` + eslint clean.
- Real-browser smoke against `?mock=bench` `bench:windowed-1000t`: 60 wheel events over the mounted window, zero page errors.
- No persisted-format, provider-bytes, or cache-contract changes.

#### Test plan
- [x] `pnpm test:transcript`
- [x] `transcript-nested-update-fuse` regression (fails without the fix, passes with it)
- [x] `tsc --noEmit`, eslint
- [x] Browser smoke on the 1000-turn windowed fixture

Generated with [Devin](https://devin.ai)


SivanCola:
Superseded by #10209. The virtualized geometry path fixed here has been removed, the React #185 scenario is covered by the new natural-flow chat regression suite, and the independent CI/release changes are already present on `main-v2`. Closing this PR while retaining it as historical context for the original crash.

PR #10241, opened 2026-09-13
refactor: align sessions with DSH ownership / 对齐 DSH 会话所有权与持久化机制

SivanCola:
## Summary

Complete the DSH-style session ownership cutover for the four blockers found in review. Runtime business state now commits to one typed in-memory `Session`, persistence is a separate write-behind binding with the existing fixed 200 ms window and semantic checkpoints, and clients attach through scoped bindings to a host-owned runtime.

The final writable codec is `reasonix.session.linear/v3.1`. Existing legacy transcripts and retired `events/v3` / `linear/v3` stores remain import sources and are never rewritten in place.

Merge requires green current-head CI and no unresolved review blockers. Native packaging and signing are release-stage validation and are outside this PR gate.

## Review fixes

1. **Stop no longer waits for session commits or persistence.** `Activity` owns an immutable cancellation handle, completion signal, commit eligibility gate, and one-shot 15-second supervisor. Cancellation reaches the activity before interaction teardown, status events, UI callbacks, or flush work. A batch that already acquired eligibility completes atomically; later business batches are rejected.
2. **Controllers cannot destroy shared runtimes.** `RuntimeOwner`, `PreparedRuntime`, and `ClientBinding` separate host authority from client attachment. Prepare/publish validates an exact instance, binding failure only releases the caller's binding, and stale cleanup cannot unregister a successor.
3. **Legacy and preview histories are resolved together.** Continue now freezes the legacy head, paired event store, sidecars, and owned artifacts before choosing a source. Proven prefixes are imported once; divergent work fails closed. Process-local tool recovery and UI timing metadata do not become migration authority or restored authorization. An importer now returns `ErrWriterOwned` immediately when another runtime still owns the preview instead of waiting without a bound deadline.
4. **Session listing no longer replays every transcript.** The catalog reads manifests, log revisions, and a rebuildable metadata projection. Rebuilds are bounded, cancellable, shared per session, generation-fenced against deletion, and incrementally consume only an uncovered tail. Warm list reads do not touch event bodies.

## Additional cutover work

- Split `Session` from `SessionHandle` through `PersistenceBinding`; accepted and durable receipts remain distinct.
- Validate physical commit identity, generation, sequence continuity, and event schema while serialized with append/sync. Uncertain writes are reconciled without consuming a later queued batch.
- Preserve the DSH checkpoint positions before model calls and top-level tool bodies. Todo, approvals, messages, and turn completion do not add per-fact fsync barriers.
- Rebind Desktop settings and model replacement through prepare/publish ownership. A failed replacement restores the prior legacy import source and leaves accepted work running.
- Keep forks and rewinds as independent session directories with copied owned artifacts and exact event boundaries.
- Upgrade Desktop to RPC v5 and require Serve `session-ownership-v1` in addition to the existing execution/session-event capabilities.
- Keep #10223's live file-observation and permission behavior; this change does not restore evidence gates, `complete_step`, or todo completion gates.

## CI lifecycle fixes

- Compare immutable Desktop binding identities without recursively reading live Controller locks.
- Drain fsnotify events/errors independently of registration and shutdown, including errors sent during close. Filesystem scans observe cancellation at directory/file boundaries.
- Apply the existing test-only watcher exclusion to Windows `.test.exe` binaries as well as Unix `.test` binaries. Production builds retain discovery invalidation, and watcher-specific tests explicitly opt in.
- Release skill stores on failed builds and through extension runtime ownership; close Desktop fixture controllers before removing their session directories.
- Permit legitimate zero-duration model-switch stages. Outcome and aggregate duration assertions remain. No CI suite, race check, timeout, or global budget was weakened.

## Verification

The preceding head completed all checks successfully except `desktop-windows-go`: eight fixtures passed their functional assertions but retained session ownership locks during directory cleanup. The current commit binds every Controller in the affected fixture files to test cleanup. Those eight scenarios and telemetry reset pass locally with `-race`; Desktop lint and generated inventory checks also pass. Native Windows CI is pending on the new head. No production behavior or test assertion was removed.

Current head: `af67e0a7a` (runtime fixes in `59ae5774a`, inventory refresh in `268a9d233`, and remaining Windows fixture cleanup in the current head).

Passed locally:

- `go test ./internal/boot ./internal/skill -count=1 -timeout=3m`
- `go test -race ./internal/skill ./internal/boot -count=1 -timeout=5m`
- Root full-suite run: all packages passed except the stale generated inventory; after regeneration, `go test ./tools/desktopinventory -count=1` passed
- Focused Desktop race tests covering binding replacement, autosave, model switching, inbox, blank sessions, checkpoints, and restored history
- Repeated skill race tests, including deterministic error-draining shutdown and cancelled directory discovery
- `go vet ./internal/skill ./internal/boot`
- `golangci-lint` for skill/boot and the complete Desktop module
- `go run ./tools/repolint` and `git diff --check`

The full Desktop race suite passed on the same Desktop identity fix before the subsequent fixture cleanup additions. Current-head GitHub Linux, Windows, Desktop, app-memory, and CodeQL results remain required; earlier green checks do not establish the new commit as green. Native packaging and signing remain release-stage work.

## Security, performance, and compatibility

Session IDs are validated as one path component and all create/open/import/export/delete paths stay under the storage root with directory ownership fencing. Existing-session lookup resolves the physical directory through `os.Root.OpenRoot`, so protocol input does not remain a path value inside the storage stack. The previous head passed CodeQL without alert dismissal or suppression; the new head must pass again. No dependency, network, credential, telemetry, or sandbox expansion is introduced. Unknown required events and complete-record corruption fail closed; torn tails remain preserved for diagnosis.

The ordinary hot path performs a short in-memory commit and enqueues an immutable batch. Disk work remains behind the 200 ms fixed window or an explicit semantic checkpoint. Catalog list work is bounded by manifest/cache reads; full transcript reduction is off the list path.

The v3.1 codec intentionally prevents older preview binaries from writing these sessions. Old data remains readable through isolated adapters and migrates to a new immutable session ID before execution.

## Scope and cache impact

No global repolint limit was increased. The Desktop inventory was regenerated after responsibility-based file extraction.

Documentation-impact: updated - bilingual DSH alignment design and generated Desktop migration inventory describe v3.1 ownership, import, and query behavior.
Cache-impact: medium - the branch intentionally changes todo/tool and reconstructed model-history serialization once at upgrade; runtime state, timestamps, session IDs, paths, and durability progress remain outside the stable system prefix.
Cache-guard: `scripts/cache-guard.sh` passes, including provider-byte bounds and same-version dialogue/tool-loop hit-rate guards.
System-prompt-review: reviewed - no recurring dynamic session or persistence state was added to the system prompt.

Related overlap reviewed but not adopted as contributor commits: #10197, #10061, #10206, #10229, #10155, #10057, and #9866.


internal/sessionv3/service.go
```diff
@@ -0,0 +1,705 @@
...
+	runtime.cancel = nil
+	runtime.activity = ""
+	runtime.phase = RuntimeIdle
+	runtime.revision++
+}
+
+// Cancel sends the signal while holding only the small runtime mutex. It does
+// not flush, publish UI state, or wait for a callback.
+func (r *Runtime) Cancel() bool {
+	r.mu.Lock()
```
SivanCola: [P1] Send cancellation independently of the commit mutex. Activity.Append holds this same mutex while calling the writable handle (lines 158–169). If that append waits for the Store lock or backend, Cancel cannot reach r.cancel(), so the new session-level Stop can still hang. TestReviewCancelSignalsBeforeRuntimeCommitLock reproduces the failure by holding that lock and observing the owned context remain uncancelled. Use the captured activity cancellation capability before any storage/state lock, then reconcile the phase and supervision separately, as in DSH.

internal/control/session_binding_v3.go
```diff
@@ -0,0 +1,416 @@
...
+	if err != nil {
+		return sessionv3.SessionRef{}, err
+	}
+	prepared := agent.NewSession("").CloneWithMessages(candidate.Session().Snapshot().Projection.ModelMessages)
+	old, err := c.publishV3Runtime(candidate, prepared, true)
+	if err != nil {
+		// A newly opened candidate is safe to close only when it was not the
+		// currently published controller runtime.
+		if current == nil || candidate != current {
+			_ = service.CloseRuntime(context.Background(), candidate)
```
SivanCola: [P1] A failed client binding must not dispose a service-owned shared runtime. service.Open can return an existing target used by another client; candidate != current only compares against this controller. If domain restoration fails, this branch closes and unregisters that shared target. TestReviewOpenFailureDoesNotCloseOtherRuntime reproduces it with two published runtimes and an invalid typed Plan payload on the second. Apply DSH prepare/publish/disposer ownership: only discard an unpublished candidate owned by this attempt; release a client binding independently of closing the global runtime.

internal/sessionv3/service.go
```diff
@@ -0,0 +1,705 @@
...
+// ContinueLegacy freezes one legacy head, publishes its deterministic final
+// session, then attaches that exact session. It never writes the source and it
+// does not accept the caller's pending submission; hosts enqueue the unchanged
+// submission only after this method returns the new immutable identity.
+func (s *Service) ContinueLegacy(ctx context.Context, sourcePath, headID string) (*Runtime, MigrationResult, error) {
+	filesystem, ok := s.persistence.(*FilesystemPersistence)
+	if !ok {
+		return nil, MigrationResult{}, errors.New("sessionv3: persistence does not support legacy migration")
+	}
+	result, err := migrateLegacyHeadForHost(ctx, sourcePath, filesystem.Root, headID)
```
SivanCola, line 454: [P1] Complete paired legacy/prototype source selection before enabling the cutover. ContinueLegacy migrates the transcript alone; ContinuePrototype is separate and has no end-to-end production continuation caller. A newer event sidecar can therefore be ignored. Blanket rejection is insufficient while existing rebuild paths still create sidecars (full Desktop regression demonstrated that). Freeze and validate both sources under the appropriate owners, select only provable history, preserve both originals and retire the old producers in the same switch.

internal/sessionv3/query.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
+		return SessionPage{}, err
+	}
+	for i := range page.Sessions {
+		info := &page.Sessions[i]
+		info.Ref = SessionRef{HostID: q.hostID, SessionID: info.SessionID}
+		if info.Error != "" {
+			continue
+		}
+		snapshot, snapshotErr := q.Snapshot(context.Background(), info.Ref)
```
SivanCola: [P2] Catalog listing currently replays every cold session in full and drops caller cancellation with context.Background(). Query.Snapshot accumulates all commits before projecting them, so a page of 50 long sessions scales with their entire histories despite sparse event-page indexes. Pass the caller context through List and build/reuse a verifiable event-sequence metadata projection; stream replay when rebuilding it instead of retaining complete raw logs.

SivanCola (commented):
**Final review outcome: resolved and merged.** The original HOLD covered independent activity cancellation, host-owned client bindings, paired-source import, and cold catalog queries. Each finding now has a resolved review thread and regression evidence. Final head af67e0a7a2dfd963fe1e42267e882643cab3ea6b completed 39 successful checks and 5 explicit skips, including Windows Desktop, race, app-memory, and CodeQL. No new PR-specific security alerts remain; four repository alerts are inherited from the unchanged base and were not suppressed. Merged into main-v2 as 36852da4dcf4c90ffad2c20fbb5f9ca20b644823. Packaging and signing remain release-stage work; this merge is not a release.

internal/sessionv3/service.go
SivanCola: Fixed in 2107f60f2. Activity validation now happens outside the runtime lock, the final gate covers only eligibility plus the atomic in-memory commit, and Runtime.Cancel reaches the captured activity through an atomic pointer before status, storage, or UI work. The activity also owns its completion signal and one-shot 15-second supervisor. TestCancelSignalsWithoutRuntimeMutex and the session/control race suites pass.

internal/control/session_binding_v3.go
SivanCola: Fixed in 2107f60f2. The host now owns exact RuntimeOwner capabilities while controllers receive ClientBinding capabilities. Failed attachment or controller replacement releases only that binding; exact-instance fencing prevents an old disposer from unregistering a successor. TestClientBindingOwnsDetachButNotRuntimeClose, TestOldRuntimeDisposerCannotCloseSuccessor, and the shared-runtime regressions pass.

internal/sessionv3/service.go
SivanCola: Fixed in 2107f60f2. ContinueImported freezes the selected legacy head and paired retired event store before publication, selects only equal/proven-prefix structured histories, and fails closed on divergent work. The v3.1 target preserves both source originals and never restores activity, approval, or tool-recovery authority. Desktop settings/model rebuilds release the old sidecar writer before the unified freeze and restore it on failure. Paired-source, frozen-source, and Desktop continuation regressions pass.

internal/sessionv3/query.go
SivanCola: Fixed in 2107f60f2. List/Stat now read manifest, file revision, and lightweight catalog metadata; missing metadata is rebuilt by a bounded shared task that streams commits and honors its host context. Warm listing reads no event bodies, tail refresh consumes only uncovered records, and deletion generations fence late cache publication. TestWarmCatalogListDoesNotReadEventBodies and metadata generation tests pass.

PR #10244, opened 2026-09-13
feat(config): opt a model into the tool-action policy / 按模型启用工具执行策略

boscocp:
## Summary

**简述**：`ModelActionPolicy` 已经编译进二进制，但自 live comparison 套件评估之后就
无法从生产装配到达。本 PR 通过 `model_overrides.<model>.action_policy = true` 逐模型
开启：仅当某个模型显式选择时，才在核心策略之后追加该段落。默认关闭，不做任何基于
model id 的推断，未启用时 provider 可见的 prefix 保持逐字节一致。

`ModelActionPolicy` (formerly `KimiActionPolicy`) has been compiled in but
unreachable from production assembly since the live comparison suite evaluated
it. `docs/OPENCODE_TOOL_RECOVERY_IMPLEMENTATION.md` records why it was left
unreachable rather than deleted: +13% tokens bought one avoided failure in 120
trials, which does not pay for itself on a frontier model.

Small open-weight models have the opposite balance. They narrate plans instead
of calling tools, and report work they never performed. So the switch is **per
model, off by default, and never inferred from the model id** — the paragraph
costs prompt tokens on every turn, and deciding that trade stays the operator's
choice.

```toml
[[providers]]
name = "ollama-local"
model_overrides = { "qwen3:8b" = { action_policy = true } }
```

### What the diff actually does

- `ProviderModelOverride` gains `action_policy` (`*bool`). Nil leaves the policy
  off. `boot.build` appends it right after `appendCorePolicies`.
- **Bug found on the way in:** `normalizedModelOverrides` dropped overrides it
  judged empty, so an `action_policy`-only entry would have been silently
  discarded during load — the key would have done nothing when it was the sole
  entry. `modelOverrideEmpty` now accounts for it, and
  `TestActionPolicyOnlyOverrideSurvivesLoad` is the guard.
- `AppliesModelActionPolicy` **derives** the answer from the resolved model's
  override rather than caching it in a `ProviderEntry` field. `ProviderEntry`
  gains no scalar — `REASONIX.md`'s struct-state rule names "fixing a boundary
  case by adding one more `bool`" as exactly the move to avoid. It also makes
  the state unable to disagree with its source: `legacyDeepSeekProviderWideProjection`
  needs no reset line, because it already clears `ModelOverrides`.

### File moves, and why they are in this PR

Two extractions out of `internal/config`, both so the per-model capability keys
live together instead of inside the whole-config type and its two renderers:

- `renderModelOverrides`, `renderModelOverride`, `modelOverrideEmpty` →
  `render_model_overrides.go`
- `ProviderModelOverride`, `applyModelOverride`, `modelOverrideForModel` →
  `model_override.go`

They are not cosmetic. `config.go` and `load.go` are both far over the 800-line
ceiling and carried on the repolint ratchet; adding the new key pushed
`config.go` 10 lines and `load.go` 1 line past their recorded budgets.
`REASONIX.md` says *"Never widen the baseline to land a change — fix the code"*,
so the baseline is **untouched** and the code moved instead. Both new files sit
well inside their own budgets.

## Verification

- `internal/boot/action_policy_effect_test.go` — the effect test at the final
  boundary, in the `internal/boot/effect_test.go` pattern: it asserts what
  reaches the assembled system prompt through the real `boot.Build`, not that a
  helper returns a string.
  - A default build's prompt is **identical across two builds** and contains no
    trace of the paragraph — this is the cache-stability guard.
  - An opted-in build appends **exactly one** copy, after the core policies.
- `internal/config/model_action_policy_test.go` — the load-path guards,
  including the `normalizedModelOverrides` regression above.
- `GOTOOLCHAIN=go1.26.6 make lint` → `golangci-lint` 0 issues at the pinned
  version, `repolint: clean (1231 baselined findings)` with the baseline
  untouched, on the current head rebased onto `main-v2` `b21deef0`. (The `GOTOOLCHAIN` pin matters locally: the pinned staticcheck
  panics building IR for the Go 1.27 stdlib. CI resolves it via
  `go-version-file: go.mod`.)
- `go test ./internal/config/ ./internal/boot/ ./internal/agent/ ./internal/tool/builtin/`
  — pass. `gofmt -l .` and `go vet ./...` — clean.
- `scripts/check-cache-impact.sh` and `scripts/check-docs-impact.sh` both pass
  against this body.

### Deliberately not done

- **No model-id inference.** No `qwen`/`llama`/`kimi` prefix match decides this.
  A family heuristic would put the paragraph in the prefix of models whose
  operator never asked for it, and the +13% token cost is real.
- **No default change.** An absent key leaves the provider-visible prefix
  byte-identical, which is the whole reason the switch is opt-in.
- **No claim that the paragraph helps.** The measurement that would support that
  is a suite-corpus A/B on a local model, which this PR does not include. It
  makes the arm reachable; it does not assert the outcome.

## Cache impact

Cache-impact: low/conditional - default builds are unchanged; an opted-in model gets one new stable system-prompt paragraph, causing a one-time prefix transition and recurring prompt-token cost.
Cache-guard: `internal/boot/action_policy_effect_test.go` asserts the default build's system prompt is unchanged and identical across two `boot.Build` calls, and that an opted-in build appends exactly one copy at a fixed position, so the opted-in prefix is itself stable across turns.
System-prompt-review: requesting a maintainer pass on the paragraph's placement immediately after `appendCorePolicies` and before the `sessioncontext.PolicyBlock()` tail, and on the enabled-path cost this accepts: turning the key on invalidates that model's prior prefix once, then holds a new stable prefix, and adds the paragraph's tokens to every request for that model.

## Documentation impact

Documentation-impact: updated - `docs/GUIDE.md` and `docs/GUIDE.zh-CN.md` document the `model_overrides.<model>.action_policy` key and its default-off rationale; `reasonix.example.toml` carries a commented example.


SivanCola:
Thanks — the per-model opt-in is a sensible boundary, and the provider-effect test is useful. I found three items that should be addressed before approval.

1. **Preserve `ActionPolicy` through legacy model-override merges.**

   `ProviderModelOverride` now has the new pointer field, but `mergeProviderModelOverride` still copies only reasoning, effort, vision, context-window, and output-token fields. `mergeLegacyDeepSeekModelConfiguration` calls this helper while canonicalizing legacy official DeepSeek providers, so an explicitly configured `action_policy = true` or `false` can be silently dropped during migration.

   A focused test currently fails with:

   ```text
   action_policy was dropped by model-override merge
   ```

   Please give the field the same tri-state merge semantics as the other pointer override:

   - source `nil`: leave the destination unchanged;
   - source `true` or `false`: copy an independent bool pointer to the destination.

   Please also deep-copy `ActionPolicy` in `cloneModelOverrideMap`, alongside `Vision`, so cloned config entries do not share pointer state. Add a regression that exercises the real legacy DeepSeek canonicalization path and proves that explicit `true` and explicit `false` both survive migration and render/load round trips.

2. **Rebase onto the latest `main-v2` and restore the repolint ratchet.**

   The PR currently has merge-base `36852da4`, while current `main-v2` is `f3e8479f`. The latest-base merge tree is conflict-free and its focused tests pass, but `make lint` fails because the added `boot.build` line exceeds the current exact baseline:

   ```text
   internal/boot/boot.go: file-size is 2021 over its 2020 budget
   internal/boot/boot.go: function-size is 1714 over its 1713 budget
   ```

   Please rebase and keep the policy insertion without widening the ratchet merely to accommodate this line — for example, fold it cleanly into the existing policy-composition boundary or make a useful extraction.

3. **Describe the cache impact as conditional rather than `none`.**

   The default/unset path is byte-identical, which the effect test covers. However, an opted-in model intentionally receives an additional provider-visible system-prompt paragraph: enabling it invalidates that model's prior prefix once, establishes a new stable prefix afterward, and adds the paragraph's prompt tokens to every request. Please change the PR metadata to something like:

   ```text
   Cache-impact: low/conditional - default builds are unchanged; an opted-in model gets one new stable system-prompt paragraph, causing a one-time prefix transition and recurring prompt-token cost.
   ```

   The fixed text and its placement after core policies look appropriately scoped; the requested system-prompt review should explicitly acknowledge this enabled-path cost.

Latest-base verification performed:

- `GOTOOLCHAIN=go1.26.6 go test ./internal/config/ ./internal/boot/ ./internal/agent/ ./internal/tool/builtin/` passed
- explicit `action_policy = true/false` render/load round trip passed in a focused review test
- legacy model-override merge retention failed as described above
- `git diff --check` and `gofmt` passed
- `GOTOOLCHAIN=go1.26.6 make lint` failed on the two current ratchet checks above

Once these items are fixed and the updated head has full code CI, I can reassess approval.

boscocp:
Thanks for the thorough pass — all three were right, and all three are fixed. Head is now `f14cea37`, rebased onto current `main-v2` (`b21deef0`).

**1. `ActionPolicy` now survives the legacy model-override merges.**

You were right that the field was inert in the migration path: `mergeProviderModelOverride` never copied it, so `mergeLegacyDeepSeekModelConfiguration` dropped an explicit `true` *or* an explicit `false` while canonicalizing the official DeepSeek providers.

Rather than paste a second tri-state block, I pulled the semantics out into one helper in `internal/config/model_override.go`:

```go
// overrideBool folds one tri-state override. A nil next leaves the current
// value alone — absent is not a decision — and a set one is copied into an
// independent pointer so merged and cloned entries never share state.
func overrideBool(current, next *bool) *bool
```

`mergeProviderModelOverride` and `cloneModelOverrideMap` now fold both pointer overrides through it. `Vision` already had exactly these semantics inline, so its behavior is unchanged — the point is that the next pointer override can't be added with the wrong shape, which is how this one got in.

Regressions, both in `internal/config/model_action_policy_test.go`:

- `TestLegacyDeepSeekCanonicalizationKeepsExplicitActionPolicy` — drives the real path, not a helper: it opts `deepseek-v4-flash` in and `deepseek-v4-pro` explicitly out on the legacy providers, asserts both are eligible via `officialLegacyDeepSeekProviders`, runs `ensureDeepSeekOfficialProvider`, then checks the migrated entry, that the opt-in actually reaches `AppliesModelActionPolicy` (and the opt-out does not), and finally a `RenderTOML` → `LoadForRootReadOnly` round trip of both states.
- `TestCloneModelOverrideMapCopiesActionPolicy` — mutating a clone must not reach through to the source.

Confirmed failing against the pre-fix code, with the two lines you'd expect:

```text
--- FAIL: TestLegacyDeepSeekCanonicalizationKeepsExplicitActionPolicy
    after migration: action_policy was dropped, want explicit true
--- FAIL: TestCloneModelOverrideMapCopiesActionPolicy
    clone shares the action_policy pointer with the source
```

**2. Rebased, and the ratchet is untouched.**

The branch is on `b21deef0` now. The budget was the right thing to push back on: I folded the policy into the existing composition boundary instead of adding a line to `boot.build`. `appendCorePolicies` takes the resolved entry and returns `config.ApplyModelActionPolicy(prompt, entry)`, so the call site changed rather than grew:

```go
-	sysPrompt = appendCorePolicies(sysPrompt)
+	sysPrompt = appendCorePolicies(sysPrompt, entry)
```

`internal/boot/boot.go` is 2821 lines on this head and 2821 on `main-v2` — byte-for-byte the same size — so neither the file-size nor the `build` function-size finding can fire. `internal/config/config.go` and `render.go` moved *down* (2315→2262 and 1761→1713) via the two extractions, which is what pays for the new key there. No baseline file is in the diff.

**3. Cache-impact is now conditional, and the review request names the enabled-path cost.**

Agreed — `none` described the default path and quietly ignored the one that matters to whoever turns the key on. The PR body now reads:

```text
Cache-impact: low/conditional - default builds are unchanged; an opted-in model gets one new stable system-prompt paragraph, causing a one-time prefix transition and recurring prompt-token cost.
```

`System-prompt-review:` now asks for a pass on that cost explicitly — one-time prefix invalidation for that model, a new stable prefix afterward, and the paragraph's tokens on every request — alongside the placement question. `Cache-guard:` also spells out that the effect test pins the opted-in paragraph's position, so the enabled prefix is stable across turns rather than merely present.

Verification on `f14cea37` (base `b21deef0`):

- `GOTOOLCHAIN=go1.26.6 go test ./internal/config/ ./internal/boot/ ./internal/agent/ ./internal/tool/builtin/` — pass
- `GOTOOLCHAIN=go1.26.6 make lint` — golangci-lint 0 issues, `repolint: clean (1231 baselined findings)`, baseline untouched
- `gofmt -l` and `git diff --check` — clean
- `scripts/check-cache-impact.sh` and `scripts/check-docs-impact.sh` — pass against the updated body
- both new regressions confirmed failing against the pre-fix code

Would appreciate another look when you have a moment.

boscocp:
Rebased onto current `main-v2` (`62862f48a`). Head is now `dd2420724`; the three commits are unchanged in content, and the fixes from your pass are all still in place.

One thing worth flagging, because it lands on the part you reviewed: `7fc2fd46e` has since created `internal/config/model_override.go` upstream — the same file split this PR makes. I kept the split and reassembled onto yours rather than the other way around:

- `ProviderModelOverride` is upstream's, `ReasoningDefaults` included, plus this PR's `ActionPolicy` field.
- `applyModelOverride` and `modelOverrideForModel` moved into that file carrying upstream's current bodies — the `reasoning*Automatic` flags and the `explicitModelReasoning` call — not the pre-rebase ones.

There was also a collision git does not report as a conflict. This PR had extracted the inline-table renderers into `render_model_overrides.go`; upstream extracted the same three functions into `model_override_render.go`. Different filenames, so the merge is textually clean and the build then fails with `renderModelOverrides redeclared in this block`. I dropped this PR's file and added `action_policy` to upstream's instead — five lines, and a smaller diff than before.

The interaction worth a second pair of eyes, since it is the kind that fails silently: `modelOverrideForModel` now returns `explicitModelReasoning(ov)`. That peels only reasoning fields, and `reasoningSnapshotHash` digests three named fields rather than the whole struct, so an added `ActionPolicy` neither perturbs the digest nor gets peeled. `go build ./...`, `go vet`, and `go test ./internal/config/... ./internal/boot/...` are green.

PR #10245, opened 2026-09-13
fix(agent): detect a prompt the provider silently truncated / 识别被服务端静默截断的提示词

boscocp:
## Summary

**简述**：Ollama 在提示词超出其上下文长度时**不报错**：返回 `200 OK`，静默丢弃超出的
部分，并在 `usage` 里报出截断后的 `prompt_tokens`。Reasonix 已有的溢出恢复阶梯只能由可解析的 400/413/422 触发，
因此完全看不到这一类退化——用户看到的只是"模型很笨"。本 PR 从每个响应已经携带的
token 计数中识别它，拒绝据此校准，并在发生的那一轮警告一次。

### The defect, measured

Stock Ollama 0.33.3, `OLLAMA_CONTEXT_LENGTH` unset and no Modelfile `num_ctx`, so
it serves its 4096 default. `context_window = 131072` configured in Reasonix.
Reading an 87 KB file on `main-v2`:

```
· 4042 tok · in 3974 (0 cached / 3974 new) · out 68
-> read_file {"path":"big.go"}
· tool output truncated: 67289 of 99001 bytes elided
· 4579 tok · in 2051 (0 cached / 2051 new) · out 2528
```

The second request reports **2051** prompt tokens for a transcript far larger
than that, and the model answers:

> It looks like you've shared a snippet of rule definitions (Rule 181 and the
> surrounding cases), but I'm not sure exactly what you'd like me to do with
> them…

It never received the question. Reasonix reported nothing and kept displaying the
configured 131072 window.

Confirmed at the protocol level with `curl`, same server:

| endpoint | body carried `options.num_ctx: 32768` | context allocated |
| --- | --- | --- |
| `/v1/chat/completions` (what Reasonix uses) | yes | **4096** — silently ignored |
| `/api/chat` (native) | yes | **32768** — honored |

So the window cannot be raised over the transport Reasonix uses, and `extra_body`
lands those keys at the same level `/v1` already ignores. A ~12,000-token prompt
returns `200`, `error: None`, `usage.prompt_tokens: 2051`. This matches Ollama's
documented behaviour: the FAQ states "By default, Ollama uses a context window
size of 4096 tokens", and the `/v1` surface has no `num_ctx` field, so the value
has to be set on the server via `OLLAMA_CONTEXT_LENGTH` or a Modelfile.

### Which runtimes this actually covers

The signal is protocol-level rather than vendor-specific, so no Ollama-only code
path is added. But the *claim* is deliberately narrow, because the runtimes do
not behave alike:

| Runtime | Prompt over the limit | Visible today? |
| --- | --- | --- |
| Ollama | `200 OK`, silently truncated | no — this PR |
| LM Studio | `truncateMiddle` / `rollingWindow` truncate silently, and an OpenAI-compatible client cannot select the policy per request | no — covered by the same signal |
| llama.cpp server | HTTP 400 `the request exceeds the available context size` | yes |
| vLLM | HTTP 400 `the engine prompt length ... exceeds the max_model_len` | yes |

The last two already fail loudly and are **not** claimed here. Worth noting
separately: llama.cpp's message carries no token numbers, so
`ParseContextLimitError` will not produce a `ContextLimitError` from it and it
surfaces as a generic 400. That is a different gap and not touched by this PR.

For scale: this repo's own golden baseline pins `ToolSchemaTokens: 3523`
(`internal/boot/testdata/golden/prefix_shape.json`), built with a stub base
prompt and no memory or skills. On a 4096-token window **the tool schemas alone
do not fit**, which is why the first thing to go is the model's ability to call
tools.

### Why it compounds

`calibratedPromptTokens` accepts a tokens-per-character density between `0.05`
and `2`. A truncated reading is not filtered out of that band, it is *inside* it:

- **moderate** (20k chars reported as 2051 tokens → 0.10) becomes the
  calibration. Every later prompt is then estimated at a fraction of its true
  size, `compactTrigger()` fires later, the next prompt is bigger, and more of it
  is dropped — the estimator learns from its own damage;
- **severe** (0.038) falls below the floor and is discarded, leaving the fallback
  density in place as though the whole prompt had landed.

Neither path tells anyone.

### The change

Detection is arithmetic on what the response already carries: a provider
reporting far fewer prompt tokens than the characters on the wire can encode
truncated them. Cache fields do not confound it — `CacheHitTokens` is documented
as a subset of `PromptTokens`, not a deduction from it.

The check sits in `setPromptTokenCalibration`, the one place the observed count
meets the shape Reasonix believes it sent, so a truncated reading is refused as
calibration **on sight** rather than learned from. Refusing it is strictly safe:
the fallback density is larger, so estimates get more conservative, not less.

Two guards keep it quiet on healthy sessions:

- a request must be large enough to demonstrate a ceiling;
- a reported size below 512 tokens is a provider reporting meaningless usage, not
  a runtime serving a tiny window. This is what the stub providers in the
  existing suite report (`PromptTokens: 10`), and without it they would be read
  as 10-token ceilings.

A calibration that is itself below the density floor does not suppress
detection — the absolute floor still applies underneath it. The warning is
claimed with a compare-and-swap, so concurrent turns produce one warning per
session rather than one per racing turn.

The notice carries a wire-stable `NoticeCode` so frontends localize by code
rather than by matching English text, per `internal/event/notice_codes.go`.

### After

Identical prompt, same server, this branch:

```
-> read_file {"path":"big.go"}
! The provider accepted this request but silently dropped part of the prompt, so
  the model did not receive the full instructions or tool definitions. Raise the
  server's own context length (for Ollama: OLLAMA_CONTEXT_LENGTH, or a Modelfile
  num_ctx), then set context_window to match.
· 3989 tok · in 2051 (0 cached / 2051 new) · out 1938
```

On the turn it happens, once per session — the third turn (`in 4082`) stays
silent.

### Deliberately not done

- **No window clamp.** Feeding the observed ceiling to `learnContextBudget` was
  implemented and then removed: it makes every later prompt that cannot fit fail
  admission outright. Verified — the session dies with
  `context exceeds provider limit and compaction failed: truncated view still 21307 >= 2051`.
  That converts a session that degrades into one that stops, which is a larger
  behavior change than detection and needs its own evidence. Warning without
  clamping also keeps the cost of a wrong reading at one line of text.
- **No native-endpoint probe.** Reading `/api/show` or `/api/ps` would tie the
  detection to Ollama; the usage arithmetic needs no vendor knowledge.
- **No preset change.** `newCatalogPreset` never sets `ContextWindow`, so the
  shipped `ollama-local` and `lmstudio` presets arrive with `ContextWindow == 0`
  — which `internal/agent/agent.go` documents as *disabling compaction*. That is
  a real and separate defect; happy to send it as its own PR.
- **`context_window = 0` still disables compaction.** `SPEC.md` documents it as a
  user choice and this PR does not override it.

## Verification

- `internal/boot/effect_truncation_test.go` — effect tests at the final boundary
  through the real `boot.Build` assembly, per REASONIX.md:
  - a truncating provider produces the notice **exactly once across two turns**,
    at `LevelWarn`, naming the accepted prompt size;
  - an **honest** provider whose reported tokens track the request produces
    none — the no-false-positive guard.
- `internal/agent/prompt_truncation_test.go` — the measured 2051/54000 case; the
  0.10-density case the absolute floor alone misses; five honest shapes that must
  not trip it (dense CJK, steady, mild drift, short request, no usage); the
  calibration is not poisoned; no window clamp; 10 tokens is not a ceiling.
- `GOTOOLCHAIN=go1.26.6 make lint` → `golangci-lint` 0 issues at the pinned
  version, `repolint: clean (1197 baselined findings)`, **baseline untouched**.
- `go test ./internal/agent/ ./internal/boot/ ./internal/config/ ./internal/event/ ./internal/i18n/ ./internal/tool/builtin/ ./internal/control/`
  — pass. `gofmt -l .` and `go vet ./...` — clean.
- The golden baseline test passes unchanged, which is the prefix-stability guard.
- `go test -race ./internal/agent/ ./internal/boot/ ./internal/i18n/ ./internal/event/`
  — pass, including the concurrent-warning test.
- Merges cleanly with #10244, which touches a different region of the same two
  docs files.
- End-to-end before/after above, on stock Ollama 0.33.3 with a model carrying no
  `num_ctx`.

> Local reproduction note: `make lint` needs `GOTOOLCHAIN=go1.26.6` on a machine
> whose system Go is 1.27 — the pinned staticcheck panics building IR for the
> 1.27 stdlib. CI is unaffected; it resolves the toolchain via
> `go-version-file: go.mod`.

## Cache impact

Cache-impact: none - detection reads the usage already returned in each response; no system prompt, tool schema, or request field changes. The golden baseline test passes unchanged.
Cache-guard: internal/boot/effect_truncation_test.go asserts an honest provider emits nothing, and internal/boot golden_baseline_test.go pins the provider-visible prefix.
System-prompt-review: requesting a maintainer pass - the only internal/boot change is a new effect test; no prompt assembly path is touched and golden_baseline_test.go passes unchanged.

## Documentation impact

Documentation-impact: updated - docs/GUIDE.md and docs/GUIDE.zh-CN.md gain a "self-hosted runtimes" section covering the runtime's context length versus the model's, the silent-truncation symptoms, and `ollama ps` as the check.


goransh-walia:
This looks good to me at a glance — thanks for putting it together.

boscocp:
Thanks for taking a look — appreciated.

Since it was a glance: the part most worth a close read is the detection boundary itself, i.e. what the agent treats as evidence that the provider truncated the prompt rather than that the model simply answered short. That's the judgment call in this PR; everything else follows from it.

The branch is still `9bc4a6c4` and mergeable against current `main-v2` — happy to rebase if you'd rather re-verify against the latest tree. Whenever you have time for a closer pass, I'd welcome it.

Issue #10192, opened 2026-09-12
[Bug]: v1.38.7 - update CheckUpdate blocks the bridge 1.4-3.5s and re-fires on window events (hidden/visible, maximise)


Aresitoo:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.38.7 (build `036c7c50c5c1`)

### What happened?

`update CheckUpdate` blocks the bridge for **1.4-3.5 s per call** and is re-fired on window events (`[view] hidden` / `[view] visible` toggles, `window IsMainWindowMaximised`), so switching to/from the app window stalls the UI and floods the bridge with multi-second calls.

Measured in a crash-dump breadcrumb window (the same session in which a React #185 crash fired — see #10186):

```
[bridge] update CheckUpdate done ms=1403
[bridge] window IsMainWindowMaximised done ms=1480
[bridge] update CheckUpdate done ms=3324
[bridge] update CheckUpdate done ms=1620
[bridge] window IsMainWindowMaximised done ms=773
[bridge] update CheckUpdate done ms=2814
[bridge] window IsMainWindowMaximised done ms=98
[bridge] update CheckUpdate done ms=2548
[bridge] window IsMainWindowMaximised done ms=1765
[bridge] update CheckUpdate done ms=3501
```

Over the same ~50 s the window toggled visibility 8 times (`[view] hidden` -> `[view] visible`), and `CheckUpdate` ran on nearly every toggle.

Why this looks like a defect rather than a slow network day:

- The calls are **serialized on the bridge** and each holds it for seconds, so the UI cannot respond while a check is in flight — which is what the timings show.
- The trigger is a **local UI event** (window visibility / maximise), not a user-initiated update check, so alt-tabbing a few times costs the full stall each time.
- In this session the burst immediately preceded a React error #185 renderer crash (#10186), consistent with the commit-cascade framing in #10177: each bridge completion re-renders state, and rapid completions feed the cascade.

Trigger note for #10177: the crash it fixes is described as "stream + scroll"; a third path observed here is window visibility flapping plus this `CheckUpdate` churn, with no streaming, scrolling or hydration involved.

### Steps to reproduce

1. Run Reasonix Desktop v1.38.7 (stable) on Windows with a session open.
2. Switch away from the app window and back a few times (or minimise/restore) so `[view] hidden` / `[view] visible` toggles repeatedly.
[...]

Issue #10186, opened 2026-09-12
[Bug]: v1.38.7 渲染层崩溃 Minified React error #185（Maximum update depth exceeded）；长会话下流式输出停摆


fluoxetine616:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.38.7

### What happened?

两个问题（第二个可能相关、也可能独立，请官方判定）：

1)【渲染层崩溃】窗口捕获 [window.error]: Minified React error #185（= "Maximum update depth
exceeded"，React hooks setState 无限更新循环；见 react.dev/errors/185）。崩溃栈指向应用自身
bundle（index-QvpapSZM.js:2:589949 → dispatchSetState → enqueueConcurrentHookUpdate）。
发生时机：应用启动约 4 分钟后；崩溃前 tab.hydrate ancillary startup 耗时 27351ms
（其中 ancillary context 14600ms、ancillary effort 26496ms），期间窗口有 [view] hidden→visible
切换；最后一条 breadcrumb 为 CheckUpdate done。崩溃所在的 tab 为普通小会话（非大会话）。
重启后一切正常，会话数据无损。

2)【流式输出停摆】长会话中，任务执行正常（后端持续产出、会话事件持续落盘），但界面流式
输出/思维链区域停止刷新；退出重进后内容完整可见（从事件流重放）。2026-09-12 当天因该问题
重启客户端 7 次以上；其中两次关闭时日志出现 desktop/shutdown timed out after 10000 ms。
本机背景：存在 2026-07 创建的超长会话（经 18 次 recovery 重建、events.jsonl 达 24MB、
上下文曾达 1,107,745 tokens 触发 compaction failed）。

期望行为：流式输出持续可见；渲染层不崩溃。

### Steps to reproduce

1. Windows 11 上运行 Reasonix Desktop v1.38.7（stable），数据目录含较大历史
   （本机多个长会话，最大 events.jsonl 24MB）
2. 启动应用（自动恢复工作区 tab，本机 tab id: tab_80132a7482d4baf7876724c427c0da3e）
3. 等待 hydrate 完成（约 27 秒；ancillary effort 约 26.5 秒），期间窗口发生 hidden→visible
4. 约 4 分钟后渲染层抛 React #185 崩溃
（另一问题复现：在长会话中持续运行任务较长时间后，界面流式输出停止刷新）

### OS / platform

Windows 11

### Relevant logs or output

```shell
[window.error]

[...]

PR #9870, opened 2026-09-07
feat(i18n): add Spanish (es-419) locale support

GuidoMaxier:
## Summary

Adds Spanish (es-419, neutral Latin American) interface support to both layers, rebased onto the current `main-v2` (v1.38.7).

- **CLI/kernel**: full Spanish catalogue in `internal/i18n/messages_es.go` (field-identical to the English baseline, including the `ReadStatus*`, `OperationNeedsUser`, `SearchModelUnavailable` and `ProviderErrNotFound` fields added upstream after the original revision), `ArgLanguageEs` in every locale, `setLanguage`/`normalize` mapping for `es`/`es-ES`/`es-419`/POSIX forms, a selectable `/language es` option (CLI + slash completion), and `config.SetLanguage` persistence for `es`. `SetLanguage`/`SetReasoningLanguage` live in `internal/config/edit_language.go` to stay inside the repolint file-size ratchet.
- **Desktop**: complete `desktop/frontend/src/locales/es.ts` dictionary — **3,234 keys, verified 1:1 against `en.ts`** (same key set, same order, same `{placeholder}` names) — plus the locale wiring (`Locale` union, `SPINNER_WORDS.es`, `detectLocale`/`preloadLocale`, settings picker with an "Español" option, tray locale). Auxiliary per-locale dictionaries are retyped to `Record<Exclude<Locale, "es">, ...>` and keep their existing `?? en` fallbacks until Spanish copy lands there.
- **Bundle budgets**: the gate now enumerates the `es-*.js` lazy chunk (its filter previously matched only `zh`/`zh-TW`, so the Spanish chunk went unmeasured) and budgets it at its measured size: 58,071 B gzip (56.710 KiB), 56.8 KiB ceiling. The zh/zh-TW dialect ceilings are unchanged.
- English remains the system-prompt and model-behavior language; the new catalogues only affect client-side UI strings (i18n scope is documented as such in the package).

## Verification (run locally on this rebased revision)

- `gofmt`, `go vet ./...` clean
- `golangci-lint run ./...` (CI pin) -> 0 issues
- `go run ./tools/repolint` -> clean
- `go test ./internal/i18n/ ./internal/config/ ./internal/control/ ./internal/cli/ ./internal/tool/builtin/` pass
- desktop/frontend: `tsc --noEmit` 0 errors, `eslint "src/**/*.{ts,tsx}"` clean, `pnpm build` passes (vite + bundle budgets, including the measured `es-*.js` chunk)
[...]

PR #10250, opened 2026-09-13
fix(desktop): coalesce the update checks fired by window focus and visibility

GuidoMaxier:
`subscribeToUpdateRefresh` in `desktop/frontend/src/components/UpdateBanner.tsx` subscribed `focus` and `visibilitychange` to a bare `refresh()`, so every toggle of the app window started a new `CheckUpdate`.

That check holds the bridge for seconds — 1.4–3.5 s in #10192's own measurements, 28.5 s in one of the breadcrumb windows reported on #10186 — so switching between windows queued one multi-second call per toggle. `useUpdater` only guards against superseding a check that is *in flight* (`isBusyOperation`); once a check finishes, the next window event starts another.

## Change

Throttle the window-event path to 5 minutes. The hourly interval keeps its own schedule, so the periodic check is unaffected.

- `UPDATE_REFRESH_MIN_EVENT_INTERVAL_MS = 5 * 60 * 1000`, exposed as an optional third parameter so callers and tests can tune it.
- `visibilitychange` for a hidden window is still ignored, as before.
- No Go side and no bridge-contract change: `CheckUpdate` itself behaves exactly as it did.

## Test

`desktop/frontend/src/__tests__/updater-shared-state.test.tsx` already pinned the old behaviour:

```ts
window.dispatchEvent(new Event("focus"));
document.dispatchEvent(new Event("visibilitychange"));
ok(scheduledRefreshes === 2, "visible focus and visibility changes schedule update refreshes");
```

Two events produced two refreshes — the defect written down as an expectation. It now asserts coalescing inside the throttle window, a new refresh once that window elapses, and that cleanup still removes both listeners.

## Verification

- `pnpm exec tsx src/__tests__/updater-shared-state.test.tsx` → 54 passed, 0 failed
- `pnpm typecheck` → 0 errors
- `pnpm test:typecheck` → 0 errors
- `pnpm lint:hooks` → clean

Refs #10192

Documentation-impact: none - the change is a client-side event throttle in `UpdateBanner.tsx`; the user-facing updater documentation describes when checks happen (at launch and periodically), and none of it documents a per-window-event refresh, so the existing docs remain correct.


GuidoMaxier:
**CI evidence.** This is a fork pull request, so the checks of this PR stay in `action_required` until a maintainer approves them. I ran the same commit inside my fork instead: **30 jobs green** on run [`34790889950`](https://github.com/GuidoMaxier/DeepSeek-Reasonix/actions/runs/34790889950), including `desktop-frontend`, `lint`, `lint-code`, `desktop`, `desktop-browser`, `desktop-browser-group (transcript / app-settings-motion)`, `test` (ubuntu/macos/windows), `race`, `desktop-go`, `desktop-go-race`, `desktop-macos`, `desktop-windows` and every `windows-*` job.

The only red job is `desktop-windows-go` (`reasonix/desktop`), which fails on the Windows runner of the fork with:

```
warning: bash sandbox requested but unavailable on this host; refusing to run unconfined.
The native Windows restricted-token/AppContainer sandbox is unavailable.
```

That job is not related to this diff. An unrelated branch (`feat/es-spanish`, the es-419 locale work in #9870) run in the same fork failed the **same single job** with the same message while every other job passed in both runs — see [run `34788173613`](https://github.com/GuidoMaxier/DeepSeek-Reasonix/actions/runs/34788173613).

Local gates on this revision: `pnpm exec tsx src/__tests__/updater-shared-state.test.tsx` 54 passed / 0 failed, `pnpm typecheck` and `pnpm test:typecheck` 0 errors, `pnpm lint:hooks` clean.

PR #10301, opened 2026-09-15
fix(cli): stop the headless run dropping a reasoning-only answer / headless 运行不再丢弃只有推理的回答

boscocp:
**简述**：思考型模型可以用非空 reasoning、空可见消息结束一轮——这是被刻意接受的形态。转录渲染器对它至少会打印一个 "thinking" 标记，加 `--show-thinking` 还会打印推理正文。headless 两者都没有：`runOutputSink.Emit` 只保留 `e.Text`，于是 `-p` 什么都不打印，`--output-field json` 给出 `result: ""` 同时 `subtype: "success"`、`is_error: false`、退出码 0。调用方无法把它与「这一轮什么都没产出」区分开。本 PR 让 `result` 在可见消息为空时回退到本轮推理文本——这正是本仓库对同一形态在别处已采用的办法——并用 `result_from_reasoning` 标明来源，避免把推理冒充成模型从未输出的可见文本。

## The defect

`internal/cli/run_output.go`, before:

```go
if e.Kind == event.Message {
	s.final = e.Text
}
```

`event.Message` is documented as carrying both: "Text holds the full answer and
Reasoning the full chain-of-thought" (`internal/event/event.go`). The sink keeps
one and drops the other.

For a turn that answered in the reasoning channel, the surfaces disagree about
whether anything happened:

| surface | what the caller gets |
|---|---|
| `reasonix run "…"` (transcript) | a dim `▎ thinking` marker; the reasoning body only under `--show-thinking` (`internal/agent/textsink.go:66`) |
| `reasonix -p "…"` | nothing at all, exit 0 |
| `--output-format json` | `"result": ""`, `"subtype": "success"`, `"is_error": false` |
| `--output-format stream-json` | the reasoning, already, unconditionally (`internal/eventwire/wire.go:81`) |

The transcript at least signals that a thinking phase ran. The two sinks this PR
touches signal nothing, and nothing distinguishes them from a turn whose answer
was genuinely empty. `stream-json` shows the asymmetry is not a policy about
withholding reasoning from automation - that surface carries it already.

### Measured

A local provider, one turn, `--output-format json`. The turn's entire output was
a malformed tool call the model wrote as prose in the reasoning channel rather
than emitting as a tool call:

```
seq  68  text=   0 chars  reasoning=  392 chars
seq 348  text=   0 chars  reasoning= 1680 chars
seq 367  text=   0 chars  reasoning=    84 chars
   reasoning: '<|tool_call>call:bash{command:<|"|>mkdir -p docker<|"|>}<tool_call|>'
```

Those 84 characters were the only thing the turn produced. The run reported
`outcome: success`, `is_error: false`, `result: ""`, exit 0, and wrote no files.
A script gating on the exit code would have recorded the work as done.

## The fix

Fall back to the turn's reasoning when the visible message is empty, and say so.

That fallback is not a new idea here — it is the remedy this repository already
applies to the same provider shape in the goal evaluator
(`internal/boundedllm`, #9679/#9785) and in the compaction summarizer (#9300),
both of which accumulate reasoning when the visible content is empty. This
applies it at the one surface where the alternative is printing nothing.

`result_from_reasoning` is additive and `omitempty`, so an ordinary answer's
JSON is byte-identical to before. It exists because silently promoting
chain-of-thought into `result` would tell the caller the model emitted visible
text it never emitted.

The answer and its provenance are a single value rather than two fields beside
each other:

```go
type finalAnswer struct {
	text          string
	fromReasoning bool
}
```

They are only ever correct together — a `fromReasoning` left over from an
earlier turn would misreport where a later visible answer came from — and
`repolint` independently asked for the grouping, since a thirteenth scalar field
on `runOutputSink` is over its ceiling.

### What this does not do

- **It does not reclassify a reasoning-only turn as an error.** That shape is
  accepted on purpose (#6618) and `RequireVisibleFinal` already exists for the
  callers that need the stricter contract (`internal/agent/agent.go`), set by
  the guardian and by subagent tasks. Turning it on for `reasonix run` would
  reverse a deliberate decision, not fix a defect.
- **It does not touch the Plan/Goal side.** #8023 owns extending the
  visible-final contract to Controller Plan and Goal workflows, and #9679/#9785
  own the evaluator. This PR touches `internal/cli` only; none of those three
  changes a file in that package, so they do not overlap.
- **It does not gate the fallback on `--show-thinking`.** `buildRunSink`
  (`internal/cli/run_sink.go:29`) takes that flag and passes it only to the
  transcript sink; `newRunOutputSink` never receives it, so on `-p` and the
  structured formats it has never had an effect. Gating on it would leave `-p`
  without the flag printing nothing, which is the defect. The flag stays
  honoured where it is honoured today, and untouched here.
- **It does not change the transcript renderer.** There `--show-thinking` is a
  real user choice that is respected, and a marker already tells the reader a
  thinking phase ran. Printing the chain-of-thought there without the flag would
  override a decision the user made.
- **It does not change `events-jsonl`.** That surface is deliberately
  content-free and carries no result text at all.
- **It does not trim or summarise the reasoning.** `result` already carries
  whatever the model produced; a 48 KB visible answer was equally possible
  before this change.

## Verification

**Teeth check.** With the fallback removed, so the sink behaves as it does on
`main-v2`:

```
--- FAIL: TestTextOutputFallsBackToReasoningWhenNoVisibleText (0.00s)
    run_output_reasoning_only_test.go:33: text output = "", want the reasoning-only turn rather than nothing
--- FAIL: TestJSONResultCarriesReasoningAndSaysSo (0.00s)
    run_output_reasoning_only_test.go:49: result = "", want the reasoning-only turn
FAIL	reasonix/internal/cli	0.018s
```

Those two empty strings are the measured defect, reproduced without a provider.

Five tests, covering the fallback and the cases that must not move:

```
TestTextOutputFallsBackToReasoningWhenNoVisibleText   -p prints the turn instead of nothing
TestJSONResultCarriesReasoningAndSaysSo               result carries it, result_from_reasoning is true
TestVisibleTextIsUnchangedByTheFallback/text only     ordinary answer untouched, field omitted
TestVisibleTextIsUnchangedByTheFallback/text beside reasoning
                                                      visible text still wins when both are present
TestLaterVisibleTurnClearsTheReasoningProvenance      a later visible turn resets the provenance
TestErrorResultIsNotMarkedAsReasoning                 an error Result is the run's message, never the model's
```

The two "unchanged" cases assert `result_from_reasoning` is absent from the
encoded JSON, not merely false, so `omitempty` is part of the contract under
test.

Full gates:

```
GOTOOLCHAIN=go1.26.6 go test ./internal/... ./cmd/...   all pass
GOTOOLCHAIN=go1.26.6 golangci-lint run ./...            0 issues
GOTOOLCHAIN=go1.26.6 go run ./tools/repolint            clean (1228 baselined findings)
gofmt -l internal/cli/                                  clean
go vet ./internal/cli/                                  clean
git diff --check origin/main-v2                         clean
```

Cache-impact: none - `internal/cli/run_output.go` renders the run's result after
the turn has finished. No prompt text, tool schema, system prefix or
`promptCacheKey` input is touched, and the provider-visible request is
unchanged.

Documentation-impact: updated - `docs/CLI.md` and `docs/CLI.zh-CN.md` both
document the structured result object, so `result_from_reasoning` is added to
the sample and to the field notes in each. The two files are kept in step.


boscocp:
Rebased onto current `main-v2` (`62862f48a`); head is now `076802e0a`. The change itself is unchanged.

The conflict was with `9eb3ccc50`, which added the authentication metadata (`error_code`, `authentication_status`, `recovery_actions`) to the same `runResult` literal this PR reformats. The two changes are independent — that one records why a run failed, this one records where the answer came from — so the resolution is the union of both field sets, with nothing dropped on either side.

`go build ./...` and `go vet ./internal/cli/...` are clean, and `go test ./internal/cli/...` passes, including the five tests this PR adds.

Issue #10226, opened 2026-09-13
[Bug]: Inline base64 images bloat session.events.jsonl to 146 MB, making the session/project unopenable


AerospaceWalker:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

1.38.7

### What happened?

A single session in my ComfyUI project grew to **146 MB**, and now clicking that project in the desktop app **fails to open the session** — it silently falls back to the previously active tab (appears to "jump to an old conversation"). The project folder and images themselves are fine.

**Root cause** (from analyzing the session files):

`projects/d--software-comfyui-ai工作/sessions/20260913-053823.494093500-session.events.jsonl`
= **146 MB, but only 1447 lines**:

| content | lines | size | share |
|---|---|---|---|
| lines containing `"images":["data:image/png;base64,..."]` | 76 | **141.0 MB** | **96%** |
| lines without images (actual conversation) | 1371 | 5.3 MB | 4% |

So **~96% of the session file is inline base64 PNG payloads returned by tools** (image-generation outputs). Every tool result stores the image as a full base64 data URI inside the append-only event log. After ~76 generated images in one session, the events file balloons and the session becomes unloadable.

Other files in the same session are also large: `...session.jsonl` ≈ 30 MB, `...session.transcript-projection.json` ≈ 528 KB.

**Expected:** tool-returned images should not be inlined in full as base64 in the session event log. They should be stored as **file references** (path + hash/thumbnail) or capped/truncated, so a single session cannot grow to hundreds of MB and break loading.

**Impact:** the affected session/project can't be opened in the desktop UI, and the on-disk session data is effectively impossible for the user to prune safely.

**Environment:** Windows 11 (10.0.26200.9445), Reasonix desktop 1.38.7 (v2 — Go rewrite, main-v2).

### Steps to reproduce

1. Create/open a project that generates images via a tool (e.g. ComfyUI image generation driven by Reasonix).
2. In a single session, generate several images; each tool result returns an image.
3. Observe that each returned image is written inline into `session.events.jsonl` as a full base64 data URI (`"images":["data:image/png;base64,iVBOR..."]`).
4. After ~76 generated images, the session's `events.jsonl` reaches ~146 MB.
5. Restart the app and click that project/session → the session fails to load and the UI falls back to the previously active tab (looks like it "jumps to an old conversation").

[...]

Issue #10031, opened 2026-09-09
[Bug]: Image validation failure leads to HTTP 400 from API and subsequent poisoning of session


zukunftsfirmaDE:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

reasonix v1.36.0

### What happened?

[We are new to Reasonix] 

When we started to instrument Reasonix and run our testsuite, we encountered the issue that passing a signature valid, but defective png file (attached), it would poison the session and not be recoverable without clearing. See steps to reproduce.

We therefor with the help of AI diagnosed the issue and I will attach a patch that rejects invalid images. There might be other paths that need to be fixed, we deliberately only fixed what we could actually test afterwards (OpenAI API path).

The patch requires no documentation, and has no impact on Cache. Build tests have passed. 

The AI provided this summary of its work:

A provider rejects the entire chat request when any attached image cannot be decoded, so one unreadable screenshot surfaced as `Malformed request (HTTP 400) ... This is likely a bug` and the turn was lost.
 
- Images were validated only as data-URL *shape* (`ClassifyImage`/`ParseImageDataURL`); the bytes were never decoded. The check now happens at OpenAI-compatible request serialisation, which is the single point that user attachments, MCP tool results and replayed session history all pass through.
- 
- Undecodable attachments are withheld and replaced by a short provider-visible notice, so the model continues and can ask for the file again.

[artifact.zip](https://github.com/user-attachments/files/32008022/artifact.zip)

### Steps to reproduce

1. Run a test suite with an signature valid but defective png file
2. reasonix will pass this on to the model, the harness will respond with 400
3. reasonix will fail with:

error: modelname · Chat Completions: Malformed request (HTTP 400): the request body was rejected. This is likely a bug — please report it if it persists.                                                                                                                                                                                                                                                                                                                                    
{"object":"error","message":"An exception occurred while loading IMAGE data at index 0: Error while loading data ImageData(url='data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAIAAAACCAIAAAD91JpzAAAAEUlEQVR4nGP4z8A...: unrecognized data stream contents when reading image file","type":"BadRequestError","param":null,"code":400} 

4. And keep attaching the file to following request, basically rendering the session unusable


### OS / platform

Debian 13 Linux

### Relevant logs or output

```shell
[...]

PR #10342, opened 2026-09-15
Price official vision with v41 tokens and offload over-budget images / 按 v41 计价官方视觉并在超预算后省略图片

SivanCola:
Official DeepSeek vision sessions priced images as free in compact/window estimates, encoded new attachments to a 1568 long-edge instead of the published v41 grid, and had no durable omission when a request exceeded the route image budget.

## Fix

- Port the DeepSeek v41 image-token calculator and request dimensions (14px patch, 3:1 downsample, 544×544 pricing floor, 1024 cap, 4096 long-edge).
- Encode **new** attachments to that grid with JPEG 85/75/60 toward 2 MiB. Stored historical data URLs are left unchanged.
- Add optional `image/offload` events that rewrite `ModelMessages` only. UI history keeps the original images. Identity is `messageId` + image index (Reasonix has no DSH surface `seq`).
- Check retained input images against 600 / Files 128 MiB / inline 20 MiB with count quantum 20 and byte quanta 64/10 MiB. On overflow, record the oldest occurrences, rebuild the request, and retry without a provider HTTP retry. Compaction summaries use the same recovery.
- Do not drop images that still fit. Tool schemas and the system prompt are unchanged.

## Cache

Cache-impact: medium - accurate v41 image tokens can trigger compaction earlier on image-heavy sessions (compaction resets the prefix). Offload itself is a one-time prefix change only after a route budget failure, then stable. Tool schemas and the system prompt are unchanged. Nine in-budget images are no longer omitted proactively, which is better for cache than keep-8.

Cache-guard: go test ./internal/provider/ -count=1 -run 'DeepSeekImage|RequiredImage|CheckRetained|ImageOffload'; go test ./internal/agent/ -count=1 -run 'ImageOffload|EstimatedPromptTokensAdds|SamplingRecoversImage|PromptCalibrationDoesNotDouble'

## Compatibility

| Field or format | Old-data behavior | New-reader behavior | Previous-reader behavior | Conclusion |
| --- | --- | --- | --- | --- |
| Session envelope / `storageRevision` | Unchanged | Opens as before | Opens as before | No format upgrade |
| `image/offload` (`optional: true`) | Absent | Applied to `ModelMessages` | Unknown optional events are skipped; images may be resent | Old binaries can open the session |
| `image-offloaded` sentinel | Absent | `ClassifyImage` → not on the wire | Unknown ref dropped | Safe |
| Historical data URLs | Unchanged | Sent as stored | Sent as stored | Compatible |
| New attachment encoding | Old 1568 long-edge | v41 grid + 4096 + JPEG ladder | N/A | New writes only |

## Verification

- `go test ./internal/provider/ ./internal/session/ ./internal/agent/ ./internal/control/ -count=1`
- `go vet` on those packages
- `go run ./tools/repolint`

Documentation-impact: none - CHANGELOG Unreleased records v41 vision pricing/encoding and budget-triggered image/offload; no docs/*.md described the old 1568 long-edge or free image tokens


wiglif9999:
## Images accumulate in a session and brick it — the image budget has no count cap and no in-app recovery

**Version:** Reasonix Desktop **v1.38.7** (build `036c7c50c5c1`), Windows 11
**Related:** #10226 (inline base64 bloat → session unopenable), #10457, #4210 (size-only downscale), #10031 (HTTP 400 poisoning)

### Reproduction — image **count**

1. In one long session, attach screenshots over several turns (one pasted/attached image each).
2. From the **5th stored image** on, **every** request fails with `more than 4 images` and the session can no longer be continued.
   - The 4-image budget is evaluated over the **whole history**: stored images stay in every subsequent request — it is *not* a per-message limit.
   - Measured locally: 5 stored images (`"images": ["data:image/png;base64,…"]` on 5 messages) → session unusable; another session had **18** stored images.
3. There is **no in-app way out**: no indicator (`x/4 images used`), no "drop/offload oldest image" action.
   Recovery is only possible with the **app closed** by editing the session files externally — and even then `<id>.jsonl` is regenerated from `<id>.events.jsonl`, so the event log has to be reset as well.

### Same root cause, second failure mode — image **size**

Inline base64 images are persisted **twice** (`.jsonl` plus the authoritative `.events.jsonl`). One local session reached

```
events.jsonl = 136,054,883 bytes
refusing unsafe event-log replay … resource=encoded_bytes … limit=134,217,728
```

→ just over the replay budget → the session can no longer be opened at all (fail-closed; files are left untouched, so nothing is lost, but the user is locked out of the session).

### Proposal (extends this PR's `image/offload`)

1. Evaluate the image budget by **count** as well as by tokens, and make **offload** the automatic behaviour for over-budget images.
2. **Offload after the turn that consumed an image** — an attached image is only needed for the request(s) that read it; afterwards the transcript text is enough.
3. When offloading, replace the image with a **short text summary produced by the vision model**. The `vision_model` / text-to-vision machinery already exists (used for text-only models) and could be reused for the session's own vision model: the information survives, no image slot is consumed, and `events.jsonl` stays small.
4. UI: show `x/4 images used` and a one-click "offload oldest images".
5. Store images out-of-band (file + reference/summary, cf. #10346) instead of inline base64 inside the event log.

### Local workaround (for reference)

With the app closed: (a) drop the oldest images from `<id>.jsonl` and reset the sidecars so the logs are rebuilt from it, or (b) replace each `images` entry with a vision-model text summary and clear the field. Both are only workarounds — a supported path (in-app, or `reasonix doctor session …`) would be much better.

Not a duplicate of #10226 (that is the bloat/size chain) or #10031 (HTTP 400 poisoning): this is the **count** budget plus the missing in-app recovery.

PR #10414, opened 2026-09-16
Show local history before runtime activation / 本地会话历史先于运行时显示

SivanCola:
Local session navigation currently waits for runtime/controller setup before making history readable. This change starts a bounded canonical history read as soon as the target tab identity is available, keeps local draft input mounted, and leaves send/control actions behind runtime readiness. Runtime failure preserves an already-readable transcript; remote transition behavior remains unchanged.

The transcript store retains bounded resident windows across single-surface navigation. Lazy body reads now allow one generation handoff within the same resident owner, reject close/reopen completions, and wait for replacement pages before reading their references. Navigation diagnostics distinguish identity, history, runtime, composer, and paint milestones.

### Validation

- Frontend production/test typechecks, production build, repository lint, lifecycle, transcript, and remote suites pass locally. The full CI frontend runner completed all 433/433 commands on the latest code.
- Exact-head local verification also passed the headless transcript reader browser benchmark, root Go suite, and Desktop Go suite (`go test -timeout=20m ./...`).
- Live PR CI on `62ba7e693` completed with 41 successful and 7 intentionally skipped checks, including both transcript browser modes, all three App memory shards and aggregate, Desktop frontend/Go/race/macOS/Windows jobs, Windows Go groups, lint, and CodeQL.
- Focused navigation coverage: 100 tab-switch, 42 topic-activation, and 13 same-path rebind assertions. Deterministic content tests cover binding replacement, Follow replacement, close/reopen, pending-page replacement, and bounded repeated churn.
- Browser integration passed: 20 alternating local session switches had a renderer first-readable-paint P95 of 100.1 ms against a 300 ms gate. It uses the browser mock backend, not Electron with the native backend.

### Draft / review hold

The first CI run exposed a same-path session-generation regression. Commit 2b6a75aef fences resident owners with the shared session binding key before early reads or Follow installation and stops the old follower on replacement.

The next CI run exposed two validation-owner defects rather than product retention: bounded transcript retries shared a browser process with the retained 1000-turn primary page, and App memory screening classified a settled listener decrease (346 to 343) as retention. Commit a83a74aaa gives retries a dedicated browser process and makes the retention classifier directional while preserving growth and unsettled-tail blockers. Deterministic regressions cover both mechanisms, and the replacement matrix is green.

This PR does not yet establish the full native fast-switch acceptance target:

[...]

PR #10409, opened 2026-09-16
Decouple durable local session operations from runtime tabs / 本地会话持久化操作与运行时解耦

SivanCola:
## Summary

Decouple durable local session operations from open runtime tabs across canonical and legacy sessions.

- resolve sidebar operations by explicit canonical `SessionRef (hostId + sessionId) > sessionPath > topicId`; `topicId` is only the legacy/topic compatibility fallback
- make canonical titles use title-event CAS and legacy titles use opaque `title_revision` tokens
- run cold AI title generation through target-scoped provider-only boot, including extension providers
- add explicit-target rename, archive, restore, move, delete, bounded history/search/content/location, fork, and full-history copy APIs
- adopt legacy move/copy through the existing migration and workspace journals without opening a controller
- make fork/copy publication idempotent and lifecycle-guarded
- keep per-target frontend loading/error state and prevent background operations from replacing the active transcript or draft
- emit target-bound metadata/lifecycle events while retaining compatibility project-tree notifications

## Compatibility

- canonical `session/title` payloads are unchanged; the latest title event sequence is the CAS token
- legacy `BranchMeta` adds optional `title_revision`; older files remain readable
- mixed-version writers that preserve an unknown old revision while performing A→B→A cannot be detected universally
- provider-visible title prompt, three-user-message input policy, main conversation messages/tools, and cache prefix are unchanged
- remote sessions keep their existing paths; unsupported local routing returns a structured capability error rather than entering the local resolver

## Verification

Passed locally:

- `go test ./...`
- focused Desktop target/title/lifecycle/history/fork/copy suites
- `go test ./desktop/internal/workspacestate` equivalent from the Desktop module
- focused race tests for legacy title metadata and canonical title/fork/copy paths
- Desktop generated contract and command-owner freshness tests
- frontend typecheck, hooks lint, production build, transcript suite, session-title tests, project-tree archive behavior test
- exact Desktop `golangci-lint`, repository `repolint`, and regenerated Desktop inventory freshness
- Playwright sidebar AI rename scenario (per-target loading, reverse completion, sanitized errors, stable transcript/draft)

[...]

PR #10423, opened 2026-09-16
Fix canonical history reuse across reopened tabs / 修复重开会话的历史缓存复用

SivanCola:
## Summary

Follow up on #10414 to close the two remaining P1 cache-identity and reopen gaps identified during review.

- derive canonical tab history fingerprints from the canonical session read model (`snapshotSequence` + `storageGeneration`) instead of legacy branch-sidecar metadata;
- keep canonical transcript residents under stable `SessionRef + sessionGeneration` ownership while treating tab IDs as ephemeral bindings;
- rebind a warm resident when the same canonical session reopens under a newly allocated tab ID;
- increment the binding generation and fence stale follower events and lazy-content completions during rebinding;
- retain path-only legacy behavior without enabling cross-tab reuse.

## Root cause

Canonical history windows and `TabMeta` were using different identity domains, so a valid warm resident could be rejected even when the canonical session had not changed. Separately, the renderer keyed resident ownership by `tabId`, but single-surface navigation prunes the old tab and allocates a new tab ID when reopening the same session. That deleted or orphaned the only reusable resident.

## Verification

- `go test ./...` — passed
- `go run ./tools/repolint` — passed (`1230` baselined findings)
- `cd desktop/frontend && pnpm test:ci-contract` — passed
- `cd desktop/frontend && pnpm test:transcript` — passed (`361` tests)
- `cd desktop/frontend && node scripts/run-ci-tests.mjs` — passed (`435/435` commands)
- `cd desktop/frontend && pnpm build` — passed, including bundle budgets
- `cd desktop/frontend && pnpm exec tsc --noEmit -p tsconfig.test.json` — passed
- focused tab-switch rebind tests — passed (`13` tests)
- focused tab-switch hydration tests — passed (`100` tests)
- `cd desktop && go test -timeout=20m ./...` — passed
- `cd desktop/electron && pnpm smoke` — passed against the real Electron shell and built Go service

The deterministic regressions cover canonical-versus-legacy fingerprint mismatch, missing canonical identity as a safe miss, backend pruning plus new tab-ID allocation, warm `A → B → A` reuse with zero history reads, fingerprint invalidation, same-path/different-session isolation, stale follower events, stale lazy-content completions, and resident-count bounds.

## CI follow-up

The first exact-head Windows `Q-Z` shard exposed a pre-existing test-ordering race in `TestRestoreGlobalTopicSessionReindexesProjectTree`: catalog publication could precede completion of the asynchronous legacy branch-meta write, so Windows could reject the immediate archive move with a sharing violation. The test now waits for the migrated sidecar identity it actually depends on before invoking `TrashTopic`. The focused case passed 20 consecutive local runs.

The replacement Windows `E-H` shard then showed that `TestEnsureBlankTabConcurrentModelSwitchKeepsLastSelection` used a five-second wall-clock bound for a full controller rebuild even though it validates last-click ordering rather than latency. Its deterministic handoff now uses the desktop suite's 30-second lifecycle bound and reports an unexpected early completion directly. The focused concurrency case passed 10 consecutive local runs. The Windows partition contract remains green, and no test or platform coverage was removed.

## Base reconciliation

- merged current `main-v2` through #10409 into this branch without rebasing or force-pushing;
- resolved the generated Desktop migration inventory conflict by regenerating both inventory files rather than editing generated output manually;
- retained the canonical fingerprint and stable transcript-rebinding semantics while extracting the new fingerprint helper and restore regression into appropriately owned files;
- verified the reconciled tree with focused Desktop regressions, the full Desktop Go suite, all `435/435` frontend CI commands, the production frontend build, root `go test ./...`, Desktop inventory validation, and clean repository lint.

## Documentation impact

Documentation-impact: updated - Regenerated the Desktop migration inventory for the new transcript binding helper and shifted source locations.

## Compatibility and scope

- No persisted format changes.
- No provider-visible prompt, request, tool schema, or cache-prefix changes.
- No cross-process payload changes.
- Native Electron `A → B → A` latency/P95 is still unmeasured; this PR proves the ownership and reuse mechanism deterministically but does not claim native latency qualification.


SivanCola (commented):
Post-reconciliation self-review completed on exact head `9bdfade86` against `41d2ae37f`.

No actionable correctness, security, concurrency, compatibility, or malicious-code findings remain. The merge preserves the two intended invariants: canonical fingerprints stay in the canonical read-model identity domain, and canonical transcript residents are owned by stable session identity while stale tab-bound completions are generation-fenced. Path-only compatibility remains tab-scoped; no persisted format, provider-visible prompt/request, or cross-process payload changed.

The generated Desktop inventory was regenerated after merging #10409, repository lint is clean, the worktree is clean, all exact-head checks are terminal and green, and the two original P1 review threads on #10414 have been replied to and resolved. Native Electron `A -> B -> A` P95 is still unmeasured and is not claimed by this PR.

PR #10431, opened 2026-09-16
fix(agent): detect perseveration loops and retry once with a nudge / 检测 perseveration（无意义重复）循环并在提示后重试一次

BuGlessRB:
## Summary

Add a client-side guard for perseveration (= mindless repetition): a model stuck emitting the same short block of text or reasoning without ever calling a tool. The existing guards key on tool rounds or on silence, so a chatty loop looked healthy while it burned the entire output budget.

The guard watches the rolling tail of each streamed text/reasoning delta and fires when a short block repeats unmistakably (byte-identical, >=8 repeats, >=160 bytes, non-whitespace). On detection it cuts the stream and settles the turn cleanly. The final-response path then appends a `[retrying (N) avoiding perseveration]` nudge and retries once by default; a second consecutive loop with no successful response in between stops and waits for the user. The retry budget is configurable via `Options.MaxPerseverationRetries` (nil = default 1, `*0` = no retry, `*N` = N retries).

Surfaces a `perseveration_loop` notice (localized en/zh/zh-TW) and documents the guard in the guide. Several test fixtures that used `strings.Repeat` as a size shortcut now trip the guard and were switched to non-repeating filler.

## Verification

- `go test ./internal/agent/ -run 'Perseveration|StreamAbortsOnPerseverationLoop|RunRetriesOnceThenStopsOnRepeatedLoop'`
- `go build ./...`
- `go run ./tools/repolint` (clean)

## Cache impact

Cache-impact: none - the provider-visible system prefix and tool schemas are byte-identical; the guard only inspects streamed deltas and appends a retry nudge to the turn tail, never the cached prefix.
Cache-guard: `go test ./internal/agent/ -run 'Perseveration|StreamAbortsOnPerseverationLoop|RunRetriesOnceThenStopsOnRepeatedLoop'`.

## Documentation impact

Documentation-impact: updated - documented the perseveration guard in docs/GUIDE.md and docs/GUIDE.zh-CN.md


baeut:
### Independent field report: single-rune CJK perseveration reproduces on v1.38.11

Confirming this failure shape on the current release, with two factors that may be worth folding into the detector design. Posting here because this PR targets the exact shape (same short block of text/reasoning repeated without ever calling a tool); #8085 covers the same class from a different angle.

**Environment**

- Reasonix Desktop v1.38.11, Windows 11
- `deepseek/deepseek-v4-flash` (OpenAI-compatible endpoint)
- `[agent]` config: `temperature = 0.0`, `reasoning_language = "zh"`, `context_window = 1000000`, `compact_ratio = 0.8`
- Affected session: `events.frames` ≈ 10.5 MB (long-running session)

**Observed**

- Reasoning channel degenerated into **verbatim repetition of a single CJK character** (「好」) — hundreds of consecutive occurrences
- **No tool call and no visible text** in the affected attempts; only the stop button ended it
- Silent from the outside: `service.log` contains no repetition / degeneration / loop-guard events at all, so the turn looks "healthy" until the user notices nothing is happening

**Two factors I have not seen mentioned, either of which may change which traces trip**

1. **`temperature = 0.0` as an amplifier.** With greedy decoding, once the argmax lands on a high-frequency connective, every subsequent step can re-select it — there is no sampling noise to break out, so the stream self-reinforces into an absorbing state. If the thresholds were tuned on default-temperature traces, the `temperature = 0` case may behave differently from what those traces suggest.

2. **`reasoning_language = "zh"` and the one-rune unit.** Single-character CJK tokens give none of the whitespace / casing / word-choice variation that breaks English repetition naturally, and the repeated unit is exactly 1 rune — the `one-rune` case already raised against #8085's `repetitionSegmentMinRunes`. For this PR the corresponding question is whether a single-rune repeat reaches `>=8 repeats / >=160 bytes` before a meaningful amount of output is burned: 「好」 is 3 bytes in UTF-8, so ~53 occurrences would meet the byte threshold — it would fire, but only after ~53 repeated units. A shorter byte floor or a distinct-unit ratio over the window would catch it earlier without touching the sentence logic.

**Not filing a separate issue**

Since both PRs cover this shape and neither is merged, a new issue would only add noise. Happy to provide the session frames (`events.frames`, codec `reasonix.session.linear/v4`) or a redacted reasoning excerpt as a fixture if a single-rune CJK case is not already covered by the tests.

BuGlessRB:
I just checked the other PR.  This PR should deal just fine with any language tripping it, English, ZH, or otherwise.  The detection mechanism in this one probably eats less CPU and has less false triggers.  The downside is that it will cost a bit more tokens, but the total cost should still be negligible.  What this PR does different is that it leaves the session so far in there, mostly to not throw away reasoning from before.  It could be argued that exactly the repeating tail should be expunged.  Then again, if you expunge, that might exactly be the reason it might come back.  So I'm not certain what the best strategy is here.  The easiest way to trip it, is to have a long context (>700K), which seems to increase the likelihood of it tripping the endless repetition.

PR #10461, opened 2026-09-17
Fix Windows SQLite URI drive-letter authority / 修复 Windows SQLite URI 盘符解析

jf200s:
## Summary

Two Windows-only SQLite open failures shared one root cause: DSNs built as `url.URL{Scheme: "file"}` without the Windows drive-letter slash, so SQLite parses the drive letter as a URI authority: `SQL logic error: invalid uri authority: C: (1)`.

- `topicstate.BackupExisting` produced `file://C:/...` on every Windows upgrade backup. The workspace-state v1 upgrade runs this backup first, so the upgrade deadlocked: the state file never left v1, and every session open and send retried the broken backup, surfacing as "Send failed" / "session history load failed" (30+ occurrences in one afternoon on a real Windows machine).
- `sessioncatalog.Inspect` built its read-only DSN the same way. `sql.Open` is lazy, so `Inspect` set `StateReady` while every count query failed silently; `reasonix doctor sessions` has shown zeroed catalog counts on Windows since bf19574d3 (2026-08-10).

Fix: route both through `projectiondb.DiskFileDSN` (now exported), which already handles Windows drive paths (`file:///C:/...`) and is what every runtime open path uses. macOS and Linux DSNs are byte-identical to before.

## Issues

None filed yet; diagnosed from Windows desktop service logs (31 `invalid uri authority` occurrences on 2026-09-17).

## Verification

- `go test ./internal/topicstate ./internal/sessioncatalog ./internal/projectiondb` pass; `go vet`, `gofmt`, `go run ./tools/repolint` clean.
- New regression test asserts the backup DSN keeps no URI authority on Windows; `DiskFileDSN` unit test updated for the export.
- Real Windows 11 machine, packaged build from this branch: workspace-state upgrade completed (v1 to v3, previously permanently stuck), zero `invalid uri authority` errors across repeated restarts and sends.
- `reasonix doctor sessions` on the same catalog, before vs after: revision 0 / indexed 0 (silently wrong) becomes revision 15123 / indexed 184.

## Documentation impact

Documentation-impact: none - no documentation describes these SQLite DSN internals; both changes restore intended backup and catalog-inspect behavior

## Cache impact

Cache-impact: none - no cache-sensitive prompt, provider, or tool surfaces touched; only SQLite DSN construction changed
Cache-guard: go test ./internal/topicstate ./internal/sessioncatalog ./internal/projectiondb
System-prompt-review: N/A


jf200s:
Closing in favor of #10459, which landed the same root-cause fix with a more complete approach: the centralized internal/sqliteuri package (drive letters + UNC), fail-closed catalog inspection, and the legacy upgrade fixture work. Verified the Windows machine that hit 31x 'invalid uri authority: C:' over this weekend: after #10459's approach the workspace-state upgrade completes and the send/history failures stop. Happy to help verify on real Windows hardware if useful.

PR #10467, opened 2026-09-17
Test recovery visibility instead of marker timing / 验证恢复结果而非标记时序

SivanCola:
The `main-v2` push CI failed in `TestCoveredRecoveryCopyBecomesVisibleAfterMigratedParentDeletion` because it asserted that two internal migration markers were absent immediately after deletion. The background catalog worker can validly recreate signature-bound markers before `deleteSession` returns, so that assertion races with successful reconciliation.

This removes the transient marker assertion and keeps the observable contract: deleting the covering parent must expose the recovery copy with the correct topic identity in the project tree. That final assertion still fails if marker invalidation or reconciliation is broken.

Validation:

- focused test passed 20 consecutive runs
- focused race test passed 5 consecutive runs
- `gofmt` and `git diff --check`


SivanCola:
This timing-only test fix has been incorporated into #10465 so the release candidate and its full Desktop matrix are validated in one pass. The product-level assertions remain: the recovered copy must become visible with the correct TopicID and appear in the project tree.

Issue #10422, opened 2026-09-16
[Bug]: CLI writes sessions into sessions-v4 (v1.38.8+) but /resume only lists legacy *.jsonl, so new sessions never appear


gaozheng2001:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

first bad: v1.38.8; still present on main-v2 (a4b83056a). Last good: v1.38.7

### What happened?

From v1.38.8 on, the CLI binds its session to the v4 store: a new conversation is written to
<state root>/projects/<slug>/sessions-v4/<session-id>/{manifest.json,events.frames} and no `.jsonl`
path is allocated any more. But the CLI's own pickers — `/resume`, `/resume <n>`, `--resume <query>`,
`--continue` — still enumerate the legacy directory with `agent.ListSessions`, which only looks at
`*.jsonl`. Result: every conversation created with 1.38.8+ is invisible in `/resume` while the data is
perfectly intact on disk, and users reasonably conclude their history was lost.

`reasonix serve` already handles this correctly (GET /sessions merges `service.Query().List(ctx,"",100)`
into the same list as the legacy rows), so this is a CLI wiring gap, not a storage problem.

Expected: the CLI pickers should list sessions from the same service the writer uses (SessionService /
SessionRef), or at least keep a compatibility `.jsonl` checkpoint next to the v4 store so existing
pickers keep working.

Code path — write side (v4):
- internal/cli/cli.go: `SessionService: cliSessionService(sessionDir)` (always passed by the CLI profile).
- internal/cli/session_service.go: `session.RootForLegacyDir(sessionDir)` → `<slug>/sessions-v4`
  (RootForLegacyDir is defined in internal/session/api.go).
- internal/boot/boot.go: `ExclusiveSession: sessionService != nil`.
- internal/control/sessionpath.go (`EnsureSessionPath`): in the exclusive branch it calls
  `BindFreshSession(...)` and returns, so `agent.NewSessionPath` (the legacy `.jsonl` path) is never used.
- internal/config/paths.go: `SessionStoreDir()`/`ProjectSessionStoreDir()` → `sessions-v4`.

Code path — read side (legacy only):
- internal/cli/resume.go:24/73/100 → `agent.ListSessions(dir)`; internal/agent/save.go:2124 documents
  "every non-empty *.jsonl session under dir".
[...]

Issue #10421, opened 2026-09-16
[Bug]: /resume reports "in use by another Reasonix window or process" for failures that are not lease conflicts (e.g. unsupported event-log schema)


gaozheng2001:
### Version line

v2 — Go rewrite (1.x), main-v2 (active development)

### Exact version

desktop-v1.37.0 (CLI built from tag v1.37.0); the same wording path also exists on main-v2 (a4b83056a)

### What happened?

`/resume` (TUI) and the resume picker report **every** failure of `commitSessionSwitch` with the lease
wording "this session is in use by another Reasonix window or process; close the other Reasonix window
or process first", even when the failure has nothing to do with a lease — e.g. the session's event log
uses a newer schema than this build can read.

Real case hit in production: a session whose `.events.jsonl` is schema 2 on a build that only supports
schema 1 (v1.37.0). `/resume` said "in use by another Reasonix window or process", so I spent a long time
hunting for a second Reasonix process: checked `*.jsonl.lease.lock` files (none for that session),
probed them with `flock -n` (all held only by the two live shells' own current sessions), compared the
two running `reasonix` pids, restarted the window (no change), and switched workspaces (worked fine).
The actual cause was only visible via the startup path, which prints the truth:

  session event log for <session>.jsonl uses schema 2; this build supports up to 1

Expected instead: when the error is not `agent.ErrSessionLeaseHeld`, the TUI should print the real error
(e.g. `m.notice("resume: " + err.Error())`), or at least a different message for load failures. Being told
to "close the other Reasonix window or process" is an action that cannot possibly help.

Code path (current main-v2):
- internal/cli/resume.go:268 and internal/cli/resume_picker.go:133 —
  `m.notice("resume: " + sessionLeaseHeldNotice(err))` with no error-type check.
- internal/cli/session_lease.go:55 — `sessionLeaseHeldNotice` = `control.SessionInUseMessage(err) + "; " + control.SessionLeaseCloseHint`.
- internal/control/session_lease_keeper.go:630 — `SessionInUseMessage` returns the generic fallback whenever
  the error is not a `*agent.SessionLeaseError` (`!errors.As(err, &leaseErr) || leaseErr.Info == nil || leaseErr.Info.PID <= 0`).
  A real lease conflict renders differently: "this session is in use by another Reasonix process (pid N on host, since HH:MM)".
- The startup path is correct because it only substitutes lease wording for lease errors:
[...]

PR #10515, opened 2026-09-18
fix(cli): resume only the v4 store and migrate legacy sessions at startup / 仅从 v4 存储恢复并在启动时迁移旧会话

BuGlessRB:
Problem: after the CLI moved session writes to sessions-v4, the resume surface
(--continue, --resume, the pickers, /resume) still enumerated only the legacy
JSONL directory, so sessions created in the current format were invisible to
resume.

Fix: make resume/restore/selection read only the canonical sessions-v4 store,
and convert every older session once at startup.

- New internal/cli/resume_native.go: a native locator encodes a v4 identity for
  the path-shaped listing/grouping/completion/picker surfaces; resumeRows lists
  via session.Service.Query, hides empty sessions and managed session-events
  mirrors, and resume binds with Controller.OpenSession.
- resolveSessionQuery matches a session id, title, or preview; --copy duplicates
  through session.Service.CopySession.
- Drop the legacy lease/load/takeover path from run, the chat TUI, and serve
  startup; deprecate --takeover; remove the now-dead lease/switch helpers.
- New internal/cli/legacy_session_migration.go: a self-contained, temporary
  startup bridge that imports legacy JSONL and sessions-v3 into sessions-v4,
  preserving source bytes in the v4 target's legacy/ evidence directory, then
  deletes the sources once the target is verified (re-runs are no-ops).
- Preserve the source's own created/updated times on the migrated target so
  --continue does not pick a freshly imported session over a genuinely newer v4
  one.

Verification: go build ./..., gofmt, and tools/repolint clean; internal/cli and
internal/session tests green. The 3 transcript-viewport/mouse-mode failures are
pre-existing on main-v2.

Documentation-impact: updated - docs/CLI.md and docs/CLI.zh-CN.md describe the
v4-only resume store and the one-time legacy migration at startup.


gaozheng2001:
Thanks — this looks like exactly the right fix, and it closes a correctness gap that is still open three releases later. Could it get a review soon?
@esengine @SivanCola

Related reports, all the same root cause (the CLI writes to the v4 store, but its resume surfaces still enumerate the legacy JSONL directory):

- #10422  — "CLI writes sessions into `sessions-v4` (v1.38.8+) but `/resume` only lists legacy `*.jsonl`, so new sessions never appear". Traced to`internal/cli/resume.go` → `agent.ListSessions` → `ListSessionOrder` → `os.ReadDir` + `store.IsSessionTranscriptName` (`.jsonl` only); `internal/cli` had zero `sessions-v4` references on `main-v2@161da6c35`. `serve` already feeds `SessionService.Query().List` into `GET /sessions`, so the same machine lists these sessions over HTTP but not in the TUI.
- #10488 — 1.38.10: CLI cannot restore sessions via `/resume`. Same symptom, still reproducible two releases after the first report.
- #10405 — `chat --resume <query>` exits with an error when the query matches multiple sessions; this is the same `resolveSessionQuery` path that this PR rewrites.
- #10421 (mine) — `/resume` reports a lease conflict for failures that are not lease conflicts. Removing the CLI lease/takeover path here should retire that whole class of misreport.

Impact, stated plainly: since v1.38.8 the CLI has no visible history for any session it created itself — the data is intact on disk (`sessions-v4/<id>/{manifest.json,events.frames}`), users just cannot reach it, and v1.38.7 is the last release whose CLI can list its own sessions. That is why reports are being filed under `data-loss`, even though nothing is actually lost.

One review concern on the migration bridge, precisely because these reports are data-loss-adjacent:

- `internal/cli/legacy_session_migration.go` deletes the source `.jsonl` / sessions-v3 files once the v4 target verifies. The direction is right, but please consider leaving the sources in place by default (or an opt-out / short retention window) rather than deleting them in the same release that first introduces the reader. Migration non-convergence is not hypothetical here — see #10489 and #10509 for the Desktop v5 migration, a different root cause but the same failure mode, where a half-finished migration left workspaces permanently read-only. Deleting the only other copy at that moment is unrecoverable.

If a manual reproduction check helps: I can build this branch and exercise `--continue` / `/resume` / `--resume <query>` against an existing multi-project `sessions-v4` tree (13 sessions across 2 projects) and report back.

PR #10426, opened 2026-09-16
Add expandable sessions and composer workspace context / 添加可展开会话与工作区上下文

SivanCola:
## Problem

The desktop sidebar hides older conversations behind a time filter and pages only at the project level. Custom groups therefore cannot independently reveal older sessions, and the rendered tree can diverge from navigation, diagnostics, and current-session visibility.

New empty sessions also inherit their workspace implicitly. Project and branch selection live elsewhere in the desktop, so starting work in another folder or branch requires leaving the composer. A first pass at adding this context row kept the creation accent on the lower composer, which produced a broken U-shaped outline instead of one continuous control.

## Result

- remove the sidebar time-filter entry and the alternate time/recent organization modes
- show 5 conversations per project or custom group, then expand in batches of 10 with independent collapse state
- keep an active conversation visible as an extra row without consuming the normal window
- use isolated 50-result search pagination and restore ordinary window state after search
- preserve pinned, remote, archive, drag-order, runtime-status, and navigation behavior
- remove the disabled `Archive all chats` placeholder while keeping the `Created time` and `Updated time` sort choices unchanged
- align the three project-header controls with the project-row action columns
- migrate the two retired local UI preference keys without changing session or project data
- add a ZCode-style project and Git branch context row to new empty sessions
- support project search and switching, opening another folder, remote connection, and working without a project
- support branch search, checkout, branch creation, and a keyboard-contained Git history dialog
- hide the context row after the session has content, preserving the compact follow-up composer
- render the context row and composer inside one complete rounded frame with shared focus and running accents

## Implementation

The frontend owns one keyed list controller for visible limits, pagination state, request generations, retries, and the final row projection. The Desktop API adds optional group and pinned filters to `ListProjectTopics`; omitted fields retain the previous unfiltered behavior. The catalog applies group membership before sorting and pagination through `json_each`, binds cursors to the full list identity and membership revision, and uses direct topic lookup for active-session fallback.

The composer context control is lazy-loaded with its styles, so it does not add to the initial app-shell JavaScript or render-blocking CSS. It reuses the existing project creation, remote connection, branch switching, and Git history bridges. Request generations prevent stale project and branch responses from updating a changed workspace, and the active branch remains visible when it is temporarily absent from the branch-list response.

The workspace row and input card share a single frame owner. The child surfaces keep only their internal separator; theme-level composer borders, shadows, top corners, and animated pseudo-elements are suppressed inside that frame so the orange status line remains continuous.

[...]

PR #10544, opened 2026-09-19
fix(desktop): derive the draft composer workspace from the draft / 草稿 composer 的工作区上下文改从草稿本身推导

jf200s:
## Problem

With a draft surface open, the composer workspace bar showed the backing tab's previous project (creating a draft in BigWarGame while an animation-project session tab was active showed the old project), and when no tab backed the draft the bar disappeared entirely, leaving no way to see or switch the draft's target workspace.

## Root cause

- The composer workspace context was built only from the active tab; the draft surface override patched scope and workspaceRoot but kept the tab's workspaceName, so the checkmark (root comparison) and the label (explicit name) could disagree.
- With no backing tab the base context was undefined and the bar never rendered. 1.38.10 predates draft surfaces and was unaffected.

## Fix

- Make the base context draft-aware: when a draft surface is open and project-scoped, scope and workspaceRoot come from the draft and the label derives from the draft root (same basename rule as the backend tab meta and the draft title bar). One construction now serves tabs and drafts; the draft override stays as-is.
- Keep the context alive for a draft surface even when the backing tab view has left the hero state (a draft is its own empty session).

## Related

- #10426 introduced the composer workspace context; this PR makes the same context draft-aware. #10507 later touched the same file for interaction identities; no overlap with this change.

## Verification

- tsc --noEmit, eslint on the changed file
- footer-decision-overflow (19), navigation-surface-transition (36), navigation-surface-lifecycle, composer-workspace-context suites pass
- Manual: draft with an animation-project tab behind it shows the picked project in the bar; draft without any backing tab keeps the bar

Documentation-impact: none - behavior fix only; no docs describe the draft workspace bar's selection state.


jf200s:
Closing in favor of #10551, which adopts this fix as part of the broader draft landing rework — the draft-aware composer context (`view.hero || input.draft?.surface`) and the draft-derived workspace label are identical to this PR's change. Verified on Windows before opening. Thanks for picking it up.

PR #10585, opened 2026-09-20
build: generate Windows ARM64 installer v1.38.10-3 [do not merge]

SivanCola:
Build-only PR used to generate and validate the Windows ARM64 installer for `v1.38.10-3` from exact `main-v2` commit `ec82bb3251550b03b43b83418a716f73fe758ffd`.

Do not merge. The packaging job runs on the native `windows-11-arm` runner, builds `windows/arm64`, installs and starts the result, verifies upgrades and data preservation, then uploads the installer and SHA-256. This PR will be closed after the artifact is downloaded and verified. It does not create a tag or release.

Local validation:
- `node --test scripts/ci-workflow.test.mjs`
- `go run ./tools/repolint`
- `git diff --check`


goransh-walia:
Clean and minimal — appreciate that it stays scoped to the actual fix.

